PR #3, opened 2011-05-19
[Bug 658133] - Provide means to hide locked questions and default to doing so unless the user has the ability to lock threads

sdaugherty:
- Added parameter showlocked to determine if locked question should be shown.
- Default this parameter to yes if the user has the permission to lock threads.
- Default this parameter to no if the user does not, or if the user is not logged in.
- Added UI to toggle hiding locked threads.


Bonniemarie216 (approved):

Bonniemarie216 (approved):

Bonniemarie216 (approved):

Bonniemarie216 (commented):
Security Updates

Bonniemarie216 (commented):
Security Updates

PR #19, opened 2011-06-02
Diff revision picker (translate and review translation pages)

rlr:
r?


robert8210 (commented):
image

robert8210 (commented):
So far it's running pretty good.  Good Job

PR #2220, opened 2014-11-14
[bug 1013471] Fix for Accidental deletion of profiles

anushbmx:
Added a checkbox which needs to checked to activate the "Close Account and delete all my profile information"

With checkbox unchecked 
screenshot from 2014-11-14 16 06 14

Checkbox checked ( Delete Buttons becomes active )
screenshot from 2014-11-14 16 06 20

### DIFF
--- a/kitsune/sumo/static/js/ui.js
+++ b/kitsune/sumo/static/js/ui.js
@@ -7,6 +7,14 @@
     initFolding();
     initAnnouncements();
 
+    $('#delete-profile').click(function(){
+      if($("input[type='checkbox'].acknowledge:checked").length){
+        $(".delete-profile-btn").removeAttr('disabled');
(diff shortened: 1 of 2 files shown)


rlr:
This still doesn't address @rehandalal's comment in the bug. I agree that a confirm modal is better. If you are worried about js not being loaded yet, then you can render the HTML with the button disabled and enable it once JS runs.

anushbmx:
@rlr and @mython sorry I was a little busy for a few days will come fix it in a couple of days :)

mythmon:
@anushbmx this PR is quite old at this point. I'm going to close it for now. The master branch has changed a lot since then, so it will be important to rebase or rewrite this against master. Please open a new PR when you work on this next.

PR #2263, opened 2014-12-11
[Bug 1083405 ] Question/Document suggestion search API.

mythmon:
WIP

### DIFF
--- a/kitsune/search/api.py
+++ b/kitsune/search/api.py
@@ -1,0 +1,98 @@
+import itertools
+
+from django.conf import settings
+
+from rest_framework.decorators import api_view
+from rest_framework.response import Response
+
+from kitsune.products.models import Product
+from kitsune.questions.models import QuestionMappingType
+from kitsune.search import es_utils
+from kitsune.sumo.api import GenericAPIException
+from kitsune.wiki.models import DocumentMappingType
+
+
+@api_view(['GET', 'POST'])
+def suggest(request):
+    text = request.body or request.GET.get('q')
+    locale = request.GET.get('locale', settings.WIKI_DEFAULT_LANGUAGE)
+    product = request.GET.get('product')
+    max_questions = request.GET.get('max_questions', 10)
+    max_documents = request.GET.get('max_documents', 10)
+
+    errors = {}
+    try:
+        max_questions = int(max_questions)
+    except ValueError:
+        errors['max_questions'] = 'This field must be an integer.'
+    try:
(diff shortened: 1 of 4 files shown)


mythmon:
I had a silly mistake in my tests: the API filters to only show questions with helpful answers, but I wasn't making any answers at all for my test questions.

After I figured that out, this was pretty easy to figure out.

kitsune/search/api.py
```diff
@@ -0,0 +1,98 @@
...
+from rest_framework.response import Response
+
+from kitsune.products.models import Product
+from kitsune.questions.models import QuestionMappingType
+from kitsune.search import es_utils
+from kitsune.sumo.api import GenericAPIException
+from kitsune.wiki.models import DocumentMappingType
+
+
+@api_view(['GET', 'POST'])
```
rlr: Does it need POST?

kitsune/search/api.py
```diff
@@ -0,0 +1,98 @@
...
+        wiki_f &= es_utils.F(product=product)
+        questions_f &= es_utils.F(product=product)
+
+    mapping_types = [QuestionMappingType, DocumentMappingType]
+    query_fields = itertools.chain(*[cls.get_query_fields() for cls in mapping_types])
+    query = {}
+    for field in query_fields:
+        for query_type in ['match', 'match_phrase']:
+            key = '{0}__{1}'.format(field, query_type)
+            query[key] = text
```
rlr: One difference between this and normal search is that boosts aren't being applied. Just worth noting, not sure if it's needed.

rlr:
Otherwise this LGTM!

kitsune/search/api.py
mythmon: This can read from the body of the text, to support long user input. Although it is possible to have request bodies on GETs, a lot of http libraries don't like it. POST is here simply to allow the nicer use case on dumb libraries. Elasticsearch does the same thing.

kitsune/search/api.py
mythmon: I think this is blocking buddyup stuff a bit, so I'd rather land it and play with boosts later. Sound good to you?

rlr:
R+

mythmon:
c3dc6db [Bug 1120480] Use correct serialization for users in Questions API.

mythmon:
Oops. This was actually c8ee9c9 [Bug 1083405] Add suggested articles and questions API.

c3dc6db was #2305.

PR #2286, opened 2014-12-29
[bug 1111801] Fix git checkout on Windows

willkg:
Windows file system doesn't like the really long filename we had for
this test and will balk on git checkout. So we change the test to create
the file first, then run the test rather than have the file in the repo
contents.

r?


mythmon:
I thought you were going to do more work here, but I'm fine with it as is. This fixes being able to clone on Windows, the test will pass on Linux, and fail on Windows. I think that's good. Sorry for the confusion!

r+

willkg:
Landed! a8e4777 [bug 1111801] Fix git checkout on Windows

PR #2289, opened 2015-01-01

safwanrahman:
This will make Users Add Mozillians.org profile link into SUMO profile. :)


safwanrahman:
@mythmon I can not understand why this can not pass! :(
In my local machine, its working perfectly! 
firefox_screenshot_2015-01-02t00-03-48 590z

PR #2290, opened 2015-01-02
Update from 2014 to 2015

safwanrahman:
As year is changed, we should change this also!
Happy New Year!


rehandalal:
lgtm. thanks! i'll merge.

rehandalal:
b6e70e72025a88ffb59576a2678b8db40c90695a

PR #2291, opened 2015-01-03

safwanrahman:
I hope this will pass and fix the issue


safwanrahman:
@mythmon I can not undersand why this Travis Ci Builds has been failed. I have done the migration and Its working fine in my local machine.
firefox_screenshot_2015-01-03t01-25-05 693z

kitsune/users/migrations/0006_auto__add_field_profile_mozillians.py
```diff
@@ -1,22 +1,23 @@
 # -*- coding: utf-8 -*-
 import datetime
 from south.db import db
-from south.v2 import DataMigration
```
mythmon: Do not edit migrations.

Migrations should never be migrated. You need to add a new migration, not change an old one. Changing migrations will cause huge problems. For example, the users 0006 migration has already been run in production. When this code goes out, the new migration will not be run.

Do no edit migrations.

PR #2292, opened 2015-01-03
Adding Open SSL and Mysql Library in requirments of Kitsune Doc

safwanrahman:
While installing Kitsune, following Library is needed. 
* Open SSL Library named libssl
* MySQL Library named libmysqlclient

So this both should be added into Requirments.

### DIFF
--- a/docs/hacking_howto.rst
+++ b/docs/hacking_howto.rst
@@ -91,7 +91,6 @@ These are required for the minimum installation:
 * libjpeg and headers
 * zlib and headers
 * libssl library
-* libmysqlclient library
 
 These are optional:
 
--- a/docs/hacking_howto.rst
+++ b/docs/hacking_howto.rst
@@ -90,7 +90,7 @@ These are required for the minimum installation:
 * libxslt and headers
 * libjpeg and headers
 * zlib and headers
-* libssl library
+* libssl and headers
 
 These are optional:
 
--- a/docs/hacking_howto.rst
+++ b/docs/hacking_howto.rst
@@ -90,6 +90,8 @@ These are required for the minimum installation:
 * libxslt and headers
 * libjpeg and headers
 * zlib and headers
+* libssl library
+* libmysqlclient library
 
 These are optional:
 


docs/hacking_howto.rst
```diff
@@ -90,6 +90,8 @@ These are required for the minimum installation:
 * libxslt and headers
 * libjpeg and headers
 * zlib and headers
+* libssl library
+* libmysqlclient library
```
willkg: This line is covered by the MariaDB 5.5 server and client headers line.

Why do we need libssl? Also, it's probably not that you need libssl, but rather that you need the headers.

safwanrahman: I have installed all the headers, but Open SSL was still needed. So addedd that.
And regarding the libmysqlclient library, MariaDB dont install it.
@willkg, can you please check with a fresh installation in Ubuntu?

willkg: I'm not sure what you want me to check. Can you walk me through exactly what I'm checking?

safwanrahman: @willkg I meant to check installing Kitsune in Ubuntu.
Whie I try to install in Ubuntu, though I installed all the Requirments, it needed that 2 Library also.
So I proposed you to check again!

mythmon: @safwanrahman: when @willkg said "This line is covered by the MariaDB 5.5 server and client headers line." he means that `libmysqlclient` is "MariaD 5.5 client headers". You are just listing the explicit package name for something we already have in the list. Since this is a generic list of dependencie, not a list of packages to install, this line isn't appropriate. It would be better put in an Ubuntu-specific install page.

safwanrahman:
@mythmon, I removed the libmysql client. So is it Ok now @willkg ?
Open SSL is needed while installation :)

mythmon:
The SSL dev headers are needed in Vagrant, so they should be included here. The line should be rewritten to match the rest of the requirements though "libssl and headers".

safwanrahman:
@mythmon Corrected as per needed. Hope this can be merged this time :)

mythmon:
bcc31d6 Adding Open SSL in requirments

Thanks!

PR #2294, opened 2015-01-05
Make migrations users/0006 more independent.

mythmon:
@safwanrahman was having trouble with this: adding new fields to our profile model breaks this migration, since it calls all the bits and bobs to create a new profile as well as a new user. The fields would be missing, and things would break.

r?

### DIFF
--- a/kitsune/users/migrations/0006_add_migration_user.py
+++ b/kitsune/users/migrations/0006_add_migration_user.py
@@ -3,16 +3,17 @@ import datetime
 from south.db import db
 from south.v2 import DataMigration
 from django.db import models
+from django.contrib.auth.hashers import make_password
 
 class Migration(DataMigration):
 
     def forwards(self, orm):
         """Adds a user to be used for migrations."""
         from django.contrib.auth.models import User
-        # Setting password to None makes it an unusable account.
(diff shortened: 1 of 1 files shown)


kitsune/users/migrations/0006_add_migration_user.py
```diff
@@ -3,16 +3,17 @@
 from south.db import db
 from south.v2 import DataMigration
 from django.db import models
+from django.contrib.auth.hashers import make_password
 
 class Migration(DataMigration):
 
     def forwards(self, orm):
         """Adds a user to be used for migrations."""
         from django.contrib.auth.models import User
```
rehandalal: you can probably ditch this line too

safwanrahman:
Thanks @mythmon for this patch. I was having problem with Pull Request #2291 . I hope this patch fix the issue.

kitsune/users/migrations/0006_add_migration_user.py
mythmon: Oh, yes.

mythmon:
Feedbacked.

rehandalal:
lgtm r+

mythmon:
193aea1 Make migrations users/0006 more independent.

PR #2295, opened 2015-01-05

safwanrahman:
This Will make a User/Contributor to add mozillians profile link in their profile


rlr:
I wonder if we should ask for the Mozillians username instead of the URL. What do you think?

safwanrahman:
I think link is Better. So if anyone want to see their Mozillians profile, they justclick on the Link and go!
If we add the Mozillians Username, it will be time consuming to find out anyones profile!

Or we Can do as following
User will add only Mozillians URL, but it will make a link redirect to their Mozillians Profle. What do you think?

rlr:
@safwanrahman if they provide the username, then we can create the URL and link them properly because we know what the URLs look like. I just think it's less confusing that way.

Also, travis doesn't like something.

mythmon:
Travis failed because of a linting error:

```
$ scripts/lint.sh
kitsune/users/models.py:79:34: E128 continuation line under-indented for visual indent
```

safwanrahman:
@rlr @mythmon I think its Ok now. and Travis pass also. 
r?

kitsune/users/templates/users/mobile/profile.html
```diff
@@ -80,6 +80,10 @@
           <header>{{ _('Facebook') }}</header>
           <a class="detail" rel="nofollow" href="{{ profile.facebook }}">{{ profile.facebook }}</a>
         {% endif %}
+        {% if profile.mozillians %}
+          <header>{{ _('Mozillians') }}</header>
+          <a class="detail" rel="nofollow" href="http://mozillians.org/u/{{profile.mozillians}}">{{ profile.mozillians }}</a>
```
rlr: You should link straight to https and save the redirect.

rlr: Also, it looks like they like the trailing slash: https://mozillians.org/en-US/u/safwan/

Add a `/` to the end and save another redirect.

kitsune/users/templates/users/profile.html
```diff
@@ -125,6 +125,12 @@ <h2 class="location">
                 <a rel="nofollow" href="{{ profile.facebook }}">{{ profile.facebook }}</a>
               </li>
             {% endif %}
+            {% if profile.mozillians %}
+              <li>
+                <label>{{ _('Mozillians') }}</label>
+                <a rel="nofollow" href="http://mozillians.org/u/{{profile.mozillians}}">{{ profile.mozillians }}</a>
```
rlr: same here

rlr:
Some minor comments. I like it this way!

rlr:
Awesome. Gonna test it now.

rlr:
works great. r+

rlr:
24b03fbe523a55c361c53ff3d5393e9711caa8d3 [Bug 1015564] Add Mozillians.org profile link into SUMO profile

rlr:
Thank you @safwanrahman !

PR #2296, opened 2015-01-06
[bug 1104909] Address IntegrityErrors with image uploads

rehandalal:
We should also probably have a cron for cleaning these up.

r?

### DIFF
--- a/kitsune/gallery/utils.py
+++ b/kitsune/gallery/utils.py
@@ -1,3 +1,5 @@
+from datetime import datetime
+
 from django.conf import settings
 from django.core.exceptions import PermissionDenied
 from django.core.files import File
@@ -62,4 +64,5 @@ def check_media_permissions(media, user, perm_type):
 
 
 def get_draft_title(user):
-    return u'Draft for user %s' % user.username
+    return u'Draft for user %s. Created at: %s' % (user.username,
+                                                   datetime.now())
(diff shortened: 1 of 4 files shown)


rlr:
Can you explain how this solves the integrity errors? I was expecting something more involved than some relatively simple UI changes.

kitsune/gallery/utils.py
```diff
@@ -62,4 +64,5 @@ def check_media_permissions(media, user, perm_type):
 
 
 def get_draft_title(user):
-    return u'Draft for user %s' % user.username
+    return u'Draft for user %s. Created at: %s' % (user.username,
+                                                   datetime.now())
```
rehandalal: The error occurs because we used to just call the draft image "Draft for user <username>" and if the image never got uploaded and the draft never got deleted, this image would persist. Causing integrity errors when the user tries to upload a new image because it can't create the intermediate draft step.

rehandalal:
@rlr see comment above.

kitsune/gallery/utils.py
rehandalal: This adds a timestamp to make sure that the drafts have a unique name.

rlr: ah! that makes sense. OK let me just do some quick testing to double check things work.

rlr:
yeah, I think we'll have more images get left behind this way. Definitely need that cron job. File that in a new bug?

r+

rehandalal:
I filed the follow up bug

rehandalal:
0e8d9b435736b2ce07f78ef15338112ef7e09c35

PR #2299, opened 2015-01-07
[Bug 1116619] Add "Needs Change" Comment in "Edit Artcle" Page

safwanrahman:
This will make "Needs Changes comment" showing in the Edit Article page.

The UI will be as following
rnqh8za

### DIFF
--- a/kitsune/wiki/templates/wiki/edit.html
+++ b/kitsune/wiki/templates/wiki/edit.html
@@ -22,6 +22,12 @@
 
     <article id="edit-document" {% if locked %}class="locked"{% endif %} data-slug="{{ document.slug }}">
       <h1>{{ _('<em>Editing</em> {title}')|fe(title=document.title) }}</h1>
+      {% if document.needs_change %}
+        <p>
+          {% trans comment=document.needs_change_comment %} <b>Needs Change:</b> {{comment}} {% endtrans %}
+        </p>
+      {% endif %}
+
       {% if document_form %}
         <div id="document-form">
           <details class="h2"{% if disclose_description %} open="open"{% endif %}>
--- a/kitsune/wiki/templates/wiki/edit.html
+++ b/kitsune/wiki/templates/wiki/edit.html
@@ -25,7 +25,7 @@
       {% if document.needs_change %}
         <p>
           {% trans comment=document.needs_change_comment %}
-          <label>Needs Change:</label> {{comment}} 
+            <label>Needs Change:</label> {{comment}} 
           {% endtrans %}
         </p>
       {% endif %}
--- a/kitsune/wiki/templates/wiki/edit.html
+++ b/kitsune/wiki/templates/wiki/edit.html
@@ -24,7 +24,7 @@
       <h1>{{ _('<em>Editing</em> {title}')|fe(title=document.title) }}</h1>
       {% if document.needs_change %}
         <p>
-          {% trans comment=document.needs_change_comment %} <strong class="needs-change>">Needs Change:</strong> {{comment}} {% endtrans %}
+          {% trans comment=document.needs_change_comment %} <strong>Needs Change:</strong> {{comment}} {% endtrans %}
         </p>
       {% endif %}
 
--- a/kitsune/wiki/templates/wiki/edit.html
+++ b/kitsune/wiki/templates/wiki/edit.html
@@ -24,7 +24,9 @@
       <h1>{{ _('<em>Editing</em> {title}')|fe(title=document.title) }}</h1>
       {% if document.needs_change %}
         <p>
-          {% trans comment=document.needs_change_comment %} <strong>Needs Change:</strong> {{comment}} {% endtrans %}
+          {% trans comment=document.needs_change_comment %}
+          <label>Needs Change:</label> {{comment}} 
+          {% endtrans %}
         </p>
       {% endif %}
 
--- a/kitsune/wiki/templates/wiki/edit.html
+++ b/kitsune/wiki/templates/wiki/edit.html
@@ -24,7 +24,7 @@
       <h1>{{ _('<em>Editing</em> {title}')|fe(title=document.title) }}</h1>
       {% if document.needs_change %}
         <p>
-          {% trans comment=document.needs_change_comment %} <b>Needs Change:</b> {{comment}} {% endtrans %}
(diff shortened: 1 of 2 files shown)


kitsune/wiki/templates/wiki/edit.html
```diff
@@ -22,6 +22,12 @@
 
     <article id="edit-document" {% if locked %}class="locked"{% endif %} data-slug="{{ document.slug }}">
       <h1>{{ _('<em>Editing</em> {title}')|fe(title=document.title) }}</h1>
+      {% if document_needs_change %}
+        <p>
+          {% trans comment=_(document.needs_change_comment) %} <b>Needs Change:</b> {{comment}} {% endtrans %}
+        </p>
+      {% endif %}
```
mythmon: You should simply use `{% if doc.needs_change %}` instead of introducing a new variable name.

kitsune/wiki/templates/wiki/edit.html
```diff
@@ -22,6 +22,12 @@
 
     <article id="edit-document" {% if locked %}class="locked"{% endif %} data-slug="{{ document.slug }}">
       <h1>{{ _('<em>Editing</em> {title}')|fe(title=document.title) }}</h1>
+      {% if document_needs_change %}
+        <p>
+          {% trans comment=_(document.needs_change_comment) %} <b>Needs Change:</b> {{comment}} {% endtrans %}
```
mythmon: Don't put the needs_change comment into `_`. `_` is only for translating static strings (or very special database driven variables with special tags). Even if it did need translated, you are translating it twice here, once with `_` and again with `{% trans ... %}`.

safwanrahman: :) Got it. Fixing!

kitsune/wiki/templates/wiki/edit.html
safwanrahman: I thought Adding the Variable makes the HTML Code looks nice! no matter. Correcting as per you told! :+1:

safwanrahman:
@mythmon Have corrected as per you mentioned. now r?

kitsune/wiki/templates/wiki/edit.html
```diff
@@ -22,6 +22,12 @@
 
     <article id="edit-document" {% if locked %}class="locked"{% endif %} data-slug="{{ document.slug }}">
       <h1>{{ _('<em>Editing</em> {title}')|fe(title=document.title) }}</h1>
+      {% if document.needs_change %}
+        <p>
+          {% trans comment=(document.needs_change_comment) %} <b>Needs Change:</b> {{comment}} {% endtrans %}
```
mythmon: The parenthesis are not needed.

mythmon:
One more thing, then this is ready to go.

safwanrahman:
@mythmon Removed the parenthesis. Hope its ready to go now!

safwanrahman:
@rlr this one? r?

kitsune/wiki/templates/wiki/edit.html
```diff
@@ -22,6 +22,12 @@
 
     <article id="edit-document" {% if locked %}class="locked"{% endif %} data-slug="{{ document.slug }}">
       <h1>{{ _('<em>Editing</em> {title}')|fe(title=document.title) }}</h1>
+      {% if document.needs_change %}
+        <p>
+          {% trans comment=document.needs_change_comment %} <b>Needs Change:</b> {{comment}} {% endtrans %}
```
rlr: `<b/>` isn't a good tag to use because it implies a certain style and that should be defined in CSS. `<strong/>` is better.

But I think it would be best to be consistent with the form below. Something like:

```
<label>{{ _('Needs Change:') }}</label>
<p>{{ comment }}</p>
```

Then make sure that the label is bolded via CSS.

mythmon:
@safwanrahman, your branch seems to have picked up some stray test failures from a broken commit that was on master. Can you rebase this on the current master? Those test failures are my fault.

safwanrahman:
@mythmon Thanks for the Comment. Yes. I have rebased it and now it seems fine :)

safwanrahman:
@rlr I have tried with CSS file changing, but that seems Critical for localization. I thing using <storng/> is better way. While localizing, localizers are habituated to see <strong/>
I would not like  to remove {{comment}} tag outside of Localizer Concern. So there is no other way without using the <strong/>

rlr:
@rehandalal @mythmon thoughts on the semantics if the HTML here ^^'

I personally think it's best to avoid HTML in l10n strings wherever possible.

mythmon:
Because this is inline content, it is important to get it right for RTL languages. Because of that, I would do 

```jinja
{% trans comment=document.needs_change_comment %}
    <label>Needs Change:</label> {{comment}}
{% endtrans %}
```

And use a CSS rule like 

```css
#edit-article label {
    font-weight: bold;
}
```

safwanrahman:
@mythmon The Fact is, this change will not affect in RTL language because **Need changes comment** will be show only in en-US. There is no chance to show it in other locals. So I think its not important to make it right for RTL.

I am trying to do as you said, but it seems that all the CSS is broken if do that. I think need to write all the CSS of this page if try to implement that. Current time, its sharing CSS style of other page.

safwanrahman:
@mythmon @rlr I have made the Changes as you mentioned in comment. 
Now it maches all the needs. r?

kitsune/wiki/templates/wiki/edit.html
```diff
@@ -22,6 +22,14 @@
 
     <article id="edit-document" {% if locked %}class="locked"{% endif %} data-slug="{{ document.slug }}">
       <h1>{{ _('<em>Editing</em> {title}')|fe(title=document.title) }}</h1>
+      {% if document.needs_change %}
+        <p>
+          {% trans comment=document.needs_change_comment %}
+          <label>Needs Change:</label> {{comment}}
```
mythmon: This line needs indented, since it is nested in the {% trans %} block. After that, I think this is ready to go.

safwanrahman:
@mythmon @rlr , Have corrrected the Indent. Now ready to go? ;)
r?

mythmon:
Awesome, thanks.

b52361f [Bug 1116619] Add "Needs Change" Comment in "Edit Artcle" Page

PR #2301, opened 2015-01-07
[bug 1118783] Change .save() for .update()

rlr:
This change fixes the task for me when running from the admin node. I am still puzzled as to how the save() was failing silently though.

r?


mythmon:
That's weird. Do you know why didn't run into this before?

r+ once Travis likes it.

rlr:
9b82e2f73c2e6424e1004995d7dadec8ffafb30a [bug 1118783] Change .save() for .update()

PR #2303, opened 2015-01-09
Split lint and main suite apart for Travis runs.

mythmon:
This should be clearer when inspecting the output.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,16 +2,22 @@ sudo: no
 language: python
 python: "2.6"
 node: "0.10"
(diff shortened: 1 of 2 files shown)


mythmon:
This made Travis do what I wanted. Now the lint run is shown as a separate build, which should make it easier to tell when just the linter fails.

r?

rlr:
that's awesomesauce! r+

mythmon:
a3e5edc Split lint and main suite apart for Travis runs.

PR #2304, opened 2015-01-12
[bug 1119920] Truncate search Record text at 255 chars.

rlr:
r?

### DIFF
--- a/kitsune/search/tasks.py
+++ b/kitsune/search/tasks.py
@@ -122,6 +122,9 @@ def index_chunk_task(write_index, batch_id, chunk):
     except Exception:
         rec.text = u'{0}: Errored out {1} {2}'.format(
             rec.text, sys.exc_type, sys.exc_value)[:255]  # Truncate at 255 chars.
+
+        log.exception('Error indexing a chunk')
+
         # Some exceptions aren't pickleable and we need this to throw
         # things that are pickleable.
         raise IndexingTaskError()
--- a/kitsune/search/tasks.py
+++ b/kitsune/search/tasks.py
@@ -82,7 +82,10 @@ def reconcile_task(write_index, batch_id, mapping_type_name):
 
     except Exception:
         rec.text = u'{0}: Errored out {1} {2}'.format(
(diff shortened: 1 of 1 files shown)


kitsune/search/tasks.py
```diff
@@ -121,7 +121,7 @@ def index_chunk_task(write_index, batch_id, chunk):
 
     except Exception:
         rec.text = u'{0}: Errored out {1} {2}'.format(
-            rec.text, sys.exc_type, sys.exc_value)
+            rec.text, sys.exc_type, sys.exc_value)[:255]  # Truncate at 255 chars.
```
willkg: If you truncate this this way, then you're going to be losing important parts of the error. Any idea how much over 255 the text is? Should we send the error in an email instead?

rlr: I don't know how big of an error it was. I was going to try this first and see what it looks like. Then increase the column size if that was needed.

rlr: oh, we should probably `log.exception()`. That goes to Sentry right?

willkg: `log.exception()` might be a good idea. I don't know where it goes, though.

Also, I think I'd opt for increasing the size of the column instead.

rlr: Actually, why dont we just re-raise instead of `raise IndexingTaskError()` ?

willkg: The note says that some exceptions aren't pickleable. Pretty sure the problem is that this can be executing in celery-land, so it's important to throw pickleable things otherwise celery can't pickle it as a return value and then things get hosed.

rlr:
Given that `log.exception()` goes to Sentry, I think we are safe truncating at 255. Even if we doubled or quadrupled it, those nasty ES exceptions still wouldn't fit.

Thoughts?

willkg:
Ok. I'm game.

rlr:
08594ba3c17f088e97346eccba53deeed31a8f18 [bug 1119920] Truncate search Record text at 255 chars.

PR #2305, opened 2015-01-12
[Bug 1120480] Use correct serialization for users in Questions API.

mythmon:
It was pointed out that it's a lot more useful for API consumers to have both the username (for API reference) and the display name (for showing to users) for every user. This makes that the case for every time a user is referenced in the question and answer serializers.


mythmon:
Oops, I merged this too soon. I can back this out if we need to.

rlr:
r+

mythmon:
This was c3dc6db [Bug 1120480] Use correct serialization for users in Questions API.

PR #2308, opened 2015-01-13
[bug 1087854] Refresh Firefox button on product page

rehandalal:
r?


kitsune/sumo/static/js/products.js
```diff
@@ -1,20 +0,0 @@
-(function($, BD) {
```
rlr: yay! KILL KILL KILL

rlr:
The bug says "If they already have the latest version of Firefox, we display a Refresh Firefox button along "download a fresh copy" link.". So you probably need to hide that when they aren't on the latest version of Firefox.

rehandalal:
phew, fixed ^

kitsune/sumo/static/less/products.less
```diff
@@ -14,6 +14,7 @@
   font-family: 'Open Sans','Droid Arabic Naskh','Mplus Fulah',sans-serif;
   position: absolute;
   right: 0;
+  text-align: center;
```
rlr: I *think* this screwed up the download button. See http://cl.ly/image/3Z1S183d1412

left is localhost and right is prod.

rlr:
Other than the style issue. LGTM

kitsune/sumo/static/less/products.less
rehandalal: yep that's the issue. should be fixed.

rehandalal:
fixed ^

rlr:
r+! :fire:

rehandalal:
a58feaebd4c9cb5a97ef1f97fd74354894bd8dca

PR #2312, opened 2015-01-14
[Bug 1121627] Use the right query for search suggestions API.

mythmon:
I had these lines in the function while I was pulling my hair out trying to debug things. Removing them makes documents work again, and makes it so we aren't always searching for "emails".

This was pretty dumb of me.

r?


rlr:
r+!

mythmon:
a5b2491 [Bug 1121627] Use the right query for search suggestions API.

PR #2315, opened 2015-01-14
Respect new lines in forum blockquotes as well.

mythmon:
No bug for this, because it is a tiny CSS change.

The background change is the same as the one I did in questions, and makes nested block quotes clearer when there is more than two levels.

Related: https://support.mozilla.org/en-US/forums/contributors/710984#post-63662

r?


rlr:
r+

mythmon:
87e91de Respect new lines in forum blockquotes as well.

PR #2316, opened 2015-01-19
Upgrade peep to 2.1.1

safwanrahman:
Upgraded to Peep 2.1.1 which is supported pip 6.0.6
* Have tested by installing pip 6.0.6
r?

@mythmon We should make travis-ci run the latest version of pip also! :)

### DIFF
--- a/scripts/peep.py
+++ b/scripts/peep.py
@@ -27,9 +27,11 @@ from os import listdir
 from os.path import join, basename, splitext, isdir
 from pickle import dumps, loads
 import re
+import sys
 from shutil import rmtree, copy
 from sys import argv, exit
 from tempfile import mkdtemp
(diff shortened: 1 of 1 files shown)


mythmon:
I installed things with this, using both pip 1.5.6 and 6.0.6. Both worked fine, and the tests work.

r+

I don't think we need to make Travis run with pip 6.y.z. Peep is well tested against versions of pip, we don't need to duplicate that in our tests.

mythmon:
bb20936 Upgrade peep to 2.1.1

BReduardokramer:
Ok. I will upgrade peep when done.
Em 19/01/2015 19:11, "Safwan Rahman" <notifications@github.com> escreveu:

PR #2318, opened 2015-01-20
[bug 971009] Upgrade to celery v3.1.17

rlr:
Here is the fjord commit for reference: https://github.com/mozilla/fjord/commit/9dff1949b093fc9e192ace41f876f6becf802f55

r?

### DIFF
--- a/requirements/default.txt
+++ b/requirements/default.txt
@@ -7,9 +7,8 @@ anyjson==0.3.1
 # sha256: SjoIXs8fzSc2VzU4_6EU8fQzGzu73Wk4Hm4XLEnJdQ8
 Babel==0.9.6
 
-# billiard: tags/v3.3.0.9
-# sha256: JHw2XY75uDAZW1znrmH9ufpEYQJalYhyMNKpcPEmrpM
-https://github.com/celery/billiard/archive/5652fe69e0a0623cea8f9f0c2810e001d4ad53df.tar.gz#egg=billiard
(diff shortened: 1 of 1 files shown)


requirements/default.txt
```diff
@@ -7,9 +7,9 @@ anyjson==0.3.1
 # sha256: SjoIXs8fzSc2VzU4_6EU8fQzGzu73Wk4Hm4XLEnJdQ8
 Babel==0.9.6
 
-# billiard: tags/v2.7.3.34
-# sha256: YKE9uzfN1UL1JUZXDT451i6erJQVPRDdfCT_poTAVy0
-https://github.com/celery/billiard/archive/f24a943cb4b6992717fa8d1db7e220d576377631.tar.gz#egg=billiard==2.7.3.34
```
mythmon: Instead of updating to a different github tag, can we switch this to something from pypi, like `billiard==3.3.0.9`? I'd like to eventually get rid of all the requirements that start with `http://`. Same goes for the rest of the requirements you updated (where possible).

rlr:
I'll fix those dependencies today.

rlr:
fixed :point_up:

mythmon:
Nice. r+

mythmon:
Er, wait. Do we need the 

```python
import djcelery
djcelery.setup_loader()
```

that Input has?

rlr:
We already had that :rabbit:

mythmon:
Oh, good. Carry on then.

PR #2319, opened 2015-01-21
[bug 1123727] Minimal version of KB articles

rehandalal:
r?


kitsune/bundles.py
```diff
@@ -255,6 +261,22 @@
...
+            'js/libs/modernizr-2.6.1.js',
+            'js/browserdetect.js',
+            'js/mobile/ui.js',
+            'js/analytics.js',
+            'js/libs/jquery.cookie.js',
+            'js/libs/jquery.lazyload.js',
+            'js/showfor.js',
+            'js/ajaxform.js',
+            'js/mobile/wiki.js',
+            'js/wiki.metrics.js',
```
mythmon: Woah, that's a lot of JS. Maybe we should try and slim this down some day.

kitsune/wiki/views.py
```diff
@@ -166,6 +166,11 @@ def document(request, document_slug, template=None, document=None):
     else:
         document_css_class = ''
 
+    if request.MOBILE and 'minimal' in request.GET:
```
mythmon: I think this should be available even when the request isn't from mobile. Embeddable versions of SUMO seem generally useful. If we don't do that, we should also make sure that BuddyUp's embedding method is properly marked as mobile.

kitsune/sumo/templates/mobile/minimal.html
```diff
@@ -0,0 +1,85 @@
+{% from "includes/common_macros.html" import list_view_item %}
+{% from "includes/common_macros.html" import favicon with context %}
+{% from "wiki/includes/document_macros.html" import show_for with context %}
```
mythmon: This isn't used.

kitsune/sumo/templates/mobile/minimal.html
```diff
@@ -0,0 +1,85 @@
+{% from "includes/common_macros.html" import list_view_item %}
```
mythmon: This isn't used.

kitsune/sumo/templates/mobile/minimal.html
```diff
@@ -0,0 +1,85 @@
...
+{% from "includes/common_macros.html" import favicon with context %}
+{% from "wiki/includes/document_macros.html" import show_for with context %}
+{% from "search/basic-form.html" import mobile_nav_form with context %}
```
mythmon: This isn't used.

kitsune/sumo/templates/mobile/minimal.html
```diff
@@ -0,0 +1,85 @@
...
+<div class="wrapper slide-on-exposed">
+  <section id="content">
+    {% block content %}{% endblock %}
+  </section>
+
+  <footer>
+    {% block footer %}{% endblock %}
+  </footer>
+
+  <ul id="notifications">
```
mythmon: I don't think we should include notifications here.

kitsune/wiki/templates/wiki/mobile/document-minimal.html
```diff
@@ -0,0 +1,54 @@
+{% extends 'mobile/minimal.html' %}
+
+{% from "includes/common_macros.html" import list_view_item %}
+{% from "wiki/includes/document_macros.html" import contributor_list, document_title, document_messages, document_content  %}
+{% from "wiki/mobile/includes/document_macros.html" import vote_form with context %}
+
+{% set title = _('{t} | {c}')|f(t=document.title, c=document.get_category_display()) %}
+{% set headline = title %}
```
mythmon: This isn't used.

kitsune/wiki/templates/wiki/mobile/document-minimal.html
```diff
@@ -0,0 +1,54 @@
...
+
+{% if products.all()[0] and topics.all()[0] %}
+  {% set return_url = url('products.documents', product_slug=products.all()[0].slug, topic_slug=topics.all()[0].slug) %}
```
mythmon: return_url isn't used.

mythmon:
There is a bunch of stuff in these templates that goes unused. I stopped looking for it. I think we should strip these templates down as much as we can, and not include things that don't get used.

kitsune/wiki/views.py
Rik: Just looking at the minimal parameter will also make it easier for us to work on BuddyUp on desktop. We won't have to change our UA ;)

rehandalal:
i stripped a bunch of stuff out. i had started on the desktop version but started breaking things so i pulled back. i can file a follow up to do this for desktop. i don't want to spend a ton of time on desktop for now.

mythmon:
+1 on passing on the desktop version. The suggestion earlier was to simply serve the mobile-minimal template any time the minimal version was requested, regardless of mobile/desktop. We can pass on this for now though, because it is easy to work around.

The new stripped down version looks good. Once this has the frame headers, I think it is ready to go.

rehandalal:
header issue should be resolved by this ^

rehandalal:
i pushed this to dev and Rik verified that it is working.

mythmon:
Awesome! r+

PR #2320, opened 2015-01-21
[bug 1114031] Pass on q= param when filtering.

rlr:
Oops. Pretty sure I F'd this up when removing other useless params from search.

STR from the bug: `When I visit https://support.mozilla.org/en-US/search?q=test and click on "HELP ARTICLES ONLY", it redirects to https://support.mozilla.org/en-US/search?w=1 instead of keeping the part "q=test" and adding "&w=1".`

r?

### DIFF
--- a/kitsune/search/templates/search/results.html
+++ b/kitsune/search/templates/search/results.html
@@ -24,7 +24,7 @@
       {% if not advanced %}
         {# Don't show filtering to advanced search users #}
 
-        {% set base_url = url('search') %}
+        {% set base_url = url('search')|urlparams(q=q) %}
 
         {% if product %}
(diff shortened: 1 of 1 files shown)


mythmon:
looks good to me. r+

rlr:
47199c747ef45681dc9007cfc32e1947fbf74e2c [bug 1114031] Pass on q= param when filtering.

BReduardokramer:
Fixing it could help others. I not fixed all today, could be only 2, 3.
Actually was found kitsunes from here. I fix it for this "bug." Thanks.
Em 26/01/2015 14:32, "ricky rosario" <notifications@github.com> escreveu:

PR #2321, opened 2015-01-21
Add UA of Win10 and Newer mac versions

philipp-sumo:
add UA of win10 and newer mac versions


mythmon:
[This document from Microsoft](https://msdn.microsoft.com/en-us/library/ie/hh869301%28v=vs.85%29.aspx) says that when Windows 10 comes out it's UA will say `Windows NT 10.0`, instead of 6.4. We should support both.

philipp-sumo:
thanks for your diligence, mike!

mythmon:
Great. Next time you make a change like this, you should put the "add UA of win10 and newer mac versions" in the "subject" of the commit, instead of the default value. It's not very clear when the PR and commit subjects are just "Update aaq.js". Just something to keep in mind next time.

philipp-sumo:
Yup, I'm still not that familiar with github, sorry about that. 
If I remember correctly from last time, the corresponding tags to the UA detection would need to be created manually also...

mythmon:
Huh. I don't know about the tagging right now. Do you think this can merge now? Or do we need to figure out that thing about tagging first?

philipp-sumo:
i think this pull request can go ahead now, but to take effect during the AAQ as per https://bugzilla.mozilla.org/show_bug.cgi?id=931274#c7 the 'Windows 10', 'Mac OS X 10.9' and 'Mac OS X 10.10' tags need to be created manually (i don't know how/where to do that - is it in some admin panel?)

PR #2322, opened 2015-01-23
Remove 2.6 stuff

mythmon:
I ran around and got rid of anything that looked like Python 2.6. Python 2.6 is no longer supported.

r?


rehandalal:
lgtm. r+

rehandalal:
also yay!

PR #2323, opened 2015-01-23
[Bug 1083417, 1083423, 1083431] Notifications

mythmon:
This will cover Bugs [1083417](https://bugzilla.mozilla.org/show_bug.cgi?id=1083417), [1083423](https://bugzilla.mozilla.org/show_bug.cgi?id=1083423), and [1083431](https://bugzilla.mozilla.org/show_bug.cgi?id=1083431).

Todo

* [x] Actions backend.
* [x] Send actions from Questions.
* [x] Notifications (read/unread tracking, mostly)
* [x] API to consume notifications.

### DIFF
--- a/kitsune/notifications/README.rst
+++ b/kitsune/notifications/README.rst
@@ -1,0 +1,107 @@
+Notifications
+=============
+
+Notifications in SUMO are an extension of the ideas in
+`django-activity-stream <https://github.com/justquick/django-activity-stream>`_.
+The package name for which is ``actstream``.
+
+There are three primary kinds of objects: ``actstream.models.Action``,
+``actstream.models.Follow`` and ``kitsune.notifications.models.Notification``.
+
+Action
+------
+
+In theory, one Action is created anytime something happens that someone might
+want to be notified about. In practice, not everything will have actions right
+away, but these will be added over time.
+
+Actions have a few important fields:
+
+* ``timestamp`` - When the action happened.
+* ``actor`` - All actions have an actor. This is generally the user that performed
+  an action. Internally this is a ContentTypes GenericForeignKey, so it could
+  point to any object.
+* ``verb`` - This is a string representing what the user did. Things like "asked",
+  "created", or "edited" are appropriate.
+* ``action_object`` (optional) - This is the object that was acted with. If an
+  object was created, this is the created object. This might not always be
+  applicable. Internally this is a ContentTypes GenericForeignKey.
+* ``target`` (optional) - This is the place the action happened. If the object
+  created belongs to a group, this might be the group. Internally this is a
+  ContentTypes GenericForeignKey.
+
+Examples
+~~~~~~~~
+
+If Bob creates a question, an action will be created with these properties::
+
+  action(bob, verb='asked', action_object=the_new_question)
+
+Since questions don't belong to a larger group, there is no Target.
+
+If Sarah comes along and answers Bob's question, an action will be created
+with these properties::
+
+  action(sarah, verb='answered', action_object=the_new_answer, target=the_question)
+
+Since Answers are groups together in relation to a Question, the target should
+be the Question in question.
+
+Follow
+------
+
+A Follow object represents the desire of one user to be alerted about a class
+of Actions. They have three important fields:
+
+* ``user`` - The user that will be notified.
+* ``follow_object`` - The object being followed.
+* ``actor_only`` (default True) - If actions that involve ``follow_object``, but
+  where ``follow_object`` is not ``action.actor`` should notify ``user``.
+
+A user will be alerted about an action if any of these properties are true:
+
+* ``action.actor == follow.follow_object``
+* ``action.action_object == follow.follow_object and not follow.actor_only``
+* ``action.target == follow.follow_object and not follow.actor_only``
+
+Examples
+~~~~~~~~
+
+In the question above, Bob was automatically registered as following the
+Question with an ``actor_only = False`` Follow object. So when Sarah answers
+the question, generating the action
+
+::
+
+  action(sarah, verb='answered', action_object=the_new_answer, target=the_question)
+
+Bob will be notified, because ``target`` matches the ``action_object`` of one
+of his Follow objects.
+
+It will be common for users to follow objects like this. It will be uncommon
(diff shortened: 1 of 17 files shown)


mythmon:
I'm a bit puzzles by those test failures. I didn't really touch the models or migrations. Can anyone take a look at this and see if there is anything obvious I'm missing?

I expect this is actually something wrong with the monkey patching and import order or something.

rlr:
Huh. weird. It looks like it's not respecting that the migration depends on the other one:

```
    depends_on = (
        ('products', '0008_enable_product_forums'),
    )
```

wtf

mythmon:
I don't think that's it. My theory is that someone in early start up is causing the models to load before the migrations run. I don't think it has run any migrations yet. But I don't really know what to do about that.

mythmon:
Oh. On second thought, I think that thing I just said is totally wonky. I bet you are right and it isn't doing the depends_on stuff right.

rehandalal:
I was able to fix the issue by adding:
```
    depends_on = (
        ('questions', '0013_populate_question_locale_products'),
    )
```
to `0009_auto__del_field_product_questions_enabled.py` in the `products` app.

mythmon:
Hey, nice! I'm still confused how I tripped that here, but adding that dependency fixed it for me too. Of course now I have lots of test failures to fix, but I'm unblocked. Thanks both of you!

rehandalal:
Yeah it's really weird, not sure why it suddenly started acting up.

FYI: The tests all passed for me locally though.

mythmon:
I think all of my test failures are from the most recent commit. One small thing that broke so many things.

mythmon:
I think this is ready now.

r?

kitsune/notifications/models.py
```diff
@@ -1,13 +1,77 @@
...
+            content_type=ContentType.objects.get_for_model(instance.action_object).pk,
+            object_id=instance.action_object.pk,
+            actor_only=False)
+
+    # execute the above query, iterate through the results, get every user
+    # assocated with those Follow objects, and fire off a  notification to
+    # every one of them. Use a set to only notify each user once.
+    users_to_notify = set(f.user for f in Follow.objects.filter(query))
+    notifications = [Notification(owner=u, action=instance) for u in users_to_notify]
+    Notification.objects.bulk_create(notifications)
```
rlr: Should this be in a celery task? There seems to be a lot going on here.

kitsune/notifications/models.py
```diff
@@ -1,13 +1,77 @@
...
     created = models.DateTimeField(default=datetime.now)
     push_url = models.CharField(max_length=256)
+
+
+actstream.registry.register(User)
+
+
+@receiver(post_save, sender=Action, dispatch_uid='action_create_notifications')
+def add_action_for_new_question(sender, instance, created, **kwargs):
+    """When an Action is created, notify every user following something in the Action."""
```
rlr: I am still trying to wrap my head around all the users that see the notification. From the test it looks like a user can follow another user and get notifications for everything that user does? Is that right? What can a user follow?

kitsune/notifications/tests/test_api.py
```diff
@@ -38,3 +47,68 @@ def test_cant_register_for_other_users(self):
...
+        act = Action.objects.order_by('-id')[0]
+        self.notification = Notification.objects.get(action=act)
+
+    def test_mark_read(self):
+        eq_(self.notification.is_read, False)
+        self.client.force_authenticate(user=self.notification.owner)
+        req = self.client.post(reverse('notification-mark-read', args=[self.notification.id]))
+        eq_(req.status_code, 204)
+        n = Notification.objects.get(id=self.notification.id)
+        import q
```
rlr: gah!

kitsune/notifications/tests/test_api.py
```diff
@@ -38,3 +47,68 @@ def test_cant_register_for_other_users(self):
...
             'creator': ["Can't register push notifications for another user."],
         })
+
+
+class TestNotificationSerializer(TestCase):
+    def test_correct_fields(self):
+        follower = profile()
+        followed = profile()
+        q = question(creator=followed.user, save=True)
+        follow(follower.user, followed.user)
```
rlr: I don't understand this. Is the notification below getting sent to the follower? Is this just a test case or is sumo becoming twitter with followers and all that?

kitsune/notifications/models.py
mythmon: Until just now I (for some reason) thought that signal handlers were async, and so they wouldn't need celery tasks. I've learned I'm wrong. I'll fix this.

kitsune/notifications/tests/test_api.py
mythmon: Because of that giant signal handler that should be a celery task, creating the action below will create a Notification, because `follower` is following `followed`. The idea is that an `Action` object represents something a user did. But in order to track read/unread and who should be notified, `Notification` objects are created, one for every following user.

This test probably relies on too many things, and I should make the Notification object directly instead of trying to get one from the Action system.

mythmon:
I changed that big signal handler to a celery task, and wrote a bunch of prose about wtf is going on here. Hopefully that is helpful.

kitsune/notifications/README.rst
```diff
@@ -0,0 +1,107 @@
...
+
+Notifications have only a few properties:
+
+* ``action`` - The associated action.
+* ``owner`` - The associated user.
+* ``read_at`` - The datetime the user read the notification (or None).
+* ``is_read`` (calculated field) - Boolean calculated from ``read_at``.
+
+Notifications also don't have any direct alerting properties, but they are used
+by other systems to alert users in some way.
```
rlr: this is awesome! That clarifies everything for me.

rlr:
This is neat. Can it replace tidings long term? Plans to have different ways of notifying user: emails (individual per notification and daily digest), push, on the web site, etc.?

I think this is good to land. If there are issues, they will surface quickly when the app starts integrating...

r+

mythmon:
I expect this to replace tidings long term, yes, and definitely different ways of notifying users. I hadn't thought about daily digest emails, but that's a good idea too.

I'm going to rebase this and clean it up a bit.

PR #2327, opened 2015-01-28
[Bug 1119886] API: Better taken fields and filtering.

mythmon:
This should enable queries like `/api/2/question/?is_taken=0`, and `/api/2/question/?taken_by=mythmon` as well as adding a lot better tests for all of this.

r?


rehandalal:
yay! r+ :sheep:

PR #2329, opened 2015-01-28
[bug 1121778] Password reset API

rehandalal:
r?

### DIFF
--- a/kitsune/users/api.py
+++ b/kitsune/users/api.py
@@ -3,10 +3,13 @@ import re
 from datetime import datetime, timedelta
 from string import letters
 
-from django.contrib.auth.models import User
 from django.conf import settings
+from django.contrib.auth.models import User
+from django.contrib.auth.tokens import default_token_generator
+from django.contrib.sites.models import get_current_site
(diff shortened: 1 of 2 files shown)


kitsune/users/api.py
```diff
@@ -365,3 +369,41 @@ def delete_setting(self, request, user__username=None):
...
+                html_template='users/email/pw_reset.html',
+                context_vars=c,
+                from_email=None,
+                to_email=profile.user.email)
+
+            return mail
+
+        email_utils.send_messages([_make_mail(profile.locale)])
+
+        return Response({'success': True})
```
mythmon: The idiom I've been using for this is `return Response('', status=204)`. 204 means "success, but no body". I think it would be could to be consistent.

rehandalal:
updated ^

mythmon:
This looks good. Can you add a test for it too?

rehandalal:
done ^

mythmon:
:+1:

mythmon:
Though it seems like now there is a merge conflict, which is probably also why Travis isn't running.

PR #2333, opened 2015-01-29
[Bug 1048935] Enable Irish (bonus: Gallego)

vesper-mozilla:
Both locales redirect to FAQ pages that will be localized by contributors.

### DIFF
--- a/kitsune/settings.py
+++ b/kitsune/settings.py
@@ -113,6 +113,7 @@ SUMO_LANGUAGES = (
     'fi',
     'fr',
     'fy-NL',
+    'ga-IE'
     'gl',
     'gu-IN',
     'ha',
@@ -173,8 +174,6 @@ FXOS_LANGUAGES = [
     'en-US',
     'es',
     'fr',
-    'ga-IE',
-    'gl',
     'ha',
     'hi-IN',
     'hr',
--- a/kitsune/settings.py
+++ b/kitsune/settings.py
@@ -113,6 +113,7 @@ SUMO_LANGUAGES = (
     'fi',
     'fr',
     'fy-NL',
+    'gl',
     'gu-IN',
     'ha',
     'he',
@@ -172,6 +173,8 @@ FXOS_LANGUAGES = [
     'en-US',
     'es',
     'fr',
+    'ga-IE',
+    'gl',
     'ha',
     'hi-IN',
     'hr',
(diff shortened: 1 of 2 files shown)


mythmon:
Why are there two commits here with the same exact text?

kitsune/settings.py
```diff
@@ -113,6 +113,8 @@
     'fi',
     'fr',
     'fy-NL',
+    'ga-IE'
```
mythmon: Missing comma on this line.

vesper-mozilla:
>Why are there two commits here with the same exact text?
Because I messed up with the first one, I should have rolled back - apologies for that, I can close this request and open up a new, cleaner one.

> Missing comma on this line.
That's how you can tell a real developer from a code-muncher ;-) I'll fix it.

mythmon:
Instead of closing this PR and making a new one, you can instead fix the problems and force-push to this branch. Something like `git push -f origin l10n-updates`.

mythmon:
Also, I didn't notice the missing comma. Travis (our CI bot) did. That's how you earned the :heavy_multiplication_x: on your commit. :smile:

vesper-mozilla:
Thank you for the *very* useful tip on the command to use.

I got another failure, but I don't think it's my fault this time:

> Traceback (most recent call last):  File "/home/travis/build/mozilla/kitsune/kitsune/sumo/tests/test_googleanalytics.py", line 35, in test_visitors_by_locale    eq_(52, len(visits))  File "/home/travis/virtualenv/python2.7.9/lib/python2.7/site-packages/nose/tools.py", line 31, in eq_    assert a == b, msg or "%r != %r" % (a, b)AssertionError: 52 != 54

mythmon:
(I edited your comment to use a code block instead of a block quote. You can see what I did by clicking the pencil icon in the top right of your comment.)

If you look at [the test that failed](https://github.com/mozilla/kitsune/blob/master/kitsune/sumo/tests/test_googleanalytics.py#L26), it is a bit obtuse but what it is doing is counting the number of locales. It's kind of a funky test that maybe needs cleaned up, but nonetheless, your change did break the test. you added two locales, the number it was expecting to get went up by two. An easy fix.

vesper-mozilla:
<Metallica>Follow the test that failed!</Metallica>

OK, so my change broke a test, which showed up as broken in a different test. I get that now. I'll update the test_googleanalytics.py accordingly and we'll see what Travis says.

FWIW, I didn't have that problem the last time I added a locale.

mythmon:
This looks good to me now, and Travis likes it. r+

safwanrahman:
@mythmon, I think this commit should be squashed!

PR #2335, opened 2015-01-29
[Bug 1124191] Show Twitter Username instead of link in SUMO profile

safwanrahman:
Please merge the Two commits separately. Or it will make conflict!
r?

### DIFF
--- a/kitsune/users/migrations/0009_Change_Twitter_link_to_Twitter_handle.py
+++ b/kitsune/users/migrations/0009_Change_Twitter_link_to_Twitter_handle.py
@@ -1,0 +1,113 @@
+# -*- coding: utf-8 -*-
+import datetime
+from south.db import db
+from south.v2 import DataMigration
+from django.db import models
+
+class Migration(DataMigration):
+
+    def forwards(self, orm):
+        """Remove Twitter.com link and keep Username only"""
+        for profile in orm.Profile.objects.exclude(twitter=None):
+            profile.twitter = profile.twitter.replace("https://www.twitter.com/","")
+            profile.twitter = profile.twitter.replace("https://twitter.com/","")
+            profile.twitter = profile.twitter.replace("http://www.twitter.com/","")
+            profile.twitter = profile.twitter.replace("http://twitter.com/","")
+            profile.twitter = profile.twitter.replace("#!/","")
+            profile.twitter = profile.twitter.replace("@","")
+            profile.twitter = profile.twitter.split('/')[0]
+            profile.twitter = profile.twitter[0:15]
+            profile.save()
+
+
+    def backwards(self, orm):
+        """Add twitter.com/ URL before the the Twitter username"""
+        for profile in orm.Profile.objects.exclude(twitter=None):
+            profile.twitter = "https://www.twitter.com/" + profile.twitter
+            profile.save()
+
+    models = {
+        u'auth.group': {
+            'Meta': {'object_name': 'Group'},
+            u'id': ('django.db.models.fields.AutoField', [], {'primary_key': 'True'}),
+            'name': ('django.db.models.fields.CharField', [], {'unique': 'True', 'max_length': '80'}),
+            'permissions': ('django.db.models.fields.related.ManyToManyField', [], {'to': u"orm['auth.Permission']", 'symmetrical': 'False', 'blank': 'True'})
+        },
+        u'auth.permission': {
+            'Meta': {'ordering': "(u'content_type__app_label', u'content_type__model', u'codename')", 'unique_together': "((u'content_type', u'codename'),)", 'object_name': 'Permission'},
+            'codename': ('django.db.models.fields.CharField', [], {'max_length': '100'}),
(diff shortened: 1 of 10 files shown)


kitsune/users/migrations/0009_Change_Twitter_link_to_Twitter_handle.py
```diff
@@ -0,0 +1,111 @@
+# -*- coding: utf-8 -*-
+import datetime
+from south.db import db
+from south.v2 import DataMigration
+from django.db import models
+
+class Migration(DataMigration):
+
+    def forwards(self, orm):
+        "Remove Twitter.com link and keep Handle only"
```
mythmon: Docstrings should be in triple quotes, like `""""this"""`.

kitsune/users/migrations/0009_Change_Twitter_link_to_Twitter_handle.py
```diff
@@ -0,0 +1,111 @@
...
+            profile.twitter = profile.twitter.replace("https://twitter.com/","")
+            profile.twitter = profile.twitter.replace("http://www.twitter.com/","")
+            profile.twitter = profile.twitter.replace("http://twitter.com/","")
+            profile.twitter = profile.twitter.replace("#!/","")
+            profile.twitter = profile.twitter.replace("@","")
+            profile.twitter = profile.twitter.split('/')[0]
+            profile.save()
+
+    def backwards(self, orm):
+        "Write your backwards methods here."
```
mythmon: This docstring should be removed or updated.

kitsune/users/models.py
```diff
@@ -71,8 +72,8 @@ class Profile(ModelBase, SearchMixin):
                            verbose_name=_lazy(u'Biography'))
     website = models.URLField(max_length=255, null=True, blank=True,
                               verbose_name=_lazy(u'Website'))
-    twitter = models.URLField(max_length=255, null=True, blank=True,
-                              verbose_name=_lazy(u'Twitter URL'))
+    twitter = models.CharField(max_length=15, null=True, blank=True, validators=[TwitterValidator],
+                               verbose_name=_lazy(u'Twitter Handle'))
```
mythmon: This should probably say "Twitter Username", since username is a more common term.

kitsune/users/validators.py
```diff
@@ -0,0 +1,6 @@
+from django.core.validators import RegexValidator
+
+
+TwitterValidator = RegexValidator(r'^[\w]+$',
+                                  message='Please enter correct Twitter Handle',
+                                  code='Invalid name')
```
mythmon: There should be tests for this.

safwanrahman:
@mythmon, The Data Migration Updated. Check if the Migration is Ok! :)

safwanrahman:
@mythmon, Updated with Space! :)

kitsune/users/validators.py
safwanrahman: :( Sorry, Failed to make any Test for this!

mythmon:
You should not add the tests to the `ProfileFormTestCase` class, because you are testing separate things. you should make a new class in the same file.

Also, although this approach of having a set of data, and iterating through it is nice, it makes it hard to tell what tests pass and what tests fail. It would be better to do the more verbose version where each case is its own test function, like below. It's more typing, but in the long run it makes the tests nicer to work with.

```python
class TwitterValidatorTestCase(TestCase):

    def setUp(self):

    def test_valid(self):
        TwitterValidator('a_valid_name')

    @assert_raises(ValidationError)
    def test_too_long(self):
        # Twitter usernames can not be more than 15 characters.
        self.assertRaises(ValidationError, lambda: TwitterValidator('x' * 16))

    # And so forth.
    ...
```

safwanrahman:
Thanks @mythmon for guiding about Test! I hope this time this should pass!

kitsune/users/validators.py
```diff
@@ -0,0 +1,6 @@
+from django.core.validators import RegexValidator
+
+
+TwitterValidator = RegexValidator(r'^[\w]+$',
```
mythmon: I don't think this regex is right, because it doesn't allow numbers or underscores. You should add two tests for those cases and *then* edit this to make the tests pass.

mythmon: Hmm. Actually, I tested it, and I guess `\w` does match digits and underscores. It would still be good to add a test for it.

safwanrahman:
@mythmon, I have Updated the Test which will Test Underscore and Number also! Hope it ready to merge now!
r?

mythmon:
This works, but it doesn't deal with `@` signs on users names like we talked about. That's fine for now, I'd like to land this now. But I think we should fix that next. Can you squash this down? Then I'll merge it.

kitsune/users/validators.py
safwanrahman: With the help of @mythmon, success to make the test! :+1:

safwanrahman:
@mythmon Thanks. Hope to fill another Bug to deal with the "@" signs and work on that.
Sure, we can fix that next. I have squashed the commits into one and I hope this time this can be merged! :)

mythmon:
Awesome, thanks Safwan!

PR #2336, opened 2015-01-30
[Bug 1112921] Allow SessionAuthentication on APIs.

mythmon:
Previously this was disallowed for cross-origin concerns. However, I learned that this also checks CSRF, so I'm not worried about cross-origin anymore.

r?


rlr:
That makes sense regarding csrf. r+

PR #2346, opened 2015-02-04
[Bug 1083434] Send notifications to SimplePush.

mythmon:
I don't really have a good way to test this. I can't run a simple push server, and I don't think anything but a Firefox OS phone can run the client side of this. I could probably code up a simple test instance, but I don't really want to. I wrote tests that assert at least we call the URL that is registered.

r?

### DIFF
--- a/kitsune/notifications/decorators.py
+++ b/kitsune/notifications/decorators.py
@@ -1,0 +1,11 @@
+notification_handlers = set()
+
+
+def notification_handler(fn):
+    """
+    Register a function to be called via Celery for every notification.
+
+    This may be used as a decorator or as a simple function.
+    """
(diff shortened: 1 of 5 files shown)


mythmon:
Docs about Simple Push are here, though they are very light on the server side: https://developer.mozilla.org/en-US/docs/Web/API/Simple_Push_API

rlr:
Test it on stage/prod? This all looks good to me. If it is broken, only celery tasks should be affected so no biggie.

r+

mythmon:
We probably can't even test this until BuddyUp writes some stuff. Chicken and egg, I guess. We can push this to stage and stare at it for a bit to make sure nothing blows up.

PR #2347, opened 2015-02-05
[Bug 1128096] Don't recreate virtualenvs every deploy.

mythmon:
This was really slow with peep since peep doesn't cache downloads.
Switching back to this should save a lot of time when deploying.

### DIFF
--- a/scripts/update/deploy.py
+++ b/scripts/update/deploy.py
@@ -115,9 +115,11 @@ def update_info(ctx):
 @task
 def setup_dependencies(ctx):
     with ctx.lcd(settings.SRC_DIR):
-        # Blow away the virtualenv and start again.
-        ctx.local('rm -rf virtualenv/*')
-        ctx.local('virtualenv-2.7 --no-site-packages virtualenv')
+        # Creating a virtualenv tries to open virtualenv/bin/python for
+        # writing, but because virtualenv is using it, it fails.
+        # So we delete it and let virtualenv create a new one.
+        ctx.local('rm -f virtualenv/bin/python')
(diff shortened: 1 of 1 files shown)


mythmon:
Note: This code is what was there previously. It was changed in 25d7434c6163e152a43eefcf785b6a264e750bbb.

scripts/update/deploy.py
```diff
@@ -115,9 +115,11 @@ def update_info(ctx):
...
 def setup_dependencies(ctx):
     with ctx.lcd(settings.SRC_DIR):
-        # Blow away the virtualenv and start again.
-        ctx.local('rm -rf virtualenv/*')
-        ctx.local('virtualenv-2.7 --no-site-packages virtualenv')
+        # Creating a virtualenv tries to open virtualenv/bin/python for
+        # writing, but because virtualenv is using it, it fails.
+        # So we delete it and let virtualenv create a new one.
+        ctx.local('rm -f virtualenv/bin/python')
+        ctx.local('virtualenv --no-site-packages virtualenv')
```
rlr: A few lines down we are still using `virtualenv-2.7`(`virtualenv-2.7 --relocatable virtualenv`) but here you changed it back to `virtualenv`. Not sure if on purpose or by accident.

mythmon: Oops, that is a mistake. Nice catch, thank you.

mythmon:
Fixed.

rlr:
r+!

mythmon:
I'm not going to merge this right now, because we may have to do some surgery on master. But this should merge before the next time we deploy.

PR #2348, opened 2015-02-05
Correcting the Backwards migration

safwanrahman:
The past backwards migration should also add links to the Users profile who would not add twitter username. But, this should not be happen. The backwards migration should add twitter URL only the field who have added twitter username.
r?

### DIFF
--- a/kitsune/users/migrations/0009_Change_Twitter_link_to_Twitter_handle.py
+++ b/kitsune/users/migrations/0009_Change_Twitter_link_to_Twitter_handle.py
@@ -1,0 +1,115 @@
+# -*- coding: utf-8 -*-
+import datetime
+from south.db import db
+from south.v2 import DataMigration
+from django.db import models
+
+class Migration(DataMigration):
+
+    def forwards(self, orm):
+        """Remove Twitter.com link and keep Username only"""
+        for profile in orm.Profile.objects.exclude(twitter=None):
+            profile.twitter = profile.twitter.replace("https://www.twitter.com/","")
+            profile.twitter = profile.twitter.replace("https://twitter.com/","")
+            profile.twitter = profile.twitter.replace("http://www.twitter.com/","")
+            profile.twitter = profile.twitter.replace("http://twitter.com/","")
+            profile.twitter = profile.twitter.replace("#!/","")
+            profile.twitter = profile.twitter.replace("@","")
+            profile.twitter = profile.twitter.split('/')[0]
+            profile.twitter = profile.twitter[0:15]
+            profile.save()
+
+
+    def backwards(self, orm):
+        """Add twitter.com/ URL before the the Twitter username."""
+        # Backwards is different because this should not add link to
+        # the Null field and the field where there is no entry
+        for profile in orm.Profile.objects.exclude(twitter=None).exclude(twitter=''):
+            profile.twitter = "https://www.twitter.com/" + profile.twitter
+            profile.save()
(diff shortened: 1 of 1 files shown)


mythmon:
You used a different set of filters here than on the forward migration. Why the difference?

safwanrahman:
@mythmon, I set different filter because of following reason
# :: twitter__isnull=True 
is used because we should not add twitter url to the field where the nill
# :: twitter__exact='' 
is used because we should not add twitter URL to the users profile who did not add twitter username to their profile

And its different from forwards because, in forwards not necessary to exclude null fields and empty fields. It just run everywhere!

mythmon:
You might want to work on the formatting of that message. Remember, this is Markdown.

Also, you used `.exclude(twitter=None)` in the forwards migration and `.exclude(twitter_isnull=True)` in the second. How are these different? Python's None and MySQL's null are generally mapped to eachother, so I don't see why you made this change. Consistency is very important, because when someone else needs to understand why this is the way it is, they will ask these same questions. Only then, they won't be able to ask you why you did it this way.

Either way, you should add comments about why these are different to the source code.

safwanrahman:
Ok mythmon, will format the message.
I think using .exclude(twitter=None) is better way. I am doing that. also adding in message
I think using both in python method is better way! Its working as expected!
:: .exclude(twitter=None).exclude(twitter=''):

So doing that!

kitsune/users/migrations/0009_Change_Twitter_link_to_Twitter_handle.py
```diff
@@ -21,8 +21,9 @@ def forwards(self, orm):
 
 
     def backwards(self, orm):
-        """Add twitter.com/ URL before the the Twitter username"""
-        for profile in orm.Profile.objects.exclude(twitter=None):
+        """Add twitter.com/ URL before the the Twitter username."""
+        """Should no add link to the Null field and the field where there is no entry"""
```
mythmon: You have two docstrings here. Either they should be combined, if you meant to have a longer doc string, or you should turn the second one into a comment (like `# Should not add link...`). Docstrings `"""like this"""` should not be used for comments.

safwanrahman:
@mythmon, corrected the things. Now must be ready to merge!

kitsune/users/migrations/0009_Change_Twitter_link_to_Twitter_handle.py
```diff
@@ -21,8 +21,9 @@ def forwards(self, orm):
 
 
     def backwards(self, orm):
-        """Add twitter.com/ URL before the the Twitter username"""
-        for profile in orm.Profile.objects.exclude(twitter=None):
+        """Add twitter.com/ URL before the the Twitter username."""
+        # Backwards is different because this should not add link to the Null field and the field where there is no entry
```
mythmon: This line is too long. The linter won't catch it because the linter doesn't lint migrations, but you should wrap this to be no longer than 99 characters. In particular, it is bad when a line is too long for the GitHub UI.

mythmon:
One tiny formatting change, and then this is ready to merge.

safwanrahman:
@mythmon, Now it seems good. Can you please merge?
:)

mythmon:
There were merge conflicts, since this was based on a rather old master branch. I cherry-picked it onto master and landed it as 4566f9086615a77968da4e373b2b1728da177725

PR #2350, opened 2015-02-05
[Bug 1123420, 1122539] Refactor and improve search suggestion APIs.

mythmon:
This makes the search suggestion do two separate queries for data, which fixes that goofy thing they did to try and get enough of both document types. It also switches to pulling the questions from the DB and using the API serializer. This increases consistency with the rest of the API, and provides some data to BuddyUp that is needed for their UX that isn't in ES.

r?

### DIFF
--- a/kitsune/search/api.py
+++ b/kitsune/search/api.py
@@ -1,5 +1,3 @@
-import itertools
-
 from django.conf import settings
 
 from rest_framework.decorators import api_view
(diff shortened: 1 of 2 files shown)


rlr:
I like the refactor. Tests are happy. r+

kitsune/search/api.py
```diff
@@ -36,60 +35,72 @@ def suggest(request):
...
+
+    return Response({
+        'questions': _question_suggestions(searcher, text, locale, product, max_questions),
+        'documents': _document_suggestions(searcher, text, locale, product, max_documents),
+    })
+
 
-    questions_f = es_utils.F(
+def _question_suggestions(searcher, text, locale, product, max_results):
+    search_filter = es_utils.F(
```
willkg: You might want a:

```
if max_results == 0:
    return []
```

That way if the max is 0, it's (almost) a no-op.

willkg: Bah--I just noticed this got merged already. I can do a PR with this change if you like.

mythmon: That sounds like a good change, go for it! Thanks for looking at this.

willkg:
The "max_* thing could be 0" thing is in a PR: https://github.com/mozilla/kitsune/pull/2352

PR #2351, opened 2015-02-06
[Bug 1130320] Adding Azerbaijani [az]

vesper-mozilla:


rehandalal:
The visitor data by locale needs to be updated in `test_googleanalytics.py`.

You'll want to add a new line here:
https://github.com/vesper-mozilla/kitsune/blob/adding-locales/kitsune/sumo/tests/test_googleanalytics.py#L135

Something like
```python
[u'/az/', u'41'],
```

That will make the tests pass.

rehandalal:
lgtm. merging!

vesper-mozilla:
Awesome, thank you - when is support.mozilla.org/az/ going to work, you reckon?

mythmon:
I deployed this to prod a few hours ago. However, it seems to still be redirecting /az/ to the simple wiki page. I'm not sure what is going on here.

rlr:
That behavior seems correct. The problem is that the simple wiki page isn't localized yet. But the locale has to be activated in order for that to happen. chicken and egg.

mythmon:
Oh, I read the diff wrong. I thought we had removed it from `SIMPLE_WIKI_LANGUAGES`. Ok, everything looks fine to me here too.

vesper-mozilla:
Thank you! Yes, the /az users will see the (as of yet untranslated) FAQ page. Sending the localizers in there nao! ;-)

PR #2352, opened 2015-02-06
[bug 1122539] Handle the 0 case

willkg:
If max_questions or max_documents (or both) are 0, then we should do the
least amount of work and just return [].

r?

### DIFF
--- a/kitsune/search/tests/test_api.py
+++ b/kitsune/search/tests/test_api.py
@@ -74,6 +74,13 @@ class SuggestViewTests(ElasticTestCase):
         req = self.client.get(reverse('search.suggest'), {'q': 'emails'})
         eq_([q['id'] for q in req.data['questions']], [q1.id])
 
+    def test_max_results_0(self):
+        self._make_question()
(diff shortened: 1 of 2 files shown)


kitsune/search/tests/test_api.py
```diff
@@ -74,6 +74,13 @@ def test_it_works(self):
         req = self.client.get(reverse('search.suggest'), {'q': 'emails'})
         eq_([q['id'] for q in req.data['questions']], [q1.id])
 
+    def test_max_results_0(self):
+        q1 = self._make_question()
```
rlr: linter failed in travis because `q1` isn't used.

willkg: My bad. I'll fix that.

willkg:
^^^ Fixes the lint issue.

mythmon:
It seems like you only did this for the questions function, not the documents function?

willkg:
Oops--you're right. I thought it got called twice. I'll do another fix later today.

PR #2354, opened 2015-02-06
[bug 1122539] Handle the 0 case for documents

willkg:
* handle when max_documents is 0 in a way that's almost a no-op
* add testing infrastructure to test that documents show up in search
  results and that the code does the right thing when max_documents=0

r?


mythmon:
Eee, thanks! I was only testing questions because I couldn't remember all the things that needed done to generate documents in tests. Thanks for adding those!

PR #2356, opened 2015-02-09
[Bug 1130415] API for users with the most solutions in the last week.

mythmon:
This turned out to be more efficient than I expected, so there is no need to cache like I expected. I tried to do it all in one query. I'm pretty sure it would be possible in SQL, but I couldn't figure out how to get the ORM to do it. Oh well, this is pretty quick too.

r?

### DIFF
--- a/kitsune/users/api.py
+++ b/kitsune/users/api.py
@@ -373,7 +373,7 @@ class ProfileViewSet(mixins.CreateModelMixin,
         profiles = Profile.objects.filter(user__username__in=username_to_count.keys())
         result = ProfileFKSerializer(instance=profiles, many=True).data
 
-        # Pair up the profiles and the solutino counts.
+        # Pair up the profiles and the solution counts.
         for u in result:
             u['weekly_solutions'] = username_to_count[u['username']]
 
--- a/kitsune/users/api.py
+++ b/kitsune/users/api.py
@@ -8,7 +8,7 @@ from django.contrib.auth.models import User
 from django.contrib.auth.tokens import default_token_generator
 from django.contrib.sites.models import get_current_site
 from django.core.exceptions import ValidationError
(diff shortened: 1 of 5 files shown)


rlr:
travis no likey.

kitsune/users/api.py
```diff
@@ -346,6 +348,39 @@ def generate(self, request, **kwargs):
...
+            )
+
+        # Turn that list into a dictionary from username -> count.
+        username_to_count = {u['creator__username']: u['creator__count'] for u in raw_counts}
+
+        # Get all the profiles mentioned in the above.
+        profiles = Profile.objects.filter(user__username__in=username_to_count.keys())
+        result = ProfileFKSerializer(instance=profiles, many=True).data
+
+        # Pair up the profiles and the solutino counts.
```
rlr: typo solutino

kitsune/users/api.py
```diff
@@ -346,6 +348,39 @@ def generate(self, request, **kwargs):
...
+        # It looks like [{'creator__username': 'bob', 'creator__count': 12}, ...]
+        # This uses ``username`` instead of ``id``, because ``username`` appears
+        # in the output of ``ProfileFKSerializer``, whereas ``id`` does not.
+        raw_counts = (
+            Answer.objects
+            .exclude(solution_for=None)
+            .filter(created__gt=start)
+            .values('creator__username')
+            .annotate(Count('creator'))
+            )
```
rlr: If this ends up being slow, you can use the contribution data. It should be very similar to this but filtering solutions only: https://github.com/mozilla/kitsune/blob/master/kitsune/community/utils.py#L17

mythmon:
Fixed :point_up:

rlr:
r+!

PR #2359, opened 2015-02-10
[bug 1085456] Support Forum metrics email report.

rlr:
I wanted some feedback here before going further.

Instead of hard-coding users and email addresses, I went with two groups. The group of users to track and the group of users to send the report to. Is that cool?

If that's cool, should we create the groups in a migration? I'm pretty sure we've always created new groups in the admin, but I'm game either way.

Does this need a test? It seems like a throwaway script. I'd probably need to refactor it to make it testable. Actually, I can probably poke at the email with django's test infra.

f?

### DIFF
--- a/kitsune/questions/cron.py
+++ b/kitsune/questions/cron.py
@@ -1,15 +1,18 @@
 import logging
+import textwrap
 import time
-from datetime import datetime, timedelta
+from datetime import date, datetime, timedelta
 
 from django.conf import settings
+from django.contrib.auth.models import Group
+from django.core.mail import send_mail
 from django.db import connection, transaction
 
 import cronjobs
 
 from kitsune.questions import config
 from kitsune.questions.models import (
-    Question, QuestionVote, QuestionMappingType, QuestionVisits)
+    Question, QuestionVote, QuestionMappingType, QuestionVisits, Answer)
 from kitsune.questions.tasks import (
     escalate_question, update_question_vote_chunk)
 from kitsune.search.es_utils import ES_EXCEPTIONS, get_documents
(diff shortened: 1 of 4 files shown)


mythmon:
I like the idea of using groups for this instead of having a hard coded list. That makes less work for us later.

Since this requires the groups to exist or else it fails, I think we should use a migration to add them (though they can be populated manually).

If it is easy to test, something like "running the cronjob sends an email" that would be cool, but I don't think it is needed, especially if it needs refactoring.

Looks good!

rlr:
OK, writing a test was worthwhile as I found a typo/bug.

Now this is ready for r?

kitsune/questions/tests/test_cron.py
```diff
@@ -87,3 +91,47 @@ def test_escalate_questions_cron(self, submit_ticket):
...
+        report_user.groups.add(report_group)
+
+        # An unanswered question that should get reported
+        question(created=datetime.now() - timedelta(days=2), save=True)
+
+        # An answered question that should get reported
+        q = question(created=datetime.now() - timedelta(days=2), save=True)
+        answer(question=q, save=True)
+
+        # An question answer by tracked user that should get reported
```
mythmon: Typo

kitsune/questions/tests/test_cron.py
```diff
@@ -87,3 +91,47 @@ def test_escalate_questions_cron(self, submit_ticket):
...
+        q = question(created=datetime.now() - timedelta(days=2), save=True)
+        answer(creator=tracked_user, question=q, save=True)
+
+        # More questions that shouldnt get reported
+        q = question(created=datetime.now() - timedelta(days=3), save=True)
+        answer(creator=tracked_user, question=q, save=True)
+        q = question(created=datetime.now() - timedelta(days=1), save=True)
```
mythmon: Nit: this doesn't need to be assigned to q, it won't be used.

scripts/crontab/crontab.tpl
```diff
@@ -44,6 +44,7 @@ HOME = /tmp
 30 3 * * * root {{ rscripts }} scripts/l10n_completion.py --truncate 30 locale media/uploads/l10n_history.json media/uploads/l10n_summary.json
 30 3 * * * {{ cron }} send_postatus_errors
 30 1 * * * {{ cron }} reindex_users_that_contributed_yesterday
+11 1 * * * {{ cront }} report_employee_answers
```
mythmon: cront?

rlr:
fixord ^^

kitsune/questions/cron.py
```diff
@@ -151,3 +153,56 @@ def escalate_questions():
...
+        question__in=questions,
+        creator__in=tracked_users).values_list('question_id').distinct().count()
+
+    email_subject = 'Support Forum answered report for {date}'.format(date=day_before_yesterday)
+    email_body = """
+    Date: {date}
+    Number of questions asked: {num_questions}
+    Number of questions answered: {num_answered}
+    Number of questions with answer from 'Support Forum Tracked' users: {num_answered_by_tracked}
+    """.format(
```
mythmon: I just realize this is probably going to look really weird in a plain text email, because it will have lots of extra space. Or does `send_email` deal with that?

mythmon:
Besides the white space thing, this looks good. r+ once that is fixed.

PR #2360, opened 2015-02-11
[bug 1128477] Tags API for questions

rehandalal:
f?

### DIFF
--- a/kitsune/questions/api.py
+++ b/kitsune/questions/api.py
@@ -7,12 +7,14 @@ from django.db.models import Q
 from rest_framework import serializers, viewsets, permissions, filters, status
 from rest_framework.decorators import action
 from rest_framework.response import Response
+from taggit.models import Tag
 
 from kitsune.products.api import TopicField
 from kitsune.questions.models import (
     Question, Answer, QuestionMetaData, AlreadyTakenException,
     InvalidUserException, QuestionVote, AnswerVote)
 from kitsune.sumo.api import DateTimeUTCField, OnlyCreatorEdits, GenericAPIException
+from kitsune.tags.utils import add_existing_tag
 from kitsune.users.api import ProfileFKSerializer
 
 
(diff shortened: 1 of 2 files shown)


kitsune/questions/api.py
```diff
@@ -289,6 +302,29 @@ def take(self, request, pk=None):
 
         return Response(status=204)
 
+    @action(methods=['POST'], permission_classes=[permissions.IsAuthenticated])
+    def add_tag(self, request, pk=None):
+        question = self.get_object()
+        tag = request.DATA['tag']
```
mythmon: If you do this, it will give a 50x error if the user doesn't pass a tag. It is better to do something like in `delete_metadata` to check that it is the data, and if not, give a 400 bad request error. 500s are bad because the are completely opaque to users.

kitsune/questions/api.py
```diff
@@ -289,6 +302,29 @@ def take(self, request, pk=None):
...
+
+        try:
+            canonical_name = add_existing_tag(tag, question.tags)
+        except Tag.DoesNotExist:
+            if request.user.has_perm('taggit.add_tag'):
+                question.tags.add(tag)
+                canonical_name = tag
+            else:
+                raise GenericAPIException(403, 'You are not authorized to create new tags.')
```
mythmon: Nice.

kitsune/questions/api.py
```diff
@@ -289,6 +302,29 @@ def take(self, request, pk=None):
...
+
+        return Response(canonical_name)
```
mythmon: It would probably be good to reuse the `QuestionTagSerializer` from above. Maybe something like

```
tag = Tag.objects.get(name=tag_name)
return Response(QuestionTagSerializer(instance=tag).data)
```

mythmon:
This is looking good. I made some comments that are mostly about friendliness of the API, and consistency. I didn't test this though.

rehandalal:
Ready to go, with tests. 

r?

mythmon:
r+, but this seems to have a merge conflict. Want to fix that, and merge this?

PR #2363, opened 2015-02-11
[Bug 1125536] Convert all rate limiting to a standard helper.

mythmon:
This converts all the rate limiting we have to use two related functions `kitsune.sumo.decorators.ratelimit` and `kitsune.sumo.utils.is_ratelimited`. These work largely like their django-ratelimit versions, and in-fact use django-ratelimit themselves, except they are simplified and specialized to our needs. In particular, they respect a permission for bypassing rate limits, so we can have a global switch for this, and they all log when a user is over the rate limit.

The rate limit overages are recorded in `kitsune.journal.models.Record`, which is loosely based off `fjord.journal.models.Record`, which @willkg wrote. I modified it a lot, since I don't think we needed everything Fjord needed.

This isn't done yet

* [x] Make new `ratelimit` decorator.
* [x] Make new `is_ratelimited` function.
* [x] Add "logging".
* [x] Add the permission some how (I think [this](https://github.com/mozilla/fjord/blob/master/fjord/analytics/migrations/0002_add_analyzers_group.py#L10) will be useful).
* [x] Test the permissions
* [x] Test the logging

f?

### DIFF
--- a/kitsune/forums/views.py
+++ b/kitsune/forums/views.py
@@ -7,7 +7,6 @@ from django.shortcuts import get_object_or_404, render
 from django.views.decorators.http import require_POST
 
 from authority.decorators import permission_required_or_403
-from ratelimit.helpers import is_ratelimited
 from statsd import statsd
 
 from kitsune import forums as constants
@@ -20,7 +19,7 @@ from kitsune.forums.forms import (ReplyForm, NewThreadForm, EditThreadForm,
 from kitsune.forums.models import Forum, Thread, Post
 from kitsune.sumo.helpers import urlparams
 from kitsune.sumo.urlresolvers import reverse
-from kitsune.sumo.utils import paginate, user_or_ip
+from kitsune.sumo.utils import paginate, is_ratelimited
 from kitsune.users.models import Setting
 
 log = logging.getLogger('k.forums')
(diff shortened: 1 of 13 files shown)


mythmon:
All the boxes above are checked. I think this is ready to go. r?

rlr:
but that test fails. gah!

kitsune/forums/views.py
```diff
@@ -171,10 +165,7 @@ def reply(request, forum_slug, thread_id):
                 post_preview = reply_
                 post_preview.author_post_count = \
                     reply_.author.post_set.count()
-            elif (_skip_post_ratelimit(request) or
-                  not is_ratelimited(request, increment=True, rate='15/d',
-                                     ip=False,
-                                     keys=user_or_ip('forum-post'))):
+            elif not is_ratelimited(request, 'forum-post', '15/d'):
```
rlr: yay! so much cleaner

kitsune/journal/admin.py
```diff
@@ -0,0 +1,17 @@
+from django.contrib import admin
+
+from kitsune.journal.models import Record
+
+
+class RecordAdmin(admin.ModelAdmin):
```
rlr: Love having this app. Is the admin the only way to look at the data for now? Is that good enough?

mythmon: Yes, the admin is the only way to look at these. I think that is good enough for now.

Do you think the model captures enough information? The Fjord one has a lot more details, but I didn't think we needed them.

mythmon:
Oops, I forgot to git add the migration that creates the permission. That is why the test failed.

kitsune/journal/admin.py
rlr: I think this works. I am thinking that we could put a json string in the msg if we need to capture some sort of structured data. That can then be processed offline.

mythmon:
:fireworks:

rlr:
yay :boom:

PR #2364, opened 2015-02-12
[Bug 1132544] Don't use bulk create for notifications.

mythmon:
I added a failing test, then poked at things until the tests passed. Yay TDD! Luckily, I was able to track this down quickly. I was expecting something harder.

Is there a way to make bulk creating Notification objects an error? I don't think we will ever want that.


rehandalal:
oh makes sense! 

/me waits for travis.

rehandalal:
yay! r+

PR #2365, opened 2015-02-13
Allow filtering notifications by is_read.

mythmon:
Now one can add `?is_read=1` to notification queries to only see unread notifications, which like likely save on bandwidth.

r?


rlr:
LGTM

PR #2366, opened 2015-02-13
Modify ratelimit logging to include anonymous users' ip addresses.

mythmon:
I noticed a lot of the rate limits that get hit have blank usernames in the message. I am guessing this is because they are anonymous users and we are rate limiting based on IP address. This changes the message to include the IP address in that case.

r?


rehandalal:
cool. r+

PR #2367, opened 2015-02-13
[Bug 1132115] Prevent double inserts when using the API to create users.

mythmon:
This is a bit of a hack. It will make moving to DRF 3.0 even more annoying since it is "off the beaten path". However, this behavior is actually more similar (in spirit) to DRF 3.0, so maybe it isn't so bad. It also fixed the bug and unblocks BuddyUp.

r?

### DIFF
--- a/kitsune/users/tests/test_api.py
+++ b/kitsune/users/tests/test_api.py
@@ -1,5 +1,6 @@
 import json
 from datetime import datetime, timedelta
+from random import random
 
 import mock
 from django.contrib.auth.models import User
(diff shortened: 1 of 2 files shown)


kitsune/users/tests/test_api.py
```diff
@@ -327,3 +328,40 @@ def test_request_password_reset(self, get_current):
...
+
+        username = 'kris-{}'.format(random())
+        url = reverse('user-list')
+        res = self.client.post(url, {
+            'username': username,
+            'password': 'testpass',
+            'email': 'kris@example.com'
+        })
+
+        import q
```
rlr: whaaaaaa

willkg: I'm going to write @mythmon a pre-commit hook that looks for this.

willkg: Hrm... I was going to do this, but it turns out to be "hard" without reworking our pre-commit hook to run a directory of hooks so that we can add arbitrary checks in there.

Thoughts on that?

mythmon: I've thought about writing that hook too, but it inevitably ends up harder than I want it to be. I think it would be good to have a directory of hooks system eventually, but that is hard and non-standard. Another idea: I think one of pep8, pyflakes, or flake8 supports some sort of plugin interface, which would probably be a good way to make this. I think this is how mccabe works.

Or I could just do better at not committing this...

mythmon:
:point_up: I force pushed to remove the `import q`.

rehandalal:
yay! works as expected.

rehandalal:
@mythmon can't merge

PR #2369, opened 2015-02-13
[bug 1133093] Select avatar size for user API

rehandalal:
tiny r?

### DIFF
--- a/kitsune/users/api.py
+++ b/kitsune/users/api.py
@@ -189,7 +189,9 @@ class ProfileSerializer(serializers.ModelSerializer):
         ]
 
     def get_avatar_url(self, profile):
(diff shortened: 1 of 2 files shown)


mythmon:
I thought you said this wouldn't work?

Is there an easy way to test this? Does the gravatar url change much, or would the only way to tell be to download the image and look at it? If it is easy, it would be nice to have tests for this so refactoring is easier in the future.

rehandalal:
I originally had issues because I was using `getattr` but context is a dict, so that was silly.

I'll add tests. This changes the `?s={size}` in the URL.

rehandalal:
test included ^

mythmon:
Nice.

PR #2372, opened 2015-02-14
[Bug 1130069] Show "Facebook Username" instead of Facebook URL

safwanrahman:
This will do as following:
* Let user insert their Facebook URL but show only the Facebook username in profile
* Get the username from the URL by using Facebook Graph API
* Validate that a user entered a correct URL by checking with Facebook Graph API
* If a user enter invalid URL, it will check with graph API and dont let user add url that is not valid.

r?

### DIFF
--- a/kitsune/users/views.py
+++ b/kitsune/users/views.py
@@ -1,4 +1,5 @@
 import os
+import requests
 from ast import literal_eval
 from datetime import datetime
 
@@ -326,13 +327,31 @@ def profile(request, template, username):
         raise Http404('No Profile matches the given query.')
 
     groups = user_profile.user.groups.all()
+
+    if user.profile.facebook:
+        # Get Facebook Username only from Facebook profile URL by using Facebook Graph API
+        # More about Facebook Graph API can be get from
+        # here: https://developers.facebook.com/docs/graph-api
+        facebook_graph_url = user.profile.facebook.replace(
+            'www.facebook.com', 'graph.facebook.com').replace(
+            'facebook.com', 'graph.facebook.com').replace(
+            'profile.php?id=', '')
+        facebook_profile_data = requests.get(facebook_graph_url).json()
+
+    try:
+        facebook_profile_data['username']
+        facebook_username = facebook_profile_data['username']
+    except:
+        facebook_username = ''
+
     return render(request, template, {
         'profile': user_profile,
         'groups': groups,
         'num_questions': num_questions(user_profile.user),
         'num_answers': num_answers(user_profile.user),
         'num_solutions': num_solutions(user_profile.user),
-        'num_documents': user_num_documents(user_profile.user)})
+        'num_documents': user_num_documents(user_profile.user),
+        'facebook_username': facebook_username})
 
 
 @login_required
--- a/kitsune/users/views.py
+++ b/kitsune/users/views.py
@@ -1,5 +1,5 @@
 import os
-import requests
+import re
 from ast import literal_eval
 from datetime import datetime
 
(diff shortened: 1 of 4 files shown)


safwanrahman:
@mythmon I can not make understand why its not passing Travis, and making 7 Errors. Can you please check?

kitsune/users/views.py
```diff
@@ -326,13 +327,31 @@ def profile(request, template, username):
...
+        facebook_graph_url = user.profile.facebook.replace(
+            'www.facebook.com', 'graph.facebook.com').replace(
+            'facebook.com', 'graph.facebook.com').replace(
+            'profile.php?id=', '')
+        facebook_profile_data = requests.get(facebook_graph_url).json()
+
+    try:
+        facebook_profile_data['username']
+        facebook_username = facebook_profile_data['username']
+    except:
```
mythmon: Having just `except:` without any guard is really bad. You should change this to something like `except KeyError:`.

mythmon: A better way to do all of this is simply `facebook_username = facebook_profile_data.get('username', '')`, which tries to get the value, and if it values returns a default value (`''` in this case).

kitsune/users/views.py
```diff
@@ -326,13 +327,31 @@ def profile(request, template, username):
...
     groups = user_profile.user.groups.all()
+
+    if user.profile.facebook:
+        # Get Facebook Username only from Facebook profile URL by using Facebook Graph API
+        # More about Facebook Graph API can be get from
+        # here: https://developers.facebook.com/docs/graph-api
+        facebook_graph_url = user.profile.facebook.replace(
+            'www.facebook.com', 'graph.facebook.com').replace(
+            'facebook.com', 'graph.facebook.com').replace(
+            'profile.php?id=', '')
```
mythmon: This string of replace calls is pretty brittle, and you have it duplicated in the Validator as well. Instead you should use a regular expression to extract the information you want from the string, and you should re-use that regular expression between this function and `FacebookValidator`.

mythmon:
I don't think I'm very comfortable making an HTTP request every time this page is loaded. I would much rather we store the username on the model instead, so we can do these http requests and url transformations only once, instead of every time the page loads.

rehandalal:
Also what is the advantage of just showing the Facebook username? Unlike Twitter, Facebook doesn't use the username for anything other than the URL (AFAIK).

safwanrahman:
@rehandalal In the new Design of User profile, Facebook Username will be show in the new UI. And the UI looks pretty well with the Facebook Username rather than URL. This is the advantage of just showing Facebook username instead of link.
Yes, I accept that Facebook Username is not used except URL, but I think its better to show the username rather than the link as it looks good with the UI :)

Latest Forum Mockup made by Derek: https://drive.google.com/file/d/0B6mUQoSvwR91S0pENjlFakVROWs/view
firefox_screenshot_2015-02-16t15-26-23 020z

safwanrahman:
Its Working like below:
* Let user insert their Facebook URL but show only the Facebook username in profile
* Get the username from the URL by checking the pattern.
* Validate that a user entered a correct username by checking the username with strict Validator

@mythmon, I think this time its as expected. :smile:
Yes, we will store the URL but we will show the only username to the user.
And I have added a strict Validator which will not let user add wrong data after all! So the database will be clean. :) :sunglasses:

safwanrahman:
@rehandalal As we discussed in IRC about adding the Facebook Username in the form, It seems that it need to change the UI of "Edit Profile". I hope you will like the latest commit.

mythmon:
This way of doing it will confuse users, as they will enter a url, but see a username. They should simply enter a username, and we can build the urls for them. Letting users enter arbitrary urls is very error prone.

willkg:
This is over a month old. I think we should close it out.

mythmon:
The code here is over two months old. Even when all the problems are fixed, it is really out of date. If you still want to work on this, please fix the problems above, and file a new PR.

PR #2373, opened 2015-02-17
[bug 1085456] Breakout counts for each user in metrics email.

rlr:
r?

### DIFF
--- a/kitsune/questions/cron.py
+++ b/kitsune/questions/cron.py
@@ -186,9 +186,11 @@ def report_employee_answers():
     num_answered = questions.filter(num_answers__gt=0).count()
 
     # Total number of questions answered by user in tracked_group
-    num_answered_by_tracked = Answer.objects.filter(
-        question__in=questions,
-        creator__in=tracked_users).values_list('question_id').distinct().count()
+    num_answered_by_tracked = {}
+    for user in tracked_users:
+        num_answered_by_tracked[user.username] = Answer.objects.filter(
+            question__in=questions,
+            creator=user).values_list('question_id').distinct().count()
 
     email_subject = 'Support Forum answered report for {date}'.format(date=day_before_yesterday)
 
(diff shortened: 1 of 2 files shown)


kitsune/questions/cron.py
```diff
@@ -186,9 +186,11 @@ def report_employee_answers():
...
 
     # Total number of questions answered by user in tracked_group
-    num_answered_by_tracked = Answer.objects.filter(
-        question__in=questions,
-        creator__in=tracked_users).values_list('question_id').distinct().count()
+    num_answered_by_tracked = {}
+    for user in tracked_users:
+        num_answered_by_tracked[user.username] = Answer.objects.filter(
+            question__in=questions,
+            creator=user).values_list('question_id').distinct().count()
```
mythmon: You can get this data in one query, instead of n: https://github.com/rlr/kitsune/blob/master/kitsune/users/api.py#L370-L374

kitsune/questions/cron.py
```diff
@@ -196,13 +198,15 @@ def report_employee_answers():
...
     email_body = email_body_tmpl.format(
         date=day_before_yesterday,
         num_questions=num_questions,
-        num_answered=num_answered,
-        num_by_tracked=num_answered_by_tracked)
+        num_answered=num_answered)
+
+    for username, count in num_answered_by_tracked.items():
+        email_body += 'Number of questions answered by {username}: {count}\n'.format(
+            username=username, count=count)
```
mythmon: Should this become a template? It is fine now, but I worry about the future.

rlr: I was under the impression that this email was a temporary thing that should go away in the future. That's why I tried to keep it self contained. But you never know..

mythmon: Oh, in that case, self contained sounds like a good idea.

kitsune/questions/cron.py
rlr: So... I tried but it's tricky. We want the number of questions that have an answer by a tracked user (not the number of answers itself, since a question could have multiple answers by the user). I tried this:

```
    num_answered_by_tracked = (Question.objects
        .filter(id__in=questions, answers__creator__in=tracked_users)
        .values('creator__username')
        .annotate(Count('creator')))
```

But that returns and groups the question creator, not the tracked user. And I got stuck there.

mythmon: Oh, you're right, this is different. I imagine it is possible, but it probably isn't worth the trouble, since this is a cronjob and not an api call.

PR #2374, opened 2015-02-18
[bug 1133165][bug 1133163][bug 1133162] Treejack snippets for product pages

rlr:
r?


mythmon:
The code looks fine, but the snippets are weird. They are going to ask every person to click this link, and then use a cookie to not ask again. This is only temporary, right?

rlr:
:point_up: updated to be only for en-US and activated with a waffle switch.

mythmon:
+1

PR #2375, opened 2015-02-18
[Bug 1133897] Refactor /api/2/question/?involved= to use a subquery.

mythmon:
This should increase the speed of the query. In my testing, the speed up was about 3x.

r?


rlr:
:ship: r+

PR #2376, opened 2015-02-19
[Bug 1134219] Update South to 1.0.2 and s/migrations/south_migrations/.

mythmon:
This is an easy upgrade. I don't expect anything to fail. However, the Travis tests will be particularly useful here, since they start with an empty DB and do all the migrations.

Changes:

* Upgrade requirements/default.txt to have the new south SHA and version.
* Rename all the files, which is the bulk of the diff.

r?


rlr:
It works! r+

PR #2377, opened 2015-02-19
[bug 1133788] Allow add and delete of Locale instances.

rlr:
This is one option to solve the problem. The other option is using migrations as we have been when we were adding locales ourselves. But I don't think we can expect the l10n manager to provide those for us in a PR.

Is it time to make locale management fully admin/db driven?

r?

### DIFF
--- a/kitsune/wiki/admin.py
+++ b/kitsune/wiki/admin.py
@@ -84,14 +84,5 @@ class LocaleAdmin(admin.ModelAdmin):
     readonly_fields = ('locale',)
     search_fields = ('locale',)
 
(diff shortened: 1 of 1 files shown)


mythmon:
I think it is time to make this fully admin/db driven.

In the meantime, what does this do? If I go in and and a locale code that doesn't exist, what happens? What if I delete one that is in use?

rlr:
It does validate that the locale being created exists in settings.py. If you delete a locale, the locale team will be nuked. Which means the locale leaders and reviewers would lose some of their privileges. The site would keep working though.

mythmon:
Alright. Lets do this.

PR #2378, opened 2015-02-19
[bug 1133788] Make locale field writable.

rlr:
Huh. I lied about the locale field being limited to the values in settings.py. I fixed that. And removed it from readonly.

r?

### DIFF
--- a/kitsune/wiki/models.py
+++ b/kitsune/wiki/models.py
@@ -1101,7 +1101,7 @@ class ImportantDate(ModelBase):
 # the M2M tables, etc.
 class Locale(ModelBase):
     """A localization team."""
-    locale = models.CharField(max_length=7, db_index=True)
+    locale = LocaleField(db_index=True)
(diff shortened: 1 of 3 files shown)


kitsune/wiki/models.py
```diff
@@ -1101,7 +1101,7 @@ class ImportantDate(ModelBase):
 # the M2M tables, etc.
 class Locale(ModelBase):
     """A localization team."""
-    locale = models.CharField(max_length=7, db_index=True)
+    locale = LocaleField(db_index=True)
```
mythmon: The bits in the DB won't change, but I think South still wants a migration for stuff like this. It should make one with `./manage.py schemamigration wiki --auto`.

rehandalal:
lgtm. r+ when Travis blesses this with a smile!

kitsune/wiki/models.py
rehandalal: oh!

PR #2381, opened 2015-02-20
Caching dependencies for Travis

rehandalal:
ignore this pull for now... i'm just trying some stuff...


mythmon:
While you are in here, the pip cache is actually useless now, since peep doesn't use it. Maybe someday we will fix peep to use caches, but until then you can probably get rid of that part.

rehandalal:
^ r?

PR #2382, opened 2015-02-23
Tag autogenerated users using a settings.

mythmon:
This uses the user settings as an arbitrary key value store to record that a user was auto generated. Most of these changes are to get the test to work nicely.

r?


kitsune/users/tests/test_api.py
```diff
@@ -101,6 +101,9 @@ def test_no_duplicate_emails(self):
     def test_users_without_emails_are_active(self):
         del self.data['email']
         serializer = api.ProfileSerializer(data=self.data)
+        import q
```
rehandalal: d'oh

rlr: GAH!

mythmon:
Fixed ^

rehandalal:
r+

PR #2384, opened 2015-02-24
[Bug 1120568] Use rendered content for output of question/answer apis.

mythmon:
This required copy/pasting a couple methods from DRF's `WritableField` class, but at least I didn't have to modify the viewset.

r?

### DIFF
--- a/kitsune/sumo/api.py
+++ b/kitsune/sumo/api.py
@@ -7,6 +7,7 @@ import pytz
 from rest_framework import fields, permissions, relations, serializers
 from rest_framework.exceptions import APIException
 from rest_framework.filters import BaseFilterBackend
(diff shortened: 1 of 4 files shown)


kitsune/sumo/api.py
```diff
@@ -85,6 +86,104 @@ def to_native(self, value):
...
+
+        class FooSerializer(serializers.ModelSerializer):
+            content = SplitSourceField(read_source='content_parsed', write_source='content')
+
+            class Meta:
+                model = Foo
+                fields = ('id', 'content')
+
+    The normal field parameter ``source`` is no longer allowed. Instead use
+    ``read_source`` and ``write_source``.
```
rlr: That's clever cool :+1:

PR #2385, opened 2015-02-24
[bug 1135058] Filter question by solver in API

rehandalal:
r?

### DIFF
--- a/kitsune/questions/api.py
+++ b/kitsune/questions/api.py
@@ -199,7 +199,7 @@ class QuestionFilter(django_filters.FilterSet):
         question_user_solved = (
             Question.objects.filter(solution__creator__username=username).values('id'))
 
-        return queryset.filter(Q(id__in=question_user_solved))
+        return queryset.filter(id__in=question_user_solved)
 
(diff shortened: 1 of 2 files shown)


kitsune/questions/api.py
```diff
@@ -191,6 +195,13 @@ def filter_is_solved(self, queryset, value):
             solved_filter = ~solved_filter
         return queryset.filter(solved_filter)
 
+    def filter_solved_by(self, queryset, username):
+        question_user_solved = (
+            Question.objects.filter(solution__creator__username=username).values('id'))
+
+        solved_by_filter = Q(id__in=question_user_solved)
```
rlr: Is this that same trick @mythmon did to force a subquery instead of a join? Or why didnt you just use the queryset above?

mythmon: This is the same trick. It's weird to do `f = Q(...); return foo.filter(f)`, though.

mythmon:
`res` is a strange variable name, and there is the issue with using Q, otherwise LGTM.

PR #2389, opened 2015-02-25
[bug 1128007] Upgrade jingo to master tip.

rlr:
For reference: https://github.com/jbalogh/jingo/compare/8a6fe63ded02970888c27fd527a3c5f9e3061119...fce09043542e71d972cc97d2ae703740741f80d0

r?

### DIFF
--- a/requirements/default.txt
+++ b/requirements/default.txt
@@ -182,9 +182,9 @@ https://github.com/html5lib/html5lib-python/archive/f5855f3d0ad6d4202e2b5d02626d
 # sha256: OeqMam2fWVwXehYTT8SamQrY04J1jL9GnIZZZi8vUas
 httplib2==0.9
 
-# jingo: tags/v0.6.1
-# sha256: YSPU9TzrM590ox_yUwL-1QBGuOEFuYihlhhLZVZ9Rb8
-https://github.com/jbalogh/jingo/archive/8a6fe63ded02970888c27fd527a3c5f9e3061119.tar.gz#egg=jingo==0.6.1
+# jingo: master
+# sha256: 5Yt-sicAUr_gY1Y-SwW6W_mvvxyBuUe2gzPpoB-3VbQ
+https://github.com/jbalogh/jingo/archive/fce09043542e71d972cc97d2ae703740741f80d0.tar.gz#egg=jingo==0.6.1.1
 
 # jingo-minify: master~6
 # sha256: BnWzKRSSTYBMnjFYlCF28MqhIPeyN5jjSoJyXkYQt6o


requirements/default.txt
```diff
@@ -182,9 +182,9 @@ https://github.com/html5lib/html5lib-python/archive/f5855f3d0ad6d4202e2b5d02626d
 # sha256: OeqMam2fWVwXehYTT8SamQrY04J1jL9GnIZZZi8vUas
 httplib2==0.9
 
-# jingo: tags/v0.6.1
-# sha256: YSPU9TzrM590ox_yUwL-1QBGuOEFuYihlhhLZVZ9Rb8
-https://github.com/jbalogh/jingo/archive/8a6fe63ded02970888c27fd527a3c5f9e3061119.tar.gz#egg=jingo==0.6.1
+# jingo: master
+# sha256: 5Yt-sicAUr_gY1Y-SwW6W_mvvxyBuUe2gzPpoB-3VbQ
+https://github.com/jbalogh/jingo/archive/fce09043542e71d972cc97d2ae703740741f80d0.tar.gz#egg=jingo==master
```
mythmon: I don't think you should have `==master`. You should just take that off and make it `#egg=jingo`.

willkg: Plus you need a version number even if it's a fake one otherwise it won't update later.

rlr:
:point_up: changed to look like fjord (fake version)

mythmon:
@willkg does this look good to you? I'm not familiar with this fake version thing.

willkg:
I'd do 0.6.0.1 so that if jingo ever did do a release, the likelihood that the new released version was less than the one you have in the requirements file is pretty minimal.

@mythmon You were the one who showed me the fake version thing.

willkg:
For some backstory, what's going on is that peep/pip will see the github url and needs to extract package name and version from the url so that it can figure out whether the url resource is different version than the one already installed.

So if the github url points to a released version, then we should use that version number. If the github url points to some arbitrary commit, then I think the easiest thing to do is use the most recent released version parent-wise from that commit and then put a `.1` in the `z` place in a `w.x.y.z` version identifier. Then if we have to later pick a later commit, we change that to a `.2`. Etc.

Pep 440 (pretty sure that's the most recent one--there's been a lot of work in this area) says that two parts or three parts are the way to go. We're going to use four parts which is supported for version comparisons, because the likelihood of jingo doing a four-part version release is like 0.

https://www.python.org/dev/peps/pep-0440/#final-releases

For the record, we screwed this up pretty bad with fjord, but it's not a big deal since we're only using peep for contributor dev environments. I keep meaning to write up a blog post with all the things we've learned about requirements files and how those things affect updates, development, convenience, etc.

mythmon:
I showed you the `egg=foo==x.y.z` thing, but I have only ever used it with true versions (at least, on purpose). That is, when the thing being downloaded was really version x.y.z, but for whatever reason Pip/Peep were unable to determine that, which often happened with tagged commits from GitHub or other tarballs. I haven't ever tested it with downloads that don't have an "intended" version number.

That being said, the `0.6.1.1` idea sounds swell. It should convince Pip/Peep to do what we want, it shouldn't break if Jingo releases a new version, and it gives us a clear way forward if we need to do this again (`0.6.1.2`).

r+

willkg:
Ahhh... got it. I was really puzzled when you asked me something I thought you showed me, but it's all clear now.

PR #2391, opened 2015-02-26
[Bug 1132052] Only suggest solved questions in suggestion API.

mythmon:
r?

### DIFF
--- a/kitsune/search/api.py
+++ b/kitsune/search/api.py
@@ -54,7 +54,8 @@ def _question_suggestions(searcher, text, locale, product, max_results):
         model='questions_question',
         question_is_archived=False,
         question_is_locked=False,
-        question_has_helpful=True)
+        question_has_helpful=True,
+        question_is_solved=True)
 
     if product is not None:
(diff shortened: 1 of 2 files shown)


kitsune/search/api.py
```diff
@@ -54,7 +54,8 @@ def _question_suggestions(searcher, text, locale, product, max_results):
         model='questions_question',
         question_is_archived=False,
         question_is_locked=False,
-        question_has_helpful=True)
+        question_has_helpful=True,
```
rlr: should we remove this has_helpful filter? I'm not sure if it's possible that an a question gets solved but doesn't have any answers marked helpful.

atopal: I see no upside to filter for solved threads that also have helpful votes, so we should get rid of this filter.

mythmon:
The Travis failure is because I forgot to update the tests. Fix coming.

mythmon:
Fixes ^

rlr:
:hot: :rocket:

PR #2393, opened 2015-02-27
[Bug 1136586] Fix validation for users api.

mythmon:
There were three main issues here:

* Lack of tests
* Using an `EmailField` for the email field on the serializer.
* Removing the idea of creating users without emails (and that they don't need validated). No one was using this.

r?


mythmon:
I did not do the good thing of unifying the validation between the views and the API and the models. Mainly because there was already pretty good validation in place in the API.

PR #2398, opened 2015-03-02
Remove "Truncating..." output from l10n_completion.

mythmon:
I get totally unhelpful emails every PST-morning that just say "Truncating...". I'd like to not. There were also lint errors the pre-commit hook wouldn't let me not deal with.

r?


rlr:
yay! less mail ftw. r+

PR #2400, opened 2015-03-06
[bug 1134215] Update django-tidings to master tip for django 1.7.

rlr:
r?


mythmon:
The [changes in Tidings](https://github.com/erikrose/django-tidings/compare/b68cbb1044e7020d3ed02f17dcce39ae53db5414...ea57792f82496a9d78fec3f9b0db74964494b42b) don't look like they will affect us, so this seems safe to me.

r+

PR #2401, opened 2015-03-08
Switch from jingo-minify to django-pipeline

rehandalal:
I've been exploring switching to Bower to handle all our front end dependencies and one of the first issues we run into is that jingo minify needs to land my changes that fix relative url's in CSS files. Since that pull has been opened for a few years and the general maintenance of jingo minify is not the greatest I decided switching to django-pipeline was a better option. The switch was way simpler than I expected.

Some things to note:
- `settings_local.py` will need to be updated on each of the server before we deploy this and then probably again to remove the jingo-minify stuff once we have finished deploying/testing this.
- We are switching from `cleancss` to `cssmin` since I didn't feel like writing a django-pipeline compiler for `cleancss`.

r?


willkg:
Funny this--@mythmon and I were just talking about switching to gulp for everything or something like that.

rehandalal:
/me nods. that could be interesting, but i think that would likely be a much larger change. i think while we figure out what that looks like this is a pretty good change. it already cuts out the double-refresh thing that i find super annoying while developing.

willkg:
OMG. I hate the double-hard-refresh thing.

willkg:
This is failing the Travis thing because it can't find lessc. Just as an eff why eye.

rehandalal:
@willkg it was but i fixed that in my second commit. the qunit tests page was broken but i just fixed it ^

willkg:
Just as an eff why eye, I'm so glad you're doing that. I'm so going to do this for fjord, too.

willkg:
Also, I'm just chatting right now. I can review this on Monday (i.e. 24 hours from now) if other people don't get to it first.

willkg:
I think we convert this PR into a sumodev message board. :+1:

rehandalal:
haha! i was actually just about to do this for fjord!

willkg:
Ditching WTFinder is totally unacceptable. r-

scripts/update/deploy.py
```diff
@@ -48,7 +48,6 @@ def update_assets(ctx):
     with ctx.lcd(settings.SRC_DIR):
         ctx.local("python2.7 manage.py nunjucks_precompile")
         ctx.local("python2.7 manage.py collectstatic --noinput")
-        ctx.local("LANG=en_US.UTF-8 python2.7 manage.py compress_assets")
```
mythmon: If we don't do this at compile time, does it happen at request time? Or does `collectstatic` now do the compress_assets part?

rehandalal: `collectstatic` handles everything

mythmon:
Looking at pipeline, I like it. It seems to have pluggable compilers and minifiers, so that's really cool. It has an es6 compiler that uses Babel, which I'm a big fan of, and in particular Babel supports the jsx compilation I'm going to need to do for my React stuff. This makes me much less gung-ho about using Gulp for all this compilation stuff, since pipeline seems to be able to handle it all.

All these changes look fine, I have a small question about when compilation happens, but besides that, yay! The tests are certainly a problem, but I bet that is fixable.

willkg:
So... I spent some time figuring out what was going on with the tests.

On my machine, kitsune tests take:

* 4m46s: master branch
* 22m: the changes in this PR

22m is crazy. I want to do this for Input, too, because I like jingo-minify like I like ham and cheese sandwiches drenched in used motor oil so I spent some time looking into it.

After some experimentation, I discovered that if `PIPELINE_ENABLED = False`, then the jinja tag triggers compiling of every file for every test. If we change that to `PIPELINE_ENABLED = True`, then this doesn't happen. Code in question is here (we're using 1.4.6):

https://github.com/cyberdelia/django-pipeline/blob/4c678d2fe033ae1dc23f552462cd382b399ee5dc/pipeline/templatetags/pipeline.py#L36

This sort of makes sense. `PIPELINE_ENABLED` defaults to `not DEBUG`. Django runs the tests with `DEBUG = False`. By that logic, seems like we should be running with `PIPELINE_ENABLED = True`. I think we need to explicitly set `PIPELINE_ENABLED = True` in `settings_test.py` because of the way it gets the default value:

https://github.com/cyberdelia/django-pipeline/blob/4c678d2fe033ae1dc23f552462cd382b399ee5dc/pipeline/conf.py#L9

Pretty sure that's the only environment where it could potentially pick up the wrong value.

I didn't know whether this means we have to run collectstatic before a test run or not (we do for Input (which is using jingo-minify)). I removed `static/` and ran the tests and didn't have problems. Ergo, I conclude we don't have to run collectstatic before running the test suite.

Also, on a positive note, the tests with this branch and `PIPELINE_ENABLED = True` run in 4m on my machine--that's 46s faster. So that's nice.

And that's all I have to say about that.

rehandalal:
updated ^

/me crosses finger.

rehandalal:
tests pass in a normal amount of time. thanks for looking into that @willkg!

r?

kitsune/messages/templates/messages/base.html
```diff
@@ -1,7 +1,7 @@
 {% extends "base.html" %}
 {% from "messages/includes/macros.html" import message_tabs %}
-{% set styles = ('jqueryui/jqueryui', 'messages') %}
-{% set scripts = ('messages', 'libs/jqueryui') %}
+{% set styles = ('jqueryui', 'messages') %}
+{% set scripts = ('messages', 'lib-jqueryui') %}
```
mythmon: I think this should be just `jqueryui`?

rehandalal: good spot!

rehandalal:
fixed ^

mythmon:
The changes look good. I'm going to test it out.

mythmon:
This works for me. Lets see how it does on stage. r+

mythmon:
I deployed 5df4075 to stage, and checked that it worked with cache busting. It did, so that's cool. It also cache busted a bunch of our other assets, including a sprite sheet we were manually cache busting. Nice! I think this is ready to go.

willkg:
WOO HOO!!!!111!!!11!! :balloon: :+1: :fireworks:

PR #2402, opened 2015-03-08
Use Bower for front-end dependencies

rehandalal:
We only care about d5b1de6 onwards for this PR.

Notes:
- I put the libraries that it was okay to upgrade or that I could find a matching version for in here. 
- We should make plans to upgrade the others.
- I also want to clean up what we're using (we have soooo much JavaScript).
- I think I'm gonna throw JS Hint in with some appropriate rules and make that part of the Travis & pre-commit linting process.
- We should also probably stop writing non reusable JavaScript as much as possible.

r?


mythmon:
A big +1 to all of this. I especially like the part about JS Hint.

What do you have in mind about "stop writing non reusable JavaScript as much as possible"? It sounds like a good idea, I'm just not sure about the specifics.

kitsune/bundles.py
```diff
@@ -1,150 +1,260 @@
...
+            'bower/moment/moment.js',
+            'bower/jquery/jquery.min.js',
+            'bower/jquery/jquery-migrate.js',
             'js/libs/jquery.cookie.js',
             'js/libs/jquery.placeholder.js',
             'js/templates/macros.js',
             'js/templates/search-results-list.js',
             'js/templates/search-results.js',
-            'js/libs/nunjucks-slim.js',
+            'bower/nunjucks/browsers/nunjucks-slim.js',
```
mythmon: "browsers" here should be "browser".

rehandalal:
@mythmon update the pull. this is the error i was getting with PipelineCachedStorage. It's due to a bad import in the selectivizr tests. I'm gonna see if there's an easy way to ditch the test files from the package.

rehandalal:
think we're gonna have to use something like https://github.com/blittle/bower-installer

rehandalal:
updated ^

mythmon:
More errors :frowning:.

mythmon:
I think you forgot to remove selectivizr from bower.json?

rehandalal:
d'oh. fixed ^

rehandalal:
uh oh! will fix the selenium stuff.

rehandalal:
fixed ^

mythmon:
Wooh! So is this ready to land now? Do you want to test it on stage?

I've been using this in my community hub branch for a while now, so I'm pretty sure the dev side of things works. I'm excited for it to land so I can move forward more with that.

r+

PR #2403, opened 2015-03-09
Updating the Doc by adding new requirments

safwanrahman:
As New requirements libffi and adding "python" before the installation!
r?


mythmon:
Thanks!

PR #2405, opened 2015-03-10
[Bug 1141436] Fix visible html tags

pollti:
See https://bugzilla.mozilla.org/show_bug.cgi?id=1141436 (Share Article template is not showing properly in Mobile View) for further information.


rlr:
Awesome, thank you! :balloon: :tada:

PR #2406, opened 2015-03-10
Bug 1141669 - add collapsible sections to article (en-us/de)

philipp-sumo:


mythmon:
I double checked that the German case was done properly. This looks good. Travis seems to be having trouble right now, so I'm running the tests manually. Once those pass, I'll merge this.

PS: For future PRs, we usually make commit/PR titles like `[Bug 1234] Blah blah blah` instead of `Bug 1234 - Blah blah blah`. Not a big deal, but it is good to be consistent for next time. :)

mythmon:
Tests pass!

PR #2409, opened 2015-03-11
fix bug 1120497 by moving the text

pollti:
See also:
http://pollti.bplaced.net/Screenshot%20%2849%29.png
http://pollti.bplaced.net/Screenshot%20%2850%29.png
https://bugzilla.mozilla.org/show_bug.cgi?id=1120497

### DIFF
--- a/kitsune/sumo/static/less/products.less
+++ b/kitsune/sumo/static/less/products.less
@@ -304,7 +304,7 @@
        display: block;
        height: 79px;
        margin-left: -50px;
-       padding: 10px 20px 8px 73px;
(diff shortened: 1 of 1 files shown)


mythmon:
Looks good to me, thanks!

In the future, it is helpful to make pull requests (and commits) with titles like `[Bug 1120497] Move text in download Thunderbird button`. We have tools that can read that kind of format. Also, you can embed images directly in PRs with a syntax like `german screenshot`

pollti:
Thanks, I'll try to keep it in mind.

PR #2410, opened 2015-03-12
[Bug 619799] Different error message for files exceeding maxsize

pollti:
This should throw another error message if a file is too large.
See [bug 619799](https://bugzilla.mozilla.org/show_bug.cgi?id=619799) for further information.

### DIFF
--- a/kitsune/sumo/static/js/gallery.js
+++ b/kitsune/sumo/static/js/gallery.js
@@ -22,6 +22,7 @@
     CONSTANTS.messages['file'] = {
         'server': CONSTANTS.messages['server'],
         'invalid': gettext('Invalid image. Please select a valid image file.'),
+        'toolarge': gettext('Image too large. Please select a smaller image file.'),
         'cancelled': gettext('Upload cancelled. Please select an image file.'),
         'deleted': gettext('File deleted. Please select an image file.'),
         'del': gettext('Delete this image')};
@@ -59,8 +60,8 @@
      * Summary:
      * init: initialize DOM events, dispatch on draft vs new upload
      * validateForm: prevent form submission if data is not valid
-     * isValidFile: boolean result for each type of file, checks size and
-     *              extenson
+     * isValidFile: boolean result for each type of file, checks extension
+     * isNotTooLarge: boolean result for each type of file, checks size
      * startUpload: validates before upload and initiates progress
      * uploadComplete: once upload is done, dispatch to success or error
      * uploadSuccess: successful upload is shown in a preview
@@ -105,6 +106,10 @@
                     beforeSubmit: function($input) {
                         if (!self.isValidFile($file)) {
                             self.uploadError($file, 'invalid');
+                            return false;
+                        }
+                        if (!self.isNotTooLarge($file)) {
+                            self.uploadError($file, 'toolarge');
                             return false;
                         }
                         return self.startUpload($file);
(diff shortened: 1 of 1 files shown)


kitsune/sumo/static/js/gallery.js
```diff
@@ -203,11 +208,15 @@
             var file = $input[0].files[0],
                 type = $input.attr('name');
             var file_ext = file.name.split(/\./).pop().toLowerCase();
-            return (in_array(file_ext, CONSTANTS.extensions[type]) &&
-                    file.size < CONSTANTS.max_size[type]);
+            return (in_array(file_ext, CONSTANTS.extensions[type]));
+        },
+        isNotTooLarge: function ($input) {
```
mythmon: It is odd that this is `isNotTooLarge`, and you only ever call it like `!isNotTooLarge(file)`. It is like a double negative. It would make more sense to have it be `isTooLarge()`, and call it directly. I'm also not convinced it needs to be a separate function, since it is only a couple lines and isn't re-used anywhere.

kitsune/sumo/static/js/gallery.js
```diff
@@ -203,11 +208,15 @@
...
+            var file = $input[0].files[0],
+                type = $input.attr('name');
+            return (file.size < CONSTANTS.max_size[type]);
         },
         /*
-         * Fired when upload starts, if isValidFile returns true
+         * Fired when upload starts, if isValidFile and isNotTooLarge return true
```
mythmon: This comment shouldn't reference `isNotTooLarge`, because that function is entirely a part of `isValidFile`. Sine `startUpload` doesn't make any direct reference to `isNotTooLarge`, this is kind of confusing.

kitsune/sumo/static/js/gallery.js
```diff
@@ -256,8 +265,8 @@
             self.uploadSuccess($input, iframeJSON, options.filename);
         },
         /*
-         * Fired after upload is complete, if isValidFile is true, and server
-         * returned succes.
+         * Fired after upload is complete, if isValidFile and isNotTooLarge
+         * are true, and server returned succes.
```
mythmon: Same here, `isNotTooLarge` is an implementation details of `isValidFile`, not something that needs documented on this function.

kitsune/sumo/static/js/gallery.js
```diff
@@ -309,7 +318,7 @@
                   .html(message);
         },
         /*
-         * Fired if isValidFile is false or server returned failure.
+         * Fired if isValidFile or isNotTooLarge is false or server returned failure.
```
mythmon: Same here, `isNotTooLarge` is an implementation details of `isValidFile`, not something that needs documented on this function.

mythmon:
Thanks for the patch! This is pretty well done, despite how many comments I left. It is functional, and I just needs a bit better organization.

pollti:
I did the following here:
* reverted the comment changes
* deleted the function isNotTooLarge
* changed the beforeSubmit function to include the following test:
```javascript
if ($file[0].files[0].size >= CONSTANTS.max_size[$file.attr('name')]) {
   self.uploadError($file, 'toolarge');
   return false;
}
```
An alternative way to do this would be to write it the following way…
```javascript
var file2 = $file[0].files[0],
    type = $file.attr('name');
if(file2.size < CONSTANTS.max_size[type]);
```
…but we should find another name for file2 then ;-)

mythmon:
Gah. I'm sorry. I totally misread your original patch. This area of the code is pretty convoluted, so I messed up. I thought that your `isNotTooLarge` function was called from inside `isValidFiles()`, which is why I said it shouldn't be its own function. That's not true, so the way you had it before was better. So sorry.

If you prefer this way, we can keep it. If you'd prefer to go back to what was there before, which I'm pretty sure is actually fine except the function name is awkward, then an easy way to do that would be:

```shell
# Undo the last two changes
$ git reset --hard 3a6335c
# Update the branch. Since this is destructive, you need to force it,
$ git push --force
# now make your changes as normal
```

Again, sorry for my confusion.

mythmon:
Thanks!

PR #2411, opened 2015-03-12
[WIP] New Community Hub Stuff

mythmon:
This is still WIP, but I would like to get some feedback on this and if it is something we want to keep doing on other parts of the site.

Right now this relies on client side JSX compilation, which is bad. I have some work in another branch that integrates this in with django-pipeline and bower and browserify and babel and it is all complicated and unfinished and I'm going to have to write a bunch of docs about it.


kitsune/bundles.py
```diff
@@ -163,9 +169,19 @@
         ),
         'community': (
             'js/libs/jquery-1.10.1.min.js',
+            'js/libs/jquery.placeholder.js',
+            'js/main.js',
```
mythmon: These two shouldn't be here.

kitsune/bundles.py
```diff
@@ -163,9 +169,19 @@
...
             'js/community.js',
             'js/select.js',
         ),
+        'community-new': (
+            'js/libs/jquery-1.10.1.min.js',
+            'js/libs/jquery.placeholder.js',
+            'js/libs/underscore.js',
+            'js/libs/moment-2.8.3.js',
+            'js/libs/react/react.js',
+            'js/libs/react/JSXTransformer.js',
```
mythmon: Most of this shouldn't be here either, since it is duplicated from common.

mythmon:
I merged in @rehandalal's django-pipeline and bower stuff. It makes things nicer and *also* production ready, but it unfortunately means this PR is huge and complicated right now. The commits you should be looking at here are the commits I authored, excluding the merge commit. Everything rehan authored is in one of his PRs, which I hope will land soon. Once they land, I'll rebase this to not be so crazy.

willkg:
I would have cheated by pushing @rehandalal's branch to our master repository, then generating the PR against that.

mythmon:
That's a really good idea. I'm going to try to do that.

mythmon:
This is a mess. At @willkg suggestion, I've made #2413 which is a PR against the bower branch, so it will be a lot easier to look at.

PR #2412, opened 2015-03-13
[Bug 1037895] show for didn't work in article reviews

pollti:
added missing argument for ```show_for()```


mythmon:
Nice catch!

PR #2414, opened 2015-03-14
[Bug 1071625] Remove the broken "Show for…" editing toolbar item

pollti:
I removed the "Show for…" toolbar item and renamed the function ```Marky.allButtonsExceptShowfor()``` to ```Marky.allButtons()```


mythmon:
Thanks!

PR #2418, opened 2015-03-17
[Bug 1135653] Add tags for BuddyUp categories.

mythmon:
@rehandalal r?

### DIFF
--- a/kitsune/tags/south_migrations/0001_add_buddyup_tags.py
+++ b/kitsune/tags/south_migrations/0001_add_buddyup_tags.py
@@ -14,8 +14,9 @@ class Migration(DataMigration):
         ]
 
         for category in categories:
-            name = 'category:{}'.format(category)
-            orm['taggit.tag'].objects.create(name=name, slug=name)
+            orm['taggit.tag'].objects.create(
+                name='category:{}'.format(category),
+                slug='category{}'.format(category))
 
     def backwards(self, orm):
         orm['taggit.tag'].objects.filter(name__startswith='category:').delete()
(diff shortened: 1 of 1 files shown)


kitsune/tags/south_migrations/0001_add_buddyup_tags.py
```diff
@@ -0,0 +1,47 @@
...
+    def forwards(self, orm):
+        categories = [
+            'email', 'clock', 'camera', 'audio', 'performance', 'radios', 'sms', 'calls',
+            'browser', 'storage', 'geolocation', 'keyboard', 'system-updates', 'language',
+            'marketplace', 'other'
+        ]
+
+        for category in categories:
+            name = 'category:{}'.format(category)
+            orm['taggit.tag'].objects.create(name=name, slug=name)
```
rehandalal: so this creates the slug with the colon in it. i don't think that should cause any issues, but it's different than how the slug is when it's created by the API.

API creates:
```python
tag.name = 'category:clock'
tag.slug = 'categoryclock'
```

kitsune/tags/south_migrations/0001_add_buddyup_tags.py
```diff
@@ -0,0 +1,47 @@
...
+
+    def backwards(self, orm):
+        orm['taggit.tag'].objects.filter(name__startswith='category:').delete()
```
rehandalal: If we decide to keep the slug with the colon, we should change this to `slug__startswith` since we'd  never have a slug starting with 'category:' coming from the API (or anywhere else the user can create tags).

kitsune/tags/south_migrations/0001_add_buddyup_tags.py
rehandalal: hm... actually i want to do some digging into whether this works.

rehandalal: sadly this breaks filtering by tag in the questions list view

mythmon:
Removed the colon from slugs ^

PR #2430, opened 2015-03-25
[Bug 1139127,1139128,1143941,1143945] Community top contributors

mythmon:
This covers a lot of bugs, and a lot of functionality. I could split this up into multiple smaller PRs, but they would all have to be in a linear order, and I think it would just be more frustrating in the long run.

Some broad architecture notes

* The new pages are at http://localhost:8000/en-US/community/top-contributors/l10n/new and http://localhost:8000/en-US/community/top-contributors/questions/new They are not linked from anywhere just yet.
* They are written as a React component, driven by a controller, embedded in our normal template.
* Most of the new JavaScript is written in [JSX][] and [ES6][]. Babel (the Node tool) covers transpiling these to ES5 (modern-ish JS). The main ES6 features uses are arrow function `(foo, bar) => foo + bar`, classes, and modules.
* Browserify is used to make the modules work.
* This won't yet work on our lowest support platforms (I'm thinking <IE8). That will come later with some shims and shams and polyfills. This is because Babel and Browserify target ES5 by default.
* Files that export a single React component are named like `ComponentName.jsx`. They use `export default class ComponentName extends React.Component`.
* Files that export multiple things or no things are named with lower case.
* Files that end in `.browserify.js`. Are treated as ES6 and as Browserify entry points for the module system. They and all the files they import will be packaged as a single file, compiled to ES5.

React docs: http://facebook.github.io/react/docs/getting-started.html

Details of the feature in particular:

The main view loads a barebones view which simply renders the React component. It also embeds the needed data into the page via script tags. This avoids and extra round trip. All filtering on the client happens by modifying the querystring parameters, and essentially reloading the page. React is very efficient, and analyzes the before and after of the DOM, and only re-renders the parts that actually changed. It is magic. Both the backend view and the frontend renderer read from the query string, so this is the one source of truth for what data should be shown on the page.

I expect this to be a high touch review. I imagine I will have questions to answer, and that there needs to be more comments in some places. I also expect we will want to change some things about this, or at least do them differently in the future.

[jsx]: http://facebook.github.io/jsx/
[es6]: https://github.com/lukehoban/es6features

r?


mythmon:
More helpful links

* [Moment docs](http://momentjs.com/docs/) - Used for datetime handling.
* [Pikaday docs](https://github.com/dbushell/Pikaday) - The calendar component in the date range picker.
* [The UX mockups](https://drive.google.com/file/d/0B6mUQoSvwR91S0pENjlFakVROWs/view)\
* [The React announcement of class based components](http://facebook.github.io/react/blog/2015/01/27/react-v0.13.0-beta-1.html), used extensively in the PR. (The blog post talks about React 0.13 beta, but this has now been released in 0.13 stable, which we are using).

kitsune/community/api.py
```diff
@@ -0,0 +1,330 @@
...
+from rest_framework.response import Response
+
+from kitsune.questions.models import AnswerMetricsMappingType
+from kitsune.sumo.api import GenericAPIException
+from kitsune.users.models import UserMappingType
+from kitsune.wiki.models import RevisionMetricsMappingType
+
+
+# This should be the higher than the max number of contributors for a
+# section.  There isn't a way to tell ES to just return everything.
```
willkg: Actually, there is with ElasticUtils--you can do `.everything()`:

https://github.com/mozilla/elasticutils/blob/master/elasticutils/__init__.py#L1566

It does an ES request for the count, then does another ES request to get *count* number of things.

I don't know if that helps you or not.

Also, "This should be *the* higher..." should probably be "This should be higher...".

mythmon: Ooh, nice. I stole that comment, and the general approach of getting metric objects, and then getting user objects, from some stuff r1cky wrote for the first iteration of this feature.

kitsune/community/api.py
```diff
@@ -0,0 +1,330 @@
...
+        # everything to be a list. This converts it into a plain single dict.
+        self.filter_values.update(dict(self.request.GET.items()))
+
+        f = F()
+
+        for key, value in self.filter_values.items():
+            method_name = 'filter_' + key
+            if not hasattr(self, method_name):
+                raise GenericAPIException(400, 'Unknown filter field {}'.format(key))
+            filter_method = getattr(self, 'filter_' + key)
```
willkg: This doesn't matter a whole  lot, but it's very slightly better to do this:
```
filter_method = getattr(self, 'filter_' + key, None)
if filter_method is None:
    ...
```

The Python docs say that hasattr just turns around and calls getattr, catches the exception and returns True/False accordingly:

https://docs.python.org/2/library/functions.html#hasattr

So given that you want to do something with the value, it's slightly better to just use getattr.

Having said that, it doesn't meaningfully matter one way or the other. Mostly I mention it in case it's interesting.

kitsune/community/api.py
```diff
@@ -0,0 +1,330 @@
...
+class TopContributorsQuestions(TopContributorsBase):
+
+    def get_default_filters(self):
+        filters = super(TopContributorsQuestions, self).get_default_filters()
+        filters['ordering'] = '-answer_count'
+        return filters
+
+    def get_filters(self):
+        f = super(TopContributorsQuestions, self).get_filters()
+        f &= F(by_asker=False)
```
willkg: Why not do this in `get_default_filters`?

kitsune/community/api.py
```diff
@@ -0,0 +1,330 @@
...
+
+        # For ever user object found, mix in the metrics counts for that user,
+        # and then reshape the data to make more sense to clients.
+        data = []
+        for u in users:
+            d = combined[u['id']]
+            d['user'] = u
+            d['last_contribution_date'] = d['user'].get('last_contribution_date', None)
+            if 'user_id' in d:
+                del d['user_id']
```
willkg: Sometimes I cheat and do this:

```
d.pop('user_id', None)
```

kitsune/community/templates/community/top_contributors_react.html
```diff
@@ -0,0 +1,25 @@
...
+    {% set title = _('Top Contributors - Localization') %}
+{% elif area == 'army-of-awesome' %}
+    {% set title = _('Top Contributors - Army of Awesome') %}
+{% endif %}
+{% set crumbs = [(url('community.home'), _('Community Hub')),
+                 (None, title)] %}
+
+{% block content %}
+    <script type="application/json" name="contributor-data">{{ contributors_json|safe }}</script>
+    <script type="applicatoin/json" name="locale-data">{{ locales_json|safe }}</script>
```
willkg: I thought we avoided inline scripts. Why not stick the data in a `data-` thing on a `div` or something like that?

kitsune/sumo/static/js/main.js
```diff
@@ -28,6 +28,19 @@ window.k = window.k || {};
...
+    var qs = '';
+    _.forEach(obj, function(value, key) {
+      if (value === undefined || value === null) {
+        return;
+      }
+      qs += key + '=' + encodeURIComponent(value);
+      qs += '&';
+    });
+    qs = qs.slice(0, -1);
+    return '?' + qs;
```
willkg: This could probably use a test.

kitsune/community/api.py
```diff
@@ -0,0 +1,330 @@
...
+
+        return {
+            'results': data,
+            'count': full_count,
+            'filters': self.filter_values,
+            'allowed_orderings': [
+                'revision_count',
+                'review_count',
+            ],
+        }
```
willkg: I would write tests for these API views.

kitsune/community/api.py
mythmon: These are probably named poorly. `get_default_filters` is the default values for things users of the API may set. Users can't change this aspect of the API, it is just always true.

kitsune/community/static/js/ContributorsList.jsx
```diff
@@ -0,0 +1,28 @@
+import {Paginator} from './contributors-common.jsx';
```
willkg: Is
```
import {Paginator} from ...;
```
equivalent to:
```
import Paginator from ...;
```
?

willkg:
Seems like there's a blank `kitsune/community/static/js/QuestionsContributors.jsx` and a blank `kitsune/community/static/js/community.js`. Does that sound right?

kitsune/community/templates/community/top_contributors_react.html
mythmon: Inline *Javascript* scripts is a problem with CSP, for injection reasons. Other kinds of script tags, like JSON data, are fine, since nothing automatically executes them. xWe could do data tags, but it seemed better to do script tags because a) we can note the type of the data, and b) it is a lot of data that I don't really want to jam in as an attribute on a tag.

Also, I notice there is a typo here, "applicatoin".

It could work either way though, so I could change it to be data attributes, if you still think it should be that way.

kitsune/community/static/js/ContributorsList.jsx
mythmon: Not quite. It is equivalent to

```js
import _contributors_common from './contributors-common.jsx';
const Paginator = foo.Paginator;
```

kitsune/community/templates/community/top_contributors_react.html
willkg: Ahhh--got it. I didn't realize it was certain types of script tags and not all script tags.

kitsune/community/static/js/ContributorsList.jsx
willkg: Huh. I don't think I understand that at all. Where does foo come from?

mythmon:
`QuestionsContributors.jsx` is my fault, left over from some refactor. It seems `community.js` [has just always been empty though](https://github.com/mozilla/kitsune/blob/3703fae461d574658166c8848c9e809d3073c086/kitsune/community/static/js/community.js).

mythmon:
I wrote some docs about all this jazz.

kitsune/community/static/js/ContributorsList.jsx
mythmon: Oops, I had called it foo originally, but renamed it to be clearer (ironically, making it more confusing). It should actually be:

```js
import _contributors_common from './contributors-common.jsx';
const Paginator = _contributors_common.Paginator;
```

mythmon: More particularly, this is actually using (or rather, being inspired by) object destructing assignment. This is the intermediate, which is equivalent to both the code in the file, and the code in my last comment. This kind of destructing pulls certain names out of an object, and binds them to variables. It can also be used in a bunch of other places.

```js
import _contributors_common from './contributors-common.jsx';
const {Paginator} = _contributors_common;
```

kitsune/bundles.py
```diff
@@ -17,6 +17,15 @@
         ),
         'output_filename': 'community-min.css'
     },
+    'community-new': {
+        'source_filenames': (
+            'bower/fontawesome/css/font-awesome.css',
+            'bower/pikaday/css/pikaday.css',
```
willkg: There's no `bower/` directory, so I think these need to be de-bower-ified.

kitsune/bundles.py
```diff
@@ -282,6 +291,34 @@
...
+    'community-new-questions': {
+        'source_filenames': (
+            # This uses the minified version because it is optimized to leave
+            # out lots of debug stuff, so it is significantly smaller than
+            # just minifying react.js.
+            # TODO: Figure out how to include the full sized version in dev,
+            # because it produces much nicer error messages.
+            'bower/react/react.min.js',
+            # 'bower/react/react.js',
+            'bower/pikaday/pikaday.js',
```
willkg: There's no `bower/` directory, so I think these need to be de-bower-ified.

kitsune/bundles.py
```diff
@@ -282,6 +291,34 @@
...
+    'community-new-l10n': {
+        'source_filenames': (
+            # This uses the minified version because it is optimized to leave
+            # out lots of debug stuff, so it is significantly smaller than
+            # just minifying react.js.
+            # TODO: Figure out how to include the full sized version in dev,
+            # because it produces much nicer error messages.
+            'bower/react/react.min.js',
+            # 'bower/react/react.js',
+            'bower/pikaday/pikaday.js',
```
willkg: There's no `bower/` directory, so I think these need to be de-bower-ified.

willkg:
The title in the titlebar for http://localhost:8000/en-US/community/top-contributors/l10n/new doesn't match the title on the page:

http://bluesock.org/~willkg/images/link/1427390887.png

willkg:
Using the filters feels kind of stuttery, but it's possible that's just when it's running on my machine.

Everything else seems to work fine as far as I can tell.

I can't really speak to the React code--I have no idea what that's supposed to look like or how it works, so it's hard to review it.

Mmm... I think I've covered everything I can think of here.

docs/frontend.rst
```diff
@@ -0,0 +1,110 @@
...
+the extensions assigned to those files don't matter to Django Pipeline, and
+should be named to be clear to the reader.
+
+Browserify has been configured with the babelify and bowerify transformers, to
+be able to load ES6 files and files from Bower.
+
+.. _ES6 modules syntax: https://github.com/lukehoban/es6features#modules
+
+
+Bower
```
rehandalal: thanks for adding this. i totally dropped the ball on adding bower docs!

docs/hacking_howto.rst
```diff
@@ -197,31 +197,36 @@ For more information on ``peep``, refer to the
...
+Kitsune gets libraries and dependencies for client side code from Bower. Bower
+is installed as a part of the NPM packages in the last step. To install these
+front-end dependencies run::
 
-    If you see a "npm ERR! notarget No valid targets found." error while installing the Node
-    packages, this is due to npm-lockdown being unable to find a package that matches the hash in
-    ``lockdown.json``.
+   $ ./node_modules/.bin/bower install
+
+This will download dependencies into ``static/bower``.
```
rehandalal: ``bower_components``

kitsune/bundles.py
rehandalal: +1

kitsune/community/api.py
mythmon: I did this for one place, but I think that we still need to use this trick for facets. Let me know if there is something clever  I'm missing there too.

kitsune/bundles.py
mythmon: I think there used to be a `static/bower/`, before a bunch of rebases. That being said, I'm puzzled why this worked for me. *shrug*

rehandalal: there used to be, i killed it in favor of using the default `bower_components` directory and just adding that to `STATICFILES_DIRS`. it's probably working for you because you probably ran `bower install` before rebasing it installed there.

mythmon:
I added tests and fixed some of the issues brought up in review. ^

kitsune/community/api.py
```diff
@@ -0,0 +1,334 @@
...
+            contributor['rank'] = page_start + i
+
+        return {
+            'results': data,
+            'count': full_count,
+            'filters': self.query_values,
+            'allowed_orderings': [
+                'answer_count',
+                'solution_count',
+                'helpful_vote_count',
```
mythmon: @willkg I know that I am going to be asked to be able to sort this data by `last_contribution_date`, but I don't really see a way to do that without getting every single user object instead of paginating just the IDs. This sounds quite slow to me. Do you have any clever ideas?

willkg: So, I'm not entirely sure I understand the question. Why can't you do another ES search ordered by `last_contribution_date` and just retrieve that page of ids?

Generally, SUMO bugs tend to be under-specified, so I'd treat everything as a "first pass", land it and then create new bugs for optimization work or changes in requirements. The theory being that they need to see a prototype to help them figure out what it is they need and spending the least amount of time on the prototype gets it out there faster and "wastes" less time.

Going on that theory, I'd just leave it as is for now and push off additional work to new bugs.

willkg:
Tests look ok to me.

kitsune/community/api.py
mythmon: Huh. What you said here actually makes a lot of sense, so I can imagine being able to do this query without it being very inefficient.  I'm not sure what my problem was earlier. Yay for more sets of eyes on a problem.

That being said, I think the idea of landing this and then following up is a good one. I'll do that.

mythmon:
@rehandalal Any opinions on the front end of this? If not, I'm inclined to land this so I can get unblocked.

rehandalal:
@mythmon looks fine to me. I don't see anything glaringly obvious. Took me a bit to wrap my head around some of it.

mythmon:
@rehandalal I'd be really interested in hearing about what parts were hard to understand. There is a lot of code here, so that is understandable, but I'm curious if we can improve any of it.

I'm going to merge this as is, but I'd love to get feedback on it to improve this.

PR #2431, opened 2015-03-25
[bug 1147569] Fix paths to jQuery UI JS files.

mythmon:
When updating the paths to match bower dependencies, these got messed up. It broke anywhere we used jQuery UI. This should fix it.

NB: When trying to test this, the bad state can be a little sticky. I ended up

```shell
$ rm -r static
$ git checkout -- static
$ bower install
```

Which helped unsticky the bad state.

r?


rehandalal:
gah!

PR #2432, opened 2015-03-26
Selenium update

willkg:
1. update selenium to 2.45 which works with the current version of Firefox; without this update, selenium opens Firefox and then it sits there for 30s and eventually times out causing the test to SKIP and since there are a few of these, that's like 1m30s or so right htere
2. reworks the qunit test to be faster; after fixing selenium, then this test takes 1m+ to run and with these changes it takes like 8s now

r?

### DIFF
--- a/kitsune/sumo/tests/test_qunit.py
+++ b/kitsune/sumo/tests/test_qunit.py
@@ -1,92 +1,61 @@
-import re
-from urllib2 import URLError
-
 from kitsune.sumo.tests import SeleniumTestCase
 
 from selenium.webdriver.common.by import By
 from selenium.webdriver.support import expected_conditions as EC
 from selenium.webdriver.support.ui import WebDriverWait
 
-
-# tl;dr: translate qunit test results into nose test results.
-#
-# Nose supports tests generators. These are functions that when run
-# yield a series of tuples. The first member of each tuple is a function
-# to call, and the rest are the arguments to pass to it. In this way you
-# can have several related tests that can run, pass, and fail
-# seperately. This is really cool.
-#
-# Sadly, they do not work if you are a subclass of TestCase. This is not
-# cool. In particular, the below code relies on the setUp, setUpClass,
-# and tearDownClass methods of SeleniumTestCase.
-#
-# In order to combine these two concepts, a dummy class is created which
-# doesn't run any tests. Then a Nose test function runs the appropriate
-# setup and teardown methods on the dummy test case, and yields a test
-# for every test that Qunit lists.
+from nose.tools import eq_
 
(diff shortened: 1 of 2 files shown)


willkg:
The one bad thing about the way I sped up the qunit tests is that you no longer see all the failed tests--you only see the first failed test in each test group. That sucks a little bit, but I think the increased speed is worth it.

willkg:
Actually, now that I'm really thinking about the qunit test, I think it's written wrong. Pretty sure nose knows it's a generator function so it executes it *at collection time* which means a browser is started, the page is hit and then the test elements are collected *at collection time*, but the tests don't execute until much later. Plus cleanup is "a test".

I think we should nix the test generation stuff, have it be one test and just have it spit out errors for all the failed tests and end on a "did they all pass" assertion.

Thoughts on that?

willkg:
I slept a couple of hours and now I'm pretty sure this test shouldn't be written this way. It's got two problems:

1. creating the browser happens at collection time, but the tests that look at the browser contents execute at test time
2. cleaning up the browser happens as a test, but this breaks the requirement that tests shouldn't have to run in a specific order

I'm going to squash it into one test.

willkg:
^^^ Rewrites the qunit test completely squashing it into one nose test.

willkg:
^^^ Last round of minor tweaks so it prints some helpful things to reduce the number of times a developer might have to run it.

mythmon:
This seems like a more robust way to test qunit, and it still prints out every test failure, which was the primary reason for all the hoop-jumping.

I tested this by breaking a couple qunit tests, and seeing what happened. The output contained the expected failures. Though the output for errors is "None != None" which isn't useful, but is no worse than before, so that is ok.

The thing that is concerning to me is that in the case of a failure a lot of Selenium debug level logs get printed out, and as far as I can tell they are useless in this case. At first I tried looking through the output (which appears next to the test failure) to "see output for detail". Instead I have to scroll waaay up past two copies of the debug output (thanks, Nose), to where the test actually ran.

For reference, here is a copy of my output: https://pastebin.mozilla.org/8827272

Overall, this works. The selenium debug output makes it hard to work with though. Is it possible to make that not happen?

willkg:
Yeah... I'm not sure what to do with all the selenium output. I have similar problems with elasticsearch output. It happened before, too.

I think the easy solution is don't write failing qunit tests?

mythmon:
Ha. If only it were so easy. If you don't have any clever ideas, I say we merge this as is. If nothing else, it means we are actually running qunit tests now, which we weren't in the past, apparently.

Thanks for taking care of this!

PR #2433, opened 2015-03-26
[Bug 1145671] Use ES 1.2.4 in Travis (as well as 0.90.10)

mythmon:
Lets see if it is really just easy.


scripts/travis/setup.sh
```diff
@@ -48,9 +48,15 @@ echo "Updating product details"
 python manage.py update_product_details
 
 echo "Starting ElasticSearch"
-pushd elasticsearch-0.90.10
-  # This will daemonize
-  ./bin/elasticsearch
+pushd elasticsearch-${ELASTICSEARCH_VERSION}
+  # New version of ES are foreground by default, old ones are backgrounded by default
```
willkg: I <3 this fun change in ES.

willkg:
Travis passed this. Looks good to me. r+

scripts/travis/setup.sh
mythmon: I think it is a good change to make, it is kind of surprising when you run something big like ES and it just does... nothing (apparently). Plus, running in foreground by default helps with things like init systems and blah blah blah.

I just hate straddling the change like this.

mythmon:
Thanks!

PR #2435, opened 2015-03-27
[bug 1148082] Fix --no-skip

willkg:
nose has a skip plugin which enhances SkipTest raising beyond our
wildest imaginations. However, there are certain occasions where we want
skipped things to not skip with reckless abandon but instead don the
malevolent cloak of errorhood.

nose has a --no-skip argument for exactly this sort of sordid thing, but
it's totally busted and doesn't work.

This changes the code around so that it works correctly.

Skip the tests or don't, but please don't skip this review! r?

### DIFF
--- a/docs/tests.rst
+++ b/docs/tests.rst
@@ -48,6 +48,9 @@ Some other helpful flags are:
 ``--pdb-fail``:
   Drop into PDB on a test failure. This usually drops you right at the
   assertion.
+``--no-skip``:
+  All SkipTests show up as errors. This is handy when things shouldn't be
+  skipping silently with reckless abandon.
 
 
 Running a Subset of Tests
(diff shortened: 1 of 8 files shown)


mythmon:
Awesome! :boom: Adding this flag to my local tests makes everything angry since Selenium isn't working right, and Redis isn't running, so I have lots of skip tests locally.

To fix the bug, can you also get Travis to use this new feature? I'm not picky if this is done with an environment variable or with the command line flag. There are a couple `raise SkipTest('Not Implemented')` that have been around forever. Lets just nix those tests (unless there is a compelling reason to fix them, that is). Alternatively, we can land this and I can mess with Travis later.

willkg:
I think I have a bunch of tests that get skipped because I'm not running Redis locally with the right setup. I can look into it, though. I'll probably do that Monday.

mythmon:
If you like, we can land this part of it, and deal with removing all the Skip tests and fixing Travis in another PR. In that case, I'll add the bug to the next sprint to work on.

willkg:
I think that plan makes sense. I thought I'd get to this today, but I've almost run out of "today".

PR #2436, opened 2015-03-28
[bug 1146577] fix typo in variable

philipp-sumo:


mythmon:
That looks right to me! Can you reword this commit and force push? It is best to describe the change you are making in the commit title. Otherwise when someone looks at the git history later, this will be a very confusing bug.

philipp-sumo:
sorry for that - i've amended the commit message

mythmon:
Better, thank you.

PR #2438, opened 2015-03-30
Upgrade django-pipeline to 1.4.7

rehandalal:
r?


mythmon:
This changes the setting names for the 6to5/babel ES6 compiler. I use those in my react branch. Can we wait until that lands and then update this PR?

rehandalal:
yep, no rush on this.

mythmon:
It turns out the react branch bypasses the 6to5/babel stuff by just using Browserify. So this was always clear to land, sorry about that.

r+

PR #2439, opened 2015-03-30
Ditch vendors.less and use Autoprefixer

rehandalal:
r?


kitsune/settings.py
```diff
@@ -662,6 +662,7 @@ def JINJA_CONFIG():
 PIPELINE_CSSMIN_BINARY = path('node_modules/.bin/cssmin')
 
 PIPELINE_LESS_BINARY = path('node_modules/.bin/lessc')
+PIPELINE_LESS_ARGUMENTS = '--autoprefix="> 1%, last 2 versions, ff <= 36"'
```
rehandalal: This probably needs to be tweaked based on what browsers we need to support.

mythmon:
When I set `DEBUG=False` this gives me an unordained "there was a server problem". I'm not sure what is up there.

kitsune/settings.py
mythmon: I tested this with browserslist (the tool autoprefixer uses to parse this). This gets us IE8 and above, Chrome 37, 39, 40, and 41, all the versions of Firefox, and some other stuff. Do we need to support IE6 and IE7? Does vendor.less support them?

I think we should do `ff >= 4` instead of `ff <= 36`. That way we won't have to update it in the future. That also drops support for Firefox 3.5 and 3.6, which we don't support in other parts of the site either (I'm getting this data from ShowFor). I'd also be fine with `ff > 1` (all versions of Firefox that this tool supports), if you'd prefer to keep the older browsers too.

mythmon:
Scratch that. I'm not sure what it was, but it is fixed now. I have a question about what prefixes we should use, but besides that, looks good!

mythmon:
r+ with the `ff > 1` change.

PR #2440, opened 2015-03-30
[Bug 1133326] Cleanup survey code, and add Firefox refresh survey

mythmon:
To test: You could set up [all the rigmarole to trigger a refresh on localhost](http://kitsune.readthedocs.org/en/latest/notes.html?highlight=https#about-support-api), which is quite annoying, or you could just set the cookie manually. That's what I did.

r?

### DIFF
--- a/kitsune/sumo/south_migrations/0004_add_refresh_survey_flag.py
+++ b/kitsune/sumo/south_migrations/0004_add_refresh_survey_flag.py
@@ -1,0 +1,98 @@
+# -*- coding: utf-8 -*-
+from south.utils import datetime_utils as datetime
+from south.db import db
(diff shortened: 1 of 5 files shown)


mythmon:
After talking with Kadir, I'm going to rework this to be controlled by a waffle flag so we can throttle it easily.

mythmon:
Reworked to use a waffle flag :point_up: r?

rlr:
travis not happy

rlr:
your changes look good to me :checkered_flag:

mythmon:
Oops, silly JS error.

Also, apparently JS syntax errors will stop the CI build now, so that's cool.

PR #2444, opened 2015-03-31
[bug 1121588] Update django-taggit for django 1.7

rlr:
Latest version on pypi.

r?

### DIFF
--- a/requirements/default.txt
+++ b/requirements/default.txt
@@ -144,9 +144,8 @@ https://github.com/zyegfryed/django-statici18n/archive/6bcc799b800a7237a37e178dd
 # sha256: LTn7BrZL1DKwIiETzYeOHlosfAnsk7K7V3zw4Kq_GNk
 https://github.com/andymckay/django-statsd/archive/bb601976056d0f922c5a613d2a9786b81d179b06.tar.gz#egg=django-statsd
 
-# django-taggit: tags/0.11.2^0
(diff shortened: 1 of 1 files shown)


mythmon:
Can we start putting highlights from the changelog, or at least links to them in these upgrade PRs?

rlr:
The commits:
https://github.com/alex/django-taggit/compare/0.11.2...0.12.3

Changelog:

0.12.3 (03.03.2015)

        Specify that the internal type of the TaggitManager is a ManyToManyField

0.12.2 (21.09.2014)

        Fixed 1.7 migrations.

0.12.1 (10.08.2014)

        Final (hopefully) fixes for the upcoming Django 1.7 release.
        Added Japanese translation.

0.12.0 (20.04.2014)

        Backwards incompatible: Support for Django 1.7 migrations. South users have to set SOUTH_MIGRATION_MODULES to use taggit.south_migrations for taggit.
        Backwards incompatible: Django’s new transaction handling is used on Django 1.6 and newer.
        Backwards incompatible: Tag.save got changed to opportunistically try to save the tag and if that fails fall back to selecting existing similar tags and retry – if that fails too an IntegrityError is raised by the database, your app will have to handle that.
        Added Italian and Esperanto translations.

mythmon:
Awesome, thanks. That makes it a lot easier to know what to look out for.

PR #2445, opened 2015-03-31
[bug 1149615] Remove all user.get_profile() calls.

rlr:
It was removed in django 1.7.
Call `Profile.objects.get(user=user)` instead.

r?


mythmon:
:\ This didn't make things prettier. Maybe some day we should fix things so we don't have to use profile objects. :hourglass_flowing_sand: Some day.

PR #2447, opened 2015-04-01
[bug 1038321] Fix l10n issues in messages/views.py

willkg:
This wraps two strings in gettext and fixes a pluralization issue.

r?


willkg:
As a side note, when fixing this I discovered there's only one other use of ngettext. That really surprises me. I would have expected more since it's how you localize messages that involve plurality.

rlr:
LGTM!

PR #2448, opened 2015-04-01
Move code from messages/ __init__.py to utils.py

rlr:
Django 1.7 is a little picky about using/importing models before the app is ready. Plus this is cleaner.

r?


rlr:
oops. travis fail. looking into it

rlr:
Fixed the mocking :point_up: 

Travis should be happy now.

PR #2449, opened 2015-04-01
Make BrowserifyCompiler account for cache-busted files.

mythmon:
This will allow Browserify stuff (and so React stuff) to work in prod and stage.

r?


rlr:
LGTM.

PR #2450, opened 2015-04-02
[bug 1091905, 1150243] apply fix for bug 1091905 only to headings without subtopics

philipp-sumo:
hackish solution with :nth-child - if there's a better way to do so, feel free to change this.

### DIFF
--- a/kitsune/sumo/static/less/products.less
+++ b/kitsune/sumo/static/less/products.less
@@ -126,14 +126,18 @@
   margin-bottom: 0;
 
   &.has-image {
+    margin-top: 0px;
     margin-left: 88px;
-    display: table-cell;
-    height: 96px;
-    vertical-align: middle;
   }
 
   &.cf {
     margin-bottom: 20px;
+    
+    &:nth-child(2){
+      display: table-cell;
+      height: 96px;
+      vertical-align: middle;
+    }
   }
 
   & + .topic-title {


mythmon:
This looks really hacking, especially `display: table-cell`. @rehandalal, any better ideas here?

kitsune/sumo/static/less/products.less
```diff
@@ -126,14 +126,18 @@
...
     margin-left: 88px;
-    display: table-cell;
-    height: 96px;
-    vertical-align: middle;
   }
 
   &.cf {
     margin-bottom: 20px;
+    
+    &:nth-child(2){
```
rehandalal: nit: should be a space between `)` and `{`

kitsune/sumo/static/less/products.less
```diff
@@ -126,14 +126,18 @@
   margin-bottom: 0;
 
   &.has-image {
+    margin-top: 0px;
```
rehandalal: nit: alphabetize

rehandalal:
this could probably be done using flexbox instead of `display: table-cell;` but I haven't looked at the actual bug, so I don't know what we're fixing here.

I'll look at the bug and let you know if I can think of a better way of doing this.

philipp-sumo:
rehan, is the information in the bug reports enough to understand the problem?

kitsune/sumo/static/less/products.less
rehandalal: you can get rid of `vertical-align` and use:
```css
display: flex;
height: 96px;
flex-direction: column;
justify-content: center;
```

rehandalal:
closed in favor of #2507

PR #2451, opened 2015-04-02
[Bug 1150556] Fix browserify path.

mythmon:
thanks @rehandalal for spotting this!

r?


rehandalal:
i'm throwing caution to the wind and not waiting for travis!

PR #2452, opened 2015-04-02
[Bug 1143944] Community hub product filter

mythmon:
66% tests by weight.

r?


rlr:
I dont have all the latest data or a full index, but this looks good and the product filter shows up. What could go wrong?

PR #2454, opened 2015-04-03
[Bug 1131143] Add cohort analysis script.

mythmon:
The logic seems right to me, but I can't really tell if the numbers are right, because nothing else tells us this info. Can someone check that this approach seems right?

r?


rlr:
Only on my first coffee today, but this looks right at first glance. Not sure what's up with travis.

mythmon:
Travis had a hiccup with npm, I guess? Its fixed now.

rlr:
Is this done? What is the plan for running it?

mythmon:
Yeah, this is done. The plan for running it is that someone like like Kadir will clone kitsune and get a database set up and run the script. if this looks like it is measuring what it should be measuring, we're done here.

PR #2455, opened 2015-04-03
Add timeout to collect_tweets, and Update Twython to 3.2.0

mythmon:
I can't test this locally. I can't test it on stage.

@willkg r?

### DIFF
--- a/kitsune/customercare/cron.py
+++ b/kitsune/customercare/cron.py
@@ -11,11 +11,11 @@ from django.db.utils import IntegrityError
 import cronjobs
 from multidb.pinning import pin_this_thread
 from statsd import statsd
-from twython import Twython
 
 from kitsune.customercare.models import Tweet, TwitterAccount, Reply
 from kitsune.sumo.redis_utils import redis_client, RedisError
 from kitsune.sumo.utils import chunked
+from kitsune.twitter import get_twitter_api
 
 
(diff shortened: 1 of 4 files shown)


rlr:
LGTM

kitsune/customercare/cron.py
```diff
@@ -54,10 +54,14 @@ def collect_tweets():
...
-                    settings.TWITTER_CONSUMER_SECRET,
-                    settings.TWITTER_ACCESS_TOKEN,
-                    settings.TWITTER_ACCESS_TOKEN_SECRET)
+        t = Twython(
+            settings.TWITTER_CONSUMER_KEY,
+            settings.TWITTER_CONSUMER_SECRET,
+            settings.TWITTER_ACCESS_TOKEN,
+            settings.TWITTER_ACCESS_TOKEN_SECRET,
+            client_args={
+                'timeout': 10,
```
willkg: I'd put a comment here if only to save someone else having to look it up. This means there's a 10 second connect timeout and a 10 second read timeout.

http://docs.python-requests.org/en/latest/user/advanced/?highlight=timeout#timeouts

mythmon:
I added the comment @willkg suggested, and also consolidated everything into a single helper, so now all uses of this API have a timeout. 10s might be really long for a web request timeout, but a 10s timeout is better than no timeout at all.

PR #2460, opened 2015-04-09
[Bug 1152471] Block search engines from indexing the media gallery

pollti:
media gallery images shouldn't show up in search results


rlr:
awesome. thanks!

PR #2462, opened 2015-04-09
[Bug 1150014] All unapproved revisions comment in Review Revision

safwanrahman:
I have tested in my local. and its working fine!
r?

### DIFF
--- a/kitsune/wiki/templates/wiki/review_revision.html
+++ b/kitsune/wiki/templates/wiki/review_revision.html
@@ -16,6 +16,24 @@
       {{ _('Revision Comment: {comment}')|f(comment=revision.comment) }}
       </p>
 
+      {% if unreviewed_revisions %}
+        <p>
+          <div class="unreviewed-revision"> {{ ngettext('Unreviewed Revision:', 'Unreviewed Revisions:', unreviewed_revisions|length) }} </div>
+          <br>
+          <ul class="revision-comment">
+            {% for revision in unreviewed_revisions %}
+              <li>
+                {{ _('Revision {id} by {user}')|f(id=revision.id, user=revision.creator) }}.
+                <br>
+                {{ _('<label>Revision Comment:</label> {comment}')|f(comment=revision.comment) }}
+                <br>
+                <a href="{{ url('wiki.review_revision', document.slug, revision.id) }}">{{ _('Review Revision {id}')|f(id=revision.id) }}</a>
+              </li>
+            {% endfor %}
+          </ul>
+        </p>
+      {% endif %}
+
       {% if revision.reviewed %}
         <p>
           {{ _('This revision was already reviewed by {user} on {date}.')|fe(user=revision.reviewer, date=datetimeformat(revision.reviewed, format='longdatetime')) }}
--- a/kitsune/wiki/views.py
+++ b/kitsune/wiki/views.py
@@ -550,11 +550,14 @@ def review_revision(request, document_slug, revision_id):
     revision_contributors = list(set(
         based_on_revs.values_list('creator__username', flat=True)))
 
+    # Get Unreviewed Revisions
+    unreviewed_revisions = Revision.objects.filter(document=doc,
+                                                   is_approved=False,
+                                                   reviewed=None).order_by('-id')
+
     # Get latest revision which is still not approved. Use **latest_revision_id** to get the id.
     try:
-        latest_unapproved_revision = Revision.objects.filter(
-            document=doc,
-            is_approved=False).order_by('-id')[0]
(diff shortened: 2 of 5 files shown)


mythmon:
This needs tests.

kitsune/wiki/templates/wiki/review_revision.html
```diff
@@ -16,6 +16,23 @@
       {{ _('Revision Comment: {comment}')|f(comment=revision.comment) }}
       </p>
 
+      {% if unreviewed_revision %}
+        <p>
+        <div class="unreviewed-revision"> {{ ngettext('Unreviewed Revision:', 'Unreviewed Revisions:', unreviewed_revision|length ) }} </div><br>
```
mythmon: Since this is inside the `<p>`, it should be indented one step more. Additionally, there should not be a space around the `{{ }}`. Good job with `nggettext` by the way. The is exactly how it should be used.

kitsune/wiki/templates/wiki/review_revision.html
```diff
@@ -16,6 +16,23 @@
...
+        <ul class="revision-comment">
+          {% for revision in unreviewed_revision %}
+            <li>
+              {{ _('Revision {id} by {user}')|f(id=revision.id, user=revision.creator) }}.
+              <br>
+              {{ _('<label>Revision Comment:</label> {comment}')|f(comment=revision.comment) }}
+              <br>
+              <a href="{{ url('wiki.review_revision', document.slug, revision.id) }}">{{ _('Review Revision {id}')|f(id=revision.id) }}</a>
+            </li>
+        {% endfor %}
```
mythmon: The indentation here does not match the indentation of the opening `{% for %}` tag, on line 23.

kitsune/wiki/views.py
```diff
@@ -525,6 +525,20 @@ def review_revision(request, document_slug, revision_id):
     revision_contributors = list(set(
         based_on_revs.values_list('creator__username', flat=True)))
 
+    # Get Unreviewed Revisions
+    unreviewed_revision = list(Revision.objects.filter(
```
mythmon: Since this is multiple revisions, the name should be `unreviewed_revisions`, to denote that it is more than one item.

kitsune/wiki/views.py
```diff
@@ -525,6 +525,20 @@ def review_revision(request, document_slug, revision_id):
...
+            document=doc,
+            is_approved=False).order_by('-created'))
+    try:
+        unreviewed_revision
+    except IndexError:
+        unreviewed_revision = ''
+
+    # Remove the revision which is reviewing and get last 5 revision
+    if rev in unreviewed_revision:
+        unreviewed_revision.remove(rev)
```
mythmon: This should be done with a `.exclude(id=rev.id)` clause in the ORM query above.

kitsune/wiki/views.py
mythmon: Also, generally calling `list()` on an ORM query is a sign that something isn't quite right. If you don't do the `if rev_in unreviewed_revision` thing below, you won't need to do `list()` here. That will be more efficient, as it will only get the first 5 from the DB, and not the entire list of every revision that matches these criteria.

kitsune/wiki/views.py
```diff
@@ -525,6 +525,20 @@ def review_revision(request, document_slug, revision_id):
...
+    unreviewed_revision = unreviewed_revision[0:5]
```
mythmon: When slicing, the first argument defaults to 0. In other words, this is better written as `unreviewed_revision = unreviewed_revision[:5]`.

mythmon:
I have a lot of little minor issues with this, but that is all they are: minor issues. This is actually quite a good PR, and I just want to make sure it fits our style guidelines and will be easy to work on in the future.

The tests you should write for this should generate a document and some revisions, and then check to see if the template contains the kinds of things you expect it to. If you need help with this, I can provide some examples of other code that does similar things. Ping me in irc, and I'll go find them.

safwanrahman:
@mythmon, I have added test and corrected other things as you mentioned! :smiley: :relaxed: 
Please regard only the last commit. other commits are just for updating the branch!
r?

mythmon:
Like before, it isn't really appropriate to merge master into PRs branches. I'll check the changes. Can you fix the merges by rebasing this branch?

safwanrahman:
@mythmon I have fixed the things as per you mentioned.
r?
After r+, I will remove the master branch things and squash it into one!

mythmon:
This looks fine. Please rebase this branch to not have any merge commits in it.

safwanrahman:
@mythmon, Take a little Glance and I think its ready to merge! :)

PR #2463, opened 2015-04-10
Bring in peep.sh from fjord to deal with pip version incompatibilities.

mythmon:
@willkg r?

### DIFF
--- a/peep.sh
+++ b/peep.sh
@@ -1,0 +1,32 @@
+#!/bin/bash
+
+# Get pip version number
+PIPVER=`pip --version | awk '{print $2}'`
+ARGS=""
+
+echo "peep.sh: Using pip $PIPVER"
+
+# If we're using pip 1.5, 1.6 or 6.0, then toss on the --no-use-wheel
+# argument.
+#
+# Note: If we encounter other versions of pip here that require other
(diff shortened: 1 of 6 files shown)


peep.sh
```diff
@@ -0,0 +1,32 @@
...
+# argument.
+#
+# Note: If we encounter other versions of pip here that require other
+# things, we can toss them in.
+case $PIPVER in
+    1.5*|1.6*|6.0*)
+        echo "peep.sh: Wheel-using pip detected, so passing --no-use-wheel."
+        ARGS="--no-use-wheel"
+        ;;
+    6.1*)
```
willkg: I think given you know 6.0 works and it's covered in the previous case and you know 6.1 does not work, you can probably assume all future versions of pip won't work (for now), so you could just do:

```
6*)
```

willkg: Plus:

```
6*)
```

kind of looks like a :smiley:.

willkg:
This looks good to me. r+

peep.sh
mythmon: I suspect that future versions of pip might start working, actually, due to pypa/pip#2655. Either way, I plan on watching to see what happens with pip and peep.

PR #2468, opened 2015-04-10
Deal with unicode in l10n msgids.

mythmon:
small r?


rlr:
unicode strikes back again!

PR #2474, opened 2015-04-14
[bug 1013471] Fix for Accidental deletion of profiles

anushbmx:
Implemented the prompt for user name as mentioned by @rehandalal on bug 1013471 in #comment5 
 
What is happening

1. The delete button is disabled in HTML, it gets activated once the page is loaded so that there is no accidental deletion if JavaScript fails to load
2. Once the user clicks the button it prompt the user with a warning message and asks for username. If the username entered is correct the deletions continues else its aborted.

I hope this is fine, any way sorry for the long delay

### DIFF
--- a/kitsune/sumo/static/js/ui.js
+++ b/kitsune/sumo/static/js/ui.js
@@ -7,6 +7,16 @@
     initFolding();
     initAnnouncements();
 
+    $('#delete-profile-username-input').keyup(function(ev) {
+      var username = $('#delete-profile-username').val();
+      var inputUsername = $('#delete-profile-username-input').val();
+      if (inputUsername === username) {
+        $('#delete-profile-button').prop('disabled', false);
+      } else {
+        $('#delete-profile-button').prop('disabled', true)
+      }
+    });
+
     $(window).scroll(_.throttle(function() {
       if ($(window).scrollTop() > $('body > header').outerHeight()) {
         $('body').addClass('scroll-header');
--- a/kitsune/users/templates/users/edit_profile.html
+++ b/kitsune/users/templates/users/edit_profile.html
@@ -59,12 +59,28 @@
       </form>
     </article>
     {% if request.user == profile.user %}
-      <form method="post" action="{{ url('users.close_account') }}" data-confirm="confirm" data-confirm-text="{{ _('Are you sure you want to close your account? You cannot undo this action.') }}">
+    <p>
+      <a id="delete-profile" class="btn btn-warning">{{ _('Close account and delete all profile information') }}</a>
+    </p>
+
+    <div class="kbox" title="{{ _('Are you sure') }}" data-target="#delete-profile" data-modal="true">
+      <form method="post" action="{{ url('users.close_account') }}">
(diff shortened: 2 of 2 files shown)


kitsune/sumo/static/js/ui.js
```diff
@@ -6,6 +6,17 @@
   $(document).ready(function() {
     initFolding();
     initAnnouncements();
+    $(".delete-profile-btn").removeAttr('disabled');
+
+    $('#delete-profile').submit(function(event){
+      var username = prompt("Are you sure you want to close your account? You cannot undo this action. Please enter your username to proceed.");
```
mythmon: This string used to be localized, but it isn't anymore. You need to wrap this in a call to gettext (`_`) too,

kitsune/sumo/static/js/ui.js
```diff
@@ -6,6 +6,17 @@
...
+      if(this.username.value == username){
+        this.submit();
+      }else{
+        alert('Sorry the username entered did not match the account. Aborting...');
+        event.preventDefault();
+      }
+    });
```
mythmon: There is a couple of style problems here. 

1. You should use `===` not `==` for comparison.
2. You should put spaces around `{` and `}`. For example, line 15 should be `} else {`, not `}else`{. 
3. Same for the `if` statement. It needs a space after it.

These are little things, but keeping everything consistent is valuable.

mythmon:
I'm not sure how I feel about using `alert()` and `prompt()` for this.

mythmon:
To be clear, my problem is with the functions `prompt()` and `alert()` in particular, not the confirmation steps (those I like). Instead of `prompt()`, you should probably use [KBox](https://github.com/mozilla/kitsune/blob/04987977eff950f62da20589e8a3c49a7730c63a/kitsune/sumo/static/js/kbox.js)

anushbmx:
@mythmon I have switched this from [code]prompt()[/code] and [code]alert()[/code] to Kbox as you said..

kitsune/sumo/static/js/ui.js
```diff
@@ -7,6 +7,18 @@
     initFolding();
     initAnnouncements();
 
+    $('#delete-profile-username-input').keyup(function(ev) {
+      var username = $('#delete-profile-username').val();
+      var inputUsername = $('#delete-profile-username-input').val();
+      if (inputUsername === username) {
+        console.log('disabled removed');
```
rehandalal: get rid of this

kitsune/sumo/static/js/ui.js
```diff
@@ -7,6 +7,18 @@
...
+        $('#delete-profile-button').prop('disabled', false);
+      } else {
+        console.log('disabled added');
```
rehandalal: same ^

rehandalal:
this looks good! just make the two small changes mentioned in the comments.

anushbmx:
@rehandalal Sorry I forgot to remove the logs, All fine now.

rehandalal:
Thanks!

PR #2476, opened 2015-04-14
Update Vagrant

safwanrahman:
Changelog:
* Use 32 bit of Ubuntu 14.04 istead of 64 Bit
* Add two new Requirments which was already added into the doc
* Install pip 6.0.3 as the latest version of pip 6.1.1 has issues with peep
* Add new frontend Dependency installing
* Boot mode changed to gui for not connection lost

r?
Can anyone test this to verify? I have tried to install via vagrant and it took me overnight with my low speed internet. I have noted the Errors and fixed according to it.
I am not able to run it again in order to test! :(
**If anyone can test it, and find out if any problem happens, that would be much helpful**

### DIFF
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -19,8 +19,8 @@ CONF = _config
 MOUNT_POINT = '/home/vagrant/kitsune'
 
 Vagrant::Config.run do |config|
-    config.vm.box = "ubuntu-14.04"
-    config.vm.box_url = "https://oss-binaries.phusionpassenger.com/vagrant/boxes/latest/ubuntu-14.04-amd64-vbox.box"
+    config.vm.box = "ubuntu/trusty32"
+    config.vm.box_url = "https://atlas.hashicorp.com/ubuntu/boxes/trusty32/versions/14.04/providers/virtualbox.box"
 
     Vagrant.configure("1") do |config|
         config.vm.customize ["modifyvm", :id, "--memory", CONF['memory']]
--- a/vagrantconfig.yaml
+++ b/vagrantconfig.yaml
@@ -9,4 +9,4 @@ memory: 256
 debug_mode: false
 
 # set this to "gui" if vagrant cannot connect
-boot_mode: none
+boot_mode: gui
--- a/bin/vagrant_provision.sh
+++ b/bin/vagrant_provision.sh
@@ -30,8 +30,10 @@ debconf-set-selections <<< 'mariadb-server-5.5 mysql-server/root_password_again 
 apt-get install -y sphinx-common libapache2-mod-wsgi python-pip libmysqlclient-dev git \
                    libxml2-dev libxslt1-dev zlib1g-dev libjpeg-dev python-dev libssl-dev \
                    openjdk-7-jre-headless mariadb-server-5.5 nodejs elasticsearch redis-server \
-                   memcached
+                   memcached libssl-dev libffi-dev
 
+# Install pip 6.0.3 as the latest version is not working with peep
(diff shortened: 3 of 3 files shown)


mythmon:
Your changes look good, but Virtualbox is failing to start on my system right now. I'll debug this more later, and try this again.

Vagrantfile
```diff
@@ -19,8 +19,8 @@ CONF = _config
 MOUNT_POINT = '/home/vagrant/kitsune'
 
 Vagrant::Config.run do |config|
-    config.vm.box = "ubuntu-14.04"
-    config.vm.box_url = "https://oss-binaries.phusionpassenger.com/vagrant/boxes/latest/ubuntu-14.04-amd64-vbox.box"
+    config.vm.box = "ubuntu/trusty32"
+    config.vm.box_url = "https://atlas.hashicorp.com/ubuntu/boxes/trusty32/versions/14.04/providers/virtualbox.box"
```
willkg: This switches it to the 32-bit vm. We should use the 64-bit vm.

You can set it up so that the vagrantfile.yml file lets you specify the box name to override this one.

vagrantconfig.yaml
```diff
@@ -9,4 +9,4 @@ memory: 256
 debug_mode: false
 
 # set this to "gui" if vagrant cannot connect
-boot_mode: none
+boot_mode: gui
```
willkg: This should be `none`.

Vagrantfile
willkg: Fjord uses this url:

```
 config.vm.box_url = "http://cloud-images.ubuntu.com/vagrant/trusty/current/trusty-server-cloudimg-amd64-vagrant-disk1.box"
```

safwanrahman: @willkg I think we should use 32 bit VM as every processor does not support 64 bit VM.

willkg: I think the default should be a 64-bit operating system since that's closer to what we're using in production. If you feel strongly about supporting 32-bit operating systems, then you can do some finagling in the `vagrantfile.yml` file to support people overriding the default.

safwanrahman: @willkg I think by default it should be 32 bit as 64 Bit Computer can have 32 bit VM but 32 bit Computer may not support 64 bit. So I think for dev Environment, 32 bit would be a good choice!

vagrantconfig.yaml
safwanrahman: @willkg Having this line as null create connection Error with the VM. I could not run vagrant with having it None

willkg: Just to clarify, it needs to be ``none`` with a lowercase "n". That works for Fjord, so I'm not sure why you had problems. What was the full error you got?

safwanrahman:
I think using a 32 Bit Box would be good solution to avoid Errors

vagrantconfig.yaml
safwanrahman: It was loosing connection with the VM if had `none`

willkg:
This is getting too painful. I get that you don't agree with my comments, but I can't r+ this in its current state and the back-and-forth over little things is frustrating.

We just did another round of Vagrant fixing on Fjord. It's much easier for me to take those changes and apply them here than it is to continue with this PR.

@mythmon I'd like to close this out.

mythmon:
I haven't done a lot of work with Vagrant. @willkg has done a lot more than I have. I'll defer to him on this issue.

safwanrahman:
Ok! No problem!

PR #2479, opened 2015-04-16
[Bug 1093509] In wiki, when translating, always show approved text in LTR.

AdrianGaudebert:
@mythmon r?

This is imperfect, because it assumes that the "default" language of the app is always an LTR language. I guess it won't ever be a problem for us, since the default should always be English, but that might not be the case for other users of kitsune. Does that matter? If so, what would be a good way to know if the default language is RTL or LTR?

### DIFF
--- a/kitsune/wiki/templates/wiki/translate.html
+++ b/kitsune/wiki/templates/wiki/translate.html
@@ -171,7 +171,7 @@
             <div class="approved">
               <h3>{{ _('Approved {default_locale} version:')|f(default_locale=settings.LANGUAGES_DICT[settings.WIKI_DEFAULT_LANGUAGE.lower()]) }}</h3>
               <div id="content-or-diff" class="content">
-                <textarea readonly="readonly">{{ based_on.content }}</textarea>
+                <textarea readonly="readonly"{% if settings.WIKI_DEFAULT_LANGUAGE.lower() not in settings.LANGUAGES_BIDI %} dir="ltr"{% endif %}>{{ based_on.content }}</textarea>
               </div>
             </div>
             <div class="localized">


kitsune/wiki/templates/wiki/translate.html
```diff
@@ -171,7 +171,7 @@
             <div class="approved">
               <h3>{{ _('Approved {default_locale} version:')|f(default_locale=settings.LANGUAGES_DICT[settings.WIKI_DEFAULT_LANGUAGE.lower()]) }}</h3>
               <div id="content-or-diff" class="content">
-                <textarea readonly="readonly">{{ based_on.content }}</textarea>
+                <textarea readonly="readonly" dir="ltr">{{ based_on.content }}</textarea>
```
willkg: Might be worth adding a `WIKI_DEFAULT_DIRECTION` to match `WIKI_DEFAULT_LANGUAGE`.

mythmon: Better yet, lets use the setting `settings.LANGUAGES_BIDI`, which is a list of the locale codes that are RTL. It is provided to us by Django's i18n.

mythmon:
In practice, there are no other users of kitsune besides SUMO. Someone from Ubuntu tried to stand up another instance, but we have too much code that is specific to Mozilla to make that a reality. On the other hand, making this adaptive is pretty easy (with `settings.LANGUAGES_BIDI`) so lets do that.

safwanrahman:
What about using django-bidi-utils?
I think its better for dealing both RTL and LTR
http://django-bidi-utils.readthedocs.org/en/latest/

mythmon:
@safwanrahman django-bidi-utils hasn't been updated since October of 2013. It likely isn't compatible with Django 1.7 or above anymore, and probably isn't maintained anyways. Most of what it does is already a part of Django (`settings.LANGUAGES_BIDI`), or something we have made in Kitsune anyways. Additionally, almost all of its functionality is done with Django templatetags. We can't use those, since we don't just Django templates, but use Jingo instead.

Is there something that django-bidi-utils does that the two suggested solutions above don't do and isn't already on the site. Is it useful in this small use case of flipping the direction of the English textbox on a translation page?

kitsune/wiki/templates/wiki/translate.html
safwanrahman: Or we can use like this:
`<textarea readonly="readonly"{% if not based_on.content.is_rtl %} dir="ltr"{% endif %}>{{ based_on.content }}</textarea>`

safwanrahman:
@mythmon so we can use as I mentioned in last comment
`<textarea readonly="readonly"{% if not based_on.content.is_rtl %} dir="ltr"{% endif %}>{{ based_on.content }}</textarea>`

mythmon:
`is_rtl` doesn't exist.

safwanrahman:
@mythmon, but its working in my local machine
firefox_screenshot_2015-04-17t02-43-21 543z

safwanrahman:
I think using the If we use the code `{% if not based_on.content.is_rtl %}`, it is better. Because in that case if the `wiki.default_language` is not LTR, the `dir="ltr"` will not get affected.
Therefore default language of kitsune should  **always not** be an LTR language. So I think its better use this

AdrianGaudebert:
Reworked using ``settings.LANGUAGES_BIDI``. Thanks for the comments everyone!

mythmon:
r+

mythmon:
@safwanrahman your approach seems equally valid. I'd be fine with either one, but the one here is the one I merged. If you feel very strongly about this, you could file a new PR.

PR #2482, opened 2015-04-17
Bug 1155599 - Update to peep 2.4

edmorley:
1) Update to peep v2.4 (https://github.com/erikrose/peep/archive/2.4.zip ; https://github.com/erikrose/peep/compare/2.2...2.4).
2) Remove blacklisting of pip 6.1 from peep.sh. Now that we've updated peep, it is no longer incompatible with pip 6.1*, so we don't need to error out.
3) Clarify explanation for --no-use-wheel in peep.sh.I've also removed references to pip v1.6 (since it does not exist) and cleaned up the comments that wrapped unnecessarily.

### DIFF
--- a/peep.sh
+++ b/peep.sh
@@ -8,20 +8,19 @@ ARGS=""
 
 echo "peep.sh: Using pip $PIPVER"
 
-# If we're using pip 1.5, 1.6 or 6.0, then toss on the --no-use-wheel
-# argument.
-#
-# Note: If we encounter other versions of pip here that require other
-# things, we can toss them in.
+# Add pip arguments that vary according to the version of pip, here:
 case $PIPVER in
-    1.5*|1.6*|6.*)
+    1.5*|6.*)
+        # Pip uses the wheel format packages by default in pip 1.5+.
+        # However in pip < 6.0+ wheel support is broken, and even with pip 6.0+
+        # we intentionally don't use the wheel packages, since otherwise each
+        # package in the requirements files would need multiple hashes.
         echo "peep.sh: Wheel-using pip detected, so passing --no-use-wheel."
         ARGS="--no-use-wheel"
         ;;
 esac
 
-# Execute peep with the command line args plus the additional args
-# if any.
+# Add the version specific arguments to those passed on the command line.
 python ./scripts/peep.py "$@" $ARGS
 
 echo "peep.sh: Done!"
--- a/peep.sh
+++ b/peep.sh
@@ -14,17 +14,10 @@ echo "peep.sh: Using pip $PIPVER"
 # Note: If we encounter other versions of pip here that require other
 # things, we can toss them in.
 case $PIPVER in
-    1.5*|1.6*|6.0*)
+    1.5*|1.6*|6.*)
         echo "peep.sh: Wheel-using pip detected, so passing --no-use-wheel."
         ARGS="--no-use-wheel"
         ;;
-    6.1*)
-        echo
-        echo "Error: peep does not support pip >= 6.1."
-        echo "Please downgrade your version of pip by running:"
-        echo ""
-        echo "    pip install --upgrade 'pip<6.1'"
-        exit 1
 esac
 
 # Execute peep with the command line args plus the additional args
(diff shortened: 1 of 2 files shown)


edmorley:
I don't have permissions to do this - but I suspect it would be wise to purge the Travis cache (https://travis-ci.org/mozilla/kitsune/caches) and then retrigger the CI runs, since with a populated cache, peep skips the pip install of the packages, so presumably we're not checking that the wheels versions of the packages match the SHAs in the repos.

peep.sh
```diff
@@ -8,27 +8,17 @@ ARGS=""
 
 echo "peep.sh: Using pip $PIPVER"
 
-# If we're using pip 1.5, 1.6 or 6.0, then toss on the --no-use-wheel
-# argument.
-#
-# Note: If we encounter other versions of pip here that require other
-# things, we can toss them in.
```
willkg: I'd keep this comment in. It's helpful to future people who might fix this file.

peep.sh
```diff
@@ -8,27 +8,17 @@ ARGS=""
...
+# Custom arguments required for some versions of pip.
 case $PIPVER in
-    1.5*|1.6*|6.0*)
-        echo "peep.sh: Wheel-using pip detected, so passing --no-use-wheel."
+    1.5*)
+        # Wheel support is broken in pip < v6.0 due to:
+        # https://github.com/pypa/pip/issues/1825
+        echo "peep.sh: Pip with broken wheel support detected, so passing --no-use-wheel."
```
willkg: This changes what this section was for. We don't want to use wheels, so this section passes the appropriate flag to versions of pip that support wheels. It's not about broken wheel support.

peep.sh
edmorley: It seems a pretty redundant to me - but happy to keep it in if preferred.

willkg:
peep 2.4 was released, so we might as well upgrade to that.

With Input, rather than have to maintain two sets of hashes which requires me to look at and test two sets of things, I just maintain hashes for tarballs and explicitly don't use wheels. I'm pretty sure that's the case with SUMO, but I don't recall if @mythmon has explicitly agreed or not.

If that's true of SUMO, too, then we want to pass a "don't use wheels" flag to all versions of pip that support wheels and not just the versions that had broken wheel support.

peep.sh
edmorley: Did you see the commit messages? This is split into three commits.

edmorley: Or are you deliberately wanting to still not use wheels, even though it now works fine with pip v6+?

edmorley:
Ok makes sense - particularly if you're stuck on old versions due to IT infra. On Treeherder we've escaped pip 1.x (and python 2.6) finally :-)

edmorley:
> peep 2.4 was released, so we might as well upgrade to that.

It's not tagged or on https://github.com/erikrose/peep/releases yet

willkg:
Maybe @erikrose didn't tag it, but it's definitely out there: 

https://pypi.python.org/pypi/peep/2.4

edmorley:
I'll ask him to tag it on github, I'd rather get it from there than pypi (and seeing as the releases feature is enabled on that repo, it's non-obvious that I need to also look somewhere else, when the release isn't listed there).

willkg:
That makes total sense to me. I was just mentioning why I thought it was released.

edmorley:
PR updated with peep 2.4 (instead of peep 2.3) and also to retain the --no-use-wheels behaviour, but with clarifications to the comments explaining why it's used :-)

mythmon:
I verified that the peep.py here matches the peep.py in [erikrose/peep ](https://github.com/erikrose/peep). I like the changes to the comments in peep.sh. Thanks for putting it back to tarballs only.

I'd be fine with switching to using both wheels and tarballs, *if* there was a good way to get all the new hashes for the packages, and additionally verify that the contents of the wheels matches the contents in the tarballs (or close enough). I don't think there is anything that does that, so it would be a really long, slow, manual process. I don't want to do that. It is easier to just not use wheels.

Thank you for the work!

edmorley:
Thanks for the review & merging :-)

PR #2484, opened 2015-04-20
[Bug 1149611] Add Bing site verification.

mythmon:
Since Yahoo search is really just Bing search, I think doing this will allow us to start customizing our Yahoo search results.

small r?


rlr:
possibly. I think rankings should be the same but the UI and content presented can be different I think. Worth a shot.

PR #2486, opened 2015-04-21
[bug 793789] RTL expand icons on wiki edit

rehandalal:
r?


rlr:
I tried it with hebrew. wfm

PR #2487, opened 2015-04-21
[bug 1146342] Fix accidental deletion of gallery images

rehandalal:
steps for testing are in the bug.

small r?


mythmon:
That seems tricky. Is it possible to write a test for this?

mythmon:
The test turns out to be really difficult to write, and this is pretty easy to verify manually.

r+

PR #2488, opened 2015-04-21
Fix issues with in_review.py

willkg:
* off-by-one with to_date
* using git-log with a range when we should be using git-show

r?

### DIFF
--- a/scripts/in_review.py
+++ b/scripts/in_review.py
@@ -91,7 +91,7 @@ def print_bugzilla_stats(from_date, to_date):
     print ''
     creators = sorted(creators.items(), reverse=True, key=lambda item: item[1])
     for person, count in creators:
-        print ' %34s : %s' % (person[:30], count)
+        print ' %34s : %s' % (person[:30].encode('utf-8'), count)
     print ''
 
     # ------------------------------------------------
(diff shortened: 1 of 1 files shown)


mythmon:
LGTM. r+

By the way, I'm getting this error both on this branch and on master. It happens right after the "Everyone" header gets printed out.

```traceback
  File "./scripts/in_review.py", line 356, in <module>
    sys.exit(main(sys.argv[1:]))
  File "./scripts/in_review.py", line 352, in main
    print_all_people()
  File "./scripts/in_review.py", line 294, in print_all_people
    people = sorted(all_people, key=lambda a: a.decode('utf-8').lower())
  File "./scripts/in_review.py", line 294, in <lambda>
    people = sorted(all_people, key=lambda a: a.decode('utf-8').lower())
  File "/home/mythmon/src/kitsune/venv/lib/python2.7/encodings/utf_8.py", line 16, in decode
    return codecs.utf_8_decode(input, errors, True)
UnicodeEncodeError: 'ascii' codec can't encode characters in position 0-2: ordinal not in range(128)
```

willkg:
Hrm... I thought I fixed this, but maybe I did it wrong. I'll look into it.

willkg:
^^^ That fixes the total crap that was str/unicode handling in the code. Sorry.

mythmon:
Now the script runs to completion without error. Thanks!

PR #2489, opened 2015-04-21
Sort and organize crontab.tpl.

mythmon:
This  shouldn't have changed any of the actual cronjobs or times. I just sorted the file by the time of day, and added zero padding to the times so they are easier to scan.

r?


willkg:
I went through the right side and made sure everything matched something on the left side and I double-checked the numbers.

The one thing I'd do is add a comment stating that the entries are in chronological order. Otherwise future editors of the file won't know unless they discover it.

This looks good to me. r+

mythmon:
Added comment about ordering.

PR #2490, opened 2015-04-22
[bug 1152816] Add Firefox for iOS to AAQ

rehandalal:
r?


mythmon:
Should the iOS icon be transparent? Right now it has a white background.

rehandalal:
That's how it was in the bug where they created a logo. It's got rounded corners to match app icons on iOS. We won't see it because it's on a white background though.

mythmon:
I tried asking a question. It all seemed to work fine. You're right, this looks fine on the site. However, on Mobile, it seems that if there are no suggested articles (which there aren't for iOS), the mobile AAQ gets stuck. That shouldn't be a problem once all this lands.

Do we need to add an icon for the mobile site? Or is that generated in some way from the big one?

Lets double check with Madalina if this can go out now or needs to wait.

rehandalal:
I added the icon to the product so it should show up on mobile on production.

mythmon:
I talked with Roland and Co. about this during the Project meeting today. We shouldn't land this yet, so this PR will just have to be open for a while. It is pretty isolated, so that shouldn't be a problem in terms of merge conflicts.

mythmon:
This is a good PR, but 3 months later, this is a merge conflict, and I have no idea when iOS support is going to be "turned on". The new AAQ has a different idea of products, so this might even be obsolete by the time it is needed. I'm going to close this to tidy things up.

PR #2492, opened 2015-04-22
[Bug 1156206] Enable Tswana as a locale for SUMO

vesper-mozilla:
This should add tn as a localizable locale on SUMO and redirect users to
the FAQ page by default.


mythmon:
LGTM. r+ once Travis goes :green_apple:

mythmon:
I'm not totally sure what is up here. Go ahead and undo that change to the test. It isn't really interesting here anyways.

mythmon:
You merged master into this. That isn't a good idea, because it creates a history that is unnecessarily convoluted at no gain. It would have been better to just make your commit and push it, instead of pulling from master.

Also, in the future, it would be better to make a branch for your changes, so that you can avoid problems like this. Additionally, under our new migration system, you'll need to add a migration any time you add a locale. I'll go over how to do that with you once I get it all figured out.

Given all that, I'm going to close this PR. Sorry about that.

PR #2496, opened 2015-04-23
Show python and django version in env admin page.

rlr:
This changes the Environment admin page to show python and django version in addition to the wsgi stuff.

r?

### DIFF
--- a/kitsune/kadmin/admin.py
+++ b/kitsune/kadmin/admin.py
@@ -1,7 +1,7 @@
 import re
 import sys
 
-from django import http
+from django import VERSION
 from django.conf import settings as django_settings
 from django.contrib import admin
 from django.db import connection
@@ -9,7 +9,6 @@ from django.shortcuts import render_to_response
 from django.template import RequestContext
 from django.views import debug
 
-import jinja2
 from celery import current_app
 from redis import ConnectionError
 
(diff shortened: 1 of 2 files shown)


kitsune/kadmin/admin.py
```diff
@@ -49,10 +48,14 @@ def celery_settings(request):
 
 
 def env(request):
-    """Admin view that displays the wsgi env."""
-    return http.HttpResponse(u'<pre>%s</pre>' % (jinja2.escape(request)))
-
-admin.site.register_view('env', view=env, name='WSGI Environment')
+    """Admin view that displays env info."""
+    return render_to_response('kadmin/env_view.html', {
+        'request': request,
```
willkg: In fjord, the request gets added to the jinja environment for every template rendering. Is that not true in kitsune?

willkg:
Hey! Ur stealing my codez! r+

kitsune/kadmin/admin.py
rlr: These are django templates though. And now I am confused as to why it works on fjord and not here.

willkg: Oh, right. We do that in django-land, too:

https://github.com/mozilla/fjord/blob/master/fjord/settings/base.py#L548

PR #2499, opened 2015-04-28
[DO NOT MERGE] Introducing Cool Search.

AdrianGaudebert:
This is for @mythmon's eyes. I'm showing this as a PR because it's nicer to read. Feel free to close any time. :)

### DIFF
--- a/kitsune/coolsearch/static/js/Result.jsx
+++ b/kitsune/coolsearch/static/js/Result.jsx
@@ -1,0 +1,10 @@
+export default class Result extends React.Component {
+    render() {
+        return (
+            <article className={"result " + this.props.type}>
+                <h3><a className="title" href="">{this.props.title}</a></h3>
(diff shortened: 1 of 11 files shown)


kitsune/coolsearch/static/js/Result.jsx
```diff
@@ -0,0 +1,10 @@
+export default class Result extends React.Component {
```
mythmon: Calling this file Result might be problematic. Although it seems like everything is namespaced with the Django app, JS files all get put into the same directory. I didn't figure out how to deal with this with the community hub React stuff. You may want to call this SearchResult, or maybe namespace all these files as `kitsune/coolsearch/static/js/coolsearch/Result.jsx`.

mythmon:
There isn't much actually going on except the boot strap, but what is here looks good.

PR #2501, opened 2015-04-29
[Bug 1158595] Enable section collapsing for all KB articles with a ToC.

mythmon:
r?


rlr:
I opened a few articles and it WFM. r+

PR #2502, opened 2015-04-30
Run migrations for real when deploying.

rlr:
Almost forgot about this.

r?


mythmon:
Ooh, yeah that would be bad to forget.

PR #2503, opened 2015-04-30
Add migrations for choice changes on locale fields.

mythmon:
These are migrations that should have been added with other code changes, but weren't, because no one thought they were making migration-worthy changes.

r?


willkg:
I just wrote up bug 1160296 (https://bugzilla.mozilla.org/show_bug.cgi?id=1160296). I kind of think we should fix that bug instead of landing these migrations.

willkg:
I wrote up bug 1160296, we discussed it, then decided it wasn't a hot idea.

Therefore, r+ on this.

PR #2504, opened 2015-05-01
[Bug 1159479] Add breadcrumbs to KB articles.

mythmon:
This resolves the issue of picking which topic/product to use for a document when there are multiple choices by picking one based on display_order. It doesn't try and preserve the navigation history of a user. It only tries to give the user an idea of where in the hierarchy they are, and how  they can explore the surrounding concepts.

r?


rehandalal:
travis ^

mythmon:
Fixed ^

rehandalal:
r+

PR #2505, opened 2015-05-01
[bug 1160493] Kill the rest of south 

rlr:
I think this covers it all.

I am pretty sure we dont need to do any special deconstruct for LocaleField. TimeZoneField should probably have one but that's off in some third party lib. We should probably break that off into a separate bug if we think it's needed (I am not clear on that).

r?


mythmon:
Django's migration seem to have figured out everything about the TimeZoneField, as far as I can tell. If there is a problem there (and I don't think there is) we can come back to it.

r+

PR #2506, opened 2015-05-04
[Bug 1156206] Enable Tswana as a locale for SUMO

vesper-mozilla:
This should add tn as a localizable locale on SUMO and redirect users to
the FAQ page by default. (with no added mess like in the previous PR, apologies for that)


mythmon:
Looks good.

PR #2510, opened 2015-05-06
[bug 1161050] Invalid locales should raise HTTP 400

willkg:
This fixes the search suggest api to return an HTTP 400 if you try to
filter on an invalid locale.

Further, it fixes the questions search so that it filters on
locale--previously, it didn't.

Adds some additional tests for behavior that wasn't previously tested.

r?

### DIFF
--- a/kitsune/search/api.py
+++ b/kitsune/search/api.py
@@ -32,6 +32,15 @@ def suggest(request):
         errors['q'] = 'This field is required.'
     if product is not None and not Product.objects.filter(slug=product).exists():
         errors['product'] = 'Could not find product with slug "{0}".'.format(product)
+    if locale not in settings.SUMO_LANGUAGES:
+        if settings.NON_SUPPORTED_LOCALES.get(locale):
+            errors['locale'] = (
+                'Locale "{0}" is not supported, but has fallback locale "{1}".'.format(
+                    locale, settings.NON_SUPPORTED_LOCALES[locale])
+            )
+        else:
+            errors['locale'] = 'Could not find locale "{0}".'.format(locale)
+
     if errors:
         raise GenericAPIException(400, errors)
 
@@ -55,9 +64,10 @@ def _question_suggestions(searcher, text, locale, product, max_results):
         question_is_archived=False,
         question_is_locked=False,
         question_is_solved=True)
-
     if product is not None:
         search_filter &= es_utils.F(product=product)
+    if locale is not None:
+        search_filter &= es_utils.F(question_locale=locale)
 
     questions = []
     searcher = _query(searcher, QuestionMappingType, search_filter, text, locale)
--- a/kitsune/search/tests/test_api.py
+++ b/kitsune/search/tests/test_api.py
@@ -2,6 +2,8 @@ from nose.tools import eq_
 
 from rest_framework.test import APIClient
 
+from django.conf import settings
+
 from kitsune.search.tests.test_es import ElasticTestCase
 from kitsune.sumo.urlresolvers import reverse
 from kitsune.questions.tests import question, answer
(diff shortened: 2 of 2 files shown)


rehandalal:
r+

willkg:
@mythmon Does this implementation match what you were thinking API-wise?

kitsune/search/tests/test_api.py
```diff
@@ -65,6 +67,30 @@ def test_invalid_product(self):
...
+            'locale': 'bad-medicine',
+            'q': 'search',
+        })
+        eq_(res.status_code, 400)
+        eq_(res.data['detail'], {'locale': 'Could not find locale "bad-medicine".'})
+
+    def test_invalid_fallback_locale(self):
+        for locale, fallback in settings.NON_SUPPORTED_LOCALES.items():
+            if fallback is not None:
+                break
```
mythmon: That's a neat trick.

willkg: Now that I look at it, it should be sorted so that it's stable. `.items()` can come in any order, so this could return a differently ordered iterable for you than it does for me which is irritating if there were bugs.

I'll fix that.

willkg: For the record, this (ab)uses the fact that locale and fallback leak out of the `for`. That seems to work in Python 3, too, but I have vague memories of this misfeature being fixed.

I'll redo it so it's less "interesting".

mythmon: I don't think it is a misfeature, since "for" doesn't introduce a new scope in Python, nor is it supposed to. I'm not opposed to replacing this with something simpler though.

kitsune/search/api.py
```diff
@@ -32,6 +32,15 @@ def suggest(request):
         errors['q'] = 'This field is required.'
     if product is not None and not Product.objects.filter(slug=product).exists():
         errors['product'] = 'Could not find product with slug "{0}".'.format(product)
+    if locale not in settings.SUMO_LANGUAGES:
+        if settings.NON_SUPPORTED_LOCALES.get(locale):
+            errors['locale'] = (
+                'Locale "{0}" is not supported, but has fallback locale "{1}".'.format(
+                    locale, settings.NON_SUPPORTED_LOCALES[locale])
+            )
```
mythmon: There are a lot of locales in NON_SUPPORTED_LOCALES with a value of None, which means they redirect to WIKI_DEFAULT_LANGUAGE. For consistency the if statement should probably be `if locale in settings.NON_SUPPORTED_LOCALES` and the message should use `settings.NON_SUPPORTED_LOCALES.get(locale, 'en-US')`.

kitsune/search/tests/test_api.py
willkg: My bad--I'm remembering wrong. It's the for in list comprehensions that leaks in Python 2, but not Python 3.

https://docs.python.org/3/whatsnew/3.0.html#changed-syntax

kitsune/search/api.py
willkg: Oh, wow--that's completely not clear from the comment for that setting. I had no idea.

So is this correct:

```
if locale not in settings.SUMO_LANGUAGES:
    if locale in settings.NON_SUPPORTED_LOCALES:
        errors['locale'] = (
            'Locale "{0}" is not supported, but has fallback locale "{1}".'.format(
                locale, settings.NON_SUPPORTED_LOCALES.get(locale, settings.WIKI_DEFAULT_LANGUAGE)
            )
        )
    ...
```

mythmon: That looks correct.

willkg:
Ugh. Because this got merged, the other commits aren't making it into the PR. I suppose that makes sense, but it's irritating.

I'll open a new PR with the fixes.

PR #2511, opened 2015-05-06
1161050 locales

willkg:
Fixes for PR #2510.

r?

### DIFF
--- a/kitsune/search/tests/test_api.py
+++ b/kitsune/search/tests/test_api.py
@@ -75,7 +75,8 @@ class SuggestViewTests(ElasticTestCase):
         eq_(res.status_code, 400)
         eq_(res.data['detail'], {'locale': 'Could not find locale "bad-medicine".'})
 
-    def test_invalid_fallback_locale(self):
+    def test_invalid_fallback_locale_none_case(self):
+        # Test the locale -> locale case.
         non_none_locale_fallback_pairs = [
             (key, val) for key, val in sorted(settings.NON_SUPPORTED_LOCALES.items())
             if val is not None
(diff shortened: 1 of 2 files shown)


willkg:
@mythmon Does this look right?

kitsune/search/tests/test_api.py
```diff
@@ -75,10 +75,13 @@ def test_invalid_locale(self):
...
-        for locale, fallback in settings.NON_SUPPORTED_LOCALES.items():
-            if fallback is not None:
-                break
+    def test_invalid_fallback_locale_none_case(self):
+        # Test the locale -> locale case.
+        non_none_locale_fallback_pairs = [
+            (key, val) for key, val in sorted(settings.NON_SUPPORTED_LOCALES.items())
+            if val is not None
+        ]
+        locale, fallback = non_none_locale_fallback_pairs[0]
```
mythmon: Instead of this, which is less tricky, but still fairly complex, why not hard-code a value for `settings.NON_SUPPORTED_LOCALES` using `@settinigs_override`? That would make the test code a bit simpler.

mythmon:
The logic looks right to me. I offered a suggestion to make the tests simpler, but either way is fine.

kitsune/search/tests/test_api.py
willkg: It would, but if someone changes how that setting works, then this would be invalid and may or may not break appropriately.

This isn't really that complex. It's pretty straight-forward. If you want, I could code up `first` and then use that.

mythmon: It isn't that complex, you're right. It is fine the way it is, and you have a good point about the setting changing. Lets leave it.

mythmon:
Thanks for following up on this!

PR #2512, opened 2015-05-06
[Bug 1162223] Use correct variable name for topics in mobile document view.

mythmon:
This was a flub from #2504. I changed a variable name in the view and the desktop template, but forgot about the mobile template.

r?

### DIFF
--- a/kitsune/wiki/tests/test_views.py
+++ b/kitsune/wiki/tests/test_views.py
@@ -10,9 +10,9 @@ from nose.tools import eq_
 from pyquery import PyQuery as pq
 
 from kitsune.products.tests import product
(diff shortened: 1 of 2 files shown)


kitsune/wiki/tests/test_views.py
```diff
@@ -667,3 +668,19 @@ def test_caching(self):
...
+class MobileDocumentTests(MobileTestCase):
+
+    def setUp(self):
+        super(MobileDocumentTests, self).setUp()
+        self.doc, _ = doc_rev()
+        p = product(save=True)
+        self.doc.products.add(p)
+        self.doc.save()
+
+    def test_it_works(self):
```
rehandalal: yay!

PR #2513, opened 2015-05-08
[bug 1159483] Related documents for KB articles

rehandalal:
r?

### DIFF
--- a/kitsune/sumo/static/js/ui.js
+++ b/kitsune/sumo/static/js/ui.js
@@ -58,12 +58,14 @@
       var $this = $(this);
       var id = $this.data('close-id');
       if (id) {
-        if (localStorage.getItem(id + '.closed') === 'true') {
-          var $target = $('#' + id);
-          if ($this.data('close-type') === 'remove') {
-            $target.remove();
-          } else {
-            $('#' + id).hide();
+        if (Modernizr.localstorage) {
+          if (localStorage.getItem(id + '.closed') === 'true') {
+            var $target = $('#' + id);
+            if ($this.data('close-type') === 'remove') {
+              $target.remove();
+            } else {
+              $('#' + id).hide();
+            }
           }
         }
       }
--- a/kitsune/sumo/static/js/ui.js
+++ b/kitsune/sumo/static/js/ui.js
@@ -36,23 +36,35 @@
 
     $(document).on('click', '.close-button', function() {
       var $this = $(this);
+      var $target;
       if ($this.data('close-id')) {
-        $('#' + $this.data('close-id')).hide();
+        $target = $('#' + $this.data('close-id'));
         if ($this.data('close-memory') == 'remember') {
           if (Modernizr.localstorage) {
             localStorage.setItem($this.data('close-id') + '.closed', true);
           }
         }
       } else {
-        $this.parent().hide();
+        $target = $this.parent();
+      }
+      if ($this.data('close-type') === 'remove') {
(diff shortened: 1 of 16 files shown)


kitsune/sumo/static/js/ui.js
```diff
@@ -36,23 +36,35 @@
...
+        $target.hide();
       }
     });
 
     $('[data-close-memory="remember"]').each(function() {
-      var id = $(this).data('close-id');
+      var $this = $(this);
+      var id = $this.data('close-id');
       if (id) {
         if (localStorage.getItem(id + '.closed') === 'true') {
```
mythmon: I think this needs to be wrapped in a check for localStorage existing.

kitsune/sumo/static/js/wiki_search.js
```diff
@@ -0,0 +1,78 @@
...
+      createResultsList();
+    }
+    $resultsList.html(_.render('wiki-search-results.html', data));
+  }
+
+  function handleSearch() {
+    var $this = $(this);
+    if ($this.val().length === 0) {
+      if (searchTimeout) {
+        window.clearTimeout(searchTimeout);
```
mythmon: I think it is safe to `clearTimeout(undefined)`, so I don't think you need to have the if statement.

kitsune/sumo/static/js/wiki_search.js
```diff
@@ -0,0 +1,78 @@
+(function($, _) {
```
mythmon: Should you be calling `k.nunjucksEnv` `_`? That is pretty confusing considering that is also Underscore and usual name.

kitsune/sumo/static/less/wiki.less
```diff
@@ -615,6 +615,48 @@ textarea[name="content"] {
...
+      line-height: 16px;
+      margin-top: -8px;
+      position: absolute;
+      right: 5px;
+      text-align: center;
+      top: 50%;
+      width: 16px;
+
+      &:before {
+        content: '\2715';
```
mythmon: Can you add a comment with what this character is?

kitsune/wiki/migrations/0003_auto_20150506_2316.py
```diff
@@ -0,0 +1,21 @@
+# -*- coding: utf-8 -*-
```
mythmon: Can you either name the migration to explain what it does, add a module level docstring explaining what it does, or both? I think that is a good habit to  be in for the new migration system.

kitsune/sumo/static/js/wiki_search.js
```diff
@@ -0,0 +1,78 @@
...
+    $resultsList.html(_.render('wiki-search-results.html', data));
+  }
+
+  function handleSearch() {
+    var $this = $(this);
+    if ($this.val().length === 0) {
+      if (searchTimeout) {
+        window.clearTimeout(searchTimeout);
+      }
+      $resultsList.html('');
```
mythmon: Sometimes I get `$resultsList is undefined` from this line. I think mainly when the window closes right after I make a search.

mythmon:
This works well. Visually it is a little confusing since there isn't a lot of contrast between the results list and the other UI elements on the page. Maybe shifting it left or right a bit will help that? or add a drop shadow? We've done that with other floating parts of the site before.

rehandalal:
updated ^

mythmon:
Looks good!

PR #2514, opened 2015-05-08
[Bug 928234] Display full name of country on profile

anushbmx:
Displays user country in expanded form.

### DIFF
--- a/kitsune/users/templates/users/profile.html
+++ b/kitsune/users/templates/users/profile.html
@@ -32,11 +32,11 @@
       <h2 class="location">
         {% if profile.city and profile.country %}
           {{ _('{city}, {country}')|f(city=profile.city,
-                                      country=profile.country) }}
+                                      country=profile.get_country_display() ) }}
         {% elif profile.city %}
           {{ profile.city }}
         {% elif profile.country %}
-          {{ profile.country }}
+          {{ profile.get_country_display() }}
         {% endif %}
       </h2>
       <div class="pm-or-signout">


willkg:
Did you test this? I'm pretty sure the changes you've made won't have any effect on the template or what gets rendered.

anushbmx:
@willkg yes I did

willkg:
Well, I'm puzzled. The country variable you added to the context is never used, so I don't see how this works.

anushbmx:
@willkg its used in profiles.html 
https://github.com/mozilla/kitsune/blob/master/kitsune/users/templates/users/profile.html#L39

It will display the country name instead of two letter code for the country.

willkg:
That's not `country`--that's `profile.country`. The `country` variable you added to the context here is not used:

https://github.com/mozilla/kitsune/pull/2514/files#diff-f1ab772aa26e1d5e799a6745706230ecR332

However, I see how this is working now. It's because you're changing the model data here:

https://github.com/mozilla/kitsune/pull/2514/files#diff-f1ab772aa26e1d5e799a6745706230ecR329

That's super bad--you shouldn't do that.

I don't know how @mythmon would prefer you fix this. I'll defer to him.

mythmon:
Gah! @willkg is right, changing the data on the profile model is a really bad idea. You shouldn't be changing the view at all in this case. Instead change the template to refer to `user_profile.get_country_name()`. Making changes to model instances outside of the model is really counter to the idea of models in the first place. A model should have full control over the data, and you shouldn't be modifying them unless you plan on calling `.save()` to actually modify them (which would be a validation error in this case, so don't do that).

anushbmx:
@mythmon and @willkg I moved that to profile.html

That is a catch for me now, not to modify data in model controller :)

kitsune/users/views.py
```diff
@@ -328,6 +328,7 @@ def profile(request, template, username):
     groups = user_profile.user.groups.all()
     return render(request, template, {
         'profile': user_profile,
+        'country': user_profile.get_country_display(),
```
willkg: You're still not using this variable, so you don't need to add it to the context.

mythmon:
@anushbmx Remember that you need to comment for us to know that you changed this PR. No one gets notifications about pushing commits.

This looks good now, thanks for going through the review.

willkg:
@mythmon anush pinged me in IRC about it. I should have followed up with a comment in here. Sorry about that.

anushbmx:
@mythmon I thought @willkg would look ok. Ill make sure to comment next time.

PR #2515, opened 2015-05-08
[bug 1159487] Show related products on KB articles

rehandalal:
r?

### DIFF
--- a/kitsune/wiki/templates/wiki/document.html
+++ b/kitsune/wiki/templates/wiki/document.html
@@ -1,5 +1,5 @@
 {% extends "wiki/base.html" %}
-{% from "wiki/includes/sidebar_modules.html" import document_tools, document_notifications with context %}
+{% from "wiki/includes/sidebar_modules.html" import document_tools, document_notifications, related_products_sidebar with context %}
 {% from 'products/includes/topic_macros.html' import topic_sidebar %}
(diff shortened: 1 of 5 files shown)


kitsune/wiki/templates/wiki/document.html
```diff
@@ -83,6 +83,11 @@ <h1 class="product-title cf">
     {% set parent = doc.parent %}
   {% endif %}
   {{ document_tools(doc, parent, user, 'article', settings, include_showfor=True) }}
+  <h4>{{ _('Explore more topics') }}</h4>
   {{ topic_sidebar(product_topics[:10], [], product) }}
+  {% if related_products %}
+    <h4>{{ _('Related products') }}</h4>
```
mythmon: Can this header be a part of the macro, so we only have to include the if statement once? I don't care if the if statement goes inside or outside the macro (though I'd lean toward inside).

mythmon:
This looks quite good. I can't see the product logos locally, but I can never see any logos locally, so I'm assuming it looks fine. The tests pass locally (Travis is having major issues). I have one little problem about the template, but besides that, I think this is good to go.

Also, lets hold off merging this until #2518 gets sorted out.

rehandalal:
updated ^

PR #2517, opened 2015-05-11
[bug 974956] Create chart of Top 100 articles in L10N metrics

rehandalal:
r?


mythmon:
This looks straight forward.

PR #2518, opened 2015-05-11
[Bug 1163855] Make helptext unicode for Profile fields

mythmon:
Because apparently this matters a lot to... something? It was causing tracebacks in the API on prod. I updated the tests to find this kind of error, though I don't really understand it.

r?

### DIFF
--- a/kitsune/users/migrations/0006_auto_update_help_text_email_sent_fields.py
+++ b/kitsune/users/migrations/0006_auto_update_help_text_email_sent_fields.py
@@ -1,0 +1,21 @@
+# -*- coding: utf-8 -*-
+from __future__ import unicode_literals
+
+from django.db import models, migrations
+import kitsune.sumo.models
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('users', '0005_set_initial_contrib_email_flag'),
+    ]
+
+    operations = [
+        migrations.AlterField(
+            model_name='profile',
+            name='first_l10n_email_sent',
+            field=models.BooleanField(default=False, help_text='Has been sent a first revision contribution email.'),
+            preserve_default=True,
+        ),
+    ]
--- a/kitsune/users/migrations/0006_auto_update_help_text_email_sent_fields.py
+++ b/kitsune/users/migrations/0006_auto_update_help_text_email_sent_fields.py
@@ -15,7 +15,13 @@ class Migration(migrations.Migration):
         migrations.AlterField(
             model_name='profile',
             name='first_l10n_email_sent',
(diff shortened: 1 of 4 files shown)


mythmon:
PS: This is currently broken in prod due to today's push. It probably effects BuddyUp, and a few power-user contributors.

kitsune/users/migrations/0006_auto_update_help_text_email_sent_fields.py
```diff
@@ -0,0 +1,21 @@
...
+
+    dependencies = [
+        ('users', '0005_set_initial_contrib_email_flag'),
+    ]
+
+    operations = [
+        migrations.AlterField(
+            model_name='profile',
+            name='first_l10n_email_sent',
+            field=models.BooleanField(default=False, help_text='Has been sent a first revision contribution email.'),
```
willkg: You changed the help text on two boolean fields, but this is still a string and not unicode and there's only one boolean field represented. That seems fishy.

mythmon: Oooops. I did some surgery on this migration, and in my haste cut too deeply.

mythmon:
Added the second migration.

willkg:
I didn't run this or test it, but it looks good to me. r+

willkg:
I should run the tests. Give me ten minutes.

willkg:
I get one test failure:

```
FAIL: kitsune/dashboards/tests/test_cron.py:L10nMetricsTests.test_update_l10n_coverage_metrics
  File "/home/willkg/mozilla/kitsune/kitsune/dashboards/tests/test_cron.py", line 281, in test_update_l10n_coverage_metrics
    eq_(4, WikiMetric.objects.filter(locale='es').count())
  File "/home/willkg/.virtualenvs/kitsune/lib/python2.7/site-packages/nose/tools.py", line 31, in eq_
AssertionError: 4 != 6
```

But that could be spurious. I occasionally get test errors and then they "go away".

willkg:
Also :red_herring: should be an emoji.

willkg:
On second thought, that test is failing all the time--not sporadically. It's weird that that's related, though, right?

mythmon:
Uh oh. I think that is a separate issue from another bug. I suspect that if you run the tests on master, you'll get the same problem. I'll double check that myself, and dig into why we have that test failure.

mythmon:
Ok, I found the problem. There was a test that wasn't updated for the metrics dashboard top 100 bug. I updated it and pushed again. Now all the tests are passing for me.

Also, :red_herring: should totally be an emoji.

willkg:
Works for me! r+

PR #2520, opened 2015-05-12
[bug 1157438] Searches with product names get filtered

rehandalal:
r?

### DIFF
--- a/kitsune/sumo/static/tpl/search-results.html
+++ b/kitsune/sumo/static/tpl/search-results.html
@@ -28,13 +28,13 @@
     {% endif %}
     <ul id="product-filter" class="search-filter sidebar-nav">
       <li {{ product|class_selected(null)|safe }}>
-        <a href="#" data-instant-search-unset-params="product" data-instant-search="link" data-href="{{ base_product_url }}">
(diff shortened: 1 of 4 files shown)


kitsune/sumo/static/tpl/search-results.html
```diff
@@ -28,13 +28,13 @@
     {% endif %}
     <ul id="product-filter" class="search-filter sidebar-nav">
       <li {{ product|class_selected(null)|safe }}>
-        <a href="#" data-instant-search-unset-params="product" data-instant-search="link" data-href="{{ base_product_url }}">
+        <a href="#" data-instant-search-unset-params="product" data-instant-search-set-params="all_products=1" data-instant-search="link" data-href="{{ base_product_url|urlparams(all_products=1) }}">
```
mythmon: It seems like you specify `all_products=1` twice here, once in `data-instant-search-set-params` and once in `data-href`. Are the both needed?

mythmon: Looking at the code more closely it seems like they are used differently, and this is consistent, so :+1:.

mythmon:
This works for instant search, but not for the old-school search page (which still comes up from time to time) http://localhost:8080/en-US/search?q=thunderbird+crash

mythmon:
And to clarify, on the old-school search page the product is appropriately picked, but I can't change to "All Products".

rehandalal:
Should work on old search page too now ^

PR #2521, opened 2015-05-12
[Bug 1164008] Filter out Pocket related tweets

MarkSchmidty:
added '-pocket' to the query used to find tweets for Army of Awesome, per bugzilla bug #1164008

### DIFF
--- a/kitsune/customercare/cron.py
+++ b/kitsune/customercare/cron.py
@@ -58,7 +58,7 @@ def collect_tweets():
 
         search_options = {
             'q': ('firefox OR #fxinput OR @firefoxbrasil OR #firefoxos '
-                  'OR @firefox_es'),
+                  'OR @firefox_es -pocket'),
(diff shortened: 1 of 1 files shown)


mythmon:
Thanks for this! In the future, we generally put the bug number in the "subject" of the commit and PR, like `[Bug 1164008] Filter out Pocket related tweets` instead of in the "body" of the commit.

Once our CI testing tool approves this change, I think it is ready to merge.

mythmon:
Oh, actually, I spoke too soon. This isn't the correct way to blacklist words. The correct way is to add it as a setting here: https://github.com/MarkSchmidty/kitsune/blob/patch-1/kitsune/settings.py#L847

Can you change this PR to make the change that way instead?

MarkSchmidty:
Ah, thanks. I'm just not familiar enough with Kitsune. I'm not sure whether '#UninstallFirefox pocket' or '#UninstallFirefox OR pocket' is right here. You're the expert. I'll let you make the change.

PR #2525, opened 2015-05-13
[bug 1164205] Refactor search suggest api

willkg:
This refactors the search suggest API to use more DRF-y things including
built-in DRF-y validation.

In the process of doing this, I changed it so that you can provide the filter
parameters in the querystring OR in the HTTP request body as JSON.

r?

### DIFF
--- a/docs/api.rst
+++ b/docs/api.rst
@@ -1,0 +1,141 @@
+===
+API
+===
+
+SUMO has a series of API endpoints to access data.
+
+.. contents::
+
+
+Search suggest API
+==================
+
+:Endpoint:     ``/api/2/search/suggest/``
+:Method:       ``GET``
+:Content type: ``application/json``
+:Response:     ``application/json``
+
+The search suggest API allows you to get back kb documents and aaq
+questions that match specified arguments.
+
+Arguments can be specified in the url querystring or in the HTTP
+request body.
+
+
+Required arguments
+------------------
+
++-------------------+--------+--------------------------------------------------------+
+|argument           |type    |notes                                                   |
++===================+========+========================================================+
+|q                  |string  |This is the text you're querying for.                   |
++-------------------+--------+--------------------------------------------------------+
+
+
+Optional arguments
+------------------
+
++-------------------+--------+--------------------------------------------------------+
+|argument           |type    |notes                                                   |
++===================+========+========================================================+
+|locale             |string  |default: ``settings.WIKI_DEFAULT_LANGUAGE``             |
+|                   |        |                                                        |
+|                   |        |The locale code to restrict results to.                 |
+|                   |        |                                                        |
+|                   |        |Examples:                                               |
+|                   |        |                                                        |
+|                   |        |* ``en-US``                                             |
+|                   |        |* ``fr``                                                |
++-------------------+--------+--------------------------------------------------------+
+|product            |string  |default: None                                           |
+|                   |        |                                                        |
+|                   |        |The product to restrict results to.                     |
+|                   |        |                                                        |
+|                   |        |Example:                                                |
+|                   |        |                                                        |
+|                   |        |* ``firefox``                                           |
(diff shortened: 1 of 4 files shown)


willkg:
This curl stuff helps test:

```
curl -X GET "http://localhost:8000/api/2/search/suggest/?q=videos&max_documents=3&max_questions=0"

echo ""

curl -X GET http://localhost:8000/api/2/search/suggest/ \
     -H "Content-Type: application/json" \
     -d '
{
   "q": "videos",
   "max_documents": 3,
   "max_questions": 0
}'

echo ""

curl -X GET "http://localhost:8000/api/2/search/suggest/?q=videos" \
     -H "Content-Type: application/json" \
     -d '
{
   "max_documents": 3,
   "max_questions": 0
}'
```

willkg:
The Travis failure is from failing to download or not having simplejson which isn't related to these changes. It's either a spurious error or related to the master branch.

kitsune/search/api.py
```diff
@@ -11,49 +12,75 @@
 from kitsune.wiki.models import DocumentMappingType
 
 
+def positive_integer(value):
```
rlr: do you need to check if it's an `int` in the first place?

willkg: The serializer has it as an IntegerField, so that should do the integer validation. Pretty sure one of the existing tests covers that scenario.

rlr: Makes sense. I typed before scrolling

willkg:
^^^ Add some rough docs for the sumo search suggest api.

kitsune/search/tests/test_api.py
```diff
@@ -139,6 +140,38 @@ def test_it_works(self):
...
 
+    def test_filters_in_postdata(self):
+        q1 = self._make_question()
+        d1 = self._make_document()
+        self.refresh()
+
+        data = json.dumps({'q': 'emails'})
+        # Note: Have to use .generic() because .get() will convert the
+        # data into querystring params and then it's clownshoes all
+        # the way down.
```
rlr: not :high_heel: dance shoes?

docs/api.rst
```diff
@@ -0,0 +1,141 @@
...
+               "num_votes_past_week": ... # the number of votes in the last week
+               "num_votes": ...           # the total number of votes
+               "product": ...             # the product
+               "solution": ...            # id of answer marked as a solution if any
+               "taken_until": ...         # FIXME:
+               "taken_by": ...            # FIXME:
+               "title": ...               # title of the question
+               "topic": ...               # FIXME:
+               "updated_by": ...          # FIXME:
+               "updated": ...             # FIXME:
```
rlr: that's a shitton of fields. But I guess that's what buddy up needed... or whoever was using this API

rlr:
I ran tests. Did some manual tests via browser and curl. My machine didn't catch :fire:.

r+

docs/api.rst
willkg: It's a lot of data. I have no idea why questions produces that much data, but documents does not. I think this API endpoint needs some more thought. I was thinking that documenting it might highlight some of the curiosities.

mythmon: I decided that the API shouldn't try and give only the minimal amount of data for "top level" results that aren't the children of another object. Instead most uses of an object give the full data available to the API. The only time the API sends reduced amounts of data is when an object is referenced from another objects as a foreign key, such as the `creator` field here. It is a lot of fields, yes, but I haven't seen any evidence that it is a problematic amount of fields. We chose to optimize for a few larger API responses instead of making smaller responses and possibly needing to make many requests to get the needed information.

willkg: I think that suggests someone should write up a bug for fixing documents so the returned results are more like questions.

mythmon: :+1: I just filed [bug 1164565](https://bugzilla.mozilla.org/show_bug.cgi?id=1164565).

PR #2527, opened 2015-05-13
Revert "Merge pull request #2501 from mythmon/collapse-all-1158595"

mythmon:
This reverts the change that made all KB articles with ToCs collapse sections. This feature isn't ready. It causes real problems in the KB, and needs to bake more before we can put it back.

This reverts commit 130d08ab0930aeadb133918f35ffa17d0eb122dc.


mythmon:
This is a straight revert, and Travis likes it, so I'm going to land it.

PR #2528, opened 2015-05-13
[bug 1164565] Fix document results in search suggest API

willkg:
This switches document results to use DocumentDetailSerializer. It also
fixes DocumentDetailSerializer to include the document summary which is
on the current revision and the document url.

r?

### DIFF
--- a/kitsune/search/api.py
+++ b/kitsune/search/api.py
@@ -9,7 +9,8 @@ from kitsune.questions.models import Question, QuestionMappingType
 from kitsune.questions.api import QuestionSerializer
 from kitsune.search import es_utils
 from kitsune.sumo.api import GenericAPIException
-from kitsune.wiki.models import DocumentMappingType
+from kitsune.wiki.api import DocumentDetailSerializer
+from kitsune.wiki.models import Document, DocumentMappingType
 
 
 def positive_integer(value):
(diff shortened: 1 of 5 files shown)


kitsune/wiki/api.py
```diff
@@ -19,10 +19,13 @@ class Meta:
...
 class DocumentDetailSerializer(DocumentShortSerializer):
+    summary = serializers.CharField(source='summary', read_only=True)
+    url = serializers.CharField(source='get_absolute_url', read_only=True)
+
     class Meta:
         model = Document
-        fields = ('id', 'title', 'slug', 'locale', 'products', 'topics',
-                  'html')
+        fields = ('id', 'title', 'slug', 'url', 'locale', 'products', 'topics',
+                  'summary', 'html')
```
willkg: I had to add the summary and url fields to this serializer so that it had all the bits the previous results had.

Does this affect other APIs adversely?

kitsune/search/api.py
```diff
@@ -101,9 +102,11 @@ def _question_suggestions(searcher, text, locale, product, max_results):
...
     searcher = _query(searcher, QuestionMappingType, search_filter, text, locale)
 
-    for result in searcher[:max_results]:
-        q = Question.objects.get(id=result['id'])
-        questions.append(QuestionSerializer(instance=q).data)
+    question_ids = [result['id'] for result in searcher]
+    questions = [
+        QuestionSerializer(instance=q).data
+        for q in Question.objects.filter(id__in=question_ids)
+    ]
```
mythmon: Oh good, I was hoping you would do this.

kitsune/wiki/api.py
mythmon: I don't know of anything that is sensitive to having too many fields from the serializer, so this should be perfectly fine.

mythmon:
This looks right to me. Thank you!

PR #2530, opened 2015-05-14
[Bug 1164972] Makes Kbox overlay on top of scrollmenu in SUMO.

anushbmx:
Makes the kbox model on top of scroll menu in SUMO.

Kbox has z-index of 98, so I change the z-index value of scroll menu to be 97

How is the menu now
screenshot from 2015-05-15 00 28 08

After Changing Z-index
screenshot from 2015-05-15 00 27 45


mythmon:
This looks good, thank you. And thanks for including the screenshots, that really helped.

PR #2532, opened 2015-05-15
[Bug 1159120] Dont show warning to the contributor

safwanrahman:
Bug 1159120 - Contributor should not see that they are working on a thread
This patch will make insure that if the user is working on the question, the Warning will not show
r?


mythmon:
Nice and easy.

PR #2533, opened 2015-05-18
Delete showfor demo

mythmon:
We used this to validate the new showfor. We don't need it anymore.


willkg:
I double checked to see if this was used or referred to anywhere and it's now. r+

PR #2534, opened 2015-05-18
[Bug 1166042] Some text is missing in mobile site

orvi2014:


mythmon:
We have tools that use a specific format for putting bugs in commits and titles. I've updated the title of this PR to match the expected format.

mythmon:
Your commit message doesn't really explain what the change is. It would be much better if the commit message was something like "Add missing image caption to AoA mobile page.". Can you fix that? It really helps when looking at the history of the code later.

To do that, you can use `git rebase -i HEAD~1`, and choose to reword this commit. Then you'll have to force push to your branch. That will automatically update the PR. After you do that, make sure to comment in the PR, because just changing the commits won't notify anyone.

willkg:
You can do `git commit --amend`, too.

mythmon:
Ooh, right. `git commit --amend` is the better choice.

Anyways, I tested this out, and it doesn't look good, since the text begins immediately after the image, instead of on the next line like in the desktop version. I think you have some more work to do here.
screen shot 2015-05-18 at 13 42 53

orvi2014:
Okay I fix that issue soon . Thanks anyway .

PR #2535, opened 2015-05-18
Fix remove button on AoA tweets loaded through infinite scroll.

mythmon:
The `authed` variable is used in the template to make sure that the user should be allowed to see the button, but it wasn't being included in the infinite scroll view.

r?

### DIFF
--- a/kitsune/customercare/views.py
+++ b/kitsune/customercare/views.py
@@ -113,6 +113,7 @@ def _count_answered_tweets(since=None):
     return q.count()
 
 
+# @ssl_required
 @require_GET
 def more_tweets(request):
     """AJAX view returning a list of tweets."""
--- a/kitsune/customercare/views.py
+++ b/kitsune/customercare/views.py
@@ -126,7 +126,9 @@ def more_tweets(request):
             locale=request.LANGUAGE_CODE,
             max_id=max_id,
             filter=filter,
-            https=request.is_secure())})
+            https=request.is_secure()),
+        'authed': request.user.is_authenticated() and request.twitter.authed,
+    })
 
 
 @ssl_required
--- a/kitsune/customercare/tests/test_views.py
+++ b/kitsune/customercare/tests/test_views.py
@@ -6,11 +6,12 @@ from django.test.client import RequestFactory
 
 from mock import patch, Mock
 from nose.tools import eq_
+from pyquery import PyQuery as pq
 
 from kitsune.customercare.models import Tweet, Reply
 from kitsune.customercare.tests import tweet, twitter_account, reply
-from kitsune.customercare.views import _get_tweets, _count_answered_tweets
-from kitsune.customercare.views import twitter_post
+from kitsune.customercare.views import (
+    _get_tweets, _count_answered_tweets, more_tweets, twitter_post)
 from kitsune.sumo.tests import TestCase, LocalizingClient
 from kitsune.sumo.urlresolvers import reverse
 from kitsune.users.tests import user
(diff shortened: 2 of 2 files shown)


kitsune/customercare/views.py
```diff
@@ -126,7 +126,9 @@ def more_tweets(request):
             locale=request.LANGUAGE_CODE,
             max_id=max_id,
             filter=filter,
-            https=request.is_secure())})
+            https=request.is_secure()),
+        'authed': request.user.is_authenticated() and request.twitter.authed,
+    })
```
willkg: Why doesn't this function have the `@ssl_required` decorator?

willkg:
Also, should this have a test?

kitsune/customercare/views.py
mythmon: Why would it?

willkg: The other page that does this sort of thing does.

mythmon:
This probably should have a test. I worked a bit on writing one, but I'm stuck. I've pushed the not-yet-working test. I am getting stuck because I need to add certain properties to the request, but using `RequestFactory` gets me a request that can't call `gettext` in templates. I'm guessing it lacks a Jinja context? Any idea how to modify it to actually test the function?

mythmon:
Huh. I don't get the same error locally as Travis does. Locally I get

```
  File "/home/mythmon/src/kitsune/kitsune/customercare/tests/test_views.py", line 124, in test_remove_in_infinite_scroll
    res = more_tweets(req)
  File "/home/mythmon/src/kitsune/venv/lib/python2.7/site-packages/django/views/decorators/http.py", line 41, in inner
    return func(request, *args, **kwargs)
  File "/home/mythmon/src/kitsune/kitsune/customercare/views.py", line 131, in more_tweets
    'authed': request.user.is_authenticated() and request.twitter.authed,
  File "/home/mythmon/src/kitsune/venv/lib/python2.7/site-packages/django/shortcuts.py", line 50, in render
    return HttpResponse(loader.render_to_string(*args, **kwargs),
  File "/home/mythmon/src/kitsune/venv/lib/python2.7/site-packages/django/template/loader.py", line 178, in render_to_string
    return t.render(context_instance)
  File "/home/mythmon/src/kitsune/venv/lib/python2.7/site-packages/jingo/__init__.py", line 195, in render
    return super(Template, self).render(context_dict)
  File "/home/mythmon/src/kitsune/venv/lib/python2.7/site-packages/jinja2/environment.py", line 891, in render
    return self.environment.handle_exception(exc_info, True)
  File "/home/mythmon/src/kitsune/kitsune/customercare/templates/customercare/tweets.html", line 18, in top-level template code
    <span class="reply-button">{{ _('Reply') }}</span>
  File "/home/mythmon/src/kitsune/venv/lib/python2.7/site-packages/tower/template.py", line 10, in _gettext_alias
    context.resolve('gettext')(string, *args, **kw))
UndefinedError: 'gettext' is undefined
```

mythmon:
This turned out to be related to some sort of shared state in this test class. Running the test in isolation would break, but running the entire class worked fine.

So I moved the test to a new class instead, to avoid all the problem of this other weird class.

willkg:
Travis is failing with this error:

```
ERROR: kitsune/customercare/tests/test_views.py:MoreTweetsTests.test_remove_in_infinite_scroll
  File "/home/travis/build/mozilla/kitsune/kitsune/customercare/tests/test_views.py", line 120, in test_remove_in_infinite_scroll
    {'max_id': tw.id + 1})
AttributeError: 'Tweet' object has no attribute 'id'
```

I think that `tw.id` needs to be `tw.tweet_id`. That gets you to the next error where `MoreTweetTests` doesn't have a `user` property.

willkg:
This PR is super old. Is it in good enough condition to update or should we can it?

rehandalal:
my vote is to close it since it probably needs some rebase love and also because if it still hasn't landed we probably don't care that much.

mythmon:
Yeah, this commit is too old. Sorry for letting it die.

We should fix this at some point, because it prevents moderation on old tweets. I'll file a bug for it.

PR #2540, opened 2015-05-20
[bug 1157477] Filter search by product from product page

rehandalal:
r?


mythmon:
Looks simple enough. r+

PR #2543, opened 2015-05-21
Fix related products link

rehandalal:
oops! tiny r?


mythmon:
Oooops, r+ when Travis goes green

PR #2547, opened 2015-05-22
[Bug 1166090] Document template properties validation 

mythmon:
This makes some important, but hopefully pretty straight forward changes to the Document model, and then changes a whole lot of tests to account for the model changes.

* Added FactoryBoy powered model factories. These have some niceties over model makers, and Fjord uses them too.
* Enforce that documents with a title like "Template:Foo" are in the Template category, and vice-versa.
* Update all the tests to be ok with that, generally by switching them from model makers to model Factories.
* Make renaming a template in such a way that it is no longer a template not create redirects. The automatic redirect creation logic doesn't work when templates are required to have "Template:" prefixes *and* be in the right category.

To test:

* Try and rename a template to not have the prefix (without changing the category). It should fail with a nice error message. (There is a test for this)
* Try and change the category of a template to something else (without changing the title). It should fail. (There is a test for this)
* Try and do both at once. It should work.  (There is a test for this)
* Try and do this for a localized article. This is where most of our real problems are. (I need to add a test for this).

TODO

I realized I have a few more things to do before I want to land this. But it should be ready to start review now.

* [x] Add test for localized articles.
* [x] Add tests for normal article -> templates
* [x] ~~Figure out if landing this on a DB full of title/category inconsistencies is going to be a bad idea. I suspect it will only be a problem if the document is edited, in which case it will give a nice error to a human. But maybe not? What else calls .save() on documents? Hunt these down.~~
* [x] Wait for the templates on the site to be renamed so landing this won't break everything.

(This isn't ready to merge yet, because of the above)

f?

### DIFF
--- a/kitsune/dashboards/tests/test_readouts.py
+++ b/kitsune/dashboards/tests/test_readouts.py
@@ -595,7 +595,6 @@ class TemplateTranslationsTests(ReadoutTestCase):
         """Test the product filtering of the readout."""
         p = product(title='Firefox', slug='firefox', save=True)
         d = TemplateDocumentFactory()
-        d = document(title='Template:test', category=TEMPLATES_CATEGORY, save=True)
         ApprovedRevisionFactory(document=d, is_ready_for_localization=True)
 
         # There shouldn't be any rows yet.
--- a/kitsune/dashboards/tests/test_readouts.py
+++ b/kitsune/dashboards/tests/test_readouts.py
@@ -90,7 +90,7 @@ class L10NOverviewTests(TestCase):
     def test_counting_unready_docs(self):
         """Docs without a ready-for-l10n rev shouldn't count in total."""
         # Make a doc with an approved but not-ready-for-l10n rev:
-        d = TemplateDocumentFactory(is_localizable=True)
+        d = DocumentFactory(is_localizable=True)
         r = ApprovedRevisionFactory(document=d, is_ready_for_localization=False)
 
         # It shouldn't show up in the total:
--- a/kitsune/dashboards/tests/test_readouts.py
+++ b/kitsune/dashboards/tests/test_readouts.py
@@ -17,10 +17,11 @@ from kitsune.dashboards.readouts import (
 from kitsune.products.tests import product
 from kitsune.sumo.tests import TestCase
 from kitsune.wiki.config import (
-    MAJOR_SIGNIFICANCE, MEDIUM_SIGNIFICANCE, TYPO_SIGNIFICANCE,
-    HOW_TO_CONTRIBUTE_CATEGORY, ADMINISTRATION_CATEGORY,
-    CANNED_RESPONSES_CATEGORY)
-from kitsune.wiki.tests import revision, translated_revision, document
+    MAJOR_SIGNIFICANCE, MEDIUM_SIGNIFICANCE, TYPO_SIGNIFICANCE, HOW_TO_CONTRIBUTE_CATEGORY,
+    ADMINISTRATION_CATEGORY, CANNED_RESPONSES_CATEGORY, TEMPLATES_CATEGORY)
+from kitsune.wiki.tests import (
+    revision, translated_revision, document,
+    DocumentFactory, RevisionFactory, TemplateDocumentFactory, ApprovedRevisionFactory)
 
 
 class MockRequest(object):
@@ -76,13 +77,8 @@ class L10NOverviewTests(TestCase):
     def test_counting_unready_templates(self):
         """Templates without a ready-for-l10n rev don't count"""
         # Make a template with an approved but not-ready-for-l10n rev:
-        r = revision(document=document(title='Template:smoo',
-                                       is_localizable=True,
-                                       is_template=True,
-                                       save=True),
-                     is_ready_for_localization=False,
-                     is_approved=True,
-                     save=True)
+        d = TemplateDocumentFactory(is_localizable=True)
+        r = ApprovedRevisionFactory(document=d, is_ready_for_localization=False)
 
         # It shouldn't show up in the total:
         eq_(0, l10n_overview_rows('de')['templates']['denominator'])
@@ -94,12 +90,8 @@ class L10NOverviewTests(TestCase):
     def test_counting_unready_docs(self):
         """Docs without a ready-for-l10n rev shouldn't count in total."""
         # Make a doc with an approved but not-ready-for-l10n rev:
-        r = revision(document=document(title='smoo',
-                                       is_localizable=True,
-                                       save=True),
-                     is_ready_for_localization=False,
-                     is_approved=True,
-                     save=True)
+        d = TemplateDocumentFactory(is_localizable=True)
+        r = ApprovedRevisionFactory(document=d, is_ready_for_localization=False)
 
         # It shouldn't show up in the total:
         eq_(0, l10n_overview_rows('de')['all']['denominator'])
@@ -138,7 +130,9 @@ class L10NOverviewTests(TestCase):
         eq_(1, l10n_overview_rows('de')['all']['denominator'])
 
         t.document.parent.title = t.document.title = 'Template:thing'
-        t.document.parent.is_template = t.document.is_template = True
+        t.document.parent.category = TEMPLATES_CATEGORY
+        # is_template will be automatically set for both templates, and so will
+        # the child document's category.
         t.document.parent.save()
         t.document.save()
         # ...but not when it's a template:
@@ -229,6 +223,7 @@ class L10NOverviewTests(TestCase):
         # Update the parent and translation to be a template
         d = t.document.parent
         d.title = 'Template:Lorem Ipsum Dolor'
+        d.category = TEMPLATES_CATEGORY
         d.save()
         t.document.title = 'Template:Lorem Ipsum Dolor'
         t.document.save()
@@ -350,14 +345,10 @@ class TemplateTests(ReadoutTestCase):
         locale = settings.WIKI_DEFAULT_LANGUAGE
         p = product(title='Firefox', slug='firefox', save=True)
 
-        d = document(save=True)
-        t = document(title='Template:test', save=True)
-
-        revision(document=d, is_approved=True, save=True)
-        revision(document=t, is_approved=True, save=True)
-
-        d.products.add(p)
-        t.products.add(p)
+        d = DocumentFactory(products=[p])
+        t = TemplateDocumentFactory(products=[p])
+        ApprovedRevisionFactory(document=d)
+        ApprovedRevisionFactory(document=TemplateDocumentFactory())
 
         eq_(1, len(self.rows(locale=locale, product=p)))
         eq_(t.title, self.row(locale=locale, product=p)['title'])
@@ -366,8 +357,8 @@ class TemplateTests(ReadoutTestCase):
     def test_needs_changes(self):
         """Test status for article that needs changes"""
         locale = settings.WIKI_DEFAULT_LANGUAGE
-        d = document(title='Template:test', needs_change=True, save=True)
-        revision(document=d, is_approved=True, save=True)
+        d = TemplateDocumentFactory(needs_change=True)
+        ApprovedRevisionFactory(document=d)
 
         row = self.row(locale=locale)
 
@@ -377,8 +368,8 @@ class TemplateTests(ReadoutTestCase):
     def test_needs_review(self):
         """Test status for article that needs review"""
         locale = settings.WIKI_DEFAULT_LANGUAGE
-        d = document(title='Template:test', save=True)
-        revision(document=d, save=True)
+        d = TemplateDocumentFactory()
+        RevisionFactory(document=d)
 
         row = self.row(locale=locale)
 
@@ -388,10 +379,10 @@ class TemplateTests(ReadoutTestCase):
     def test_unready_for_l10n(self):
         """Test status for article that is not ready for l10n"""
         locale = settings.WIKI_DEFAULT_LANGUAGE
-        d = document(title='Template:test', save=True)
-        revision(document=d, is_ready_for_localization=True, save=True)
-        revision(document=d, is_ready_for_localization=False, is_approved=True,
-                 significance=MAJOR_SIGNIFICANCE, save=True)
+        d = TemplateDocumentFactory()
+        RevisionFactory(document=d, is_ready_for_localization=True)
+        ApprovedRevisionFactory(
+            document=d, is_ready_for_localization=False, significance=MAJOR_SIGNIFICANCE)
 
         row = self.row(locale=locale)
 
@@ -594,11 +585,8 @@ class TemplateTranslationsTests(ReadoutTestCase):
 
     def test_untranslated(self):
         """Assert untranslated templates are labeled as such."""
-        d = document(title='Template:test', save=True)
-        untranslated = revision(is_approved=True,
-                                is_ready_for_localization=True,
-                                document=d,
-                                save=True)
+        d = TemplateDocumentFactory()
+        untranslated = ApprovedRevisionFactory(document=d, is_ready_for_localization=True)
         row = self.row()
         eq_(row['title'], untranslated.document.title)
         eq_(unicode(row['status']), 'Translation Needed')
@@ -606,11 +594,9 @@ class TemplateTranslationsTests(ReadoutTestCase):
     def test_by_product(self):
         """Test the product filtering of the readout."""
         p = product(title='Firefox', slug='firefox', save=True)
-        d = document(title='Template:test', save=True)
-        revision(is_approved=True,
-                 is_ready_for_localization=True,
-                 document=d,
-                 save=True)
+        d = TemplateDocumentFactory()
+        d = document(title='Template:test', category=TEMPLATES_CATEGORY, save=True)
+        ApprovedRevisionFactory(document=d, is_ready_for_localization=True)
 
         # There shouldn't be any rows yet.
         eq_(0, len(self.rows(product=p)))
--- a/kitsune/sumo/tests/test_parser.py
+++ b/kitsune/sumo/tests/test_parser.py
@@ -23,9 +23,9 @@ def pq_img(p, text, selector='img', locale=settings.WIKI_DEFAULT_LANGUAGE):
     return doc(selector)
 
 
-def doc_rev_parser(content, title='Installing Firefox', parser_cls=WikiParser):
+def doc_rev_parser(content, title='Installing Firefox', parser_cls=WikiParser, **kwargs):
     p = parser_cls()
-    d = document(title=title)
+    d = document(title=title, **kwargs)
     d.save()
     r = revision(document=d, content=content, is_approved=True)
     r.save()
--- a/kitsune/wiki/models.py
+++ b/kitsune/wiki/models.py
@@ -329,16 +329,17 @@ class Document(NotificationsMixin, ModelBase, BigVocabTaggableMixin,
             if name in ('slug', 'title') and hasattr(self, name):
                 old_name = 'old_' + name
                 if not hasattr(self, old_name):
-                    prefix = TEMPLATE_TITLE_PREFIX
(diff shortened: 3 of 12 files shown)


kitsune/dashboards/tests/test_readouts.py
```diff
@@ -17,10 +17,11 @@
...
 from kitsune.wiki.config import (
-    MAJOR_SIGNIFICANCE, MEDIUM_SIGNIFICANCE, TYPO_SIGNIFICANCE,
-    HOW_TO_CONTRIBUTE_CATEGORY, ADMINISTRATION_CATEGORY,
-    CANNED_RESPONSES_CATEGORY)
-from kitsune.wiki.tests import revision, translated_revision, document
+    MAJOR_SIGNIFICANCE, MEDIUM_SIGNIFICANCE, TYPO_SIGNIFICANCE, HOW_TO_CONTRIBUTE_CATEGORY,
+    ADMINISTRATION_CATEGORY, CANNED_RESPONSES_CATEGORY, TEMPLATES_CATEGORY)
+from kitsune.wiki.tests import (
+    revision, translated_revision, document,
+    DocumentFactory, RevisionFactory, TemplateDocumentFactory, ApprovedRevisionFactory)
```
rehandalal: Yay!

kitsune/sumo/tests/test_parser.py
```diff
@@ -23,9 +23,9 @@ def pq_img(p, text, selector='img', locale=settings.WIKI_DEFAULT_LANGUAGE):
     return doc(selector)
 
 
-def doc_rev_parser(content, title='Installing Firefox', parser_cls=WikiParser):
+def doc_rev_parser(content, title='Installing Firefox', parser_cls=WikiParser, **kwargs):
     p = parser_cls()
-    d = document(title=title)
+    d = document(title=title, **kwargs)
```
rehandalal: Shouldn't this also use the new factories?

willkg:
If there are known "inconsistency" situations, then is it possible to do one or more data migrations to fix them? e.g. any document that's in the Template category that doesn't have "Template: " gets a "Template: ".

If we're not going to do that, then we probably need to build a list of "bad templates" and what needs to be fixed with them so that someone can do it by hand.

mythmon:
The list of bad templates is in [Bug 1012384 comment 13](https://bugzilla.mozilla.org/show_bug.cgi?id=1012384#c13).

I thought about doing an automatic migration for all of them, but due to the problems with renaming templates and redirects (outlined above, and in the code), I fear we might break a lot of articles. I think that renaming the problem templates will require also rewriting articles that use them. Next week I'm going to do some investigation to see how bad things would be if this landed without an automatic migration.

willkg:
Makes sense.

willkg:
I think one thing I'd do is write up a bug for switching to factory boy and then moving that work to that bug and landing it separately from all this other stuff. That'd reduce the work for this review and make it easier.

mythmon:
Moving the FactoryBoy work out of this PR significantly fewer lines of code. It would remove the factories themselves from the review. However, all the tests that were changed to use the factories needed to be changed anyways to handle templates correctly.

If you think that using the factories instead of using model makers would make this PR easier to review even at the same size, I could go and rewrite those tests again to use model makers. I don't think it is worth the effort though.

willkg:
So, I try to reduce the work it takes to review something where possible because the less there is to review, the easier it is to review and the more likely the review catches issues.

Having said that, I don't know the details here and I'm not going to be reviewing this. So either way is fine by me.

kitsune/sumo/tests/test_parser.py
mythmon: I started to refactor this function, but it ended up having too much using it. Instead I tried to make as few changes to it as possible.

mythmon:
I fixed the tests, kind of. Now we are hitting a problem that #2551 fixes. Namely we are actually pulling down a different version of autoprefixer than we asked for, and the newer version changed some things. We are using a feature that is now deprecated. It prints a deprecation warning on stderr. Django Pipeline assumes that any output at all on stderr is a fatal error, and so it fails the build.

mythmon:
Rebased on top of master to pick up more Travis fixes.

mythmon:
I rebased this on the current master, and tidied it up a bit. I think it is ready to land, and Kadir tells me that the manual work with the templates is ready for t his to land too.

r?

willkg:
I'll look at this tonight or tomorrow. If I don't, feel free to send me nasty email.

kitsune/dashboards/tests/test_readouts.py
```diff
@@ -606,11 +594,9 @@ def test_untranslated(self):
     def test_by_product(self):
         """Test the product filtering of the readout."""
         p = product(title='Firefox', slug='firefox', save=True)
-        d = document(title='Template:test', save=True)
-        revision(is_approved=True,
-                 is_ready_for_localization=True,
-                 document=d,
-                 save=True)
+        d = TemplateDocumentFactory()
+        d = document(title='Template:test', category=TEMPLATES_CATEGORY, save=True)
```
willkg: The previous line assigns a value to d, then this line assigns another value to d.

I mention that in case the first line is superfluous. If it is, you can ditch the line. If it's not, then ditch the `d =`, but probably worth adding a comment about why you're creating two since the test seems to only be about one.

kitsune/wiki/models.py
```diff
@@ -209,15 +210,30 @@ def _ensure_inherited_attr(self, attr):
...
+        if (self.category == TEMPLATES_CATEGORY and
+                not self.title.startswith(TEMPLATE_TITLE_PREFIX)):
+            # L10n: Note, do not translate "Template:" in this string.
+            raise ValidationError(_(u'Documents in the Template category must have titles that '
+                                    u'start with "Template:". (Current title is "{}")'
+                                    .format(self.title)))
+
+        if self.title.startswith(TEMPLATE_TITLE_PREFIX) and self.category != TEMPLATES_CATEGORY:
+            # L10n: Note, do not translate "Template:" in this string.
+            raise ValidationError(_(u'Documents with titles that start with "Template:" must be '
```
willkg: It's probably less of a problem (though vaguely goofy) to do this:

```
raise ValidationError(_(u'Documents with titles that start with {prefix} must be '
                        u'in the templates category. (Current category is "{category}")'
                        .format(prefix='Template:', category=self.get_category_display())))
```

Then you don't have to have the l10n comment and you don't have to worry about people not understanding it and/or not reading it.

willkg: Plus if for some bizarro reason, we decide to change the prefix, we don't then also have to change all the strings.

willkg: Even better, use `TEMPLATE_TITLE_PREFIX`.

kitsune/wiki/models.py
```diff
@@ -263,39 +279,53 @@ def save(self, *args, **kwargs):
...
         slug_changed = hasattr(self, 'old_slug')
         title_changed = hasattr(self, 'old_title')
 
-        # If the slug changed, we clear out the share link so it gets regenerated.
-        self.share_link = ''
+        templateness_changed = (
+            title_changed and
+            # If the title was a template before, and is not now, or vice-versa
+            (self.title.startswith(TEMPLATE_TITLE_PREFIX) !=
+             self.old_title.startswith(TEMPLATE_TITLE_PREFIX)))
```
willkg: `old_slug` and `old_title` are persisted after a case-insensitive check. But this checks to see if the title starts with the prefix and is case-sensitive. So if someone changes the document from "Template: blah" to "template: blah", I think this code and some of the other template-prefix-related code won't work right.

kitsune/wiki/models.py
```diff
@@ -263,39 +279,53 @@ def save(self, *args, **kwargs):
...
         # changed. Allowing redirects for unapproved docs would (1) be of
         # limited use and (2) require making Revision.creator nullable.
-        if self.current_revision and (slug_changed or title_changed):
+        #
+        # If the "templateness" changed, don't try and make a redirect.
+        # The category and the title-prefix have to stay consistent, and
+        # redirects are in the same category as their targets. There is not
+        # a clear way to construct the redirect without invalidating one of
+        # those conditions.
+        if self.current_revision and (slug_changed or title_changed) and not templateness_changed:
```
willkg: I'm not following this at all. It seems like the additional `not templateness_changed` only affects the case where the document got switched to a template or the template got switched to a document. Is that what you want? Do we want redirects for templates at all?

kitsune/wiki/models.py
```diff
@@ -263,39 +279,53 @@ def save(self, *args, **kwargs):
...
                 pass
 
-            if slug_changed:
-                del self.old_slug
-            if title_changed:
-                del self.old_title
+        if slug_changed:
+            del self.old_slug
+        if title_changed:
+            del self.old_title
```
willkg: These were dedented one level. Why?

kitsune/wiki/tests/test_models.py
```diff
@@ -313,6 +314,71 @@ def test_get_products(self):
...
+        # Change the parent
+        d_en.title = 'Template:' + d_en.title
+        d_en.category = TEMPLATES_CATEGORY
+        d_en.save()
+
+        # Now the French article can be changed too.
+        d_fr = Document.objects.get(id=d_fr.id)  # reset
+        d_fr.title = 'Template:' + d_fr.title
+        d_fr.category = TEMPLATES_CATEGORY
+        d_fr.save()
```
willkg: Do we need the additional test where if `d_en` and `d_fr` are templates and then someone changes the title and category of `d_fr` to a regular document, that raises an error? (Assuming that raises an error--I have no idea if that's valid or not.)

kitsune/wiki/tests/test_parser.py
```diff
@@ -21,9 +22,9 @@ def doc_rev_parser(*args, **kwargs):
         *args, parser_cls=WikiParser, **kwargs)
 
 
-def doc_parse_markup(content, markup, title='Template:test'):
+def doc_parse_markup(content, markup):
     """Create a doc with given content and parse given markup."""
-    _, _, p = doc_rev_parser(content, title)
+    _, _, p = doc_rev_parser(content, 'Template:test', category=TEMPLATES_CATEGORY)
```
willkg: Probably worth using `TEMPLATE_PREFIX_STRING` here.

kitsune/wiki/tests/test_parser.py
```diff
@@ -191,14 +192,12 @@ def test_template_does_not_exist(self):
...
         """Localized template is returned."""
         py_doc, p = doc_parse_markup('English content', '[[Template:test]]')
-        parent = document()
-        d = document(parent=parent, title='Template:test', locale='fr')
-        d.save()
-        r = revision(content='French content', document=d, is_approved=True)
-        r.save()
-        eq_('English content', py_doc.text())
+        parent = TemplateDocumentFactory()
+        d = TemplateDocumentFactory(parent=parent, title='Template:test', locale='fr')
```
willkg: Probably worth using the `TEMPLATE_PREFIX_STRING` here, too.

kitsune/wiki/tests/test_parser.py
```diff
@@ -297,7 +296,7 @@ def test_build_template_params_named_anonymous_numbered(self):
         eq_({'1': 'a', 'hi': 'test', '3': 'z'}, _btp(['hi=test', 'a', '3=z']))
 
     def test_unapproved_template(self):
-        document(title='Template:new').save()
+        TemplateDocumentFactory(title='Template:new')
```
willkg: It might be worth tweaking `TemplateDocumentFactory` so that it prefixes the title with `Template:` if it's not there already.

kitsune/wiki/tests/test_parser.py
```diff
@@ -339,30 +334,23 @@ def test_button_image_for_nesting(self):
...
-        d = document(title='Template:Boo')
-        d.save()
+        d = TemplateDocumentFactory(title='Template:Boo')
 
         # Twice so the second revision sees content identical to itself:
-        for i in range(2):
-            revision(document=d, content='Fine [[Template:Boo]] Fellows',
-                     is_approved=True).save()
+        for _ in range(2):
+            ApprovedRevisionFactory(document=d, content='Fine [[Template:Boo]] Fellows')
```
willkg: I think you can do this with `.build_batch()` if you wanted to.

https://factoryboy.readthedocs.org/en/latest/reference.html#factory.build_batch

kitsune/wiki/tests/test_parser.py
```diff
@@ -305,12 +304,8 @@ def test_unapproved_template(self):
 
     def test_for_in_template(self):
         """Verify that {for}'s render correctly in template."""
-        d = document(title='Template:for')
-        d.save()
-        r = revision(document=d,
-                     content='{for win}windows{/for}{for mac}mac{/for}')
-        r.is_approved = True
-        r.save()
+        d = TemplateDocumentFactory(title='Template:for')
```
willkg: Probably worth using the `TEMPLATE_PREFIX_STRING` here, too.

kitsune/wiki/tests/test_parser.py
```diff
@@ -339,30 +334,23 @@ def test_button_image_for_nesting(self):
 
     def test_direct_recursion(self):
         """Make sure direct recursion is caught on the very first nesting."""
-        d = document(title='Template:Boo')
-        d.save()
+        d = TemplateDocumentFactory(title='Template:Boo')
```
willkg: Probably worth using the `TEMPLATE_PREFIX_STRING` here, too.

kitsune/wiki/tests/test_parser.py
```diff
@@ -339,30 +334,23 @@ def test_button_image_for_nesting(self):
...
-        boo = document(title='Template:Boo')
-        boo.save()
-        yah = document(title='Template:Yah')
-        yah.save()
-        revision(document=boo, content='Paper [[Template:Yah]] Cups',
-                 is_approved=True).save()
-        revision(document=yah, content='Wooden [[Template:Boo]] Bats',
-                 is_approved=True).save()
+        boo = TemplateDocumentFactory(title='Template:Boo')
+        yah = TemplateDocumentFactory(title='Template:Yah')
```
willkg: Probably worth using the `TEMPLATE_PREFIX_STRING` here, too.

kitsune/wiki/tests/test_parser.py
```diff
@@ -339,30 +334,23 @@ def test_button_image_for_nesting(self):
...
+        ApprovedRevisionFactory(document=boo, content='Paper [[Template:Yah]] Cups')
+        ApprovedRevisionFactory(document=yah, content='Wooden [[Template:Boo]] Bats')
```
willkg: Probably worth using the `TEMPLATE_PREFIX_STRING` here, too.

kitsune/wiki/tests/test_parser.py
```diff
@@ -840,7 +828,7 @@ def test_links(self):
         eq_([d.linked_from.title for d in d2.links_to()], ['D3'])
 
     def test_templates(self):
-        d1, _, _ = doc_rev_parser('Oh hai', title='Template:D1')
+        d1, _, _ = doc_rev_parser('Oh hai', title='Template:D1', category=TEMPLATES_CATEGORY)
         d2, _, _ = doc_rev_parser('[[Template:D1]]', title='D2')
```
willkg: Probably worth using the `TEMPLATE_PREFIX_STRING` here, too.

I'll stop marking this. Point being that we've got a constant so we should use it to make life easier and reduce likelihood of spelling mistakes and all that.

willkg:
I'm done this review pass. I'll keep tabs on it and try to help get it landed today.

kitsune/wiki/models.py
mythmon: A big +1 to this and all the other places we should use the constant instead of a string.

kitsune/wiki/models.py
mythmon: Nice catch. I'm not terrible sure what to do here. My two thoughts are to either

1. make the `old_slug` and `old_title` checks no longer case insensitive.
2. Add special handling to the part that sets `old_title` to watch for case changes in the template prefix.

Do you have opinions on either of these? I'm not sure why it was case insensitive to being with, so I'm not sure if changing that would be a good idea. It would be simpler though.

kitsune/wiki/models.py
mythmon: We had redirects for templates before. I wanted to keep that behavior (to minimize change) where it made sense. It doesn't make sense to make the redirects when documents switch to templates or vice versa (for the reasons explained in the comment).

Thinking about this more though, I don't think having a template redirect would actually do anything useful. The template code doesn't follow redirects, so at best you would get some strange content in the article that includes the template. I think I'll change this to simply never add redirects for templates, which will simplify some things. Does that sound right?

kitsune/wiki/models.py
mythmon: If a document became a template through a title change, it would have a `self.old_title` that needs cleaned up. But because of the `not templateness_changed` check in the previous if block, there would be cases where these attributes were created  but not cleaned up. Dedenting them makes sure they always get cleaned up.

kitsune/wiki/tests/test_models.py
mythmon: I tried to unravel this enough to make a test case, but I didn't get very far. Then I realized that the UI for editing translation articles doesn't actually allow people to change the category of a translation document.

This is a admittedly leaving a hole in the test coverage for a case that could in theory happen. I don't think I have any good answers for what should happen here. I'd rather leave it undefined instead of codifying a possibly confusing behavior. Does that sound like a mistake?

kitsune/wiki/tests/test_parser.py
mythmon: It is there already, this was just sloppy on my part.

mythmon: Er, nevermind. It doesn't do that already. It only picks a good default value that includes the prefix. I'll modify it to add the prefix.

kitsune/wiki/tests/test_models.py
willkg: That sounds fine. If you think future you might be puzzled, then it's worth
adding a comment.

kitsune/wiki/models.py
willkg: Awesome!

kitsune/wiki/models.py
willkg: I don't know this stuff well enough to know whether there are issues with
that plan, but it sounds fine as far as I can tell.

Is there anyone you can talk to whip might have a better idea? If not, I
vote go with it and see whether the tests show any issues.

kitsune/wiki/models.py
willkg: I didn't look at it enough to know why being case insensitive helped.

I think I would add some code to the old_title case to cover the template
thing. That seems least invasive.

mythmon:
Feedbacked ^

kitsune/wiki/tests/test_parser.py
mythmon: And I've come full circle. I tried to make the factory always ensure there was the template prefix in place, but I can't figure out how to do that and also have a sequence-driven default value. I'd love to try and figure that out, but for now I'll push it off.

willkg:
I think these are still outstanding:

* https://github.com/mozilla/kitsune/pull/2547#discussion_r35366848
* https://github.com/mozilla/kitsune/pull/2547#discussion_r35364001

If they're accounted for already, I'm not sure how.

mythmon:
Those two templates were not covered at the time you commented, but should be covered now. I'll refactor that nasty if statement, and add a comment.

mythmon:
Done.

willkg:
This looks good to me as far as I can tell. The "merge" button is grey, so I suspect you'll need to tweak it some more after you fix whatever the merge issue is.

mythmon:
I merged this manually on the cli. There was an easy merge conflict in the imports of `kitsune/dashboards/tests/test_readouts.py`.

PR #2549, opened 2015-05-25
[bug 1058747] Restyle wiki translation preview

rehandalal:
r?


rlr:
changes look sane to me. I'll give it a spin locally.

rlr:
that's way better :+1:

PR #2550, opened 2015-05-27
Move static files into app namespaces.

mythmon:
We are starting to add more and more JS. I think it is a good idea to add app-specific JS in the app directories, to keep everything organized and easy to find. These app-specific static directories all get merged into one during collectstatic. That means there is the possibility for collisions. As far as I could tell, the accepted solution to this is to namespace everything in `app/static/app/js/foo.js`. Just like how we do templates.

This PR consists of two commits. The first simply moves all the files from point A to point B. It is a huge number of files, and GitHub isn't happy with that. The second commit updates settings, paths in templates, css urls, qunit configs, etc. to match the new location of the static files. There was a lot of search/replace here.

If you view that second diff locally, I have an alias `git worddiff` that expands to `git diff --color-words --word-diff-regex='[A-z0-9_\\-]+|[^[:space:]]` that I found quite helpful. It highlights intra-line changes well, and this patch has a lot of those.

I tested this by running the tests, running collectstatic, double checking qunit, poking around the site a bit while watching my network monitor, and running in DEBUG=False mode by running Kitsune in gunicorn and serving static files separately.


rlr:
I see you moved less files around. I thought we had tried that before and it broke things like `@import 'includes/details.less';` because files were no longer contained to one directory. I guess this is no longer a problem with our new django-pipeline setup?

mythmon:
Yes, since pipeline moves everything to the static directory before building, as long as we change it to use more-qualified paths like `@import '../../sumo/includes/details.less';` it will work fine.

It would be great to move things to the app they belong to, though we can do that later.

Oh, and the jenkins test failure is related to #2551. I'll rebase this once that gets landed.

willkg:
Haha--you said jenkins.

mythmon:
i lied

BReduardokramer:
Oi.  Quais os comentários para eu ajudar?
Em 27/05/2015 18:52, "Mike Cooper" <notifications@github.com> escreveu:

BReduardokramer:
Aonde vocês estão?

mythmon:
@BReduardokramer Development for kitsune uses English for communication, sorry. I used google translate to translate your message, and it seems like you are asking how you can help here. There isn't a lot that you can do here. This is mostly between me and @rlr. If you want to help Mozilla, I suggest checking out http://www.whatcanidoformozilla.org/

I rebased this branch on master to pick up the changes to CI. Travis should pass now.

mythmon:
We have other CI related problems now. I'd really like to get this merged in soon, because otherwise it is going to have too many conflicts. I don't think we should block this PR on Travis anymore (though we should run tests locally in that case). If you look at CI, this is only failing the linting, not the actual test suite.

r?

mythmon:
Rebased to fix merge conflicts. This should be ready to go now. I look forward to our new well organized future.

rehandalal:
travis has smiled on this. r+

PR #2551, opened 2015-05-27
Make sure Travis is using npm lockdown

mythmon:
The test failure in #2550 makes me think Travis isn't using lockdown correctly, and we are hiding that from the install process. This will make it a little easier to debug.

This is mainly to get Travis to run, it isn't for review just yet.

### DIFF
--- a/lockdown.json
+++ b/lockdown.json
@@ -4,14 +4,14 @@
   },
   "JSONStream": {
     "0.10.0": "74349d0d89522b71f30f0a03ff9bd20ca6f12ac0",
-    "0.7.4": "734290e41511eea7c2cfe151fbf9a563a97b9786",
-    "0.8.4": "91657dfe6ff857483066132b4618b62e8f4887bd"
+    "1.0.4": "aa86d6a89c8e77f206cd542581b563e7eee32459"
   },
   "abbrev": {
-    "1.0.5": "5d8257bd9ebe435e698b2fa431afde4fe7b10b03"
+    "1.0.5": "5d8257bd9ebe435e698b2fa431afde4fe7b10b03",
+    "1.0.6": "b6d632b859b3fa2d6f7e4b195472461b9e32dc30"
   },
   "acorn": {
-    "0.9.0": "67728e0acad6cc61dfb901c121837694db5b926b"
+    "1.2.1": "ae46ed295eea4e175a376634a8825eb6710f8058"
   },
   "acorn-babel": {
     "0.11.1-38": "6cf7b0c0791a11a05ef6020bf06ef34b30086768"
@@ -20,6 +20,7 @@
     "0.1.0": "3ca9735cf1dde0edf7a4bf6641709c8024f9b227"
   },
   "ansi-regex": {
+    "0.1.0": "55ca60db6900857c423ae9297980026f941ed903",
     "0.2.1": "0d8e946967a3d8143f93e24e298525fc1b2235f9",
     "1.1.1": "41c847194646375e6a1a5d10c3ca054ef9fc980d"
   },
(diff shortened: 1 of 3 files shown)


mythmon:
Alright, I've got it all figured out. When we upgraded Bower to 1.4.1, very sanely we removed the hashes for Bower 1.3.12. That makes sense. However, debowerify depends on a version of bower >=1.3 and <1.4. That's ok though, since this is Node land. Node just installs two copies, one for debowerify and one for us. So when we removed the hashes from lockdown.json for the old version of bower, that was bad.

So why didn't CI and production freak out? Well, it turns out Travis hasn't really been using Lockdown, as far as I can tell, and I don't think production is either. I've fixed it for Travis here, but more work needs down on prod.

So this PR does a few things:

* Reintroduced the old  hashes (and probably some new hashes, I can't really tell).
* Makes CI tell us more.
* Makes CI actually use lockdown explicitly.

Why didn't I fix this for production yet? We pass extra command line args to npm install on the server. We have to. When using lockdown explicitly like this, we don't have a way to pass those extra arguments. I suspect I'll be patching lockdown soon to deal with this.

So... r?

lockdown.json
```diff
@@ -134,23 +135,22 @@
...
-    "4.0.1": "7f4ab2b3a11c36a9274141fb7912122fc974e5f6"
+    "4.0.4": "8dae95a20ca43b3fea201faa6cfaa84ff4a0d484"
   },
   "browser-resolve": {
-    "1.8.2": "a00b06c871bee21b50d450ff94c2a423bc787729"
+    "1.9.0": "a3ab60185582063f8317fe48a1380f2ea81425ec"
   },
   "browserify": {
-    "9.0.4": "e58da3d5fe415bafa5734d1cb8e8c9d0ddb27d43"
+    "9.0.3": "f2f742b82ec5631c64b8c98a9788db0017c6517c"
```
willkg: 9.0.4 to 9.0.3? Why the down-grade?

lockdown.json
```diff
@@ -691,17 +705,16 @@
     "0.2.0": "6bc028149440e570d495ab509692aa08bd779c6e"
   },
   "lockdown": {
-    "0.0.8-dev": "075c0c1852e7cf746e4e85ccab8113b381f0b6e0"
+    "0.0.8-dev": "e6213cc58f76c4f7b62bf64189462367a87ba112"
```
willkg: Why'd the hash for this change?

scripts/travis/install.sh
```diff
@@ -16,20 +16,14 @@ if [[ $TEST_SUITE == "lint" ]]; then
...
 
 ./peep.sh install -r "requirements/default.txt"
-# Print the installed packages for the world to see.
-pip freeze
 echo
 
-
 echo "Installing Node.js dependencies"
-npm install > /dev/null 2> /dev/null
-npm list
```
willkg: Why ditch the "npm list" line?

scripts/travis/install.sh
```diff
@@ -16,20 +16,14 @@ if [[ $TEST_SUITE == "lint" ]]; then
...
+./scripts/lockdown.js
 echo
 
-
-echo "Intalling front end dependencies"
-./node_modules/.bin/bower install > /dev/null 2> /dev/null
-./node_modules/.bin/bower list -o
```
willkg: Why ditch the " ... list" line?

mythmon: When we can see the full output of the bower install command, I think it is redundant. Ditto for npm.

lockdown.json
mythmon: I'm not sure. I tried to avoid this change, but lockdown consistently insisted this was the hash. I suspect it is for the same reason that peep hashes for github zip files broke recently.

lockdown.json
mythmon: If you look at the package.json, you can see that we have been asking for `^9.0.3`, which means "anything that is sem-ver-close-enough to 9.0.3. When we did the lock the first time, it got the newest version, 9.0.4. I removed the `^` operator on the version, so it reverted back to 9.0.3.

mythmon:
@willkg any more comments here? Are my explanations sufficient, or do you think more changes are required? Or should someone else review this? I'd really like to get this merged soon, as it is blocking several things.

willkg:
I don't have any other opinions. I don't know enough about this to review it appropriately beyond the questions I had.

mythmon:
Ok. Thanks for the questions! @rehandalal said he was going to look at this. I assume he either didn't get to it, or thought you had it covered.

rehandalal:
r+

rehandalal:
We should probably have a bug for fixing this on prod, if we don't already.

PR #2553, opened 2015-06-01
[bug 1156856] Switch from ACE to CodeMirror

rehandalal:
Some notes:
- You probably want to use ?w=1 so that `markup.js` is a little easier to review.
- CodeMirror's own simple mode addon handles 99% of what we needed for the SUMO mode however I had to add the `condensed` flag for matching one of the cases.  I will submit a PR to CodeMirror to add that behavior, but I've included a slightly modified version with this pull that I will remove if/when my patch gets merged.

r?


mythmon:
The Travis failure here is because pre-commit and our requirements are conflicting. In particular, they are pulling in a newer version of flake8 than we are. I think we could just drop flake8 from our requirements.txt, since we only use it through pre-commit.

mythmon:
I tried this out. It seems to have the right highlighting, and work with the editing toolbar. It doesn't have suggestion or completion like the old one does, but that is what the other bug is for.

CodeMirror looks a lot harder to read to me, but I couldn't place exactly why. I think it is using a different font. Aesthetically, I'd prefer to use the font declaration from Ace, but that's not really a big issue, and will probably only come up looking at the two editors side-by-side.

I don't understand what condense does. Can you explain that?

rehandalal:
While tokenizing the text, if the simple mode addon sees there are more than 2 results in matches (there are multiple capturing groups) it assumes that we care about all of them and need to assign a token to each one. Which is usually the case. However in the case of the rule for `{for win, not winxp}` for which the regex is 
```js
/((,\s?)?(not\s)?((win(xp|7|8)?|mac|linux|android|maemo)|=?(fx|tb|m)\d*))+/
```
there multiple nested capturing groups within a main capturing group. In this case we only care about the main capturing group and also the way the add on deals with matches it ends up messing up the position in the stream when it sees the nested capturing groups, causing the editor to behave erratically and the tokens to be assigned incorrectly.

So `condense` basically tells it to ignore the multiple matches and that we only care about the overall match.

mythmon:
Could we instead use non-capturing groups in the regex? It makes it messier, but it means not needing to modify simple mode. I think the new regex would look something like

```javascript
/((?:\s?)?(?:not\s)?(?:(?:win(?:xp|7|8)?|mac|linux|android|maemo)|=?(?:fx|tv|m)\d*))+/
```

IOW: if the first two characters of a group are `?:` it is a group for purposes of `|` and `?` and friends, but not for captures.

```node-repl
> /(a(?:b))/.exec('ab')
[ 'ab',
  'ab',
  index: 0,
  input: 'ab' ]
```

mythmon:
Also, at that point, maybe we should be breaking this token into smaller bits, like just firefox or just os.

mythmon:
Oops, didn't mean to close.

rehandalal:
updated ^

rehandalal:
removed that extra file we don't need anymore ^

mythmon:
Travis is being slow, and is going to fail anyways until I fix our pre-commit setup. I ran the tests locally, and they all passed, and this works well in manual testing too.

r+

PR #2554, opened 2015-06-02
[bug 1119411] Change "reply" to "quote" in forums

rehandalal:
small r?

N.B.: I deleted `post-new.html` because we don't use it anywhere.


rehandalal:
rebased ^

travis likey

mythmon:
Yassss.

PR #2557, opened 2015-06-08
[Bug 1012384] Fixing Problematic Templates title

safwanrahman:
I believe this Data migration will fix all the Problematic Template Titles.
r?

### DIFF
--- a/kitsune/wiki/migrations/0004_correcting_template_title.py
+++ b/kitsune/wiki/migrations/0004_correcting_template_title.py
@@ -1,0 +1,32 @@
+# -*- coding: utf-8 -*-
+"""Migration to correct the wrong title of Template Catagory Documents"""
+from __future__ import unicode_literals
+
+from django.db import models, migrations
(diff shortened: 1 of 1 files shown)


mythmon:
Doing it this way creates a redirect for every document, which will only make the problem worse. Checkout the Document.save() method for details on how that works.

safwanrahman:
@mythmon, I have tested about the redirects. Till you don't change the slug, it will not create any redirects. As this migration onnly change the title, not slug, it will not create any redirect.

safwanrahman:
Actually the things work like below
# An English Template is created. regard the template name is "Template:test kitsune"
# Use the template at an english kb article with [T:test kitsune]
# The English template goes for localization. and they change the title by localizing the title also.
# The English KB article goes for localization. where the template [T:test kitsune] is used, the localizer don't change it.

'''So need to be mentioned that, till we don't change any title of English KB, there will not be any problem if we change the localized template's title'''

PR #2558, opened 2015-06-08
[bug 1172549] Upgrade less-plugin-autoprefixer

rehandalal:
r?


mythmon:
This happened last time too. Somehow when you ran relock, it removed some sub-dependencies, including an older version of bower that we need for bowerify.

rehandalal:
gah! weird. let me check why that's happening.

rehandalal:
reran tests and travis was happy

PR #2559, opened 2015-06-08
[bug 1159025] Invite to screen share manual flow

rehandalal:
r?


kitsune/questions/templates/questions/question_details.html
```diff
@@ -416,6 +416,13 @@ <h2 class="summary" itemprop="headline">{{ question.title }}</h2>
         </ul>
       {% endif %}
 
+      {% if user.has_perm('users.screen_share') %}
+        <form class="screen-share-form cf" method="post" action="{{ url('questions.screen_share', question.id) }}">
+          {{ csrf() }}
+          <a href="#" class="btn btn-submit"data-type="submit">{{ _('Invite to share screen') }}</a>
```
mythmon: Nit: I think you missed a space here between the class and the data-type.

kitsune/questions/views.py
```diff
@@ -1676,6 +1677,33 @@ def metrics(request, locale_code=None):
...
+
+    content = _("I invited {user} to a screen sharing session, "
+                "and I'll give an update here once we are done.")
+    answer = Answer(question=question, creator=request.user,
+                    content=content.format(user=display_name(question.creator)))
+    answer.save()
+    statsd.incr('questions.answer')
+
+    if Setting.get_for_user(request.user,
+                            'questions_watch_after_reply'):
```
mythmon: Nit: This probably doesn't need to wrap.

mythmon:
Can you add a metadata field to the question when the request for screensharing happens? Something like `question.add_metadata('screen_sharing', 'true')`. I'm sure that in the future we are going to want to do something like figure out what percentage of screen sharing questions were answered during a full moon, or something.

kitsune/questions/templates/questions/question_details.html
rehandalal: gah!

rehandalal:
updated ^

mythmon:
LGTM

PR #2560, opened 2015-06-09
[bug 1159045] Add GA tracking to instant search

rehandalal:
r?

### DIFF
--- a/kitsune/sumo/static/sumo/js/instant_search.js
+++ b/kitsune/sumo/static/sumo/js/instant_search.js
@@ -75,6 +75,7 @@
         search.setParams(params);
         search.query($this.val(), k.InstantSearchSettings.render);
         _gaq.push(['_trackEvent', 'Instant Search', 'Search', search.lastQueryUrl()]);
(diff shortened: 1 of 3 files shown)


kitsune/sumo/static/sumo/js/instant_search.js
```diff
@@ -93,6 +101,8 @@
       });
     }
 
+    _gaq.push(['_trackEvent', 'Instant Search', 'Search', $this.data('href')]);
```
mythmon: We need to track these as page views that match Google Analytics' search tracking so we can use all the search metrics.

mythmon: There is some information about this here: https://support.google.com/analytics/answer/1012264?hl=en#Post

In particular, the event should be something like `ga('send', 'pageview', '/search?q=keyword');`

rehandalal:
updated ^

mythmon:
I think this will work. It can't be worse than what we have now.

r+, though I can't merge it (looks like merge conflict?). Want to fix this up and land it?

PR #2561, opened 2015-06-09
[Bug 1172936] Don't show warnings about outdated articles and missing l10ns for archived/not maintained articles

pollti:
I changed the indentation in line 51. I guess it should be like this.


mythmon:
Looks good, and yes, the indentation change you made is better.

PR #2564, opened 2015-06-09
[Bug 1173190] Fix nunjucks template path.

mythmon:
This was missed when all the static files were moved into namespaced directories.

Joni is going to file a bug for this, and I'll make sure this gets linked up with it, once it exists.

r?


mythmon:
This fixes the problem (I tested on stage), and Travis likes it. I'm going to go ahead an merge this and get it deployed tonight, so search stops being broken.

PR #2565, opened 2015-06-10
[Bug 1149559] Show "No one has helped..." notification only on articles marked Ready For L10n

pollti:
[Show the "No one has helped translate this article yet..." notification only on articles that are marked Ready For Localization](https://bugzilla.mozilla.org/show_bug.cgi?id=1149559)

### DIFF
--- a/kitsune/wiki/templates/wiki/document.html
+++ b/kitsune/wiki/templates/wiki/document.html
@@ -52,7 +52,7 @@
     <article class="wiki-doc">
       {{ document_title(document) }}
       {{ document_messages(document, redirected_from) }}
-      {{ document_content(document, fallback_reason, request, settings, document_css_class) }}
+      {{ document_content(document, fallback_reason, request, settings, document_css_class, any_localizable_revision) }}
 
     {% set share_link = document.share_link or (document.parent and document.parent.share_link) %}
     {% if share_link %}
--- a/kitsune/wiki/templates/wiki/includes/document_macros.html
+++ b/kitsune/wiki/templates/wiki/includes/document_macros.html
@@ -41,12 +41,12 @@
   {% endif %}
 {%- endmacro %}
 
-{% macro document_content(document, fallback_reason, request, settings, css_class='') -%}
+{% macro document_content(document, fallback_reason, request, settings, css_class='', any_localizable_revision) -%}
   <section id="doc-content" class="{{ css_class }}">
     {% if not fallback_reason %}
       {{ document.html|safe }}
     {% elif fallback_reason == 'no_translation' %}
-      {% if not document.is_archived %}
+      {% if document.is_localizable and any_localizable_revision and not document.is_archived %}
         <div id="doc-pending-fallback" class="warning-box">
           {% trans help_link=url('wiki.document', 'localize-firefox-help'), document_trans_link=url('wiki.translate', document.slug) %}
             No one has helped translate this article yet. If you already know how localizing for SUMO works,
--- a/kitsune/wiki/tests/test_templates.py
+++ b/kitsune/wiki/tests/test_templates.py
@@ -209,10 +209,8 @@ class DocumentTests(TestCaseBase):
         # Marking the document and the revision ready for localization
         r2 = revision(save=True, document=r.document, content='Some text.',
                       is_approved=True, is_ready_for_localization=True)
-        r2.document.current_revision = r2
-        r2.document.is_localizable = True
-        r2.document.save()
 
+        url = reverse('wiki.document', args=[r.document.slug], locale='de')
         response = self.client.get(url)
         doc = pq(response.content)
 
--- a/kitsune/wiki/tests/test_templates.py
+++ b/kitsune/wiki/tests/test_templates.py
@@ -192,8 +192,8 @@ class DocumentTests(TestCaseBase):
 
     def test_document_fallback_no_translation(self):
         """The document template falls back to English if no translation
-        exists. The fallback message is only shown when there is any revision
-        ready for localization"""
+        exists. The fallback message is not shown if there is no revision
+        ready for localization."""
         r = revision(save=True, content='Some text.', is_approved=True)
         url = reverse('wiki.document', args=[r.document.slug], locale='fr')
         response = self.client.get(url)
@@ -206,22 +206,25 @@ class DocumentTests(TestCaseBase):
         # Included content is English.
         eq_(pq(r.document.html)('div').text(), doc('#doc-content div').text())
 
-        # Marking the document and the revision ready for localization
-        r2 = revision(save=True, document=r.document, content='Some text.',
-                      is_approved=True, is_ready_for_localization=True)
-
+    def test_document_fallback_no_translation_ready_for_l10n(self):
+        """The document template falls back to English if no translation
+        exists. The fallback message is shown if there is any revision
+        ready for localization."""
+        # Creating a revision ready for localization
+        r = revision(save=True, content='Some text.', is_approved=True,
+                     is_ready_for_localization=True)
         url = reverse('wiki.document', args=[r.document.slug], locale='de')
         response = self.client.get(url)
         doc = pq(response.content)
 
-        eq_(r2.document.title, doc('article h1.title').text())
+        eq_(r.document.title, doc('article h1.title').text())
         # Fallback message is shown.
         eq_(1, len(doc('#doc-pending-fallback')))
         # Removing this as it shows up in text(), and we don't want to depend
         # on its localization.
         doc('#doc-pending-fallback').remove()
         # Included content is English.
-        eq_(pq(r2.document.html)('div').text(), doc('#doc-content div').text())
+        eq_(pq(r.document.html)('div').text(), doc('#doc-content div').text())
 
     def test_redirect(self):
         """Make sure documents with REDIRECT directives redirect properly.
--- a/kitsune/wiki/tests/test_templates.py
+++ b/kitsune/wiki/tests/test_templates.py
@@ -192,20 +192,38 @@ class DocumentTests(TestCaseBase):
 
     def test_document_fallback_no_translation(self):
         """The document template falls back to English if no translation
-        exists."""
+        exists. The fallback message is only shown when there is any revision
+        ready for localization"""
         r = revision(save=True, content='Some text.', is_approved=True)
         url = reverse('wiki.document', args=[r.document.slug], locale='fr')
         response = self.client.get(url)
         doc = pq(response.content)
         eq_(r.document.title, doc('article h1.title').text())
 
+        # Fallback message is not shown because the revision is not ready for
+        # localization.
+        eq_(0, len(doc('#doc-pending-fallback')))
+        # Included content is English.
+        eq_(pq(r.document.html)('div').text(), doc('#doc-content div').text())
+
+        # Marking the document and the revision ready for localization
+        r2 = revision(save=True, document=r.document, content='Some text.',
+                      is_approved=True, is_ready_for_localization=True)
+        r2.document.current_revision = r2
+        r2.document.is_localizable = True
+        r2.document.save()
+
+        response = self.client.get(url)
+        doc = pq(response.content)
+
+        eq_(r2.document.title, doc('article h1.title').text())
         # Fallback message is shown.
         eq_(1, len(doc('#doc-pending-fallback')))
         # Removing this as it shows up in text(), and we don't want to depend
         # on its localization.
         doc('#doc-pending-fallback').remove()
         # Included content is English.
-        eq_(pq(r.document.html)('div').text(), doc('#doc-content div').text())
+        eq_(pq(r2.document.html)('div').text(), doc('#doc-content div').text())
 
     def test_redirect(self):
         """Make sure documents with REDIRECT directives redirect properly.
--- a/kitsune/wiki/views.py
+++ b/kitsune/wiki/views.py
@@ -125,6 +125,9 @@ def document(request, document_slug, template=None, document=None):
             # and OK to fall back to parent (parent is approved).
             fallback_reason = 'no_translation'
 
+    any_localizable_revision = doc.revisions.filter(is_approved=True,
+                                                    is_ready_for_localization=True).exists()
+
     # Obey explicit redirect pages:
     # Don't redirect on redirect=no (like Wikipedia), so we can link from a
     # redirected-to-page back to a "Redirected from..." link, so you can edit
@@ -213,6 +216,7 @@ def document(request, document_slug, template=None, document=None):
         'ga_push': ga_push,
         'breadcrumb_items': breadcrumbs,
         'document_css_class': document_css_class,
+        'any_localizable_revision': any_localizable_revision,
     }
 
     response = render(request, template, data)


mythmon:
Although this change is good, the bug was about a different field, `Revision.is_ready_for_localization`. You can check that with `document.current_revision.is_ready_for_localization`.

pollti:
Hi and thanks for your comment @mythmon. What is `document.is_localizable` about? Probably, I didn't use the correct boolean here – I'm sorry for that.
Afaik, `document.current_revision.is_ready_for_localization` checks whether the newest revision is ready for l10n or not. However, the note should be shown if [any revision](https://bugzilla.mozilla.org/show_bug.cgi?id=1149559#c2) of the article is ready for l10n, so the article can be translated.
Please correct me if I'm wrong…

mythmon:
Some articles aren't meant to be localized, like administration articles, or those written originally in another language. That is what `Document.is_localizable` is about.

This bug does talk explicitly about hiding the banner when the latest revision isn't localizable. We don't stop localizers from localizing, but we don't want to urge them to work on a task that is harder than it needs to be. A revision being not ready for localization is generally a product of it not being completely written.

I think we should do both, `Document.is_localizable` and `Revision.is_ready_for_localization`.

safwanrahman:
Good Changes @pollti 
I think you should change the test mentioned here. 
https://github.com/mozilla/kitsune/blob/master/kitsune/wiki/tests/test_templates.py#L193
Let me know if you need any support in making the changes in the test!

kitsune/wiki/templates/wiki/includes/document_macros.html
```diff
@@ -46,7 +46,7 @@ <h1 class="title">{{ document.title }}</h1>
     {% if not fallback_reason %}
       {{ document.html|safe }}
     {% elif fallback_reason == 'no_translation' %}
-      {% if not document.is_archived %}
+      {% if document.is_localizable and document.current_revision.is_ready_for_localization and not document.is_archived %}
```
safwanrahman: Actually, if you use `document.current_revision.is_ready_for_localization` this Notice will show only if the current revision is ready for localization.
The fact is if any revision of a English Document is ready for localization, the notice should show to localizer to localize them.

safwanrahman:
So from the bug, I think you should make a quarry to database to check if any revision is ready for localization. If that is, `True`, then show the warning. If `False` Dont show any warning.
I think you should write this quarry code in the views.

kitsune/wiki/models.py
```diff
@@ -632,6 +632,13 @@ def clear_cached_html(self):
         for mobile, minimal in itertools.product([True, False], repeat=2):
             cache.delete(doc_html_cache_key(self.locale, self.slug, mobile, minimal))
 
+    @property
```
safwanrahman: Remove this full code!

kitsune/wiki/templates/wiki/includes/document_macros.html
```diff
@@ -46,7 +46,7 @@ <h1 class="title">{{ document.title }}</h1>
     {% if not fallback_reason %}
       {{ document.html|safe }}
     {% elif fallback_reason == 'no_translation' %}
-      {% if not document.is_archived %}
+      {% if document.is_localizable and document.any_localizable_revision() and not document.is_archived %}
```
safwanrahman: Here should be only `any_localizable_revision` rather than document.any_localizable_revision()

safwanrahman: So this code should be `{% if document.is_localizable and any_localizable_revision and not document.is_archived %}`

safwanrahman:
you should add this code at views.
     `any_localizable_revision = document.revisions.filter(is_approved=True,
                                                                                        is_ready_for_localization=True).exists()`

maybe at here https://github.com/tpmoz/kitsune/blob/master/kitsune/wiki/views.py#L127

kitsune/wiki/views.py
```diff
@@ -125,6 +125,9 @@ def document(request, document_slug, template=None, document=None):
             # and OK to fall back to parent (parent is approved).
             fallback_reason = 'no_translation'
 
+    any_localizable_revision = document.revisions.filter(is_approved=True,
```
safwanrahman: You should pass this through `data` which is at line 206.

safwanrahman:
@pollti You should pass this through document.html to document_macros.html

kitsune/wiki/views.py
safwanrahman: I think it should be `doc` instead of `document`

kitsune/wiki/templates/wiki/document.html
```diff
@@ -52,7 +52,7 @@ <h1 class="product-title cf">
     <article class="wiki-doc">
       {{ document_title(document) }}
       {{ document_messages(document, redirected_from) }}
-      {{ document_content(document, fallback_reason, request, settings, document_css_class) }}
+      {{ document_content(document, fallback_reason, request, settings, document_css_class, any_localizable_revision) }}
```
safwanrahman: Good job @pollti as per discussion over IRC! :+1:

safwanrahman:
This is working as Expected! :smile: Thanks @pollti 

I think this should get r+
What do you think @mythmon?

**I think After getting r+, the commits should be squashed into one**

mythmon:
Looks good to me. Feel free to squash these commits. Alternatively, I can merge this as is. Let me know which you do (remember that pushing commits to this branch does update the PR, but won't notify anyone).

kitsune/wiki/tests/test_templates.py
```diff
@@ -193,7 +193,8 @@ def test_document_fallback_with_translation_english_slug(self):
     def test_document_fallback_no_translation(self):
```
safwanrahman: Hmm. I think the changes made on this test to pass, is not good enough. Because this test case check about Document fallen back to English. This should be happen whether the document has any localization revision or not. 
The only fact is, if have localization revision, it will show warning. 
I think this test case need to be customized bit more to check whether the warning showing or not!

pollti:
@safwanrahman:
Thanks for your assistance and patience!
I think the test would fail if the revision was not marked as ready for l10n because the test also checks whether this message is shown or not.
Do you think an additional test is needed here?
@mythmon:
I squashed the commits now.
Do you think an additional test is needed here?
Thanks for your assistance, too.

safwanrahman:
@pollti I know it would fail. So we need to customize this test like this
* 1. The document has **only a revision (Not ready for localization)** and it will check that English article is showing but the **Warning is not showing**
* 2. The same document has **a revision which is ready for localization** and it will check that the **Warning is showing**

I dont know Why, I think this test should cover this or work like this. If anyone think this is not much necessary, its Ok to merge! :smile:

kitsune/wiki/tests/test_templates.py
```diff
@@ -206,6 +206,13 @@ def test_document_fallback_no_translation(self):
         doc('#doc-pending-fallback').remove()
         # Included content is English.
         eq_(pq(r.document.html)('div').text(), doc('#doc-content div').text())
+        
+        r.is_ready_for_localization=True
+
```
safwanrahman: you should call `r.save()` to save the revision

kitsune/wiki/tests/test_templates.py
```diff
@@ -206,6 +206,13 @@ def test_document_fallback_no_translation(self):
         doc('#doc-pending-fallback').remove()
         # Included content is English.
         eq_(pq(r.document.html)('div').text(), doc('#doc-content div').text())
+        
+        r.is_ready_for_localization=True
+        
+        # Fallback message is not shown.
+        eq_(None, len(doc('#doc-pending-fallback')))
```
safwanrahman: Here should be `eq_(1, len(doc('#doc-pending-fallback')))` because The fallback message will show! ;)
As the revision is ready for localization

kitsune/wiki/tests/test_templates.py
```diff
@@ -206,6 +206,13 @@ def test_document_fallback_no_translation(self):
         doc('#doc-pending-fallback').remove()
         # Included content is English.
         eq_(pq(r.document.html)('div').text(), doc('#doc-content div').text())
+        
+        r.is_ready_for_localization=True
+        
+        # Fallback message is not shown.
```
safwanrahman: Then you should again get the content of the document again with `response= ........ , doc = ........ etc`

kitsune/wiki/tests/test_templates.py
```diff
@@ -192,7 +192,7 @@ def test_document_fallback_with_translation_english_slug(self):
 
     def test_document_fallback_no_translation(self):
         """The document template falls back to English if no translation
-        exists."""
+        exists. The fallback message is (not) shown."""
```
safwanrahman: `The fallback message is shown when there is any revision for localization`

kitsune/wiki/tests/test_templates.py
```diff
@@ -200,12 +200,24 @@ def test_document_fallback_no_translation(self):
         eq_(r.document.title, doc('article h1.title').text())
 
         # Fallback message is shown.
```
safwanrahman: change the comment here

kitsune/wiki/tests/test_templates.py
```diff
@@ -200,12 +200,24 @@ def test_document_fallback_no_translation(self):
...
-        eq_(1, len(doc('#doc-pending-fallback')))
+        eq_(None, len(doc('#doc-pending-fallback')))
         # Removing this as it shows up in text(), and we don't want to depend
         # on its localization.
         doc('#doc-pending-fallback').remove()
         # Included content is English.
         eq_(pq(r.document.html)('div').text(), doc('#doc-content div').text())
+        
+        r.is_ready_for_localization=True
```
safwanrahman: give a comment!

kitsune/wiki/tests/test_templates.py
```diff
@@ -200,12 +200,24 @@ def test_document_fallback_no_translation(self):
         eq_(r.document.title, doc('article h1.title').text())
 
         # Fallback message is shown.
-        eq_(1, len(doc('#doc-pending-fallback')))
+        eq_(None, len(doc('#doc-pending-fallback')))
         # Removing this as it shows up in text(), and we don't want to depend
         # on its localization.
         doc('#doc-pending-fallback').remove()
```
safwanrahman: This should be removed also

kitsune/wiki/tests/test_templates.py
```diff
@@ -200,12 +200,24 @@ def test_document_fallback_no_translation(self):
...
         eq_(pq(r.document.html)('div').text(), doc('#doc-content div').text())
+        
+        r.is_ready_for_localization=True
+        r.save()
+        
+        response = self.client.get(url)
+        doc = pq(response.content)
+        
+        eq_(r.document.title, doc('article h1.title').text())
+        # Fallback message is not shown.
```
safwanrahman: Man, after you have a revision ready for localization, the fallback message will show!

kitsune/wiki/tests/test_templates.py
```diff
@@ -200,12 +200,24 @@ def test_document_fallback_no_translation(self):
...
+        eq_(1, len(doc('#doc-pending-fallback')))
+        # Included content is English.
+        eq_(pq(r.document.html)('div').text(), doc('#doc-content div').text())
```
safwanrahman: include the `doc('#doc-pending-fallback').remove()` here because its needed here

kitsune/wiki/tests/test_templates.py
```diff
@@ -192,13 +192,27 @@ def test_document_fallback_with_translation_english_slug(self):
...
         """The document template falls back to English if no translation
-        exists."""
+        exists. The fallback message is not shown."""
         r = revision(save=True, content='Some text.', is_approved=True)
         url = reverse('wiki.document', args=[r.document.slug], locale='fr')
         response = self.client.get(url)
         doc = pq(response.content)
         eq_(r.document.title, doc('article h1.title').text())
 
+        # Fallback message is not shown.
```
safwanrahman: The comment should be `Fallback message is not shown because the revision is not ready for localization`

kitsune/wiki/tests/test_templates.py
```diff
@@ -192,13 +192,27 @@ def test_document_fallback_with_translation_english_slug(self):
 
     def test_document_fallback_no_translation(self):
         """The document template falls back to English if no translation
-        exists."""
+        exists. The fallback message is not shown."""
```
safwanrahman: The doc-string should be `The fallback message is only shown when there is any revision ready for localization`

kitsune/wiki/tests/test_templates.py
```diff
@@ -192,13 +192,29 @@ def test_document_fallback_with_translation_english_slug(self):
...
+        # Fallback message is not shown because the revision is not ready for
+        # localization.
+        eq_(0, len(doc('#doc-pending-fallback')))
+        # Included content is English.
+        eq_(pq(r.document.html)('div').text(), doc('#doc-content div').text())
+
+        # Marking the revision ready for localization
+        r.is_ready_for_localization = True
+        r.save()
+
```
safwanrahman: As you inserted `document.is_localizable` here(1), This Warning will only show if the document is marked as localizable. so in test you should also mark the document is localizable.
it should be easy
   ```
    r.document.is_localizable = True
    r.document.save()
```
https://github.com/mozilla/kitsune/pull/2565/files#diff-31e8b5522cb01ad1ac16331cc8397eebR49

**Comment by safwanrahman** [comment]
I dont know why, but changing locale for second time query make the build pass. I think its a Cache issue that the test was failing with same locale. I dont know exactly.

**Comment by safwanrahman** [comment]
Travis Pass. @mythmon r?

File: kitsune/wiki/tests/test_templates.py [hunk 19]
```diff
@@ -192,20 +192,36 @@ def test_document_fallback_with_translation_english_slug(self):
...
+        eq_(r2.document.title, doc('article h1.title').text())
         # Fallback message is shown.
         eq_(1, len(doc('#doc-pending-fallback')))
         # Removing this as it shows up in text(), and we don't want to depend
         # on its localization.
         doc('#doc-pending-fallback').remove()
         # Included content is English.
-        eq_(pq(r.document.html)('div').text(), doc('#doc-content div').text())
+        eq_(pq(r2.document.html)('div').text(), doc('#doc-content div').text())
```
mythmon: All of this really should be a new test. This test previously tested one thing: document fell back when they don't have translations. Now it it is testing two things: 1) that it falls back, and 2) if the message shows in certain situations. This should be two tests.

mythmon: Generally you should only modify old tests if the behavior under test changed. For new features you should prefer to add new tests.

pollti:
Thanks. I'll update the test(s) tomorrow.
So I keep the old test as it was and add a new one checking if the message is not shown and the content is still shown correctly?

mythmon:
That sounds right, yes.

pollti:
@mythmon I updated the tests now. However, I had to change the first test, because the behavior changed there.

safwanrahman:
good changes @pollti. Please squash into two as we made!

pollti:
@safwanrahman Since your commit is between my commits and your changes are based on my first commit, reordering doesn't work by using `git rebase`. Is there another way of doing it?

mythmon:
There isn't really another way besides `git rebase`. The commits don't need to be squashed, you can leave them as is.

As far as the tests, relative to when this started, you added more assertions to a test. Those new assertions (the `eq_` calls) should be part of a new test, not added to another. Needing to change the docstring of an existing test is a pretty good sign you should be adding a new test instead of editing one. You should revert the original test to the way it was at first, and you should add two new tests (you have already added one of them).

mythmon:
Closing in favor of #2607

pollti:
Thank you very much!

PR #2567, opened 2015-06-10
[Bug 1173414] Fix url for product icons placeholder.

mythmon:
I missed this while doing the static rearrangement because it wasn't using `STATIC_URL`. I also grepped around for anything else with this pattern, but I didn't find anything.

r?


rehandalal:
gah! r+

PR #2568, opened 2015-06-10
[Bug 919867] Add date started feature to profiles

anushbmx:
Added function for user to add when did they get involved with sumo data to profile.

Label used for filed is "Involved with SUMO from", it will allow user to select a month and year ( starting from 1998, start year in mozillians.org) 

I don't it would serve any good to have "Involved with SUMO or Mozilla from".
  1. The Label is too long.
  2. The field already exist in Mozillians.org

### DIFF
--- a/kitsune/users/forms.py
+++ b/kitsune/users/forms.py
@@ -1,4 +1,5 @@
 import re
+from datetime import datetime
 
 from django import forms
 from django.conf import settings
@@ -20,6 +21,7 @@ from kitsune.upload.forms import clean_image_extension
 from kitsune.upload.utils import check_file_size, FileTooLargeError
 from kitsune.users.models import Profile
 from kitsune.users.widgets import FacebookURLWidget
+from kitsune.users.widgets import MonthYearWidget
 
 
 USERNAME_INVALID = _lazy(u'Username may contain only English letters, '
@@ -175,11 +177,17 @@ class AuthenticationForm(auth_forms.AuthenticationForm):
 class ProfileForm(forms.ModelForm):
     """The form for editing the user's profile."""
 
+    involved_from = forms.DateField(
+        required=False,
+        label=_lazy(u'Involved with Mozilla from'),
+        widget=MonthYearWidget(years=range(1998, datetime.today().year + 1),
+                               required=False))
+
     class Meta(object):
         model = Profile
         fields = ('name', 'public_email', 'bio', 'website', 'twitter',
                   'facebook', 'mozillians', 'irc_handle', 'timezone', 'country', 'city',
-                  'locale')
+                  'locale', 'involved_from')
         widgets = {
             'facebook': FacebookURLWidget,
         }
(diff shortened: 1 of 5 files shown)


kitsune/users/models.py
```diff
@@ -99,6 +101,7 @@ def clear(self):
         self.mozillians = ''
         self.irc_handle = ''
         self.city = ''
+        self.involved_from = 'x'
```
mythmon: Why is this 'x'?

kitsune/users/models.py
```diff
@@ -67,6 +67,8 @@ class Profile(ModelBase, SearchMixin):
     # No city validation
     city = models.CharField(max_length=255, null=True, blank=True,
                             verbose_name=_lazy(u'City'))
+    involved_from = models.DateField(null=True, blank=True,
+                            verbose_name=_lazy(u'Involved with SUMO from'))
```
mythmon: This isn't correct. The bug is about the date the user started contributing to Mozilla, not SUMO. So if they were a rep for a year before they joined SUMO they could put the date when they started being a rep.

kitsune/users/forms.py
```diff
@@ -175,11 +177,17 @@ def clean(self):
 class ProfileForm(forms.ModelForm):
     """The form for editing the user's profile."""
 
+    involved_from = forms.DateField(
+        required=False,
+        label=_lazy(u'Involved with SUMO from'),
+        widget=MonthYearWidget(years=range(1998, datetime.today().year + 1),
```
mythmon: How did you come up with the year 1998? It is probably worth adding a comment with that info.

mythmon:
This needs a migration to add the new field to the model.

anushbmx:
@mythmon I have added model as well and fixed the requested changes.

kitsune/users/templates/users/profile.html
```diff
@@ -64,10 +64,15 @@ <h2 class="location">
...
       {% endif %}
 
-      {% if num_answers or num_questions or num_solutions or num_documents %}
+      {% if num_answers or num_questions or num_solutions or num_documents or profile.involved_from %}
         <section class="contributions">
           <h2>Contributions</h2>
           <ul>
+            {% if profile.involved_from %}
+              <li>
+                {{ _('Started on') }} {{ profile.involved_from.strftime('%b %Y') }}
```
mythmon: This works for English, but doesn't work for some other languages that may want to put the date somewhere besides the end of the sentence. This should be something like `{{ _('Involved with Mozilla since {date}')|f(date=...) }}`

Also, using strftime directly like this doesn't localize well. I'm not sure what the right thing to do about that is though. Can you look into that?

anushbmx: I'll check that tomorrow and update PR @mythmon

mythmon:
Please don't squash commits during review. It makes it very hard to know what has changed since the last time I looked at it.

anushbmx:
@mythmon sorry. I'll keep that in mind next time :)

mythmon:
This will work, thank you! Sorry for taking so long to get back to this.

PR #2569, opened 2015-06-10
[bug 1173526] Add FactoryBoy, and factories for Users, Documents, and Revisions.

willkg:
This only touches test code, but isn't actually used. As long as the tests pass (i.e. no import errors), then this is good to go. We can adjust the actual factories when they start getting used.

Quick r?


mythmon:
This looks fine to me. Thanks for pulling this out into a separate PR.

PR #2570, opened 2015-06-11
[bug 1167008] Add missing motidings templates

willkg:
r?

### DIFF
--- a/kitsune/motidings/templates/motidings/unsub.html
+++ b/kitsune/motidings/templates/motidings/unsub.html
@@ -15,7 +15,7 @@
       {{ csrf() }}
       <div class="submit-or-cancel">
         <input type="submit" value="{{ _('Unsubscribe') }}" />
-        <a href="{{ url('home') }}"{# Discards l10n. Better ideas? #}>{{ _('Cancel') }}</a>
+        <a href="{{ url('home') }}">{{ _('Cancel') }}</a>
       </div>
     </form>
   </article>
--- a/kitsune/motidings/tests/__init__.py
+++ b/kitsune/motidings/tests/__init__.py
@@ -1,0 +1,10 @@
+from factory.django import DjangoModelFactory
+from tidings.models import Watch
+
+
+class WatchFactory(DjangoModelFactory):
+    class Meta:
+        model = Watch
+
+    event_type = 'fooevent'
+    is_active = True
--- a/kitsune/motidings/tests/__init__.py
+++ b/kitsune/motidings/tests/__init__.py
@@ -1,3 +1,4 @@
+from factory import fuzzy
 from factory.django import DjangoModelFactory
 from tidings.models import Watch
 
@@ -8,3 +9,4 @@ class WatchFactory(DjangoModelFactory):
 
     event_type = 'fooevent'
     is_active = True
+    secret = fuzzy.FuzzyText(length=10)
--- a/kitsune/motidings/tests/test_unsubscribe.py
+++ b/kitsune/motidings/tests/test_unsubscribe.py
@@ -1,0 +1,44 @@
+from nose.tools import eq_
+
+from kitsune.motidings.tests import WatchFactory
+from kitsune.sumo.tests import TestCase
+from kitsune.sumo.urlresolvers import reverse
+
+
+class UnsubscribeTests(TestCase):
+    def test_view_get(self):
+        watch = WatchFactory()
+
+        url = reverse('tidings.unsubscribe', args=(watch.id,))
+        resp = self.client.get(url, follow=True)
+
+        eq_(resp.status_code, 200)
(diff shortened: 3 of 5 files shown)


willkg:
Not sure why travis is failing on npm stuff. We should probably just retry it.

mythmon:
This seems to happen fairly often. I'm not sure what causes it. I'll restart the job now, which should fix it.

willkg:
^^^ Travis passed.

kitsune/motidings/templates/motidings/unsub.html
```diff
@@ -15,7 +15,7 @@ <h1 class="title">{{ _('Are you sure you want to unsubscribe?') }}</h1>
       {{ csrf() }}
       <div class="submit-or-cancel">
         <input type="submit" value="{{ _('Unsubscribe') }}" />
-        <a href="{{ url('home') }}"{# Discards l10n. Better ideas? #}>{{ _('Cancel') }}</a>
+        <a href="{{ url('home') }}">{{ _('Cancel') }}</a>
```
mythmon: I don't understand this comment. Do you? Was it just a left over from when this template was copied from tidings?

kitsune/motidings/tests/__init__.py
```diff
@@ -0,0 +1,10 @@
+from factory.django import DjangoModelFactory
+from tidings.models import Watch
+
+
+class WatchFactory(DjangoModelFactory):
+    class Meta:
+        model = Watch
+
+    event_type = 'fooevent'
+    is_active = True
```
mythmon: Is there any reason this factory shouldn't have make instances with a `secret` value by default?

kitsune/motidings/tests/test_unsubscribe.py
```diff
@@ -0,0 +1,44 @@
...
+        url = reverse('tidings.unsubscribe', args=(watch.id,))
+        resp = self.client.get(url, follow=True)
+
+        eq_(resp.status_code, 200)
+        self.assertTemplateUsed('motidings/unsub.html')
+
+    def test_view_post_success(self):
+        watch = WatchFactory(
+            secret='ou812'
+        )
```
mythmon: Nit: This wrapping seems awkward to me. Is there a good reason to have it across 3 lines like this?

kitsune/motidings/tests/test_unsubscribe.py
```diff
@@ -0,0 +1,44 @@
...
+
+        url = reverse('tidings.unsubscribe', args=(watch.id,), locale='en-US')
+        resp = self.client.post(url + '?s={0}'.format(watch.secret))
```
mythmon: This works fine, but I just learned that you could get this url with `watch.unsubscribe_url()`. It even includes the query string. On the other hand, having symmetry with the doesnotexist test below is nice.

mythmon:
I have some nits/questions with the tests, but nothing that should stop this from merging.

mythmon:
To be clear: the rest is good, the templates look like they'll work, and the tests look like they are testing the right thing.

r+

kitsune/motidings/templates/motidings/unsub.html
willkg: It's talking about dropping the locale prefix. I think it should drop it because then the server will figure it out. Given that, I ditched the comment.

kitsune/motidings/tests/__init__.py
willkg: I don't really have an opinion one way or the other. It won't change any of the rest of the code I wrote.

What do you want to do?

kitsune/motidings/tests/test_unsubscribe.py
willkg: My bad--that was originally a gazillion lines.

kitsune/motidings/tests/test_unsubscribe.py
willkg: Oh, I didn't know about `unsubscribe_url()`.

I'm game for changing it. I don't feel strongly either way.

kitsune/motidings/tests/__init__.py
mythmon: I think I'd make secret a [fuzzy text value](http://factoryboy.readthedocs.org/en/latest/fuzzy.html#fuzzytext), so it is different each time. I think the case where we don't want a secret on a Watch is rarer than the case where we want a secret.

kitsune/motidings/templates/motidings/unsub.html
mythmon: :+1:

willkg:
^^^ Fixes from review comments.

willkg:
Is this good to go?

mythmon:
Thanks!

PR #2571, opened 2015-06-11
[Bug 1157024] Fixed broken edit profile page on Mobile

anushbmx:
Fixes Edit profile page on Mobile which was broke due to bad styling.

I have removed the border applied to Label and made it look like other form input used in SUMO mobile version.

### DIFF
--- a/kitsune/sumo/static/sumo/less/mobile/users.less
+++ b/kitsune/sumo/static/sumo/less/mobile/users.less
@@ -147,7 +147,7 @@
       border-radius: 3px 0 0 3px;
       box-shadow: 0 1px 2px 0 rgba(0, 0, 0, 0.1) inset;
       box-sizing: border-box;
-      display: block;
+      display: inline-block;
       float: left;
       font-family: @Moz;
       font-size: 14px;
@@ -159,11 +159,9 @@
       & + input,
       & + select,
       & + textarea,
-      & + .value,
-      & + .field-button {
+      & + .value {
         border-radius: 0 3px 3px 0;
-        margin-left: 96px;
-        width: calc(100% - 96px);
+        width: calc(~"100% - 96px");
       }
 
       &.big {
@@ -174,8 +172,7 @@
         & + textarea,
         & + .value,
         & + .field-button {
-          margin-left: 150px;
-          width: calc(100% - 180px);
+          width: calc(~"100% - 180px");
         }
       }
     }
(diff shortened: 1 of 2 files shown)


rehandalal:
this solution doesn't work as it completely changes the styling of the form and is no longer in line with the UI that was designed for this page.

kitsune/sumo/static/sumo/less/mobile/users.less
```diff
@@ -140,43 +140,22 @@
...
-      width: 96px;
 
       & + input,
       & + select,
       & + textarea,
       & + .value,
       & + .field-button {
-        border-radius: 0 3px 3px 0;
-        margin-left: 96px;
-        width: calc(100% - 96px);
```
rehandalal: this should be changed to:
```css
width: calc(~"100% - 96px");
```

kitsune/sumo/static/sumo/less/mobile/users.less
```diff
@@ -140,43 +140,22 @@
...
       &.big {
         width: 180px;
-
-        & + input,
-        & + select,
-        & + textarea,
-        & + .value,
-        & + .field-button {
-          margin-left: 150px;
-          width: calc(100% - 180px);
```
rehandalal: likewise:
```css
width: calc(~"100% - 180px");
```

rehandalal:
if you undo all your changes and make the two changes I mentioned that should fix the issue correctly.

the problem is less is computing those values within the calc expression.

anushbmx:
Ok @rehandalal I had worked sometime to fix the original design but was having issues with drop downs for country and time line so I just took a look at inputs and found this simple style. Let me try again.

anushbmx:
@rehandalal Fixed it :D it needed a little more changes. I didn't knew how to use calc() before. Thanks It was another lesson for me.

rehandalal:
this is better, but there is still an issue. you've gotten rid of the label on the password field

kitsune/users/templates/users/mobile/edit_profile.html
```diff
@@ -28,7 +28,6 @@
         <a class="field-button" href="{{ url('users.change_email') }}" data-next="{{ request.get_full_path() }}">{{ _('Change Email') }}</a>
       </div>
 
-      <label>{{ _('Password') }}</label>
```
rehandalal: we need to keep this

rehandalal:
you'll need to float a bunch of things left and adjust left margins accordingly on certain elements.

rehandalal:
While investigating I ended up fixing this locally (along with all the other `calc()`) expressions. Filed #2572. Closing this in favor of that.

anushbmx:
ok @rehandalal .

Any idea why build failed...?

rehandalal:
the test failure looks unrelated.

PR #2578, opened 2015-06-17
Add eslint file to tweak.

mythmon:
This is what I've been using for es6/jsx code, which is supported when we use Babel. I'd like to figure out how to only enable es6 on a per-file basis (or file globs). I haven't  looked into exactly how to do that though. I like these options, but I'm not set in stone on any of them, we can tweak these for sure.

Note: I'm not adding any pre-commit or CI for this. There are a lot of errors by these standards in our code base. Some day we will have to fix that.

@rehandalal r?

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -1,0 +1,44 @@
+ecmaFeatures:
+  arrowFunctions: true
+  binaryLiterals: true
+  blockBindings: true
+  classes: true
+  defaultParams: true
+  destructuring: true
+  forOf: true
+  generators: true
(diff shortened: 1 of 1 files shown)


rehandalal:
travis no likey

mythmon:
What. How could travis possibly take offense to this?

rehandalal:
dunno looked unrelated so i restarted.

rehandalal:
not sure why travis is being fussy about this. but r+ when travis doesn't hate this.

PR #2579, opened 2015-06-18
[bug 1175531] Fix the screen sharing message formatting

rehandalal:
the way we parse messages it doesn't like single line breaks or nested lists. we may want to switch to markdown? this should fix the formatting for now.

r?


mythmon:
Switching to markdown would be... interesting. It is kind of nice that we use the same markup language everywhere, but our wiki markup isn't very nice.

This change looks like it will fix the problem. :ship:

PR #2580, opened 2015-06-18
[bug 1042231] Show canned responses on L10N dash on All Products

rehandalal:
quick r?


mythmon:
I had to think hard about this one to make sure it does the right thing, but it seems right. r+

PR #2587, opened 2015-07-02
[bug 1085618] Use bulk_create in badge tests.

rlr:
I was looking at test performance for like an hour last week. I didnt find anything groundbreaking but... This is a decent speedup for only 4 tests. So it's only like 5 seconds, not much in the grand scheme of things but I figured it doesnt hurt. Might add up to that hour I spent over the next year or decade.

r?

### DIFF
--- a/kitsune/customercare/tests/test_badges.py
+++ b/kitsune/customercare/tests/test_badges.py
@@ -1,6 +1,7 @@
 from datetime import date
 
 from kitsune.customercare.badges import AOA_BADGE
+from kitsune.customercare.models import Reply
 from kitsune.customercare.tests import reply
 from kitsune.kbadge.tests import badge
 from kitsune.sumo.tests import TestCase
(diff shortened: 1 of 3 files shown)


kitsune/customercare/tests/test_badges.py
```diff
@@ -21,8 +22,12 @@ def test_aoa_badge(self):
...
 
         # Create 49 replies.
-        for i in range(49):
-            reply(user=u, save=True)
+        replies = []
+        for i in range(48):
+            replies.append(reply(user=u))
+        Reply.objects.bulk_create(replies)
+
+        reply(user=u, save=True)
```
mythmon: You should add the comment about signals to this test too.

mythmon:
Thanks for pushing forward on this. r+ with that extra comment.

rlr:
fixed ^ Waiting on travis.

rlr:
yay! merging!

PR #2589, opened 2015-07-06
[Bug 1175880] Change survey gizmo code to use SG's Beacon.

mythmon:
Now all Survey Gizmo surveys will be controlled from Survey Gizmo, instead of us having to put code on each page for each survey.

r?

### DIFF
--- a/kitsune/products/templates/products/documents.html
+++ b/kitsune/products/templates/products/documents.html
@@ -9,11 +9,6 @@
 {% set styles = ('products',) %}
 {% set search_params = {'product': product.slug} %}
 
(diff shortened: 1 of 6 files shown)


rlr:
That's awesome! I am not sure how to test this properly or if it's even possible. I grepped and you got all the instances. r+

mythmon:
I checked that it pulled in the real beacon script in the browser (yes). The real test is to put this on stage on have Kadir verify it. So I guess I'll do that.

mythmon:
I went to far with this, we shouldn't remove everything just yet. I'm going to close this, and open another one.

PR #2591, opened 2015-07-07
[bug 1170085] Autocompletion for code mirror

rehandalal:
r?


kitsune/sumo/static/sumo/js/codemirror.sumo-hint.js
```diff
@@ -0,0 +1,38 @@
...
+  else // Plain browser env
+    mod(CodeMirror);
+})(function(CodeMirror) {
+  "use strict";
+
+  var WORD = /[\w$-]+/, RANGE = 500;
+
+  CodeMirror.registerHelper("hint", "sumo", function(editor, options) {
+    var word = options && options.word || WORD;
+    var range = options && options.range || RANGE;
```
mythmon: You could rename this file to `codemirror.sumo-hints.js.es6` and use es6 default parameters like this:

```js
CodeMirror.registerHelper("hint", "sumo", function(editor, {word=/[\w$-]+/, range=500}) {
```

kitsune/sumo/static/sumo/js/codemirror.sumo-hint.js
```diff
@@ -0,0 +1,38 @@
+(function(mod) {
+  if (typeof exports == "object" && typeof module == "object") // CommonJS
+    mod(require("../../lib/codemirror"));
+  else if (typeof define == "function" && define.amd) // AMD
+    define(["../../lib/codemirror"], mod);
+  else // Plain browser env
+    mod(CodeMirror);
+})(function(CodeMirror) {
```
mythmon: I suspect we could just assume that this code is running in a plain browser env, or switch to browserify and use CommonJS. Either way, we don't need to include all of them.

rehandalal:
updated ^

kitsune/sumo/static/sumo/js/codemirror.sumo-hint.js
```diff
@@ -0,0 +1,28 @@
+(function($) {
+  "use strict";
+
+  CodeMirror.registerHelper("hint", "sumo", function(editor, options) {
+    options = $.extend({}, options, {word: /[\w$-]+/, range: 500});
```
mythmon: This will override the options passed, since `$.extend` makes later argument take precedence (according to the docs). It seemed last time like you wanted these values to be the defaults, not forced values.

rehandalal:
d'oh! fixed ^

mythmon:
I'm pretty unhappy about the style of this code. It has a lot of issues I would want a linter to catch. But we don't have a linter, and so I don't really want to r- it because of those style issues. Also, it seems like this is code that was copied from some CodeMirror plugin and modified to work for us, so modifying it to fit our style seems a bit odd.

What do you think? Should we let the style problems go until we have to fix it for a linter? or do you want to fix it now?

rehandalal:
i'd rather fix it now. i'll run it through the eslint and fix issues.

rehandalal:
linted ^

mythmon:
Beautiful, thanks!

r+

PR #2592, opened 2015-07-10
[bug 1082898] Add ready for l10n column to KB dashboard

rehandalal:
r?


mythmon:
Can we write tests for this? I think that this function already has one test, so it shouldn't be too hard to add another.

rehandalal:
updated with test ^

mythmon:
Cool, r+

PR #2595, opened 2015-07-13
[bug 1082900] Filter KB dashboard by category

rehandalal:
r?

### DIFF
--- a/kitsune/dashboards/views.py
+++ b/kitsune/dashboards/views.py
@@ -18,6 +18,7 @@ from kitsune.dashboards.utils import render_readouts, get_locales_by_visit
 from kitsune.products.models import Product
 from kitsune.sumo.urlresolvers import reverse
 from kitsune.sumo.utils import smart_int
+from kitsune.wiki.config import CATEGORIES
 from kitsune.wiki.models import Locale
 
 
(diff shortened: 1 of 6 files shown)


mythmon:
Travis had an unrelated failure. I restarted the job.

mythmon:
This time the Travis failure is an actual test failure. :boom:

rehandalal:
gah! i am silly! fixed ^

safwanrahman:
I think this button should also be in Complete overview Page. So a person can see all the document of a category in Complete overview Dashboard.
`/en-US/contributors/kb-overview`

safwanrahman:
I think the UI looks better if its like this. 
firefox_screenshot_2015-07-14t20-45-37 701z

kitsune/dashboards/views.py
```diff
@@ -210,3 +215,17 @@ def _get_product(request):
...
     return None
+
+
+def _get_category(request):
+    category = request.GET.get('category')
+
+    if category:
+        for c in CATEGORIES:
+            print c[0]
+            print category
```
mythmon: Oops!

mythmon:
I think Safwan is right about the visual part of this, it would be better if it was horizontally aligned with the subscribe button. Right now it looks kind of haphazard.

Functionality seems to work great. You've got a couple debug print statements, but besides that, the code looks good.

I think we don't need to add it to the other page right now. It makes sense, but that would be a different bug, and it makes sense to make it a different PR to keep things small.

rehandalal:
fixed ^

mythmon:
r+

PR #2596, opened 2015-07-13
[Bug 1182530] Update Django to 1.7.9.

mythmon:
The changes to simplejson are simply a rearrangement of the existing hashes, because when I tried to test the Django upgrade on a fresh venv, the conflicting library versions causes pip to reject the installation.

@willkg r?

### DIFF
--- a/requirements/default.txt
+++ b/requirements/default.txt
@@ -291,9 +291,6 @@ selenium==2.45.0
 # sha256: v445sBLQFGpK6vVBNT9bm5Ywu8tawnUFhJrIlrz_L2o
 setuptools==0.9.8
 
-# sha256: _lMoR-CPEVj4EJr9NJdUBZ4QFYMNsYjXOwHi7MD35Sg
-simplejson==3.1.2
-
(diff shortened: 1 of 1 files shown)


willkg:
I'm pretty sure we can ditch simplejson. We needed it when we were using python 2.6 since python 2.6 didn't have a json module. But python 2.7 does have a json module and I'm pretty sure most things can just use that.

mythmon:
It looks like you're right. We don't need it for Kitsune. pre-commit still uses it though, so it needs to stay in the dev dependencies. Rebased to do that (and split the commits).

willkg:
I thought pre-commit does its own virtualenv somewhere else. No?

mythmon:
The hooks get their own environments, but it still has some dependencies of its own for bootstrapping. I tested by uninstalling simplejson and running the tests and running pre-commit. The tests worked fine, pre-commit broke.

willkg:
Ahh... weird and sucky.

I'll double-check the rest of this after bedtime.

willkg:
Shas work. Tests pass. Upgrades worked for me. I think this is good. r+

willkg:
Landing this on behalf of @mythmon.

PR #2598, opened 2015-07-20
Make logo retina friendly

rehandalal:
r?


mythmon:
Oops, I didn't look closely enough at this. It does strange things to the floating header, stretching it out too wide.

PR #2599, opened 2015-07-20
Fix logo sizing

rehandalal:
oops!

small r?


mythmon:
r+ once travis goes :green_book:.

PR #2600, opened 2015-07-21
[Bug 1175880] Replace exit survey with SurveyGizmo Beacon.

mythmon:
This is a less ambitious version of #2589 which only removes the exit survey, like specified in the bug.

There isn't a great way yet to target the other kind of surveys we have (like product surveys) in the beacon, so we will need to do more work if we want to kill the rest of the surveys.

r?

### DIFF
--- a/kitsune/sumo/static/sumo/js/surveygizmo.js
+++ b/kitsune/sumo/static/sumo/js/surveygizmo.js
@@ -1,23 +1,23 @@
-;(function() {
+/* globals $:false */
+(function() {
     function launchWindow(url) {
-        var sg_div = document.createElement("div");
+        var sg_div = document.createElement('div');
         sg_div.innerHTML = (
             '<h1>You have been selected for a survey</h1>' +
             '<p>We appreciate your feedback!</p><p><a href="' + url + '">Please click here to start it now.</a></p>' +
             '<a href="#" onclick="document.getElementById(\'sg-popup\').style.display = \'none\';return false;">No, thank you.</a>'
         );
-        sg_div.id = "sg-popup";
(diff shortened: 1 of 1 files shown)


mythmon:
This shouldn't be merged directly, as the bug is marked [needs-verify]. We'll need to land this on stage for Kadir to check out after it gets r+ and before we merge it.

kitsune/sumo/static/sumo/js/surveygizmo.js
```diff
@@ -1,23 +1,23 @@
...
         site_survey: function() {
-            var surveyGizmoURL = $('body').data('surveygizmo-url') ||
-                "https://www.surveygizmo.com/s3/popup/1002970/46488f9ad4eb";
-            // This is adapted from a snipptet given by Survey Gizmo.
+            // This is adapted from a snippet given by Survey Gizmo.
             // It has been formatted to fit your screen. Bug 782809.
-            if (!$.cookie('hasSurveyed')) {
-                window.addEventListener("load", function(e) {
+            var surveyGizmoURL = $('body').data('surveygizmo-url');
+            if (surveyGizmoURL && !$.cookie('hasSurveyed')) {
```
mythmon: It is a bit hard to tell from the diff, but here I made it so there is no default survey. The former default survey is being replaced with the Beacon.

Most of the other changes in this function are simply style changes.

rehandalal:
lgtm.

I verified the code is being output to console with `?SurveyGizmoBeaconDebug=1`.

r+

mythmon:
Turns out this isn't what Kadir wanted either. I'll file another pull request.

PR #2601, opened 2015-07-22
[Bug 1180947] SUMO Universal Download Button for Non-Firefox Users

safwanrahman:
I think this should work as expected. mentioned in the bug
* User who are **not using Firefox** browser will see Firefox Download Button in Thunderbird product Page and Thunderbird articles
* User will see the button only if he is **not logged in**

r?

### DIFF
--- a/kitsune/products/views.py
+++ b/kitsune/products/views.py
@@ -7,6 +7,7 @@ from mobility.decorators import mobile_template
 from product_details import product_details
 
 from kitsune.products.models import Product, Topic
+from kitsune.sumo.utils import get_browser
 from kitsune.wiki.decorators import check_simple_wiki_locale
 from kitsune.wiki.facets import topics_for, documents_for
 
@@ -43,12 +44,17 @@ def product_landing(request, template, slug):
         else:
             latest_version = 0
 
+    user_agent = request.META.get('HTTP_USER_AGENT', '')
+    browser = get_browser(user_agent)
+    show_fx_download = (product.slug == 'thunderbird' and browser != 'Firefox')
+
     return render(request, template, {
         'product': product,
         'products': Product.objects.filter(visible=True),
         'topics': topics_for(product=product, parent=None),
         'search_params': {'product': slug},
-        'latest_version': latest_version
+        'latest_version': latest_version,
+        'show_fx_download': show_fx_download
     })
 
 
--- a/kitsune/sumo/static/sumo/less/main.less
+++ b/kitsune/sumo/static/sumo/less/main.less
@@ -121,7 +121,7 @@ body {
       }
 
       #aux-nav {
-        margin-right: 302px;
+        margin-right: 290px;
 
         > ul {
           margin-top: 4px;
@@ -1229,6 +1229,29 @@ input[type=button],
     position: relative;
     top: -2px;
   }
+
+  .firefox-download-button {
+    background: #81BC2E;
+    background: linear-gradient(to bottom, #81BC2E, #659324);
+    border-radius: 5px;
+    box-shadow: 0 1px 0 0 rgba(0, 0, 0, 0.2), 0 -1px 0 0 rgba(0, 0, 0, 0.3) inset;
+    display: inline-block;
+    line-height: 25px;
+    margin: 6px 15px 0px 15px;
+    position: relative;
+    text-shadow: 0 -1px 0 rgba(0, 0, 0, 0.25);
+    transition: all 0.5s;
+    width: 130px;
+
+    .download-title {
+      color: #FFFFFF;
+      display: inline-block;
+      font-family: "Open Sans",Arial,Helvetica,sans-serif;
+      padding: 1px 6px 1px 12px;
+      text-align: center;
+      text-transform: capitalize;
+    }
+  }
 }
 
 .html-rtl {
@@ -1395,7 +1418,7 @@ input[type=button],
   margin: 35px 0 0 0;
 
   .searchbox {
-    width: 240px;
+    width: 200px;
   }
 }
 
--- a/kitsune/sumo/utils.py
+++ b/kitsune/sumo/utils.py
@@ -1,4 +1,5 @@
 import json
+import re
 import sys
 from contextlib import contextmanager
 from datetime import datetime
@@ -328,3 +329,14 @@ def is_ratelimited(request, name, rate, method=['POST'], skip_if=lambda r: False
             Record.objects.info('sumo.ratelimit', '{key} hit the rate limit for {name}',
                                 key=key, name=name)
     return request.limited
+
+
+def get_browser(user_agent):
+    """Get Browser Name from User Agent"""
+
+    match = re.search(r'(?i)(firefox|msie|chrome|safari|trident)', user_agent, re.IGNORECASE)
+    if match:
+        browser = match.group(1)
+    else:
+        browser = None
+    return browser
--- a/kitsune/wiki/views.py
+++ b/kitsune/wiki/views.py
@@ -30,7 +30,7 @@ from kitsune.sumo.decorators import ratelimit
 from kitsune.sumo.helpers import urlparams
 from kitsune.sumo.redis_utils import redis_client, RedisError
 from kitsune.sumo.urlresolvers import reverse
-from kitsune.sumo.utils import paginate, smart_int, get_next_url, truncated_json_dumps
+from kitsune.sumo.utils import paginate, smart_int, get_next_url, truncated_json_dumps, get_browser
 from kitsune.wiki.config import (
     CATEGORIES, MAJOR_SIGNIFICANCE, TEMPLATES_CATEGORY, DOCUMENTS_PER_PAGE,
     COLLAPSIBLE_DOCUMENTS)
@@ -203,6 +203,10 @@ def document(request, document_slug, template=None, document=None):
     # The list above was built backwards, so flip this.
     breadcrumbs.reverse()
 
+    user_agent = request.META.get('HTTP_USER_AGENT', '')
+    browser = get_browser(user_agent)
+    show_fx_download = (product.slug == 'thunderbird' and browser != 'Firefox')
+
     data = {
         'document': doc,
         'redirected_from': redirected_from,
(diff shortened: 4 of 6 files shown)


kitsune/sumo/utils.py
```diff
@@ -328,3 +329,16 @@ def is_ratelimited(request, name, rate, method=['POST'], skip_if=lambda r: False
...
+
+
+def get_browser(request):
+    user_agent = request.META.get('HTTP_USER_AGENT', '')
+    user_agent = str(user_agent)
+
+    try:
+        browser = [[b.group(1), b.group(2)] for b in re.finditer(
+            r'(?i)(firefox|msie|chrome|safari)[\/\s]([\d.]+)', user_agent)][0]
+        browser = browser[0]
```
mythmon: Instead of using `.finditer` and catching the index error here, you should do something like

```python
match = re.find(r'...', user_agent)
if match:
    browser = match.group(1)
else:
    browser = None
```

kitsune/sumo/utils.py
```diff
@@ -328,3 +329,16 @@ def is_ratelimited(request, name, rate, method=['POST'], skip_if=lambda r: False
             Record.objects.info('sumo.ratelimit', '{key} hit the rate limit for {name}',
                                 key=key, name=name)
     return request.limited
+
+
+def get_browser(request):
```
mythmon: Utils functions should have docstrings that explain what they do, and the arguments they take.

Also, I think this function would be easier to test if took a user agent string instead of a request object.

kitsune/wiki/views.py
```diff
@@ -200,6 +200,12 @@ def document(request, document_slug, template=None, document=None):
     # The list above was built backwards, so flip this.
     breadcrumbs.reverse()
 
+    browser = get_browser(request)
+    if product.slug == 'thunderbird' and browser != 'Firefox':
+        fx_download_button = True
+    else:
+        fx_download_button = False
```
mythmon: Things of the form

```
if foo == bar:
    baz = True
else:
    baz = False
```

are better written as `baz = foo == bar`. Sometimes parenthesis help readability, like `baz = (foo == bar)`, but they aren't required.

kitsune/wiki/views.py
```diff
@@ -213,6 +219,7 @@ def document(request, document_slug, template=None, document=None):
         'ga_push': ga_push,
         'breadcrumb_items': breadcrumbs,
         'document_css_class': document_css_class,
+        'fx_download_button': fx_download_button,
```
mythmon: `fx_download_button` is a noun. I would expect this variable to hold a button object. Instead it holds a boolean. A name like `show_fx_download` would be more readable, mostly because it is a verb phrase.

kitsune/sumo/utils.py
mythmon: Speaking of tests, utility functions that are "pure" (have no side effects) like this one are prime candidates for testing. They are often easy to test (when made well), and since other code relies on them to work as expected, it is vital to ensure that they work using tests.

mythmon: In short: please add a test for this function.

kitsune/sumo/static/sumo/less/main.less
```diff
@@ -1229,6 +1229,31 @@ input[type=button],
...
+    width: 135px;
+    margin-top: 6px;
+    right: 6px;
+    position: relative;
+
+    .download-title {
+      color: #FFFFFF;
+      display: inline-block;
+      font-size: 77%;
+      font-family: georgia, "Bitstream Charter", serif;
```
mythmon: Using serif fonts is kind of jarring here, and feels a bit heavy to me. Maybe change this to look more like the download firefox button on https://support.mozilla.org/en-US/products/firefox ?

safwanrahman:
@mythmon Added test and made some fixes mentioned above!
r?

kitsune/sumo/utils.py
```diff
@@ -328,3 +329,16 @@ def is_ratelimited(request, name, rate, method=['POST'], skip_if=lambda r: False
...
                                 key=key, name=name)
     return request.limited
+
+
+def get_browser(user_agent):
+    """Get Browser Name from User Agent"""
+
+    user_agent = str(user_agent)
+
+    match = re.search(r'(?i)(firefox|msie|chrome|safari|trident)[\/\s]([\d.]+)', user_agent)
```
mythmon: 1. What's going on with the [\/\s] in the regex here? I don't think / needs escaped. Why the extra \?
2. The normal way to make a regex case insensitive when you want it for the whole regex is re.search(r'...', user_agent, re.IGNORECASE).

kitsune/sumo/utils.py
```diff
@@ -328,3 +329,16 @@ def is_ratelimited(request, name, rate, method=['POST'], skip_if=lambda r: False
             Record.objects.info('sumo.ratelimit', '{key} hit the rate limit for {name}',
                                 key=key, name=name)
     return request.limited
+
+
+def get_browser(user_agent):
+    """Get Browser Name from User Agent"""
+
+    user_agent = str(user_agent)
```
mythmon: I think I missed this last time. Why are you doing this? The user agent should always be a string, right?

mythmon:
I have a couple more issues with the get_browser function, but the tests look good, and this is almost ready to go.

mythmon:
Actually, this looks not so great rendered. It looks like you need some padding to the button, especially on the right.

shot_20040

kitsune/sumo/utils.py
safwanrahman: 1. Oh! Sorry, I forgot to remove it. It was used to get the version of the browser. As we dont need the version anymore, I think its ok to remove it.

2. Thanks for this. I did not have idea about this. I have corrected it in next commit.

kitsune/sumo/utils.py
safwanrahman: Yap. The user agent is always a string but I got a error while testing in my local. Maybe its otherwise. 
Also I thought that anyone may pass other things through the user agent. But didn't have much thinking. I googled and it seems that only string can be passed by header. Also corrected in next commit!

safwanrahman:
@mythmon I have corrected the function and changed the CSS style to match the UI design. Also changed the GA Event as mentioned in the bug.
r?

willkg:
One thing to think about is whether we're doing enough user agent sniffing stuff that we want a real library rather than maintaining all this stuff ourselves.

For Input, I've been looking at uap-python (https://github.com/ua-parser/uap-python). The thinking being that I'd have a utility function that runs the uap-python code to do a first pass and then modifies the results so that I can augment the result based on the things I've been doing in Input so far.

Anyhow, if the user agent parsing stuff is getting at all complicated and/or needs to be updated on a regular basis, I encourage switching to a library rather than continuing to tweak it.

safwanrahman:
@willkg I also thought about importing this library. But I think for this small work, importing a library is too much heavy. This need to check if the browser **is not Firefox.** to show the button. We actually dont need to detect the actual browser in this case. Just need to detect whether its Firefox or not. For this small job, I believe importing a library is not much necessary.

kitsune/products/views.py
```diff
@@ -43,12 +44,17 @@ def product_landing(request, template, slug):
         else:
             latest_version = 0
 
+    user_agent = request.META.get('HTTP_USER_AGENT', '')
+    browser = get_browser(user_agent)
+    show_fx_download = (product.slug == 'thunderbird' and browser != 'Firefox')
```
mythmon: You don't need to make any changes, but you should know the parenthesis here don't do anything. The just help readability. Which is a good thing, but it is important to know what is happening.

kitsune/sumo/static/sumo/less/main.less
```diff
@@ -1229,6 +1229,30 @@ input[type=button],
...
+    background: linear-gradient(to bottom, #81BC2E, #659324);
+    border-radius: 5px;
+    box-shadow: 0 1px 0 0 rgba(0, 0, 0, 0.2), 0 -1px 0 0 rgba(0, 0, 0, 0.3) inset;
+    display: inline-block;
+    line-height: 25px;
+    margin: 6px 15px 0px 15px;
+    position: relative;
+    text-shadow: 0 -1px 0 rgba(0, 0, 0, 0.25);
+    transition: top 0s;
+    transition: all 0.5s;
```
mythmon: This combination of rules is redundant. Later rules always overwrite previous rules, so the `transition: top 0s;` rule is erased. If you want this to be "all except top", then you should swap the order of the rules. Otherwise, just remove the first transition rule.

kitsune/sumo/static/sumo/less/main.less
```diff
@@ -1229,6 +1229,30 @@ input[type=button],
...
+    width: 130px;
+
+    .download-title {
+      color: #FFFFFF;
+      display: inline-block;
+      font-family: "Open Sans",Arial,Helvetica,sans-serif;
+      padding: 1% 6% 1% 10%;
```
mythmon: Since the size of the element is given in pixels, the padding should also be in pixels.

safwanrahman: Updated! :)

kitsune/sumo/static/sumo/less/main.less
safwanrahman: I think in this small button the transition is not much affected, so removed the first one

kitsune/products/views.py
safwanrahman: Yes @mythmon. I have understood it. I did not have Idea about this, thanks for your last comment regarding this. :smile:

safwanrahman:
@mythmon Updated the PR. So now it is r?
If its r+, I can squash it into one!

kitsune/sumo/static/sumo/less/main.less
```diff
@@ -1905,11 +1928,6 @@ input[type=button],
   z-index: 1000;
 }
 
-/* Survey Gizmo bottom bar */
-.sg-b-bar {
-  box-shadow: 0px -3px 3px rgba(0, 0, 0, 0.1);
-}
```
mythmon: This change is needed. This is one of the reasons why I don't like rebasing: it can easily wipe out newer changes if you aren't careful, and forces the burden of the merge conflicts on you instead of me.

mythmon:
This looks good. I'll merge it once Travis goes :green_apple:.

PR #2603, opened 2015-07-22
[bug 1186421] Add CSRF field to instant search form

rehandalal:
r?


mythmon:
I don't think this is needed. The search views don't require CSRF, right? And it still leaves the ajaxSetup function vulnerable to errors.

rehandalal:
closing in favor of #2602

PR #2604, opened 2015-07-27
[Bug 1090363] Fixes issue with navigation due to fixedheader

anushbmx:
The problem was the fixedheader overlays the navigation done by browser using the hash in end of URL.

This can be solved by adding a function on page load which does the following
1. Check if hash is present in URL.
2. If yes check if a valid element with that hash exist in the page. This is done so that we don't end up scrolling the page without any requirement.
3. if element exists we need to scroll the page a little down ( Scroll to the height of fixed header) so that the contents are not under the menu.

### DIFF
--- a/kitsune/sumo/static/sumo/js/ui.js
+++ b/kitsune/sumo/static/sumo/js/ui.js
@@ -276,11 +276,11 @@
     }
   }
 
-  function correctFixedHeader(){
+  function correctFixedHeader() {
     var headerHeight = document.querySelector('header');
     var scrollHeight = headerHeight.scrollHeight;
-    if(window.location.hash && document.querySelector(window.location.hash)){
-      window.scrollBy(0,-scrollHeight);
+    if (window.location.hash && document.querySelector(window.location.hash)) {
+      window.scrollBy(0, -scrollHeight);
     }
   }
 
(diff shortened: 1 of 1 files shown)


kitsune/sumo/static/sumo/js/ui.js
```diff
@@ -275,6 +276,14 @@
     }
   }
 
+  function correctFixedHeader(){
+    var headerHeight = document.getElementsByTagName("header");
```
mythmon: This should be `var headerHeight = document.querySelector('header');`. Using `querySelector` universally will make things a bit easier to read.

kitsune/sumo/static/sumo/js/ui.js
```diff
@@ -275,6 +276,14 @@
...
+    var scrollHeight = headerHeight[0].scrollHeight;
+    if(window.location.hash && document.getElementById(window.location.hash)){
```
mythmon: `getElementById` should be `document.querySelector('#' + window.location.hash)`, for the same reason as above.

anushbmx:
@mythmon I changed the getselector to querySelector, will rebase the commit once you review them

mythmon:
My main issue here is that it is only a partial fix. It doesn't fix the case where someone clicks on an anchor link after the page is already loaded (like clicking on a Table of Contents link). Any way you could fix that?

anushbmx:
@mythmon I have added a hashchange listner via jquery to detect change hash and trigger the extra scroll of fixed header height. Hope this solves all issues over here.

mythmon:
This works. Sometimes it lags a little bit, and there is some jitter, and sometimes it doesn't work, but it is better than what we have now.

Thanks!

PR #2605, opened 2015-07-27
[Bug 1175880] Replace exit survey with Beacon on pages with no other survey.

mythmon:
The logic is now

1. A template can ask for a custom survey by name (not by URL).
2. If no survey is provided, the survey named `beacon` is used.
3. The refresh survey is added to the list from (1) and (2). (It is always included).
4. Each survey has custom rules for if it gets shown (generally cookie and waffle based). These didn't change.

The beacon code is a copy and paste from the other two PRs for this bug. The interesting part is the change to the logic for how the survey is selected. Two things should be true:

1. All pages should have at least two surveys.
2. If a page has the beacon survey, it should have no other survey except the refresh survey.
3. All pages should have the refresh survey.

r?

### DIFF
--- a/kitsune/products/templates/products/documents.html
+++ b/kitsune/products/templates/products/documents.html
@@ -11,7 +11,7 @@
 
 {% if product.slug == 'mobile' %}
   {# Firefox for Android has a special separate survey. #}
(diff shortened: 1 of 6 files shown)


rehandalal:
lgtm. r+

PR #2606, opened 2015-07-28
[bug 1186464] Fix showfor for Win 10

rehandalal:
Probably best to check the code with `?w=0` since I did some re-indenting.

I got the Win 10 user agent for testing from here:
https://msdn.microsoft.com/en-us/library/hh869301(v=vs.85).aspx

r?


mythmon:
Some Qunit tests got mad at you. Besides that, this looks good. Thanks for cleaning up my trailing commas.

rehandalal:
fixed ^

PR #2608, opened 2015-07-28
[Bug 1175880] Style Survey Gizmo bottom bar.

mythmon:
This adds a little shadow above the survey gizmo beacon UI (which is just an unstyled white box). It makes it look a lot better on the site.

To test: Visit a page like /en-US/kb/install-older-version-of-firefox?SGINTTEST=2645123 . That will force the survey gizmo beacon to trigger, showing the UI that this patches.

r?


rehandalal:
makes sense! r+

PR #2609, opened 2015-07-28
[Bug 1187070] Don't use repr for Documents in DocumentLink.

mythmon:
I'm not sure why, but this causes some Django templates to get very confused and throw unicode decode errors. In particular, when trying to delete documents, if the document is linked to another document and one of those two documents has unicode in the title, the '%r' in the format string throws an error. Changing these to %s fixes the delete page.

I still don't know what is going on here. I tried and failed to reproduce it.

r?


rehandalal:
weird. but r+

PR #2612, opened 2015-08-03
[Bug 1183429] Add user login and registration to new AAQ.

mythmon:
The interesting part of this is mostly `UserAuthActions.es6.js`. In there I make various kinds of HTTP requests to log a user in without leaving the page. The main goal being to make requests that will cause the browser to set the right cookies so that future API calls and page loads will recognize the user as logged in. User registration uses the normal create user API.

This required a bit of a nasty trick. We used to log users out as soon as they became deactivated, and not allow deactivated users to log in. Newly created users are created in a deactivated state. I removed/modified both of the safe guards to allow newly created users to log in and stay logged in.

I put forth that it's fine it's fine because the sensitive parts of the site are (or should be) guarded with `@requires_login`, which only allows authenticated, *active* users to view the page. This makes me itch. It feels bad, and it feels like there could be a better way of doing  this, but I can't think of it. I'd love to hear any other ideas.

The rest of this is pretty straight forward. I stole the user auth forms from the old AAQ, and stuck them in React components. I rewired some of the events on them (like submit) to tie into a state machine in a `UserAuthStore`. Once I figured out what to do, it was pretty easy. The hardest part was convincing the auth system to let me in.

In testing this, I created lots of users. My local DB now contains a user named mythmon, mythmon2, mythmon3, etc all the way up to mythmon30. Whee!

Flows to test:

* Asking a question as an authenticated user still works.
* Logging in in the AAQ with an inactive user doesn't work.
* Logging in with an active user with the AAQ works, and you can then actually ask the question.
* Registering a new user works, and you can actually ask the question.

> I had a bug at one point where you could register, and the login mechanism worked for the first request, but you still couldn't post the question, getting an authentication denied error.

r?

### DIFF
--- a/kitsune/questions/static/questions/js/components/SubmitQuestion.jsx
+++ b/kitsune/questions/static/questions/js/components/SubmitQuestion.jsx
@@ -3,6 +3,7 @@ import cx from 'classnames';
 import AAQStep from './AAQStep.jsx';
 import AAQActions from '../actions/AAQActions.es6.js';
 import {questionEditState} from '../constants/AAQConstants.es6.js';
+import {authStates} from '../../../users/js/constants/UserAuthConstants.es6.js';
 
 export default class SubmitQuestion extends AAQStep {
   handleSubmit() {
@@ -14,7 +15,8 @@ export default class SubmitQuestion extends AAQStep {
   }
 
   enabled() {
-    return this.props.questionState === questionEditState.VALID;
+    return this.props.questionState === questionEditState.VALID &&
+      this.props.userAuth.state === authStates.LOGGED_IN;
   }
 
   render() {
@@ -32,11 +34,6 @@ export default class SubmitQuestion extends AAQStep {
                 onClick={this.handleSubmit.bind(this)}>
           {buttonTexts[this.props.questionState]}
         </button>
-        <pre>
-          {JSON.stringify(this.props.question, null, 2)}
-        </pre>
-        <pre>{this.props.questionState}</pre>
-        <pre>{this.props.validationErrors}</pre>
       </div>
     );
   }
@@ -46,4 +43,5 @@ SubmitQuestion.propTypes = {
   question: React.PropTypes.object.isRequired,
   questionState: React.PropTypes.object.isRequired,
   validationErrors: React.PropTypes.object.isRequired,
+  userAuth: React.PropTypes.object.isRequired,
 };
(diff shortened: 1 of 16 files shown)


kitsune/questions/static/questions/js/components/SubmitQuestion.jsx
```diff
@@ -32,11 +34,6 @@ export default class SubmitQuestion extends AAQStep {
                 onClick={this.handleSubmit.bind(this)}>
           {buttonTexts[this.props.questionState]}
         </button>
-        <pre>
-          {JSON.stringify(this.props.question, null, 2)}
-        </pre>
-        <pre>{this.props.questionState}</pre>
-        <pre>{this.props.validationErrors}</pre>
```
mythmon: Removing this debug info.

kitsune/questions/static/questions/js/components/UserAuth.jsx
```diff
@@ -1,97 +1,49 @@
...
+import {authStates} from '../../../users/js/constants/UserAuthConstants.es6.js';
+import LoginForm from '../../../users/js/components/LoginForm.jsx';
+import RegisterForm from '../../../users/js/components/RegisterForm.jsx';
 
 export default class UserAuth extends AAQStep {
   constructor(props) {
     super(props);
-    this.state = {
-      signingIn: false,
-    };
```
mythmon: The state held by `signingIn` was moved to the state machine in `UserAuthStore.`

kitsune/questions/static/questions/js/components/UserAuth.jsx
```diff
@@ -1,97 +1,49 @@
...
+      return null;
     }
   }
 }
 UserAuth.propTypes = {
-  loggedIn: React.PropTypes.bool.isRequired,
+  userAuth: React.PropTypes.object.isRequired,
 };
-
-class RegisterForm extends React.Component {
```
mythmon: This form moved, largely unchanged, to `kitsune/users/static/users/js/components/RegisterForm.jsx`.

kitsune/users/static/users/js/actions/UserAuthActions.es6.js
```diff
@@ -1,9 +1,132 @@
...
+
+export function login(username, password, {inactive=false}={}) {
+  const csrfElem = document.querySelector('input[name=csrfmiddlewaretoken]');
+  const csrf = csrfElem ? csrfElem.value : null;
+
+  Dispatcher.dispatch({
+    type: actionTypes.AUTH_LOG_IN_OPTIMISTIC,
+  });
+
+  $.post('/en-US/users/login', {
```
mythmon: I used jquery's ajax function here because it serializes form data nicely. In a perfect world I'd probably have a utility function to do that formatting for me, and  use `fetch` instead.

rehandalal:
stepped through each flow. works as expected. r+

PR #2613, opened 2015-08-04
[Bug 1189285] Adding dsb and hsb as FAQ locales

vesper-mozilla:
This should add dsb and hsb as FAQ locale to SUMO.
(https://bugzilla.mozilla.org/show_bug.cgi?id=1189285).

I expect Travis to dislike something missing in the tests, as usual ;-)


mythmon:
Looks good.

It would probably be worth it to set up a local kitsune environment for yourself so that you can run these tests without having to wait for Travis. Let me know if you need some help doing that.

vesper-mozilla:
I know that a local environment under Windows is a bother to set up, so I may have to set up a VM with Linux on it for testing.

If you point me towards any docs other than [the Kitsune docs](https://kitsune.readthedocs.org/en/latest/hacking_howto.html?highlight=environment#setting-up-an-environment), I can get started on this after I come back from PTO.

mythmon:
The kitsune docs you linked to are the only thing we have. We have tried in
the past to document how to get a basic system up and running, but we can't
usually keep up with the maintenance required, so we just leave that to irc
and research.

PR #2614, opened 2015-08-04
[Bug 1190340] Renaming sr-Cyrl to sr

vesper-mozilla:
This should change all the references for "sr-Cyrl" to "sr", to prepare
Kitsune for Pontoon.

WARNING: scripts/schema.sql still refers to sr-CYRL, as per
https://bugzilla.mozilla.org/show_bug.cgi?id=1190340#c1 - additional work may be required


mythmon:
`scripts/schema.sql` is a relic of a by-gone time, and doesn't matter anymore.

This is setting off some alarm bells in my head. I don't remember all the details, but I feel like we have made this rename before in the other direction (`sr` -> `sr-Cyrl`) and it broke things. I'll have to look into this more.

safwanrahman:
I think this will break because it does not have any migration. The Data of `sr-Cyrl` locale should remain as there and will not move into `sr`. I think the correct way to do this by adding another locale `sr`, then write a data migration to migrate the `sr-Cyrl` locale data to `sr` locale and then remove the `sr` locale.
I believe this should work.

safwanrahman:
As it needed some Data Migration works, I have written another patch #2622 based on this patch. Hope that should fix the issue! :smile:

mythmon:
Closing in favor of #2622

PR #2615, opened 2015-08-05
Remove schematic migrations.

mythmon:
We kept these around for a little while because we used them to bootstrap South. I don't think we need these anymore though.

:boom: r?


rehandalal:
yay!

rehandalal:
one failure looks related

rehandalal:
i think you just got to remove the test that checks for the migration folder.

mythmon:
Removed the tests. This will still fail because of that error with the AAQ stuff. I'll rebase this once #2616 lands.

mythmon:
^ rebased on top of master with #2616's changes.

rehandalal:
r+

PR #2616, opened 2015-08-05
[bug 1183422] Add waffle switch for new AAQ

rehandalal:
r?


kitsune/questions/migrations/0004_new_aaq_waffle_flag.py
```diff
@@ -0,0 +1,23 @@
...
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('waffle', '0001_initial'),
+        ('questions', '0003_auto_20150430_1304'),
+    ]
+
+    operations = [
+        migrations.RunPython(create_waffle_flag),
```
mythmon: This migration is easily reversible, by simply deleting the waffle flag. You can do that by passing a second function to `RunPython` for the reverse direction.

mythmon:
I tried this out locally. Both flows work nicely.

The failing test appears to be my fault. That didn't happen on my AAQ auth PR because, for some reason, Travis didn't run on my commit. I think we can just delete the test in question, and probably the middleware too. I can do that, if you prefer.

rehandalal:
i'll fix the test and add the delete method

rehandalal:
updated ^

mythmon:
r+ when we get :tea:.

PR #2617, opened 2015-08-06
[bug 1082900] Add category filter to KB dashboard overview

rehandalal:
r?


mythmon:
At the bottom of the original page, there is a "Complete overview..." link. Right now it removes the category and product filters. Maybe it should keep them instead? I could see it both ways.

Also, changing products after selecting a category clears the category. It should probably preserve the category. Luckily, changing the category after selecting a product works as it should.

rehandalal:
fixed ^

mythmon:
Nice. r+

PR #2618, opened 2015-08-07
[Bug 1180947] SUMO Universal Download Button for Non-Firefox Users

safwanrahman:
This should fix the workaround of the Bug. It was misunderstood that the download button should be shown in Thunderbid product page. Actually it should work like below
* The button **should not** show in thunderbird product page
* But the button **Should Show** in any of the **Topic Page** of Thunderbird Product

r?

### DIFF
--- a/kitsune/products/views.py
+++ b/kitsune/products/views.py
@@ -44,17 +44,12 @@ def product_landing(request, template, slug):
         else:
             latest_version = 0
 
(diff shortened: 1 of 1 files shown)


safwanrahman:
#2601

mythmon:
Yep, that works.

Thanks!

PR #2619, opened 2015-08-10
Add translations for Buddyup to jsi18n.

mythmon:
This lets Buddyup pull down its strings from our localization system. This is the same weird way that Yaocho did it.

@rehandalal r?


rehandalal:
makes sense. r+

PR #2620, opened 2015-08-14
Introducing Cool Search.

AdrianGaudebert:
@mythmon I think most of this, at least the JS side, is ready for v0.1. It is still missing API unit tests, that I am going to work on next, and I'd like to get your input on what you think is needed next.

Current features:
- [x] simple query for each document type, tabulated form
- [x] results are displayed under the form
- [x] results are fetched automatically when user types

Things to do before merge:
- [x] unit tests for API
- [x] refactoring of the API views to use Django Rest Framework
- [x] refactoring SearchForm classes

Later needed features:
- [ ] advanced filters for each document type
- [x] styling of the tabs (I'll need to fork the tab library to adapt it to our needs)
- [ ] pagination of results

Also, why do we have files in the ``/static`` folder if that one is in the .gitignore? I don't know how I ended up deleting that ``.htaccess`` file, but now I'm struggling to re-add it.


mythmon:
Travis has concerns with your code.

For linting errors, it is pretty easy to catch these before you commit. The docs have [instructions for how to set up the automatic linting tools](http://kitsune.readthedocs.org/en/latest/hacking_howto.html#install-linting-tools).

As for the .htaccess files, they are necessary for the server configuration, but everything else in that directory is generated, and so shouldn't be checked into git. You can try and restore that file with something like `git checkout HEAD~5 -- static/.htaccess`.

I'll look over the rest of this now.

bower.json
```diff
@@ -25,7 +25,8 @@
     "nunjucks": "nunjucks#1.0.5",
     "nwmatcher": "nwmatcher#1.3.4",
     "pikaday": "~1.3.2",
-    "react": "#0.13.0",
+    "react": "#0.13.3",
```
mythmon: Any particular reason you updated the version of React here?

kitsune/coolsearch/api.py
```diff
@@ -0,0 +1,249 @@
...
+from kitsune.search import es_utils
+from kitsune.search.utils import locale_or_default, clean_excerpt, ComposedList
+from kitsune.sumo.utils import smart_int
+from kitsune.wiki.models import DocumentMappingType
+
+
+EXCERPT_JOINER = _lazy(u'...', 'between search excerpts')
+
+
+class SearchView(View):
```
mythmon: This should probably subclass a pDjango Rest Framework APIView class](https://github.com/tomchristie/django-rest-framework/blob/e11f41ebc4ef088a5849771dfda5a7fba4f82904/docs/api-guide/views.md#class-based-views). That would probably reduce the amount of boiler plate you have to do, and aligns better with how the rest of the APIs on the site are built.

kitsune/coolsearch/api.py
```diff
@@ -0,0 +1,249 @@
...
+        language = locale_or_default(
+            request.GET.get('language', request.LANGUAGE_CODE)
+        )
+        lang = language.lower()
+        if settings.LANGUAGES_DICT.get(lang):
+            lang_name = settings.LANGUAGES_DICT[lang]
+        else:
+            lang_name = ''
+
+        page = max(smart_int(request.GET.get('page')), 1)
```
mythmon: I've had a lot of luck using [Django Rest Framework serializers](https://github.com/tomchristie/django-rest-framework/blob/e11f41ebc4ef088a5849771dfda5a7fba4f82904/docs/api-guide/serializers.md#serializers) for parsing parameters like this.

mythmon:
Instead of an empty `kitsune/coolsearch/models.py` you can use a [Django 1.7 application config](https://docs.djangoproject.com/en/1.8/releases/1.7/#app-loading-refactor).

kitsune/coolsearch/static/coolsearch/js/actions/ResultsActionCreator.js
```diff
@@ -0,0 +1,14 @@
...
+var ResultsActionCreator = {
+    createResults: function (results) {
+        SearchDispatcher.dispatch({
+            type: ActionConstants.RECEIVE_RESULTS,
+            results: results
+        });
+    }
+};
+
+export default ResultsActionCreator;
```
mythmon: I've been preferring a style like

```js
export function foo() {}
export function bar() {}
export functino baz() {}

export default {
  foo,
  bar,
  baz,
};
```

This makes each individual part exportable with something like `import {foo} from './utils.js';`, while also keeping the `import utils from './utils.js';` ability. It also tends to be a bit less typing, and one level less indentation.

kitsune/coolsearch/static/coolsearch/js/actions/SearchActionCreator.js
```diff
@@ -0,0 +1,23 @@
...
+var SearchActionCreator = {
+  runSearch: function (dataType, payload) {
+    var apiUrl = '/api/2/coolsearch/search/' + dataType + '/';
+
+    // Send an AJAX query to the API.
+    $.getJSON(
+      apiUrl,
+      payload,
+      function success(results) {
+        // When we receive the data, send an action to the Dispatcher to notiy
```
mythmon: typo: notiy

kitsune/coolsearch/static/coolsearch/js/actions/SearchActionCreator.js
```diff
@@ -0,0 +1,23 @@
+/* globals $:false */
+
+import ResultsActionCreator from './ResultsActionCreator.js';
+
+
+var SearchActionCreator = {
+  runSearch: function (dataType, payload) {
+    var apiUrl = '/api/2/coolsearch/search/' + dataType + '/';
```
mythmon: ES6 syntax tip: This could be written as ``var apiUrl = `/api/2/coolsearch/search/${dataType}/`;``.

kitsune/coolsearch/static/coolsearch/js/components/SearchController.jsx
```diff
@@ -0,0 +1,23 @@
...
+    }
+
+    render() {
+        React.render(
+            <div id="coolsearch">
+                <h1 className="nomargin"
+                    title="It's like Advanced Search, but cooler">Cool Search</h1>
+                <SearchForm />
+                <SearchResults
+                    data={[{title: "Hello", content: "World"}]} />
```
mythmon: Should this be an empty list to start with?

kitsune/coolsearch/static/coolsearch/js/components/SearchForm.jsx
```diff
@@ -0,0 +1,86 @@
...
+  };
+};
+
+
+export default class SearchForm extends React.Component {
+  constructor(props) {
+    super(props);
+
+    this.state = getStateFromStores();
+    this.currentForm = 'wiki';
```
mythmon: `this.currentForm` looks a lot like it belongs in `this.state`.

kitsune/coolsearch/static/coolsearch/js/components/SearchForm.jsx
```diff
@@ -0,0 +1,86 @@
...
+import SearchFormQuestion from './SearchFormQuestion.jsx';
+import SearchFormForum from './SearchFormForum.jsx';
+
+import QueryStore from '../stores/QueryStore.js';
+
+import QueryActionCreator from '../actions/QueryActionCreator.js';
+import SearchActionCreator from '../actions/SearchActionCreator.js';
+
+
+var getStateFromStores = function () {
```
mythmon: I think this should be a method of the `SearchForm` class instead of a separate function.

kitsune/coolsearch/static/coolsearch/js/components/SearchForm.jsx
```diff
@@ -0,0 +1,86 @@
...
+
+  getResults(e) {
+    if (e) {
+      e.preventDefault();
+    }
+
+    SearchActionCreator.runSearch(this.currentForm, this.state.query);
+  }
+
+  /** A debounced proxy to getResults(). */
```
mythmon: This comment isn't strictly accurate, since the debouncing happens elsewhere. In fact, I think that this method definition isn't needed. The constructor could have the same effect with

```js
this.startQuery = _.debounce(this.getResults.bind(this), 300);
```

kitsune/coolsearch/static/coolsearch/js/components/SearchFormForum.jsx
```diff
@@ -0,0 +1,32 @@
+/*
+React component that controls the forum specific part of the
+Search form.
+Shows the inputs to search the "forum" documents.
+*/
+
+import QueryActionCreator from '../actions/QueryActionCreator.js';
+
+
+export default class SearchFormForum extends React.Component {
```
mythmon: The three `SearchForm`s share a lot in common. I'd make a super class that they all inherit from.

kitsune/coolsearch/static/coolsearch/js/components/SearchResults.jsx
```diff
@@ -0,0 +1,49 @@
...
+
+  render() {
+    var results = this.state.results.map(function (result) {
+      return (
+        <SearchResult
+          key={result.title}
+          data={result}
+        />
+      );
+    });
```
mythmon: ES6 syntax tip: this could be written as

```js
var results = this.state.results.map(result => <SearchResult key={result.title} data={result}/>);
```

kitsune/coolsearch/static/coolsearch/js/dispatcher/SearchDispatcher.js
```diff
@@ -0,0 +1 @@
+export default new Flux.Dispatcher();
```
mythmon: There should only be one dispatcher. You can use the one in `kitsune/sumo/static/sumo/js/dispatcher.js`.

kitsune/coolsearch/tests/test_api.py
```diff
@@ -0,0 +1,28 @@
...
+        self.refresh()
+
+        response = self.client.get(
+            reverse('coolsearch.search_wiki'),
+            {}
+        )
+
+        assert "We couldn't find any results for" not in response.content
+        # eq_(200, response.status_code)
+        # assert 'Product One' in response.content
```
mythmon: These commented lines probably shouldn't be here. Or they should be uncommented.

mythmon:
I think it makes sense to put this in the existing search app instead of creating a new app. Eventually, I'd like the APIs built here to be the canonical APIs for searching sumo.

bower.json
AdrianGaudebert: Well at first I did it to see if it would fix the problems with that tab library I was using, but now I'm thinking, why not just use the latest version possible? What's the convention on updating deps in SUMO?

mythmon: We don't have a specific convention, but it is generally good to keep PRs as small as small as reasonable. If you aren't using anything specific to 0.13.3, I think it makes sense to move this to a separate PR.

AdrianGaudebert: Aha, actually, I just tried going back to 0.13.0 and bower complained that ``react-swipe-views`` require ``~0.13.1`` which resolves to 0.13.3.

AdrianGaudebert:
@mythmon I got stuck on Socorro things this week, and couldn't finish this. But I addressed most of your comments! What I need to do now is add unit tests for the api, merge it into the search app, and see how I can make a better use of Django Rest. 

On the "feature" side, are you happy with the current state of things for a v0.1?

mythmon:
First off, Travis failed. I think that is just a spurious failure though, so I restarted the builds.

Feature side:

1. Should there be search results even before typing a query? I think it should probably show something else when the query is blank. Maybe this can wait until a future PR.
2. I think we should ditch the "1000 results found" (or any other count of results found). This generally just makes people unhappy, and they only look at the top few results anyways. It is especially frustrating to see that when there is no way to get to the other 990 results.
3. I think having 3 separate search boxes is confusing. Instead I think there should be one search box, above the tabs and the wiki/questions/forums UI should simply change the filtering UI below the tabs. It is especially confusing to switch tabs and lose your query. 
4. The discussion forum icon is messed up. h9nydub
5. What is the "Search Mozilla Support" button for? It doesn't seem to do anything for me.

Most of these are actually fine for now. Definitely fix 4 and 5, but the rest can probably be saved for later PRs. I worry about about this as a base though. It doesn't seem to have a clear way to grow the UI or client data model to deal the many dimensions of complexity the Sumo search system can involve. Do you have any plans for what that looks like? I think more importantly, do you plan on having time to implement those? I am unsure what your time looks like going forward.

AdrianGaudebert:
@mythmon the discussion forum icon is broken on production as well: https://support.mozilla.org/en-US/search/advanced?q=firefox&author=&created=0&created_date=&updated=0&updated_date=&sortby=0&a=1&w=4 I'm not sure what the problem is though.

AdrianGaudebert:
@mythmon My latest change should fix the unit tests. However, if I use the non-minified React lib in the bundle for coolsearch, I get a JS error with React that I can't explain. I'll keep investigating tomorrow, but in the meantime if you have a clue what it could be, let me know. 

```
TypeError: node._store is undefined
```

mythmon:
That's odd. I haven't seen that error before. The minified version of React has fewer safety checks (they are compiled away by the minifier). Maybe you're hitting one of those cases?

AdrianGaudebert:
OK, I finally nailed it. The uncompressed version of React exposes a ``_store`` object for components, which contains the ``props`` sub-object. However, ``_store`` does not exist with the minified version. The solution is to use ``props`` directly from components instead. I had to re-implement my own Tabs library to figure this out, but that's good because it also allowed me to solve the styling issue. 

@mythmon could you give it another look? I'm going to start working on the next steps, but will keep them for separate pull requests.

AdrianGaudebert:
@mythmon ping?

mythmon:
Ah, sorry about letting this sit. It got lost among the churn. I have some comments about the React code.

AdrianGaudebert:
And I forgot to let you know that I addressed your comments last Friday... My bad! Mind reviewing again?

mythmon:
That's a lot cleaner thank you! Lets merge this, I'm excited to see the other changes you've made.

safwanrahman:
@AdrianGaudebert @mythmon Whats the Bug number of this change? its not mentioned in the PR! :question:

AdrianGaudebert:
@safwanrahman it's bug https://bugzilla.mozilla.org/show_bug.cgi?id=1118375, I forgot to put it anywhere, sorry!

PR #2622, opened 2015-08-14
[Bug 1190340] Renaming sr-Cyrl to sr

safwanrahman:
I think this should fix the bug.
Reference #2614 

@mythmon r?


mythmon:
I think you missed the questions models. The Question model and the QuestionLocale model probably both need updated.

How did you find this list of things to change? Do you think you got everything? Why?

safwanrahman:
@mythmon Thanks for this. I have forgotten about the Question Model as I thought this locale does not have any forum.

I have got this list of change by searching models. Whether any model have locale field or not. I believe I have got everything except the `questions` app, and I will make a migration soon. I have not find any other app's Model that contain any kind of locale field therefore I am sure that this are the only app that needs migration.

safwanrahman:
@mythmon Updated with the questions Models. r?

mythmon:
Ok. I think that this will do the right thing. It runs fine forwards and backward on my laptop, and the site didn't explode.

I want to be pretty careful with this though. I'd like to wait until [bug 1195771](https://bugzilla.mozilla.org/show_bug.cgi?id=1195771) gets fixed completely so that we can test this on stage with real data, in case anything goes wrong. Then we can test this on stage before merging it.

Thanks for the work!

mythmon:
r+. Well merge+deploy in coordination with the svn change.

PR #2624, opened 2015-08-18
[bug 1195898] Update to Django 1.7.10

willkg:
Quick r?


willkg:
@mythmon Quick r? It's the same hash as we used for Fjord.

mythmon:
Verified this is the same as the Fjord PR.

r+

PR #2625, opened 2015-08-19
Use the correct set of products in the new AAQ.

mythmon:
I had a quick chat with Madaline about the new AAQ this morning, and she pointed out we weren't doing this.

r?

### DIFF
--- a/kitsune/questions/views.py
+++ b/kitsune/questions/views.py
@@ -104,8 +104,7 @@ ORDER_BY = OrderedDict([
 def product_list(request, template):
     """View to select a product to see related questions."""
     return render(request, template, {
-        'products': Product.objects.filter(
-            questions_locales__locale__in=[request.LANGUAGE_CODE])
+        'products': Product.objects.filter(questions_locales__locale=request.LANGUAGE_CODE)
     })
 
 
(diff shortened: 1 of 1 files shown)


kitsune/questions/views.py
```diff
@@ -104,8 +104,7 @@
 def product_list(request, template):
     """View to select a product to see related questions."""
     return render(request, template, {
-        'products': Product.objects.filter(
-            questions_locales__locale__in=[request.LANGUAGE_CODE])
+        'products': Product.objects.filter(questions_locales__locale=request.LANGUAGE_CODE)
```
mythmon: I'm pretty sure this is equivalent, and not using `__in` is faster. Right? All the tests still passed locally for me, but if there was a good reason for this, easy to revert it.

willkg: I have no idea why that would be using `__in` like that. That's weird.

I think this change is fine.

kitsune/questions/views.py
```diff
@@ -468,7 +467,9 @@ def edit_details(request, question_id):
 def aaq_react(request):
     request.session['in-aaq'] = True
     to_json = JSONRenderer().render
-    products = ProductSerializer(Product.objects.filter(visible=True), many=True)
+    products = ProductSerializer(
+        Product.objects.filter(questions_locales__locale=request.LANGUAGE_CODE),
+        many=True)
```
willkg: I don't understand this change. It went from visible products to products where the locale has AAQ enabled.

Why not visible *and* products where the locale has AAQ enabled?

mythmon: This mimics the filter in the `product_list` view, which controls which products are included on https://support.mozilla.org/questions. I imagine the reasoning there is because has-AAQ-enabled acts as a finer grained version of visible, and some locales may even want to open question forums for not-visible products.

But really it is just cargo culting. If I change this, I'd change the other one too.

willkg: I don't really understand the nitty gritty, so it all sounds good to me.

willkg:
Linting failed but because of a http timeout. Assuming linting passed on your system, I think that's fine.

This looks fine to me. r+

mythmon:
I restarted the lint build, and now it seems find. Merging.

PR #2626, opened 2015-08-20
[Bug 1186697] Enable Bambara and Malagasy

vesper-mozilla:
This should enable Bambara (bm) and Malagasy (mg) on SUMO.


mythmon:
Looks good.

PR #2628, opened 2015-08-21
Upgrade vagrant provision to actually work

turtleloveshoes:
r?

PR #2631, opened 2015-08-24
Correct version of Pip in vagrant

mythmon:
@safwanrahman pointed out that we got the version wrong of Pip in #2628. Oops.

r?


willkg:
That looks good to me. r+

PR #2633, opened 2015-08-25
[Bug 1193553] Add in_aaq flag to Topics for new AAQ.

mythmon:
r?

### DIFF
--- a/kitsune/products/admin.py
+++ b/kitsune/products/admin.py
@@ -16,10 +16,9 @@ class TopicAdmin(admin.ModelAdmin):
         return obj.parent
     parent.short_description = 'Parent'
 
-    list_display = ('product', 'title', 'slug', parent, 'display_order',
-                    'visible')
+    list_display = ('product', 'title', 'slug', parent, 'display_order', 'visible', 'in_aaq')
     list_display_links = ('title', 'slug')
-    list_editable = ('display_order', 'visible')
+    list_editable = ('display_order', 'visible', 'in_aaq')
     list_filter = ('product', 'parent', 'slug')
     readonly_fields = ('id',)
     prepopulated_fields = {'slug': ('title',)}
--- a/kitsune/products/models.py
+++ b/kitsune/products/models.py
@@ -4,6 +4,7 @@ import os
 from django.conf import settings
 from django.db import models
 from PIL import Image
+from tower import ugettext_lazy as _lazy
 from uuid import uuid4
 
 from kitsune.sumo.models import ModelBase
@@ -136,7 +137,7 @@ class Topic(ModelBase):
     visible = models.BooleanField(default=False)
     # Whether or not this topic is used in the AAQ.
     in_aaq = models.BooleanField(
-        default=False, help_text='Whether this topic is shown to users in the AAQ or not.')
+        default=False, help_text=_lazy(u'Whether this topic is shown to users in the AAQ or not.'))
 
     class Meta(object):
         ordering = ['product', 'display_order']
--- a/kitsune/products/models.py
+++ b/kitsune/products/models.py
@@ -134,6 +134,9 @@ class Topic(ModelBase):
 
     # Whether or not this topic is visible in the ui to users.
     visible = models.BooleanField(default=False)
+    # Whether or not this topic is used in the AAQ.
+    in_aaq = models.BooleanField(
+        default=False, help_text='Whether this topic is shown to users in the AAQ or not.')
 
     class Meta(object):
         ordering = ['product', 'display_order']
--- a/kitsune/sumo/tests/__init__.py
+++ b/kitsune/sumo/tests/__init__.py
@@ -1,3 +1,4 @@
+# -*- coding: utf-8 -*-
 import os
 import sys
 from functools import wraps
@@ -13,6 +14,7 @@ from django.test.utils import override_settings
 from django.utils.translation import trans_real
 
 import django_nose
+import factory.fuzzy
 from nose.tools import eq_
 from selenium import webdriver
 from selenium.common.exceptions import WebDriverException
@@ -180,3 +182,10 @@ def eq_msg(a, b, msg=None):
     """Shorthand for 'assert a == b, "%s %r != %r" % (msg, a, b)'
     """
     assert a == b, (str(msg) or '') + ' (%r != %r)' % (a, b)
+
+
+class FuzzyUnicode(factory.fuzzy.FuzzyText):
+    """A FuzzyText factory that contains at least one non-ASCII character."""
+
+    def fuzz(self):
+        return u'đ{}'.format(super(FuzzyUnicode, self).fuzz())
--- a/kitsune/sumo/tests/__init__.py
+++ b/kitsune/sumo/tests/__init__.py
@@ -1,4 +1,5 @@
 # -*- coding: utf-8 -*-
+import inspect
 import os
 import sys
 from functools import wraps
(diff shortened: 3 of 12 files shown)


kitsune/products/admin.py
```diff
@@ -16,10 +16,9 @@ def parent(obj):
         return obj.parent
     parent.short_description = 'Parent'
 
-    list_display = ('product', 'title', 'slug', parent, 'display_order',
-                    'visible')
+    list_display = ('product', 'title', 'slug', parent, 'display_order', 'visible', 'in_aaq')
     list_display_links = ('title', 'slug')
-    list_editable = ('display_order', 'visible')
+    list_editable = ('display_order', 'visible', 'in_aaq')
     list_filter = ('product', 'parent', 'slug')
```
willkg: Would you ever want to filter on if they're in the aaq or not?

kitsune/products/models.py
```diff
@@ -134,6 +134,9 @@ class Topic(ModelBase):
 
     # Whether or not this topic is visible in the ui to users.
     visible = models.BooleanField(default=False)
+    # Whether or not this topic is used in the AAQ.
+    in_aaq = models.BooleanField(
+        default=False, help_text='Whether this topic is shown to users in the AAQ or not.')
```
willkg: Generally, help text should get localized. It shows up in forms generated from model forms. Should this get localized?

willkg:
There's no tests with this. Some of it is done in django-land and some in react-land. Does it make sense to verify that only products marked for `in_aaq` show up in that view?

kitsune/products/models.py
mythmon: Probably... Do you happen to know how this interacts with Django migrations being picky about helptext?

mythmon: And by "being picky" I mean how Django migrations take help text into account in the migrations and how changing the helptext creates new migrations.

mythmon:
I added some tests to the Django side of this. The changes on the React side are unrelated, and mainly make it look a little prettier. I don't think they need tests right now.

According to the Django docs, tthere should be an easier way of doing this, by investigating `request.context`. That seems to not play nicely with Jingo, and throws errors.

kitsune/products/models.py
willkg: You have to do a new migration every time you change the help text.

mythmon: And Django would correctly deal with the call to gettext?

willkg: I don't know. If it's wrapped in gettext, it should be a lazy gettext. It could be that it grabs the msgid. I don't have any of the model help text fields in Fjord gettext'd.

mythmon: Ok. I'll give it a shot.

kitsune/products/admin.py
mythmon: Since `visible` isn't one of the filter properties, I think `in_aaq` should be the same.

mythmon:
Fixes ^

kitsune/sumo/tests/__init__.py
```diff
@@ -180,3 +182,10 @@ def eq_msg(a, b, msg=None):
     """Shorthand for 'assert a == b, "%s %r != %r" % (msg, a, b)'
     """
     assert a == b, (str(msg) or '') + ' (%r != %r)' % (a, b)
+
+
+class FuzzyUnicode(factory.fuzzy.FuzzyText):
+    """A FuzzyText factory that contains at least one non-ASCII character."""
+
+    def fuzz(self):
+        return u'đ{}'.format(super(FuzzyUnicode, self).fuzz())
```
willkg: Clever!

willkg:
Travis failures, but otherwise this looks ok to me.

One thing you might do if you're doing waffle-related tests is to write a decorator for setting up the waffle value. I did that with Fjord, though looking at the code I have in Fjord, I think I could probably reduce it.

mythmon:
Grumble. These tests pass for me locally. I'll see if I can fix it.

mythmon:
Well, I figured it out. Travis tests after merging to master. On master, there is a migration that adds an product for iOS (so we could safely add it to the AAQ). But that product doesn't have a image, since it was really just a safety net for the AAQ.

The API serializers get unhappy when products don't have images. This must be the first test that exercises the serializers in a way that finds the iOS product added by the migrations.

I'll see if I can figure out how to work around this tomorrow. It might be nice if the API serializers didn't blow up when there is no image. That would make a bunch of things nicer. If that doesn't work, I'll try and give the iOS product an image. If *that* doesn't work, I'll just delete all the products before the problematic tests start.

mythmon:
Ha! I fixed it.

kitsune/products/models.py
```diff
@@ -134,6 +135,9 @@ class Topic(ModelBase):
 
     # Whether or not this topic is visible in the ui to users.
     visible = models.BooleanField(default=False)
+    # Whether or not this topic is used in the AAQ.
+    in_aaq = models.BooleanField(
+        default=False, help_text=_lazy(u'Whether this topic is shown to users in the AAQ or not.'))
```
willkg: Out of curiosity, do you want to default these all to False? Would that break what's currently in production?

kitsune/sumo/tests/__init__.py
```diff
@@ -180,3 +184,91 @@ def eq_msg(a, b, msg=None):
...
+        return _give_me_waffles
+
+    def make_flag(self):
+        """Ensure that the flag is created and has the correct values."""
+        Flag.objects.update_or_create(name=self.flagname, defaults=self.kwargs)
+
+    def restore_flag(self):
+        """Ensure that the flag is reset back to its original value."""
+        Flag.objects.filter(name=self.flagname).delete()
+        self.origflag.save()
```
willkg: I think this fails if `self.origflag` is None.

willkg: Also, I'm not sure what this does in the case where you really want to update it back to the original version since the id is a sequence. Does `.save()` here create a new `Flag` with a new id?

mythmon: Oh, this should definitely guard on `self.origflag` being None. It should look like

```python
if self.origflag is not None:
    self.origflag.save()
```

My intent here was to have an easy way to restore the old `Flag`. I suspect that calling `.delete()` and then `.save()` would keep the id intact, but maybe it doesn't. I should test that.

mythmon: Experimentation shows that `Flag.id` is cleared with I call `oldflag.delete()` and a new id is assigned after calling `oldflag.save()` again. That's unfortunate. Does this seem like something to worry about?

This decorator is different than the Fjord one in that it allows setting any field on the mocked flag (since it do `Flag.objects.update_or_create(name=self.flagname, defaults=self.kwargs)`). That's the main reason why I did the `oldflag`/`newflag` thing, not to keep ids the same. But if ids staying the same is a useful property, I bet that I could save the id separately and restore that too. Does that sound good/reasonable/safe to you?

willkg: I think that's fine.

I did notice the decorator was different in that way. The only usecase I had was to set everyone to True or False. I can't think of a usecase mine doesn't serve, but what you have is fine.

kitsune/products/models.py
mythmon: This is only used in the new AAQ, not the old one, so it isn't going to have any effect on the actively used part of production. I wanted it to be consistent with `visible`.

willkg:
I think this looks ok. r+ if travis is cool with it.

mythmon:
Awesome! Thanks for all the review on this, it turned out to be a lot bigger than I expected.

PR #2634, opened 2015-08-27
[bug 1152816] [bug 1152547] Add Firefox for iOS to AAQ

rehandalal:
r?


kitsune/questions/config.py
```diff
@@ -158,6 +158,45 @@
             }),
         ])
     }),
+    ('ios', {
```
mythmon: I'll be so happy when we can ditch all this.

rehandalal: +1

mythmon:
This works, but it is hiding a bit of a foot-gun. If we do this and then don't add a `QuestionLocale` object for iOS in English it gets all screwy with infinite redirects. Can we add a migration to add that object if it doesn't exist?

rehandalal:
updated ^

mythmon:
Travis failed, your migration conflicts with the one that landed for the `sr-Cyrl` → `sr` migration. (yay for Travis testing the result of merges!)

Besides that, it looks fine.

mythmon:
Actually, Joni *just* got back to us with a new list of topics to use. Sorry about that. She is adding them to the admin on prod now, so it should work fine.

rehandalal:
fixed ^

rehandalal:
oh, er... still have to add topics.

rehandalal:
updated ^

mythmon:
Travis is super grumpy today.

kitsune/questions/migrations/0006_ios_questionlocale.py
```diff
@@ -0,0 +1,29 @@
...
+    Product = apps.get_model('products', 'Product')
+    QuestionLocale = apps.get_model('questions', 'QuestionLocale')
+
+    p = Product.objects.get_or_create(slug='ios', defaults={
+        'title': 'Firefox for iOS',
+        'description': 'Firefox for iPhone, iPad and iPod touch devices',
+        'display_order': 0,
+        'visible': False})
+
+    QuestionLocale.objects.get_or_create(locale='en-US', product=p)
```
mythmon: Products is a many-to-many field. You'll have to check the return value of get_or_create (which is `obj, was_created`), and add `p` the new object if it was created.

rehandalal:
FIXED FOR REALZ ^

mythmon:
:laughing:

That test assumed we started with zero products, but you added one in the migration.

Man, this patch. It was supposed to be so easy!

rehandalal:
FML! if this doesn't fix it i'm going to lose it.

mythmon:
:tada:. r+, but I think we should hold of merging this until it is closer to Monday, in case we want to land something else between now and then.

PR #2640, opened 2015-09-03
[Bug 1201656] Fix slug of Firefox iOS in AAQ.

mythmon:
The slug of the Firefox for iOS product that was created in the DB was `fxios`, not `ios` like we expected. This updates the AAQ to deal with that.

There is also another problem: the data migration that created the `ios` product. Options I thought of

1. ~~Delete it.~~ Can't delete it, because it has been deployed.
2. Nullify it. That is, leave it in place but make it a no-op.
3. Add another migration to move the `ios` product to be `fxios`, but only if `fxios` doesn't exist and otherwise delete the `ios` product.

3 sounds better in theory. It also sounds hard and not very useful. I picked 2 instead.

@willkg r?

### DIFF
--- a/kitsune/questions/config.py
+++ b/kitsune/questions/config.py
@@ -158,12 +158,12 @@ products = SortedDict([
             }),
         ])
     }),
-    ('ios', {
+    ('fxios', {
         'name': _lazy(u'Firefox for iOS'),
         'subtitle': '',
         'extra_fields': ['ff_version', 'os', 'plugins'],
-        'tags': ['ios'],
-        'product': 'ios',
+        'tags': ['fxios'],
+        'product': 'fxios',
         'categories': SortedDict([
             ('install-and-update-firefox-ios', {
                 'name': _lazy(u'Install and Update'),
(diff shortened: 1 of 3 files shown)


willkg:
Does this fix the data that's not correct that's currently in all the environments or is the plan to do that by hand?

mythmon:
This is changing the code to match the data that already exists in prod (and I guess stage? I didn't check). There is also an extra product with the wrong slug floating around on stage and prod, because of the migration. I plan to delete those by hand. Does that sound like a terrible way to go about this?

willkg:
If you think you can do that confidently, I think that's ok.

mythmon:
Ok. I'm pretty sure that even if I do nothing, it won't break anything to have the extra product, and it is a pretty easy change.

Travis failed because I forgot to update a test. Fix incoming.

willkg:
I think this is fine. If you want, do a test deploy on your -stage environment before merging it and see if it's good to go.

mythmon:
Um. Huh. I was wrong about the slugs. Something else is going on, and for some reason changing this config fixes partially fixes it. Either way, this PR is wrong.

PR #2644, opened 2015-09-07
[Bug 1202377] Knowledge Base Overview page structe fixed

anushbmx:
The options were out of the grid which causes the issue. Moved them inside the grid.


rehandalal:
good spot. thanks!

PR #2645, opened 2015-09-08
[bug 1193441] Fix styling for new AAQ

rehandalal:
r?

### DIFF
--- a/kitsune/questions/static/questions/js/components/AAQApp.jsx
+++ b/kitsune/questions/static/questions/js/components/AAQApp.jsx
@@ -8,6 +8,7 @@ import TitleContentEditor from './TitleContentEditor.jsx';
 import UserAuth from './UserAuth.jsx';
 import SubmitQuestion from './SubmitQuestion.jsx';
 import TroubleshootingDataStore from '../stores/TroubleshootingDataStore.es6.js';
+import {authStates} from '../../../users/js/constants/UserAuthConstants.es6.js';
 
 export default class AAQApp extends React.Component {
   constructor(props) {
@@ -39,17 +40,30 @@ export default class AAQApp extends React.Component {
       validationErrors: QuestionEditStore.getValidationErrors(),
       questionState: QuestionEditStore.getState(),
       userAuth: UserAuthStore.getAll(),
-      troubleshooting: TroubleshootingDataStore.getAll(),
+      troubleshooting: TroubleshootingDataStore.getAll()
     };
+  }
+
+  setStep(step) {
+    this.setState({step: step});
   }
 
   render() {
     return (
       <div className="AAQApp">
-        <ProductSelector {...this.state}/>
-        <TopicSelector {...this.state}/>
-        <TitleContentEditor {...this.state}/>
-        <UserAuth userAuth={this.state.userAuth}/>
+        {(() => {
+          switch (this.state.step) {
+            case 'topic':
+              return <TopicSelector {...this.state} setStep={this.setStep.bind(this)} next="title"/>;
+            case 'title':
+              return [
+                <TitleContentEditor {...this.state}/>,
+                <UserAuth userAuth={this.state.userAuth}/>
+              ];
+            default:
+              return <ProductSelector {...this.state} setStep={this.setStep.bind(this)} next="topic"/>;
+          }
+        })()}
         <SubmitQuestion {...this.state}/>
       </div>
     );
--- a/kitsune/questions/static/questions/js/components/AAQApp.jsx
+++ b/kitsune/questions/static/questions/js/components/AAQApp.jsx
@@ -8,7 +8,6 @@ import TitleContentEditor from './TitleContentEditor.jsx';
 import UserAuth from './UserAuth.jsx';
 import SubmitQuestion from './SubmitQuestion.jsx';
 import TroubleshootingDataStore from '../stores/TroubleshootingDataStore.es6.js';
-import {authStates} from '../../../users/js/constants/UserAuthConstants.es6.js';
 
 export default class AAQApp extends React.Component {
   constructor(props) {
@@ -45,7 +44,7 @@ export default class AAQApp extends React.Component {
   }
 
   setStep(step) {
-    this.setState({step: step});
+    this.setState({step});
   }
 
   render() {
--- a/kitsune/questions/static/questions/js/components/AAQStep.jsx
+++ b/kitsune/questions/static/questions/js/components/AAQStep.jsx
@@ -4,11 +4,8 @@ import scrollTo from '../../../sumo/js/utils/scrollTo.es6.js';
 export default class AAQStep extends React.Component {
   render() {
     let heading = this.heading();
-    let style = {
-      display: this.visible() ? 'block' : 'none'
-    };
     return (
-      <div className="AAQApp__Step" style={style}>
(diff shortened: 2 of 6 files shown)


mythmon:
Travis's failure looks related.

kitsune/questions/static/questions/js/components/AAQStep.jsx
```diff
@@ -4,8 +4,11 @@ import scrollTo from '../../../sumo/js/utils/scrollTo.es6.js';
 export default class AAQStep extends React.Component {
   render() {
     let heading = this.heading();
+    let style = {
+      display: this.visible() ? 'block' : 'none'
```
mythmon: Generally instead of using CSS to hide/show content, it works really well to simply return `null`, telling React to put nothing here.

mythmon:
Oh, wow. I spent 5 minutes trying to figure out why the waffle flag wasn't working any more. Then I realized that this patch makes the product selector look identical to the old AAQ.

kitsune/questions/static/questions/js/components/AAQStep.jsx
```diff
@@ -28,15 +31,21 @@ export default class AAQStep extends React.Component {
...
     if (nextStep === null) {
-      throw 'Tried to call scrollToNextStep(), but this is the last step.';
+      throw 'Tried to call switchToNextStep(), but this is the last step.';
     }
     if (!nextStep.classList.contains('AAQApp__Step')) {
-      throw 'Tried to call scrollToNextStep(), but the next element is not an AAQStep.';
+      throw 'Tried to call switchToNextStep(), but the next element is not an AAQStep.';
     }
+    nextStep.style.display = 'block';
+    currentStep.style.display = 'none';
```
mythmon: This kind of manipulation generally isn't safe in React. If the components re-render, this will easily be erased. Anything that affects the DOM like this should be in a `render()` method.

I think a better approach would be to have a state variable that lists the current active step, and render just that one instead of manipulating the visibility of each step.

rehandalal:
feedbacked ^

kitsune/questions/static/questions/js/components/AAQApp.jsx
```diff
@@ -39,17 +40,30 @@ export default class AAQApp extends React.Component {
       validationErrors: QuestionEditStore.getValidationErrors(),
       questionState: QuestionEditStore.getState(),
       userAuth: UserAuthStore.getAll(),
-      troubleshooting: TroubleshootingDataStore.getAll(),
+      troubleshooting: TroubleshootingDataStore.getAll()
```
mythmon: Trailing commas are actually fine in ES6 code. ES5 defined trailing commas in objects and lists as OK.

kitsune/questions/static/questions/js/components/AAQApp.jsx
```diff
@@ -39,17 +40,30 @@ export default class AAQApp extends React.Component {
...
+              return <TopicSelector {...this.state} setStep={this.setStep.bind(this)} next="title"/>;
+            case 'title':
+              return [
+                <TitleContentEditor {...this.state}/>,
+                <UserAuth userAuth={this.state.userAuth}/>
+              ];
+            default:
+              return <ProductSelector {...this.state} setStep={this.setStep.bind(this)} next="topic"/>;
+          }
+        })()}
```
mythmon: Wow. That's a lot of parenthesis. I've never seen a iife put into a JSX context before, but it works pretty well.

I think I probably would have put this outside the JSX like below, but this works fine too. I wouldn't hvae thought to do it this way

```javascript
render() {
  let step;
  switch (this.state.step) {
    case 'topic':
      step = ...;    
    ...
  }

  return (
    <div className="AAQApp">
      <step/>
      <SubmitQuestion {...this.state}/>
    </div>
  }
```

kitsune/questions/static/questions/js/components/AAQApp.jsx
```diff
@@ -39,17 +40,30 @@ export default class AAQApp extends React.Component {
...
     return (
       <div className="AAQApp">
-        <ProductSelector {...this.state}/>
-        <TopicSelector {...this.state}/>
-        <TitleContentEditor {...this.state}/>
-        <UserAuth userAuth={this.state.userAuth}/>
+        {(() => {
+          switch (this.state.step) {
+            case 'topic':
+              return <TopicSelector {...this.state} setStep={this.setStep.bind(this)} next="title"/>;
```
mythmon: You could do `<TopicSelector {...this.state} nextStep={this.setStep.bind(this, "title")}/>`. Binding the argument into the parameter makes the child component simpler, since it just has a single function to call to move to the next step.

kitsune/questions/static/questions/js/components/ProductSelector.jsx
```diff
@@ -6,6 +6,10 @@ import AAQActions from '../actions/AAQActions.es6.js';
 const products = JSON.parse(document.querySelector('.data[name=products]').innerHTML);
 
 export default class ProductSelector extends AAQStep {
+  visible() {
+    return true;
+  }
```
mythmon: I don't think you need this method anymore.

kitsune/questions/static/questions/js/components/AAQApp.jsx
```diff
@@ -39,17 +40,30 @@ export default class AAQApp extends React.Component {
       validationErrors: QuestionEditStore.getValidationErrors(),
       questionState: QuestionEditStore.getState(),
       userAuth: UserAuthStore.getAll(),
-      troubleshooting: TroubleshootingDataStore.getAll(),
+      troubleshooting: TroubleshootingDataStore.getAll()
     };
   }
 
+  setStep(step) {
+    this.setState({step: step});
```
mythmon: ES6 trick: This is can be shortened to this.setState({step});. When the key name and the name of variable for the value are the same, you only have to write it once.

kitsune/questions/static/questions/js/components/AAQApp.jsx
```diff
@@ -8,6 +8,7 @@ import TitleContentEditor from './TitleContentEditor.jsx';
 import UserAuth from './UserAuth.jsx';
 import SubmitQuestion from './SubmitQuestion.jsx';
 import TroubleshootingDataStore from '../stores/TroubleshootingDataStore.es6.js';
+import {authStates} from '../../../users/js/constants/UserAuthConstants.es6.js';
```
mythmon: Do you use this import? I don't see it anywhere.

mythmon:
The registration form looks a bit off. Probably add some padding? Or get rid of the white background entirely? 

shot_20057

mythmon:
I didn't see the article suggestions as I was typing. Where did those go?

rehandalal:
updated ^

mythmon:
This looks fine, I think we should merge it this.

It has some issues though. It is really confusing with the back button. There is no way to go backwards in the process, and if you do hit back you lose all your progress. This should probably use push state to track where you are in the progress. That can be another pr/bug though. It also was a bit confusing when the submit button disappeared and didn't do anything at first. It took a few seconds to submit. The old design had some status indication, and it was easy to go back by scrolling.

Would you like to do those in this PR? or in another?

rehandalal:
Let's land this and I'll file a bug for follow up

PR #2647, opened 2015-09-11
[Bug 1203633] Switch to Mocha for JS tests.

mythmon:
Qunit can't deal with ES6. That leaves us in a rough spot testing anything written ES6, like all this new React stuff we've done. Mocha can handle ES6.

Plus, I was able to convince Mocha to run outside the browser, but still mimic being in a browser for the important bits. It makes running the tests a lot nicer. It has a downside though: We can't test browser specific issues. We only get V8 for JS, and we don't even technically have a rendering engine.

Another alternative would be to use Karma as a test runner. Karma can wrap Mocha, and provides a much nicer way to run the tests in a browser than our hacked up Qunit+Selenium. It would be more work though, and I think running things outside the browser has more pros than cons. Thoughts? Should we keep this, or put it back in a browser and use Karma?

This meant rewriting all the tests to use the Mocha style, which was kind of tedious. It makes this patch really big, so sorry about that.

r?

PS: I still want to do some more with this, but didn't want to bloat this PR even more.

* Better test harnesses. The fixtures thing here works already, but it could be better.
* Actually test the React and ES6 code.
* I had to install a lot of our "frontend" dependencies via NPM to get this to work. I'd like to finish that conversion at some point.


rehandalal:
phew!

rehandalal:
after two passes, it looks good to me and the tests pass. so let's merge!

rehandalal:
ps: this is awesome!

PR #2650, opened 2015-09-14
Home during deploys

mythmon:
Adding jsdom to the NPM dependencies, which I did in #2647, revealed an oddity in our deployment infrastructure. We don't have the `$HOME` environment variable set during deploys. node-gyp needs this to work, for some reason, and jsdom uses node-gyp to build compiled dependencies.

Without this patch, stage doesn't deploy. With this patch, it does.

I'm also game for other ways we might fix this. I just can't think of any right now.

r?

### DIFF
--- a/scripts/update/deploy.py
+++ b/scripts/update/deploy.py
@@ -14,10 +14,13 @@ from commander.deploy import task, hostgroups  # noqa
 import commander_settings as settings  # noqa
 
 
-# Setup virtualenv path.
+# Setup environment
 venv_bin_path = os.path.join(settings.SRC_DIR, 'virtualenv', 'bin')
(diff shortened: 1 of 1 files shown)


rehandalal:
Not sure how to test this other than to deploy to stage, but it lgtm.

mythmon:
I really wish we could test this script besides deploying to stage. Maybe we could do something with Vagrant or Docker? Maybe things will get better when we are on AWS in the distant fture. I've actually already deployed this to stage: I used stage to test and develop this it. Maybe I should have used -dev for that?

rehandalal:
/me nods
-dev is probably better. but if you've tested on stage already, then r+. 
/me merge

rehandalal:
d'oh page hadn't updated. this is merged already! yay

PR #2651, opened 2015-09-15
[Bug 1128948] Show "translate article" while requesting with English

safwanrahman:
I think this will fix the issue!
Contributor can localize a article by following pattern
{{ host }}/{{ locale }}/kb/{{ english_article_slug }}/edit

I believe this should fix the issue! @mythmon r?

### DIFF
--- a/kitsune/wiki/tests/test_views.py
+++ b/kitsune/wiki/tests/test_views.py
@@ -20,7 +20,7 @@ from kitsune.wiki.config import (
 from kitsune.wiki.models import Document, HelpfulVoteMetadata, HelpfulVote
 from kitsune.wiki.tests import (
     doc_rev, document, helpful_vote, new_document_data, revision,
-    translated_revision, TemplateDocumentFactory, RevisionFactory)
+    translated_revision, TemplateDocumentFactory, RevisionFactory, DocumentFactory)
 from kitsune.wiki.views import (
     _document_lock_check, _document_lock_clear, _document_lock_steal)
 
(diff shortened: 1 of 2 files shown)


mythmon:
I think this change should work, but you need to add a test for it.

safwanrahman:
@mythmon I cant understand why this Test is failing. not related my code. maybe a travis problem. Can you please check?

mythmon:
That test is a timeout when trying to download the NPM dependencies. That's generally out of your control. I'll restart the build.

kitsune/wiki/tests/test_views.py
```diff
@@ -369,6 +369,16 @@ def test_new_document_slugs(self, get_current):
                                                args=[data['slug']],
                                                locale='en-US'))
 
+    def test_translate_with_parent_locale_slug(self):
+        doc = document(locale=settings.WIKI_DEFAULT_LANGUAGE, save=True)
```
mythmon: New tests should prefer to use `DocumentFactory` if possible.

kitsune/wiki/tests/test_views.py
```diff
@@ -369,6 +369,16 @@ def test_new_document_slugs(self, get_current):
...
+        trans_doc = document(parent=doc, locale='bn-BD', slug='bn-BD_slug', save=True)
+        url = reverse('wiki.edit_document', args=[doc.slug], locale=trans_doc.locale)
+        response = self.client.get(url)
+        assert response.status_code == 200
```
mythmon: Don't use bare assert statements, use the `eq_` helpfer from nose. It produces easier to debug tests.

safwanrahman:
@mythmon updated the tests! 
Have a look! :)

safwanrahman:
tests pass! :+1: :fireworks:

mythmon:
Can you add a second negative test that tests the case where there is no matching document, both in the case that there is no English document, and in the case that there is no translated document under another slug?

safwanrahman:
@mythmon Added another 2 tests. Thanks for this
Found a bug in my code while writing tests. So changed the code also. Now its totally perfect I believe
r?

safwanrahman:
@mythmon r?

kitsune/wiki/views.py
```diff
@@ -385,8 +385,25 @@ def steal_lock(request, document_slug, revision_id=None):
...
+        # Check if the document slug is available in default language.
+        parent_doc = get_object_or_404(Document, locale=settings.WIKI_DEFAULT_LANGUAGE,
+                                       slug=document_slug)
+        # If the document is available in default language, show the user the translation page
+        # of the requested locale
+        translation = parent_doc.translated_to(request.LANGUAGE_CODE)
+        # If the document is translated into the requested locale, show them the edit article
+        # page of that translated document
+        if translation:
+            doc = Document.objects.get(locale=request.LANGUAGE_CODE, slug=translation.slug)
```
mythmon: This seems like it is just re-fetching the same document from DB again. How is `Document.objects.get(locale=request.LANGUAGE_CODE, slug=translation.slug)` different from `translation`?

safwanrahman: Oh! I did not notice it! I just gone by as usual I do to fetch a document!
Thanks :+1: 
I am fixing it!

safwanrahman:
@mythmon Fixed the issue and squashed it into one!
This is ready to merge!

mythmon:
Looks good. Once Travis passes, I'll merge this.

PR #2652, opened 2015-09-15
[Bug 1201560] Show Last Update date in Mobile view

safwanrahman:
@mythmon A small fix.
r?

### DIFF
--- a/kitsune/wiki/templates/wiki/mobile/document.html
+++ b/kitsune/wiki/templates/wiki/mobile/document.html
@@ -44,6 +44,9 @@
         {{ _('Share this article: {link}')|f(link=link|safe) }}
       </p>
     {% endif %}
+    <p>
+      Last Update: {{ latest_approved_revision.created.date() }}
+    </p>
 
     {% if not waffle.switch('hide-voting') %}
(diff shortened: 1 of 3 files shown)


kitsune/wiki/templates/wiki/mobile/document.html
```diff
@@ -44,6 +44,9 @@
         {{ _('Share this article: {link}')|f(link=link|safe) }}
       </p>
     {% endif %}
+    <p>
+      Last Update: {{ latest_approved_revision.created.date() }}
```
rehandalal: you can just change this to:
```
document.current_revision.created.date()
```

and then undo all the changes to `wiki/views.py`

safwanrahman:
Thanks @rehandalal. My think was bit critical though it can be done easily! :+1:

safwanrahman:
This should fix! :+1: r?

rehandalal:
code looks good, but looking at the page I think it should actually go under the list of people that helped write the article and it should be lighter in color `color: #888;`.

safwanrahman:
@rehandalal r?

rehandalal:
good spot on the l10n. this is great, thanks!

PR #2654, opened 2015-09-17
Remove django_qunit from settings_local.py.dist

willkg:
We stopped using qunit recently. This removes it from
settings_local.py.dist so it doesn't throw errors in new development
environments.

Quick r?

Also, we should send an email to the developer list telling people about the switch to the new JS test system and also that they should update their `settings_local.py` and remove references to `django-qunit`.


rehandalal:
makes sense, r+

BReduardokramer:
Ok. I  will fix it now. But i go to WhatsApp. Any problem, send to us an email. Thanks.

R&D <notifications@github.com> escreveu:

makes sense, r+

PR #2655, opened 2015-09-22
[Bug 1204627] smoketests

willkg:
Adds smoketest infrastructure and rewrites the language-picker test as a smoketest. This also removes the selenium test case stuff in the normal tests.

r?


smoketests/README.rst
```diff
@@ -0,0 +1,145 @@
...
+
+
+Setting up to run the tests
+===========================
+
+I suggest using a different virtual environment for these tests than the
+rest of Kitsune so you're not mixing requirements::
+
+    $ mkvirtualenv kitsune-smoketests
+    $ pip install -r requirements.txt
```
mythmon: Is mkvirtualenv a real thing? Is that part of virtualenv wrapper? I'd vote that we simply leave off the command part of this and say something like 

> I suggest using a different virtual environment for these tests than the
> rest of Kitsune so you're not mixing requirements. The requirements are
> listed in requirements.txt.

smoketests/pages/index.py
```diff
@@ -0,0 +1,24 @@
+# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this
+# file, You can obtain one at http://mozilla.org/MPL/2.0/.
```
mythmon: Do we need the license stanza in these files? Or is having `smoketests/LICENSE` god enough?

smoketests/tests/sumo/test_index.py
```diff
@@ -0,0 +1,30 @@
...
+# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this
+# file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+import pytest
+
+from pages.index import IndexPage
+
+
+class TestIndex:
```
mythmon: Should this subclass something like TestCase or object?

smoketests/tests/__init__.py
```diff
@@ -0,0 +1,7 @@
+import time
+
+
+class TestCase(object):
+    def take_a_breather(self):
```
mythmon: Cute name. I don't see where this is used though?

smoketests/pages/index.py
```diff
@@ -0,0 +1,24 @@
...
+
+class IndexPage(Page):
+    _page_title = 'Mozilla Support'
+    _url = '/'
+
+    _locale_picker = (By.CSS_SELECTOR, '.locale-picker')
+
+    @property
+    def current_language(self):
+        return self.selenium.find_element(*self._locale_picker).text
```
mythmon: Caveat: I don't have a great understanding of how mozwebqa test work. This property seems useful, and also applicable to just about any Sumo page. Should there be some SumoPage super class with useful things like that? Or is that not really how these Page classes work?

mythmon:
I have a couple nits, but nothing major. I ran through the instructions, and the tests worked fine for me. There is a merge conflict now that needs fixing.

smoketests/README.rst
willkg: Yeah... It's part of virtualenvwrapper. I'll fix the text.

smoketests/pages/index.py
willkg: It's good practice to add that header to the top of every file covered under the MPL v2.

> Q4: I want to use the Mozilla Public License for software that I have written. What do I have to do?
>
> To apply the Mozilla Public License to software that you have written, add the header from Exhibit A of the license to each source code file in your project. Sample headers for various commenting styles are available here. You may also add additional accurate notices of copyright ownership, such as the name of the copyright holder, but this is not necessary.

https://www.mozilla.org/en-US/MPL/2.0/FAQ/

smoketests/pages/index.py
willkg: Yeah, you could do that. I mostly threw enough together to move that language picker test over. I didn't want to spend much time making architecture decisions for a future I didn't think I knew much about.

smoketests/tests/__init__.py
willkg: This entire class isn't used. We can nix it.

smoketests/tests/sumo/test_index.py
willkg: Yeah... so pytest supports unittest TestCase, but it gets ornery with fixtures and (imo) doesn't work correctly. Recently, I've decided to nix the TestCase hierarchy and just have it bare.

Also, I used to subclass object because that's how you get new-style objects with Python 2 and that helps with some things. In this case, you don't need anything with new-style objects and the pytest docs have examples like this:

http://pytest.org/latest/getting-started.html#grouping-multiple-tests-in-a-class

If we ever switch to Python 3 (probably unlikely with kitsune, but who knows), everything is new-style.

Anyhow, I could go either way, so let me know what you want me to do.

willkg:
^^^ Rebased against master to fix merge issue. Also, tweaked README text as suggested and removed some dead code.

mythmon:
:+1:

PR #2656, opened 2015-09-22
[bug 1206918] Nix SQLAlchemy

willkg:
This removes sqlalchemy from the requirements file. As near as I can
tell, it's not used by kitsune at all.

Quick r?


rehandalal:
i couldn't find it anywhere either... so r+

PR #2657, opened 2015-09-22
[bug 1207323] Update setuptools to 18.3.2

willkg:
If Travis is Green, that's cool. However, the bigger concern is that this touches deployments and it's possible that our admin nodes are using something something that doesn't work with this version of setuptools. Ergo, we'll kind of be testing this by pushing it to -stage.

Anyhow.... r?


willkg:
I'll do a test deploy of this to -stage and if it's fine, then I'll merge it.

willkg:
I did a test deploy to stage and it looked fine, so I'm going to merge this.

PR #2658, opened 2015-09-22
[bug 1207330] Upgrade requests to 2.7.0

willkg:
Travis will run the tests, but I'm not sure what else we should do to make sure this is kosher for all the things that uses is. Looks like notifications, bitly and surveygizmo, but I'm not sure whether those things are tickled via the tests or not.

Ideas? Review?


willkg:
I want to wait on @mythmon before we land this in case things go south.

mythmon:
I don't think any of these things are tickled by the tests. Generally we mock out requests or design things to isolate the things we test away from the parts we use requests.

I can test bitly manually though, so I'll try that locally.

mythmon:
Bitly worked fine with this updated version. I read through the changelog and didn't see anything interesting. I think we are fine here.

PR #2659, opened 2015-09-22
[bug 1207320] Remove logilab-common and logilab-astng

willkg:
I don't see anything that uses either of these requirements, so let's
remove them.

"logilab" doesn't show up anywhere, so this should be pretty safe. If Travis turns green, we're probably fine.


willkg:
Travis is green. The sky is blue.
I'm going to land this PR. Woo hoo!

PR #2661, opened 2015-09-23
[bug 702276] Upgrade lxml to 3.4.4 and pyquery to 1.2.9

willkg:
This upgrades lxml to 3.4.4. In order to do that, I also had to upgrade
pyquery and pull in cssselect.

I think there's a bug in the cssselect handling for the :first
pseudo-element. That breaks a few of our tests. It kind of looks like
both pyquery and cssselect are dead-ish projects and aren't being
actively maintained, so at some point we might want to look at replacing
pyquery with something else like BeautifulSoup4.

I talked with Rehan about this a bit and decided that since it's just a
few tests that are busted, I'd just subclass PyQuery and provide a
.first() method.

The other thing I discovered when doing this is that the messages page
has changed significantly since 2011 and one of the tests not only fails
now, but it doesn't appear to be a valid test anyhow, so I removed it.

r?

### DIFF
--- a/kitsune/forums/tests/test_templates.py
+++ b/kitsune/forums/tests/test_templates.py
@@ -1,13 +1,13 @@
 from django.contrib.contenttypes.models import ContentType
 
 from nose.tools import eq_
-from pyquery import PyQuery as pq
 
 from kitsune.access.tests import permission
 from kitsune.forums.models import Post
 from kitsune.forums.tests import (
     ForumTestCase, forum, thread, post as forum_post)
 from kitsune.sumo.tests import get, post
+from kitsune.sumo.tests import SumoPyQuery as pq
 from kitsune.users.tests import user, group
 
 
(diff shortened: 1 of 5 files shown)


willkg:
I want to wait on @mythmon before we land this in case things go south.

mythmon:
Changes here look good. A shame about `:first`. I'm glad we didn't use it in any more complex situations.

I skimmed the changelog for lxml. The only thing I found that made me nervous was

> * Requires libxml2 2.7.0 or later and libxslt 1.1.23 or later, use lxml 3.3.x with older versions.

I checked on the admin node, and I think it has libxml2 2.7.6 and libxslt 1.1.26, so we should be good. I assume all our servers have those in common. We should watch this change carefully once it is on stage though.

willkg:
As far as I could tell, lxml gets used in the test suite and the wiki parser and that's it. So I think we just need to verify the wiki parser is working on -stage.

PR #2663, opened 2015-09-26
[bug 1116489] Tweak manage.py to be more helpful on errors

willkg:
If you don't have your virtualenv activated, then manage.py sputters and
squawks and between muscle spasms of its right eye rants about
conspiracies involving django.forms which is unintelligible gibberish,
not helpful and not friendly to kitsune developers.

This fixes that by catching the ImportError and giving you a more
helpful hint.

r?

### DIFF
--- a/manage.py
+++ b/manage.py
@@ -1,6 +1,7 @@
 #!/usr/bin/env python
 import os
 import sys
+import traceback
 
 # Now we can import from third-party libraries.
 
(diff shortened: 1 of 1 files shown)


willkg:
Is there better more helpful text to provide? I was thinking we could link to the documentation, but it's kind of all over the place and there isn't really a spot that focuses on virtual environments.

Also, when the virtual environment is activate, there's a `VIRTUAL_ENV` environment variable. We could check for that and thus differentiate between "virtual environment is not activated" and "virtual environment is activated, but one or more packages isn't installed". Is that a better thing to do?

mythmon:
This works for me. Though I had to jump through some hoops to find a way to run manage.py *without* activating my virtualenv, thanks to my environment helping out.

I think checking for `VIRTUAL_ENV` is a good idea. I have definitely had contributors who had a correct virtualenv set up, but didn't have a correct set of dependencies.

willkg:
Ok. I'll tweak it some more.

willkg:
^^^ Rebased against master and tweaked the ImportError message to account for `VIRTUAL_ENV` in the environment.

Does that look good?

willkg:
Travis failed. Restarted it because it's a bunch of ephemeral ES errors.

mythmon:
LGTM, thanks.

PR #2664, opened 2015-09-28
[bug 1206854] Make API error messages localizable

rehandalal:
r?


kitsune/users/api.py
```diff
@@ -302,7 +302,7 @@ def generate(self, request, **kwargs):
         Generate a user with a random username and password.
         """
         # The loop counter isn't used. This is an escape hatch.
-        for _ in range(10):
+        for __ in range(10):
```
willkg: I sometimes use "throwaway" instead of `_` because of the clash with gettext.

willkg:
There aren't any tests here. How do we know it works and/or how do we test it?

mythmon:
I don't think this is a very good way to fix this problem. Sumo's API in general isn't localized, so relying on strings from it in a localized UI probably isn't the best idea. I think we need a better plan for how we are going to make this issue better across the entire API before we can do this.

r-

PR #2665, opened 2015-09-30
Re-indent AAQ view and Question.auto_tag to use longer lines.

mythmon:
I was reading through these functions for [Bug 1195859](https://bugzilla.mozilla.org/show_bug.cgi?id=1195859). It made it easier to read through by re-indenting them to take advantage of longer lines, so I did that.

It followed me home, can we keep it?

### DIFF
--- a/kitsune/questions/models.py
+++ b/kitsune/questions/models.py
@@ -244,18 +244,17 @@ class Question(ModelBase, BigVocabTaggableMixin, SearchMixin):
         You don't need to call save on the question after this.
 
         """
-        to_add = (self.product_config.get('tags', []) +
-                  self.category_config.get('tags', []))
-
+        to_add = (self.product_config.get('tags', []) + self.category_config.get('tags', []))
         version = self.metadata.get('ff_version', '')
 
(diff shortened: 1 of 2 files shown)


kitsune/questions/models.py
```diff
@@ -244,18 +244,17 @@ def auto_tag(self):
         You don't need to call save on the question after this.
 
         """
-        to_add = (self.product_config.get('tags', []) +
-                  self.category_config.get('tags', []))
-
+        to_add = (self.product_config.get('tags', []) + self.category_config.get('tags', []))
```
willkg: Nit: Now that this is on one line, you can ditch the parens if you want.

willkg:
This looks fine to me. r+!

kitsune/questions/models.py
mythmon: Ah, yes. I'll do that. I basically ran around hitting `J` in vim.

mythmon:
/me waits for Travis, and then hits the :green_apple: button.

PR #2667, opened 2015-10-02
Update dennis to 0.7

willkg:
This picks up some new lint rules that SUMO has had problems with
recently.

r?


mythmon:
r+ if travis has no issues.

PR #2668, opened 2015-10-02
[Bug 1210975] Document.related_documents may be blank.

mythmon:
I couldn't figure out how to write a failing test case for this. Any ideas? @willkg ?

r?

### DIFF
--- a/kitsune/wiki/migrations/0005_auto_20151002_1359.py
+++ b/kitsune/wiki/migrations/0005_auto_20151002_1359.py
@@ -1,0 +1,22 @@
+"""Make related_documents optional on Document."""
+# -*- coding: utf-8 -*-
+from __future__ import unicode_literals
+
+from django.db import models, migrations
+import kitsune.sumo.models
+
+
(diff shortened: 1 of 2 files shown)


willkg:
I didn't test this, but it looks right to me. Essentially, it's like a 10 character change with an accompanying migration that shouldn't have any db-related effects. blank is specifically validation-related. So everything seems right here to me.

r_plus

willkg:
To answer the question, you can write tests for the admin. It's like writing any other test for a form. I think you'd have to write the post-data by hand since you're writing post-data that omits a form element. I don't usually write admin tests because my admin stuff is usually limited to just me.

@Osmose probably has better thoughts than I do on this.

mythmon:
It seems really weird to me that the models validate things differently than the admin form. I don't really want to get into testing the admin forms, so I'm going to leave this without test.

Osmose:
> I don't really want to get into testing the admin forms

Straight wisdom

PR #2669, opened 2015-10-06
[bug 1211961] Fix opensearch

willkg:
opensearch was totally busted. Partially from a lint-fixing round in
2014 and partially from upgrading to a new ElasticUtils that "dealt"
with array values.

I fixed those two issues and also redid the template so that it's not
using the Site framework because that makes things difficult to test.

However, it's still got problems:

1. it builds a different search than simple search does and since it's
   in a different place in the code, it's been unmaintained and thus
   things like templates pop up in the results; we should merge the
   two

2. none of the tests test the results of the opensearch search
   suggestions so when it breaks, we have no idea--further, there's
   a bunch of other stuff that should get tested

I didn't want to do either of those now. Instead I tested it manually
with curl. I'll look into doing them later as I work through refactoring
all the Elasticsearch code.

Relatedly, I think we should nix the opensearch altogether since few
people use it (as evidenced by the fact it was busted for sooooo long).

r?


mythmon:
Looks good to me. Thanks a lot for jumping on this!

PR #2670, opened 2015-10-08
[Bug 1149730] added delay to hover transition on search element

shanewignall:


kitsune/sumo/static/sumo/less/main.less
```diff
@@ -1846,6 +1846,8 @@ input[type=button],
   }
 
   &:hover {
+    transition-delay:250ms;
```
rehandalal: nit: could you add a space between `:` and `250ms`

rehandalal:
thanks! r+

PR #2671, opened 2015-10-08
Add pushstate handling to new AAQ.

mythmon:
1. Add an ES6-style shim for some functions in the old school `k` library.
2. Add a shim to treat the URL's query string as a Store.
3. Update the new AAQ to store it's current step in the `UrlStore`.

r?


mythmon:
@rehandalal can you take a look at this? It isn't directly related to any bug, but I think it makes things a lot less surprising to have the back buttons work.

rehandalal:
sweet. r+

PR #2672, opened 2015-10-12
[bug 1212476] Refactor simple search to share code

willkg:
* moves some things over to the SimpleSearchForm
* extracts S building code into simple_search.py:generate_simple_search
  so that it can be re-used between simple_search and
  opensearch_suggestions
* redoes opensearch_suggestions to use the new generate_simple_search

One thing this does is changes how opensearch_suggestions works.
Previously, it would do a search and show the top 5 kb articles followed
by the top 5 questions. That's what simple search used to do eons ago.

This changes that so that it does what simple search does and shows the
top 10 mixed results. This reduces the number of different kinds of
search we're doing by 1 and reduces the likelihood that
opensearch_suggestions gets totally out of sync with the other searches
like it did.

r?

### DIFF
--- a/kitsune/search/views.py
+++ b/kitsune/search/views.py
@@ -28,6 +28,7 @@ from kitsune.search.utils import locale_or_default, clean_excerpt, ComposedList
 from kitsune.search import es_utils
 from kitsune.search.forms import SimpleSearchForm, AdvancedSearchForm
 from kitsune.search.es_utils import ES_EXCEPTIONS, F, AnalyzerS
+from kitsune.search.simple_search import generate_simple_search
 from kitsune.sumo.helpers import Paginator
 from kitsune.sumo.urlresolvers import reverse
 from kitsune.sumo.utils import paginate, smart_int
@@ -59,41 +60,33 @@ EXPIRES_FMT = '%A, %d %B %Y %H:%M:%S GMT'
 
 @mobile_template('search/{mobile/}results.html')
 def simple_search(request, template=None):
-    """ES-specific simple search view.
+    """Elasticsearch-specific simple search view.
 
     This view is for end user searching of the Knowledge Base and
     Support Forum. Filtering options are limited to:
-    * product (`product=firefox`, for example, for only Firefox results)
-    * document type (`w=2`, for esample, for Support Forum questions only)
-    """
 
+    * product (`product=firefox`, for example, for only Firefox results)
+    * document type (`w=2`, for example, for Support Forum questions only)
+
+    """
     # Redirect to old Advanced Search URLs (?a={1,2}) to the new URL.
-    a = request.GET.get('a')
-    if a in ['1', '2']:
+    if request.GET.get('a') in ['1', '2']:
         new_url = reverse('search.advanced') + '?' + request.GET.urlencode()
         return HttpResponseRedirect(new_url)
 
     # JSON-specific variables
     is_json = (request.GET.get('format') == 'json')
     callback = request.GET.get('callback', '').strip()
-    content_type = (
-        'application/x-javascript' if callback else 'application/json')
+    content_type = 'application/x-javascript' if callback else 'application/json'
 
     # Check callback is valid
     if is_json and callback and not jsonp_is_valid(callback):
         return HttpResponse(
             json.dumps({'error': _('Invalid callback function.')}),
-            content_type=content_type, status=400)
+            content_type=content_type,
+            status=400)
 
-    language = locale_or_default(
-        request.GET.get('language', request.LANGUAGE_CODE))
-    r = request.GET.copy()
-
-    # TODO: Do we really need to add this to the URL if it isn't already there?
-    r['w'] = r.get('w', constants.WHERE_BASIC)
-
-    # TODO: Break out a separate simple search form.
-    search_form = SimpleSearchForm(r, auto_id=False)
+    search_form = SimpleSearchForm(request.GET, auto_id=False)
 
     if not search_form.is_valid():
         if is_json:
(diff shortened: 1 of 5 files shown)


willkg:
@mythmon What do you think about this refactoring? Does this make sense architecturally? Are there better ways to break it up and centralize code?

kitsune/search/views.py
```diff
@@ -269,13 +133,11 @@ def simple_search(request, template=None):
...
 
-        # If we know there aren't any results, let's cheat and in
-        # doing that, not hit ES again.
+        # If we know there aren't any results, let's cheat and doing that, not hit ES again.
         if num_results == 0:
             searcher = []
         else:
-            # Get the documents we want to show and add them to
-            # docs_for_page
+            # Get the documents we want to show and add them to # docs_for_page
```
mythmon: You missed the `#` when you joined these lines.

kitsune/search/views.py
```diff
@@ -346,21 +207,17 @@ def simple_search(request, template=None):
...
 
     product = Product.objects.filter(slug__in=cleaned['product'])
     if product:
-        product_titles = [_(p.title, 'DB: products.Product.title')
-                          for p in product]
+        product_titles = [_(p.title, 'DB: products.Product.title') for p in product]
     else:
         product_titles = [_('All Products')]
 
+    # FIXME: This is probably bad l10n.
```
mythmon: We should build something like postatus that shows `# FIXME`, `# TODO` and `# XXX` things. Or maybe that's a terrible idea.

kitsune/search/views.py
```diff
@@ -490,10 +346,7 @@ def advanced_search(request, template=None):
     offset = (page - 1) * settings.SEARCH_RESULTS_PER_PAGE
 
     lang = language.lower()
-    if settings.LANGUAGES_DICT.get(lang):
-        lang_name = settings.LANGUAGES_DICT[lang]
-    else:
-        lang_name = ''
+    lang_name = settings.LANGUAGES_DICT.get(lang, '') or ''
```
mythmon: I saw this, and then



It makes sense after I realized the `LANGUAGES_DICT` has `None` in it. Sumo has weird stuff in settings. It's probably fine here, especially because you didn't actually change anything. Just weird.

mythmon:
I like this. I like that Open Search shares more code with Simple Search, and also shares the behavior. I think all the bits involved are simpler than they were before, so that's great.

It seems a bit odd that the locale is a separate parameter to `generate_simpler_search`. It is very similar to the things contained in the search form. I guess this is because everything else in the form came from `request.GET` and the locale has some extra special logic? I don't see a great way to move that into the Form (if that is even a sensible thing to do). Do you have any bright ideas about that?

willkg:
I don't have any good ideas about that. Currently, it could come from `request.GET.get('language')` or from `request.LANGUAGE`. I didn't know if at some point we'd want to pull it from the user's profile. I was thinking maybe we could take what we're doing here and do the same thing with `advanced_search`, too. 

Anyhow, I think figuring out the language is beyond the scope of `generate_simple_search` at least until someone convinces me otherwise.

kitsune/search/views.py
willkg: I didn't realize that. That means I can ditch the second argument to `.get()`.

mythmon:
That sounds reasonable. This is certainly an improvement, and should make the coming ES library change one step simpler.

r+

willkg:
I'll fix a couple of things.

Also, I want to take a look at ditching `ComposedList` there and whether that reduces the amount of code.

willkg:
^^^ That does some minor code cleanup and removes `ComposedList` from simple search.

simple search is still a mess. Seems like it's doing too much still, but I'm not sure how else to unravel it into smaller parts. I might have better ideas after going through advanced search and seeing how that works.

kitsune/search/views.py
willkg: Atom has a plugin for that.

It's an interesting idea, but I'm not sure how it helps us. For things that need to get done, I'll spin off new bugs. For everything else, I'm adding notes for someone in the future who's spelunking in that code area so they are better informed about the code. They're not necessarily things I think need to get fixed.

kitsune/search/views.py
```diff
@@ -119,176 +113,41 @@ def simple_search(request, template=None):
...
-    searcher = searcher.doctypes(*doctypes)
-    searcher = searcher.filter(final_filter)
-
-    if 'explain' in request.GET and request.GET['explain'] == '1':
-        searcher = searcher.explain()
+    lang_name = settings.LANGUAGES_DICT.get(lang) or ''
 
-    documents = ComposedList()
+    searcher = generate_simple_search(search_form, language, with_highlights=True)
+    searcher = searcher[:settings.SEARCH_MAX_RESULTS]
```
mythmon: I had no idea we did this. I always thought it was curious how the search pages maxed out at 1000. Turns out *we* did that.

mythmon:
GitHub is making the diff for the ComposedList change hard to read, but I think it looks good.

I'm not 100% about the changes to move all the stuff out of the try/except block. Does the Paginator consume the ES iterator on the spot? Or is it lazy? If it is lazy, I think moving everything out might miss some potential ES errors. If it isn't lazy, then this should be good.

willkg:
Oh, that's a good point--I'll double-check. The `.count()` call is covered by `.page()`. But I suspect there's an `.execute()` that might only happen in the template when the pager starts iterating over the search.

willkg:
I said template, but I meant the part where we're building excerpts. I'll move it back into the try-block.

willkg:
^^^ That fixes the try/except thing. Sorry about that.

mythmon:
I wish it had been true. That try block is a lot bigger than I wish it was. Oh well.

r+

PR #2636, opened 2015-08-31
[Bug 1169416] Show Translation page now shows full locale name along …

skiee-mahir:

PR #2673, opened 2015-10-13
[Bug 1169416] Added support to show full locale name like "English (en-US)" in the show_translations page.

skiee-mahir:
[Bug 1169416] Updated contents of `<li>` elements under 'locales' class to show full locale name

[Bug 1169416] Updated test_show_translations_page() to test for full locale like 'English (en-US)' instead of only 'en-US', test order or the actual test is not changed

[Bug 1169416] Updated show_translations() to display full locale name, imported LOCALES

### DIFF
--- a/kitsune/wiki/views.py
+++ b/kitsune/wiki/views.py
@@ -25,6 +25,7 @@ from tower import ugettext_lazy as _lazy
 from tower import ugettext as _
 
 from kitsune.access.decorators import login_required
+from kitsune.lib.sumo_locales import LOCALES
 from kitsune.products.models import Product, Topic
 from kitsune.sumo.decorators import ratelimit
 from kitsune.sumo.helpers import urlparams
@@ -1353,13 +1354,18 @@ def show_translations(request, document_slug):
     translated_locales = []
     untranslated_locales = []
 
-    translated_locales.append(document.locale)
-    translated_locales.extend(document.translations.all().values_list(
-        'locale', flat=True))
+    # Makes sure "English (en-US)" is always on top!
+    translated_locales.append((document.locale, LOCALES[document.locale].native))
+
+    # Gets all the locale code "ONLY" for translated locales
+    # Note: en-US is not a translated locale, it's the default
+    translated_locales_code = document.translations.all().values_list('locale', flat=True)
 
     for locale in settings.LANGUAGE_CHOICES:
-        if not locale[0] in translated_locales:
-            untranslated_locales.append(locale[0])
+        if locale[0] in translated_locales_code:
+            translated_locales.append(locale)
+        else:
+            untranslated_locales.append(locale)
 
     return render(request, 'wiki/show_translations.html', {
         'document': document,
--- a/kitsune/wiki/templates/wiki/show_translations.html
+++ b/kitsune/wiki/templates/wiki/show_translations.html
@@ -11,13 +11,13 @@
       <h3>{{ _('This document has been translated to the following locales:') }}</h3>
       <ul class="locales">
         {% for locale in translated_locales %}
-          <li><a class="translated_locale" href="{{ url("wiki.document", locale=locale, document_slug=document.slug) }}">{{ locale }}</a></li>
+          <li><a class="translated_locale" href="{{ url("wiki.document", locale=locale[0], document_slug=document.slug) }}">{{ locale[1] }} ({{ locale[0] }}) </a></li>
         {% endfor %}
       </ul>
       <h3>{{ _('This document is missing translations for the following locales:') }}</h3>
       <ul class="locales">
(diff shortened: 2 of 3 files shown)


skiee-mahir:
previously worked, [here](https://github.com/mozilla/kitsune/pull/2636).

kitsune/wiki/views.py
```diff
@@ -1353,13 +1354,20 @@ def show_translations(request, document_slug):
     translated_locales = []
     untranslated_locales = []
 
-    translated_locales.append(document.locale)
-    translated_locales.extend(document.translations.all().values_list(
-        'locale', flat=True))
+    # This will hold locale codes only
+    translated_locales_code = []
+    # Makes sure "English (en-US)" is always on top!
+    translated_locales.append((document.locale, LOCALES[document.locale].native()))
```
safwanrahman: Here is the Error! :-1: 
it should be ``native`` instead of ``native()``

safwanrahman:
@mythmon I have checked the code and it seems that this should get r+
Thoughts?

kitsune/wiki/views.py
```diff
@@ -49,6 +49,7 @@
     send_reviewed_notification, schedule_rebuild_kb,
     send_contributor_notification, render_document_cascade)
 
+from kitsune.lib.sumo_locales import LOCALES
```
mythmon: A small nit: We organize imports into several groups, each sorted. This would go in the `kitsune` group right above this, and should be sorted. I think that means it would be at about line 44.

kitsune/wiki/views.py
```diff
@@ -1353,13 +1354,20 @@ def show_translations(request, document_slug):
...
-    translated_locales.extend(document.translations.all().values_list(
-        'locale', flat=True))
+    # This will hold locale codes only
+    translated_locales_code = []
+    # Makes sure "English (en-US)" is always on top!
+    translated_locales.append((document.locale, LOCALES[document.locale].native))
+
+    # Gets all the locale code for translated locales
+    # Note: en-US is not a translated locale, it's the default
+    translated_locales_code.extend(document.translations.all().values_list('locale', flat=True))
```
mythmon: This is overly verbose. You can simply say `translated_locale_codes = document.translations.values_list('locale', flat=True)`. This is also less confusing since you aren't alternating between working with `translated_locales` and `translated_locale_codes`.

kitsune/wiki/views.py
safwanrahman: Hmm, I think it should be at line 28 @skiee-mahir

mythmon: Ah, right. Sorry, I was looking at the list of imports wrong. Line 28 is right.

kitsune/wiki/views.py
skiee-mahir: I just wanted to keep the style of coding that's already in master! Fixin' that too

skiee-mahir:
@safwanrahman @mythmon please check :smile:

kitsune/wiki/views.py
```diff
@@ -1353,13 +1354,20 @@ def show_translations(request, document_slug):
     translated_locales = []
     untranslated_locales = []
 
-    translated_locales.append(document.locale)
-    translated_locales.extend(document.translations.all().values_list(
-        'locale', flat=True))
+    # This will hold locale codes only
+    translated_locales_code = []
```
mythmon: Now this line doesn't actually achieve anything, since you immediately overwrite `translated_locales_code` on line 1364.

skiee-mahir:
@mythmon I don't see any reason the test should fail! Besides that, I've already been through the pre-commit which ran the tests :+1: but why Travis is failing?

willkg:
Travis says this:

```
kitsune/wiki/views.py:53:1: F811 redefinition of unused 'LOCALES' from line 29
```

skiee-mahir:
@willkg The re run fails too :-1:

willkg:
Ahh... I see. Travis is testing against what the result of this merged to master would be. If you rebase against the current master tip, I think that'll fix this.

Also, it looks like you squashed all your commits into one and merged the original commit messages. The resulting commit message should be rewritten. See the conventions for writing git commit messages here: http://kitsune.readthedocs.org/en/latest/conventions.html#git-commit-messages

safwanrahman:
Hmm @skiee-mahir, It seems that this test has been failing because last night a patch of mine has been merged. :smile: . That patch already imported the ``LOCALES`` so your patch is failing. 
You should rebase upon master and fix it.
Let me know if you need any help in rebasing upon master!

mythmon:
This has gone over 3 weeks without any progress. Feel free to re-open when you have fixed the current issues.

PR #2675, opened 2015-10-13
[bug 1201550] Navigate back to topic from mobile document

rehandalal:
r?

### DIFF
--- a/kitsune/wiki/templates/wiki/mobile/document.html
+++ b/kitsune/wiki/templates/wiki/mobile/document.html
@@ -31,8 +31,13 @@
 {% endif %}
 
 {% block content %}
-  <div class="title-bar">{{ document_title(document) }}</div>
+  {% if breadcrumb_items[1] %}
+    <div class="filter-bar">
+      <a class="return-link" href="{{ breadcrumb_items[1][0] }}">{{ breadcrumb_items[1][1] }}</a>
+    </div>
+  {% endif %}
   <article id="wiki-doc">
+    {{ document_title(document) }}
     {{ document_messages(document, redirected_from) }}
     {{ document_content(document, fallback_reason, request, settings) }}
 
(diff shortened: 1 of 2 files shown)


kitsune/wiki/templates/wiki/mobile/document.html
```diff
@@ -31,8 +31,13 @@
 {% endif %}
 
 {% block content %}
-  <div class="title-bar">{{ document_title(document) }}</div>
+  {% if breadcrumb_items[1] %}
+    <div class="filter-bar">
+      <a class="return-link" href="{{ breadcrumb_items[1][0] }}">{{ breadcrumb_items[1][1] }}</a>
```
mythmon: Will this work with documents in sub-topics? I'm guessing that `breadcrumb_items[1]` will always be the top level topic.

rehandalal: yep. it's always the highest topic level.

mythmon: Ok. I'd expect this to want to go to the lowest level topic. That is, the last bread crumb (second to last?). The "dominant topic" language in the bug is to disambiguate the case where one article has multiple top level topics.

mythmon: I feel like I didn't explain myself very well, let me try with code:

```python
topic = TopicFactory()
subtopic1 = TopicFactory(parent=topic, order=1)
subtopic2 = TopicFactory(parent=topic, order=2)
d = DocumentFactory(topics=[subtopic1, subtopic2])
```

the breadcrumbs would be `[topic.product, topic, subtopic2, d]` in this case. The "dominant topic" mechanism was used to pick subtopic2 over subtopic1. The "back to topic" navigation should go to `subtopic2`. I think that would be `breadcrumb_items[-2]`.

rehandalal:
fixed ^

mythmon:
r+

PR #2677, opened 2015-10-14
[bug 1214396] Move sumo/api.py to sumo/api_utils.py

willkg:
This will fix some issues with bootstrapping when we start adding api
view code to sumo/api.py.

Quick r?


willkg:
This was pretty straight-forward. Renamed a couple of files, then went through the code fixing import statements.

mythmon:
Thanks!

PR #2678, opened 2015-10-14
1194899 locales api

willkg:
This is pretty straight-forward--nothing exciting here. Quick r?

### DIFF
--- a/docs/api.rst
+++ b/docs/api.rst
@@ -144,3 +144,51 @@ Using curl::
        "max_documents": 3,
        "max_questions": 0
     }'
+
+
+Locales API
+===========
+
+.. http:get:: /api/2/locales/
+
+   All locales supported by SUMO.
+
+   **Example request**:
+
+   .. sourcecode:: http
+
+      GET /api/2/locales/ HTTP/1.1
+      Accept: application/json
+
+   **Example response**:
+
+   .. sourcecode:: http
+
+      HTTP/1.0 200 OK
+      Vary: Accept, X-Mobile, User-Agent
+      Allow: OPTIONS, GET
+      X-Frame-Options: DENY
+      Content-Type: application/json
+
+      {
+         "vi": {
+            "name": "Vietnamese",
+            "localized_name": "Ti\u1ebfng Vi\u1ec7t",
+            "aaq_enabled": false
+         },
+         "el": {
+            "name": "Greek",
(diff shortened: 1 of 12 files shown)


docs/api.rst
```diff
@@ -144,3 +144,51 @@ Using curl::
...
+            "name": "English",
+            "localized_name": "English",
+            "aaq_enabled": true
+         },
+         ...
+      }
+
+   :reqheader Accept: application/json
+   :resheader Content-Type: application/json
+   :statuscode 200: no error
```
willkg: @mythmon What do you think about documenting apis this way? If we ever get around to building autodocumenting apis, I think I'd want to use this format. It's from here:

http://pythonhosted.org/sphinxcontrib-httpdomain/

kitsune/sumo/api.py
```diff
@@ -0,0 +1,31 @@
...
+        # FIXME: should we add something based on SIMPLE_WIKI_LANGUAGES?
+        locale = {
+            'name': LOCALES[lang].english,
+            'localized_name': LOCALES[lang].native,
+        }
+        try:
+            QuestionLocale.objects.get(locale=lang)
+            locale['aaq_enabled'] = True
+        except QuestionLocale.DoesNotExist:
+            locale['aaq_enabled'] = False
```
willkg: I added the bits you specified in the bug. Do we want to add something for `SIMPLE_WIKI_LANGUAGES`? Are there other things we want to add?

docs/api.rst
mythmon: I like `sphinxcontrib.httpdomain`. I think it is a good idea to use something already made for documenting APIs. +1

Also, thanks so much for writing API docs here. You're an inspiration to us all.

kitsune/sumo/api.py
mythmon: SIMPLE_WIKI_LANGUAGES could be useful. I don't know of any other properties we need, but if you have any other ideas, I think that would be fine.

mythmon: On another note, couldn't this stanza be simplified to `locale['aaq_enabled'] = QuestionLocale.objects.filter(locale=lang).exists()`? (Or am I making that up?)

willkg: I don't think I've ever used `.exists()`. I can test that now.

willkg:
^^^ That removes the FIXME and also tweaks the code to use `.exists()`.

mythmon:
Looks good!

willkg:
Thank you!

PR #2573, opened 2015-06-11
[Bug 800880] Add Fallback Local in wiki

safwanrahman:
* The Local is just for test. It will be changed upon Local mapping

r? for the rest other?

PR #2679, opened 2015-10-15
[Bug 800880] Correction to show correct message to mobile

safwanrahman:
It was missed during #2573. The ``full_locale_name`` was added in the document template, but missed at Mobile template.(1). So while user view the article from mobile, they can not see the proper locale name
r?
1. https://github.com/mozilla/kitsune/pull/2573/files#diff-a193392a22b60ae6502da2147debf34cR55


rehandalal:
good catch. r+

PR #2680, opened 2015-10-15
[Bug 1215168] KB Minimal view shows escaped share links

pollti:


mythmon:
Thanks for the quick fix! This makes the share links look right for me. I'll merge this one Travis goes green.

PR #2681, opened 2015-10-15
[Bug 1215226] Don't fail silently when sending emails.

mythmon:
Cyliang pointed out that we make sending email fail silently. That sounds like a bad thing, and makes it really hard to track down when we aren't sending notifications to users.

r?

### DIFF
--- a/kitsune/sumo/email_utils.py
+++ b/kitsune/sumo/email_utils.py
@@ -22,11 +22,13 @@ def send_messages(messages):
     if not messages:
         return
 
-    conn = mail.get_connection(fail_silently=True)
-    conn.open()
-
-    for msg in messages:
-        conn.send_messages([msg])
+    try:
+        conn = mail.get_connection()
+        conn.open()
+        for msg in messages:
+            conn.send_messages([msg])
+    finally:
+        conn.close()
 
 
 def safe_translation(f):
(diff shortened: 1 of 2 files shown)


willkg:
As penance for writing that code, I will review it.

kitsune/sumo/email_utils.py
```diff
@@ -22,11 +22,13 @@ def send_messages(messages):
...
-
-    for msg in messages:
-        conn.send_messages([msg])
+    try:
+        conn = mail.get_connection()
+        conn.open()
+        for msg in messages:
+            conn.send_messages([msg])
+    finally:
+        conn.close()
```
willkg: Nit: I'm pretty sure you don't need the `conn.close()` because it will close automatically when the function ends and things get cleaned up.

Other than that, I think this looks good. Depending on what happens next, we might want to create message retrying. If we do that, then we're going to want to keep track of which messages succeeded and which failed. We might find that `.send_messages()` should be in a try/except block so that if we hit a failure, we skip just one message rather than that message and the rest in the list. I don't know offhand, though. I think this is good for now, but we might want to file a followup bug to hone it next week depending on what happens.

mythmon: I've always been told that I shouldn't rely on implicit destruction in Python. I looked into this a bit more and it would probably be fine in this case. I can switch it back.

I definitely agree we should add re-send logic at some point in the stack. I wanted to see what happened with this before we made too many decisions though.

willkg: OK. I'm r+ on this, but not near a computer to merge it.

Can someone green-button this for me?

PR #2682, opened 2015-10-16
Correcting missing macros of mobile and minimal

safwanrahman:
This was missing from the Mobile template which cause not showing the locale name correctly and rest other thing.
r?


rehandalal:
great, thanks! r+

PR #2683, opened 2015-10-16
[bug 1215110] Refactor simple and advanced search

willkg:
* centralize duplicate code to both get it out of the way and also make
  it easier to reason about
* remove intermediary variables that aren't needed
* clean up comments
* reduce complexity

This is a work in progress. I'm doing a bunch of yak-shaving here and fiddling with things and want to make sure the way things compose makes sense.

This mostly adjusts `simple_search`. I haven't really gotten into overhauling `advanced_search`, yet. That's partially because I want to make sure the work I've done on `simple_search` makes sense. Then I'll do similar things to `advanced_search` and other search code.

Feedback?


kitsune/search/es_utils.py
```diff
@@ -886,3 +890,52 @@ def indexes_for_doctypes(doctype):
     # Import locally to avoid circular import.
     from kitsune.search.models import get_mapping_types
     return set(d.get_index() for d in get_mapping_types(doctype))
+
+
+def handle_es_errors(template, status_code=503):
```
mythmon: Moving this to a decorator is an awesome idea. :+1:

mythmon:
I skimmed through this a bit. I really like the idea of splitting up the reusable parts of the function into separate parts, like the ES errors decorator and the `request.IS_JSON` checks.

On the other hand, this feels a lot like (how I think about) class based views. the ES errors decorator and the `request.IS_JSON` parts would probably be mixins, `build_results_list` would probably be a method, etc.

On the gripping hand, converting this to CBV would probably be a much larger project, and I'm not sure it actually gains us anything besides warm fuzzies. I think we've talked about this in the past and I don't recall you ever being very enthused by CBV.

How do you feel about exploring CBV for this? Do you want to keep it this way to minimize the amount of changes needed?

(To be clear, I'm talking about classes that are views, with multiple interrelated methods, and sharing code via inheritance and mixins. I'm *not* talking about Django's CBVs. I don't know if that changes anything.)

kitsune/search/views.py
```diff
@@ -74,38 +156,31 @@ def simple_search(request, template=None):
...
 
     search_form = SimpleSearchForm(request.GET, auto_id=False)
 
     if not search_form.is_valid():
-        if is_json:
+        if request.IS_JSON:
             return HttpResponse(
                 json.dumps({'error': _('Invalid search data.')}),
-                content_type=content_type,
+                content_type=request.CONTENT_TYPE,
```
mythmon: Looking at this, I don't think it has ever correctly dealt with jsonp callbacks when there is an error. Do we even use those?

kitsune/search/views.py
```diff
@@ -54,10 +67,79 @@ def jsonp_is_valid(func):
...
+
+    :arg pages: paginated S
+    :arg is_json: whether or not this is generated results for json output
+
+    :returns: list of dicts
+
+    """
+    offset = pages.start_index()
+    results = []
+    for i, doc in enumerate(pages):
```
mythmon: Since the only place you use `i` is to calculate the rank by adding it to `offset`, you could use `enumerate`'s second argument to get `rank` directly.

```python
>>> list(enumerate('abc'))
[(0, 'a'), (1, 'b'), (2, 'c')]
>>> list(enumerate('abc', 5))
[(5, 'a'), (6, 'b'), (7, 'c')]
```

kitsune/search/views.py
```diff
@@ -230,48 +246,38 @@ def simple_search(request, template=None):
...
+@handle_es_errors(_es_down_template)
 @mobile_template('search/{mobile/}results.html')
 def advanced_search(request, template=None):
     """Elasticsearch-specific Advanced search view"""
 
-    # JSON-specific variables
-    is_json = (request.GET.get('format') == 'json')
-    callback = request.GET.get('callback', '').strip()
-    content_type = 'application/x-javascript' if callback else 'application/json'
+    request = json_markup_request(request)
```
mythmon: Should this be a decorator too?

kitsune/search/views.py
willkg: Fancy! I didn't even know it took a second argument.

willkg:
I had to take a break from this and work on Puente. I'll rebase this against master and continue working on it.

kitsune/search/es_utils.py
```diff
@@ -886,3 +890,52 @@ def indexes_for_doctypes(doctype):
...
+                    template = template(request, *args, **kwargs)
+
+                # FIXME: statsd exceptions
+                # # Cheating here: Convert from 'Timeout()' to 'timeout' so
+                # # we have less code, but still have good stats.
+                # exc_bucket = repr(exc).lower().strip('()')
+                # statsd.incr('search.esunified.{0}'.format(exc_bucket))
+
+                # FIXME: Log exceptions
+                # log.exception(exc)
```
willkg: I can reimplement these, but I'm not sure we care. I know I haven't looked at them in ages. Do we care? If not, then let's nix them. We can always re-add them later.

willkg: For some context, we used to care a lot because ES would fall over *all* the time. I'm pretty sure it doesn't do that anymore.

willkg: Oh... I just rethought this. If we nix these, then this fails silently. Nevermind. I'll readd the exception logging.

willkg: Good idea, @willkg! If you need anything from the rest of us in this conversation, just let us know.

willkg:
I think I'm done with this as far as I'm going to go. We were talking about converting the search views to a kind of CBVs, but I never looked into it and while I think it might be nice, I don't think it'll help a whole lot.

@mythmon Can you review this?

willkg:
^^^ Both of those died on npm problems that I think we're seeing with other PRs.

mythmon:
Rebasing this patch on master will fix the CI by skipping the lint checks. I'll try and review this soon.

willkg:
^^^ Rebased. There's no super rush on this. Sometime this week would be good.

kitsune/search/views.py
```diff
@@ -54,10 +54,113 @@ def jsonp_is_valid(func):
...
+                'is_solved': doc['question_is_solved'],
+                'num_answers': doc['question_num_answers'],
+                'num_votes': doc['question_num_votes'],
+                'num_votes_past_week': doc['question_num_votes_past_week']}
+
+        else:
+            summary = _build_es_excerpt(doc, first_only=True)
+            result = {
+                'title': doc['post_title'],
+                'type': 'thread'}
```
mythmon: Is this `else` clause a catch all? Or is it for a specific document type? If it is a catchall, this doesn't seem robust enough. If it is for a specific type, I think it should be an `elif doc['model'] == 'that':`' instead. Maybe even have logging or an exception for unknown models as a sort of sanity assertion?

mythmon:
That's a bunch of code. The result is simpler, but it isn't an easy diff to read. I think I'll go back after lunch and read it in it's final form instead the diff to try and understand it a bit better. Generally it looks simpler, and if nothing else it is less code.

kitsune/search/views.py
willkg: In this case, I'm just moving code around. It's for contributor forum posts which is the only other type that gets searched. I can make it explicit.

mythmon: Oops. I was comparing this new function to the removed code, and in one place there wasn't a branch for the forum posts, so I thought this was new code. If this is just moved code, don't worry about it.

mythmon:
This change doesn't break basic or advanced search in my manual testing. It moves a lot of complex logic out to named, documented functions, so that's great. It makes some of the logic around error handling easier. Also great.

It would be nice to have tests for the new extracted functions. That would make further refactoring easier, I think. On the other hand, the tests for them might be difficult to write and kind of fragile. If you think it is useful, more tests are always good, but if you think there isn't a lot of benefit, we can land without them.

willkg:
The testing thing is a good point. I'll go through the tests now and move things around and add whatever is missing.

willkg:
^^^ That adds some tests for the json bits some of which were taken from old code and moved around.

I didn't add a test for `build_results_list` since the unit tests for those views tests that pretty heavily. It didn't seem like it made a lot of sense to test it on its own and the setup would have been involved.

Incidentally, there's a `json_view` view decorator. At some point, we might want to unify that with some of the json stuff in these views. That could clean things up. It's also possible that the code is still too tangled for that to be wildly helpful. Amongst other things, the context data that gets generated differs greatly between json and html output.

Anyhow, stuff to think about some other day.

willkg:
@mythmon r? that last bit? Then I think I'm done working on this for now.

mythmon:
This looks good. Thanks for all the work on this! I hope it makes the ES migration easier.

willkg:
I'm not wholly sure whether it'll make it easier to migrate, but I figured it was worth doing regardless plus I needed to be more familiar with the code, so this seemed like a good exercise.

PR #2685, opened 2015-10-19
[Bug 1216295] Upgrade django-taggit to 0.17.1

mythmon:
I read through [the changelog](https://github.com/alex/django-taggit/blob/develop/CHANGELOG.txt). A feature was introduced in 0.16 (case insensitive tags) that we could rewrite some code to take advantage of. Besides that, I don't think anything affects us.

r?


willkg:
Looks good to me. r+

Does it make sense to write up a bug for removing the code enforcing case insensitive tags?

mythmon:
Good point. [Bug 1216622](https://bugzilla.mozilla.org/show_bug.cgi?id=1216622).

PR #2686, opened 2015-10-19
[Bug 1193443] Add more metadata and autotagging to new AAQ.

mythmon:
This makes the new AAQ collect and add metadata about user agent, browser version, etc, that the old AAQ did, and adds an API end point for auto tagging.

Originally, I wanted auto tagging to run every time that metadata was changed so that BuddyUp would be able to take advantage of this, to reduce the number of requests the new AAQ had to make, and to make sure that future clients wouldn't have to think about auto tagging. Unfortunately, there is [a race condition in Taggit](https://github.com/alex/django-taggit/issues/349) that makes that not work. In particular, both BuddyUp and the new AAQ set metadata items in parallel. This causes a tag to be applied to a question multiple times because Taggit didn't make the DB constraints needed to make `get_or_create` atomic. So no auto-auto-tagging for us.

> Note: we should probably make sure we are using `get_or_create` in an atomic way in the rest of Kitsune.

r?

### DIFF
--- a/kitsune/questions/api.py
+++ b/kitsune/questions/api.py
@@ -310,8 +310,7 @@ class QuestionViewSet(viewsets.ModelViewSet):
             serializer.save()
             return Response(serializer.data)
         else:
-            return Response(serializer.errors,
-                            status=status.HTTP_400_BAD_REQUEST)
+            return Response(serializer.errors, status=status.HTTP_400_BAD_REQUEST)
 
     @action(methods=['POST', 'DELETE'])
     def delete_metadata(self, request, pk=None):
@@ -379,6 +378,12 @@ class QuestionViewSet(viewsets.ModelViewSet):
         for tag in tags:
             question.tags.remove(tag)
 
+        return Response(status=status.HTTP_204_NO_CONTENT)
+
+    @action(methods=['POST'], permission_classes=[permissions.IsAuthenticated])
+    def auto_tag(self, request, pk=None):
+        question = self.get_object()
+        question.auto_tag()
         return Response(status=status.HTTP_204_NO_CONTENT)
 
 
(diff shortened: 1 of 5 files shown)


willkg:
Travis failed setting things up with this:

```
npm WARN engine get-stdin@5.0.0: wanted: {"node":">=0.12.0"} (current: {"node":"0.10.36","npm":"1.4.28"})
```

I restarted the build on the off chance it's an ephemeral error.

kitsune/questions/api.py
```diff
@@ -381,6 +380,12 @@ def remove_tags(self, request, pk=None):
 
         return Response(status=status.HTTP_204_NO_CONTENT)
 
+    @action(methods=['POST'], permission_classes=[permissions.IsAuthenticated])
+    def auto_tag(self, request, pk=None):
+        question = self.get_object()
+        question.auto_tag()
```
willkg: Does this throw any errors?

willkg:
Other than the travis issue and the question I have about errors, this looks ok to me. I didn't check the JS code, though.

@rehandalal Can you check the JS code?

After that, we should wait for Travis to go green and/or for us to deal with it's non-greenness.

kitsune/questions/api.py
mythmon: No, except errors with `.get()` returning too many things when Taggit gets itself into a race condition (which shouldn't happen).

`autotag` is actually really simple. It mostly just moves data from metadata to tags.

mythmon:
We are still getting that warning Node engine warning, but the builds are passing. We should probably fix the error (probably by downgrading the get-stdin library to one that supports Node 0.10), but I don't think it is related to this.

rehandalal:
JS code looks fine to me. r+

PR #2688, opened 2015-10-21
[Bug 1216472] GA Tracking for Fallback Locale

safwanrahman:
@mythmon r?


rehandalal:
code looks good, but we should probably have a test for this in the `LocalizationAnalyticsTests` class in `wiki/tests/test_views.py`

safwanrahman:
@rehandalal Done adding test. r?

rehandalal:
thank you! r+

PR #2689, opened 2015-10-21
[Bug 1217076] Add Signal to AAQ.

mythmon:
NB: Signal is kind of hush-hush for now, so this doesn't have a logo or description. We'll get a logo later, in [Bug 1217084](https://bugzilla.mozilla.org/show_bug.cgi?id=1217084).

This should be everything we need to do in the code. Last time we also added migrations to make the products we needed, but that was a big cluster, so I think we shouldn't do that again. Does that sound ok?

r?

### DIFF
--- a/kitsune/sumo/static/sumo/less/questions.less
+++ b/kitsune/sumo/static/sumo/less/questions.less
@@ -631,32 +631,39 @@ h2 {
   height: 96px;
   width: 96px;
 
+  /* Vertical position = 100px + i * 296 */
+
   &.logo-firefox {
     background-position: -100px -100px;
-  }
-
-  &.logo-firefox-os {
-    background-position: -100px -988px;
-  }
-
-  &.logo-mobile {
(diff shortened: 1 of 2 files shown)


kitsune/sumo/static/sumo/less/questions.less
```diff
@@ -631,32 +631,39 @@ h2 {
...
 
   &.logo-ios {
     background-position: -100px -1876px;
   }
 
-  &.logo-other {
-    background-position: -100px -1580px;
+  &.logo-signal {
+    /* Use temporary logo for now */
+    background-position: -100px -1580;
```
rehandalal: should be `-1580px;`

mythmon: Oops!

mythmon:
The product people changed there mind. We can't ship this until next week, so lets just close this and I'll file another one with the real logo next week.

PR #2690, opened 2015-10-26
[Bug 1218459] Improve SimplePush error handling.

mythmon:
We found out that some times the SimplePush server *doesn't* send back JSON errors (Even though it is supposed to). This should handle that and still print out the errors.

I'm deploying this to stage now to do some testing with Marcia and Kadir. I'll update this PR once that has gone through.

### DIFF
--- a/kitsune/notifications/tasks.py
+++ b/kitsune/notifications/tasks.py
@@ -8,6 +8,7 @@ import simplejson
 import requests
 from actstream.models import Action, Follow
 from celery import task
+from multidb.pinning import use_master
 from requests.exceptions import RequestException
 
 from kitsune.notifications.models import (
@@ -80,14 +81,11 @@ def _send_simple_push(endpoint, version, max_retries=3, _retry_count=0):
             logger.error('SimplePush error: %s %s', r.status_code, r.json())
 
 
-@task(bind=True)
-def add_notification_for_action(self, action_id):
-    try:
-        action = Action.objects.get(id=action_id)
-    except Action.DoesNotExist as exc:
-        # Maybe database replication is lagging a bit, retry after a few seconds.
-        raise self.retry(exc=exc, max_retries=1, countdown=30)
-
+@task(ignore_result=True)
+@use_master
+def add_notification_for_action(action_id):
+    action = Action.objects.get(id=action_id)
+    
     query = _full_ct_query(action, actor_only=False)
     # Don't send notifications to a user about actions they take.
     query &= ~Q(user=action.actor)
@@ -102,6 +100,7 @@ def add_notification_for_action(self, action_id):
 
 
 @task(ignore_result=True)
+@use_master
 def send_realtimes_for_action(action_id):
     action = Action.objects.get(id=action_id)
     query = _full_ct_query(action)
--- a/kitsune/notifications/tasks.py
+++ b/kitsune/notifications/tasks.py
@@ -4,6 +4,7 @@ from django.contrib.contenttypes.models import ContentType
 from django.db.models import Q
 
 import actstream.registry
+import simplejson
 import requests
 from actstream.models import Action, Follow
 from celery import task
@@ -48,7 +49,10 @@ def _send_simple_push(endpoint, version):
         # If something does wrong, the SimplePush server will give back
         # json encoded error messages.
         if r.status_code != 200:
-            logger.error('SimplePush error: %s %s', r.status_code, r.json())
+            try:
+                logger.error('SimplePush error: %s %s', r.status_code, r.json())
+            except simplejson.scanner.JSONDecodeError:
+                logger.error('SimplePush error (also not JSON?!): %s %s', r.status_code, r.text)
     except RequestException as e:
         # This will go to Sentry.
         logger.error('SimplePush PUT failed: %s', e)
--- a/kitsune/notifications/tasks.py
+++ b/kitsune/notifications/tasks.py
@@ -38,24 +38,46 @@ def _full_ct_query(action, actor_only=None):
     return query
 
 
-def _send_simple_push(endpoint, version):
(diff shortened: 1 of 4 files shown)


kitsune/notifications/tasks.py
```diff
@@ -48,7 +48,19 @@ def _send_simple_push(endpoint, version):
...
+            try:
+                import simplejson
+                ErrorType = simplejson.scanner.JSONDecodeError
+            except ImportError:
+                ErrorType = ValueError
+
+            try:
+                logger.error('SimplePush error: %s %s', r.status_code, r.json())
+            except ErrorType:
+                logger.error('SimplePush error (also not JSON?!): %s %s', r.status_code, r.text)
```
willkg: Uhhh... I think instead I'd just add simplejson to the requirements again.

mythmon: That would be simpler...

mythmon:
Moved simplejson to the default requirements ^

kitsune/notifications/tasks.py
mythmon: That... would be simpler.

mythmon:
This, uh... got bigger. Sorry about that.

kitsune/notifications/tasks.py
```diff
@@ -37,26 +38,56 @@ def _full_ct_query(action, actor_only=None):
...
+        try:
+            data = r.json()
+        except simplejson.scanner.JSONDecodeError:
+            logger.error('SimplePush error (also not JSON?!): %s %s', r.status_code, r.text)
+            return
+
+        if r.status_code == 503 and data['errno'] == 202 and _retry_count < max_retries:
+            return _send_simple_push(endpoint, version, max_retries, _retry_count + 1)
+        else:
+            logger.error('SimplePush error: %s %s', r.status_code, r.json())
```
willkg: Every now and then, I get tempted to write a retry decorator. I'm pretty sure I was looking at a sophisticated retryable library at one point. It's probably overkill here, but it might be nice to separate the retry logic out.

kitsune/notifications/tasks.py
```diff
@@ -37,26 +38,56 @@ def _full_ct_query(action, actor_only=None):
...
-@task(ignore_result=True)
-def add_notification_for_action(action_id):
-    action = Action.objects.get(id=action_id)
+@task(bind=True)
+def add_notification_for_action(self, action_id):
+    try:
+        action = Action.objects.get(id=action_id)
+    except Action.DoesNotExist as exc:
+        # Maybe database replication is lagging a bit, retry after a few seconds.
+        raise self.retry(exc=exc, max_retries=1, countdown=30)
```
willkg: I'm not sure I follow this. What's the situation here?

mythmon: For context: This task is run as part of a post-save Django event. We often get errors like [this one](https://errormill.mozilla.org/support/sumo-stage/group/397026/) ([pastebin](https://pastebin.mozilla.org/8850590)).

My theory is that sometimes celery events get processed faster than our database gets replicated. The object *just* got saved. This task is running on a different server than the one that the post-save event happened on, so there won't be any DB pinning. It is likely that the read-DB and the write-DB haven't synced up yet. In that case, waiting a few seconds and trying again would fix it. "A few seconds" could be more accurately described as "whatever our current DB replication lag is". I hope it is less than 30 seconds.

This is happened about 600 times on stage, so it isn't a non-issue. I deployed this PR to stage for testing by BuddyUp people, and those errors stopped happening, so I think my thinking above is correct.

kitsune/notifications/tasks.py
mythmon: This code was kind of nasty to write. Maybe isolating the retry logic would help with that. Do you have another place in mind where a retry decorator could be used? I'd be game for taking a stab at it.

kitsune/notifications/tasks.py
willkg: Can't you say "yo--use master only"? I thought we had that power.

mythmon: The comment in the code is really terse, but I'm not sure about what details should be added. It makes sense to me right now, but I've been thinking about this all day.  What information would make it clearer?

willkg: Yeah, just pin the thread:

```
from multidb.pinning import use_master

with use_master:
    blah blah blah
```

mythmon: Oh. I totally forgot that's a thing. Thanks!

kitsune/notifications/tasks.py
willkg: There's a bunch of retry code in indexing. I think that should go away, though. I forget if I have it in other places or not.

willkg:
I was thinking about how to test this or verify any of it is correct, but I don't have any good ideas short of writing a ton of code. So I think maybe land it and see what happens?

kitsune/notifications/tasks.py
mythmon: In the interest of landing this soon, I think I'm going to skip the retry decorator for now. I may re-visit it later in another PR.

mythmon:
One way to test it would be to make a small http server that responds with things like the 503 retry message this checks for, and make api requests to kitsune to ensure it hits the urls we tell it to in the way we told it to.

Also, I developed this patch while talking to BuddyUp people to figure out what was wrong, deploying code to stage, and then watching Errormill to see if it worked. So I'm pretty sure it works.

willkg:
I'm +1 on skipping the retry decorator. I was just thinking out loud in those comments.

This code looks fine to me. It'd be good to have tests for all the obvious reasons, but ... I'm inclined to just land it as is. r+

mythmon:
I added a test that tests the case when the server asks use to retry. I just touched test code, so if Travis likes it I'll merge this.

willkg:
r+

PR #2692, opened 2015-10-27
[bug 1218973] Implement buddyup merge script

willkg:
This merges buddyup.po files using a shell script rather than using
"./manage.py merge". This makes it easier to implement a Tower
replacement with fewer moving parts.

r?

### DIFF
--- a/scripts/merge-buddyup.sh
+++ b/scripts/merge-buddyup.sh
@@ -14,9 +14,15 @@ if [[ ($# -ne 1) || (! -d "$1") ]]; then
     usage
 fi
 
-POTFILE=`find $1 -type f -name "buddyup.pot"`
+LOCALEDIR="$1"
+POTFILE="${LOCALEDIR}/templates/LC_MESSAGES/buddyup.pot"
 
-for lang in `find $1 -type f -name "buddyup.po"`; do
+if [ ! -e "${POTFILE}" ]; then
+    echo "${POTFILE} does not exist. exiting."
+    exit 1
+fi
+
+for lang in $(find $1 -type f -name "buddyup.po"); do
     echo -n "working on ${lang} "
     msgmerge --update ${lang} ${POTFILE}
 done
--- a/scripts/merge-buddyup.sh
+++ b/scripts/merge-buddyup.sh
@@ -1,0 +1,22 @@
+#!/bin/bash
+
+# syntax:
+# merge-buddyup.sh locale-dir/
+
+function usage() {
+    echo "syntax:"
+    echo "merge-buddyup.sh locale-dir/"
+    exit 1
+}
+
+# check if file and dir are there
+if [[ ($# -ne 1) || (! -d "$1") ]]; then
+    usage
+fi
+
+POTFILE=`find $1 -type f -name "buddyup.pot"`
+
+for lang in `find $1 -type f -name "buddyup.po"`; do
+    echo -n "working on ${lang} "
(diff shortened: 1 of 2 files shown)


willkg:
I looked for documentation to update in regards to how buddyup strings now have to be merged, but I didn't see anything anywhere. Do we have docs for updating buddyup strings somewhere?

mythmon:
There are some docs in some wiki page about how to do the BuddyUp side of this, but there isn't anything written down about how to do the Kitsune side of this, because there wasn't anything to do except the normal Kitsune localization stuff.

scripts/merge-buddyup.sh
```diff
@@ -0,0 +1,22 @@
...
+    echo "merge-buddyup.sh locale-dir/"
+    exit 1
+}
+
+# check if file and dir are there
+if [[ ($# -ne 1) || (! -d "$1") ]]; then
+    usage
+fi
+
+POTFILE=`find $1 -type f -name "buddyup.pot"`
```
mythmon: Nit: using backticks for subshells is lame, mainly because they don't nest, I can't easily type them in code blocks, and because they are all slanty. `$(find ...)` is generally considered better style.

mythmon: Generally one would do something like 

```bash
$localeDir = $1
shift
```

To give `$1` a useful name, and also to remove it from the argument list (in case you wanted to parse other things).

mythmon:
I'm a little nervous by your use of `find` because if there were ever two buddyup.pot files it will do weird things. But it should work fine. I have some style nits but they are minor. Otherwise this looks fine.

mythmon:
PS: You could probably hard code the path to `buddyup.pot`, but I don't think I have a better idea for your use of find to get all the `buddyup.po` files. Probably don't worry about it.

scripts/merge-buddyup.sh
willkg: I was mimicking the structure of `compile-mo.sh` as much as possible to reduce differenceness in the code-base. If you don't want that, that's fine. Then we should add some code to verify that `buddyup.pot` exists in the first place and some other things.

willkg:
^^^ That hard-codes the path, adds a check for existence and removes backticks.

scripts/merge-buddyup.sh
mythmon: Oh. I never actually looked at compile-mo.sh, sorry about th at.

mythmon:
I ran this script with my local data, and it made changes like this to my svn repo:

```diff
 #: app/views/main/unsupported_locale.html:10
-msgid "Please ask your questions in English, or go to <a href=\"http://support.mozilla.org\">http://support.mozilla.org</a> for more help."
+msgid ""
+"Please ask your questions in English, or go to <a href=\"http://support."
+"mozilla.org\">http://support.mozilla.org</a> for more help."
 msgstr ""
```

Will that affect localization? I'm guessing either it won't or that this is a result of msgmerge not stripping whitespace like tower does. Do you think we should go a head with this? Or will it break things for our localizers?

willkg:
That's just wrapping of the PO file. It doesn't affect translations at all.

willkg:
This is affected by the "width" argument. We can pass in whatever width you want.

mythmon:
If localizers won't notice, this is (probably) just a one time change in SVN. So I think we can leave it at the default.

Thanks!

PR #2694, opened 2015-10-28
[bug 1183424] Support image uploads in new AAQ

rehandalal:
r?

### DIFF
--- a/kitsune/questions/api.py
+++ b/kitsune/questions/api.py
@@ -3,6 +3,7 @@ from datetime import datetime
 import actstream.actions
 import django_filters
 import json
+from django.contrib.contenttypes.models import ContentType
 from django.db.models import Q
 from rest_framework import serializers, viewsets, permissions, filters, status, pagination
 from rest_framework.decorators import detail_route
@@ -17,6 +18,7 @@ from kitsune.questions.models import (
 from kitsune.sumo.api_utils import (
     DateTimeUTCField, OnlyCreatorEdits, GenericAPIException, SplitSourceField)
 from kitsune.tags.utils import add_existing_tag
+from kitsune.upload.models import ImageAttachment
 from kitsune.users.api import ProfileFKSerializer
 from kitsune.users.models import Profile
 
@@ -375,6 +377,18 @@ class QuestionViewSet(viewsets.ModelViewSet):
     def auto_tag(self, request, pk=None):
         question = self.get_object()
         question.auto_tag()
+        return Response(status=status.HTTP_204_NO_CONTENT)
+
+    @action(methods=['POST'], permission_classes=[permissions.IsAuthenticated])
+    def attach_images(self, request, pk=None):
+        question = self.get_object()
+
+        user_ct = ContentType.objects.get_for_model(request.user)
+        qst_ct = ContentType.objects.get_for_model(question)
+
+        # Move over to the question all of the images I added to the reply form
+        up_images = ImageAttachment.objects.filter(creator=request.user, content_type=user_ct)
+        up_images.update(content_type=qst_ct, object_id=question.id)
         return Response(status=status.HTTP_204_NO_CONTENT)
 
 
--- a/kitsune/questions/api.py
+++ b/kitsune/questions/api.py
@@ -379,7 +379,7 @@ class QuestionViewSet(viewsets.ModelViewSet):
         question.auto_tag()
         return Response(status=status.HTTP_204_NO_CONTENT)
 
-    @action(methods=['POST'], permission_classes=[permissions.IsAuthenticated])
+    @action(methods=['POST'])
     def attach_images(self, request, pk=None):
         question = self.get_object()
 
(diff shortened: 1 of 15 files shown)


kitsune/questions/api.py
```diff
@@ -386,6 +388,18 @@ def auto_tag(self, request, pk=None):
         question.auto_tag()
         return Response(status=status.HTTP_204_NO_CONTENT)
 
+    @action(methods=['POST'], permission_classes=[permissions.IsAuthenticated])
```
mythmon: `permission_classes=[permissions.IsAuthenticated]` will allow any authenticated user to attach an image to a question. You probably want to leave this off and use the default permissions for this class, which are `permission_classes = [OnlyCreatorEdits, permissions.IsAuthenticatedOrReadOnly]`.

rehandalal: makes sense. i'll change this.

rehandalal:
updated ^

rehandalal:
not sure why the linter is failing...

mythmon:
Rebasing this on master will fix the lint errors.

rehandalal:
rebased

kitsune/questions/static/questions/js/components/AttachmentField.jsx
```diff
@@ -0,0 +1,64 @@
...
+      return (
+        <div className="AAQApp__AttachmentField">
+          <input type="file" name="image" size="30" title="Browse for an image to upload."
+                 onChange={this.handleChange.bind(this)}/>
+          <ul className="AAQApp__AttachmentField__Attachments">
+            {this.props.question.images.map((image) => {
+              return (
+                <Attachment image={image}/>
+              );
+            })}
```
mythmon: By the way, this can be shortened to simply

```jsx
{this.props.question.images.map(image => <Attachment image={image}/>)}
```

ES6 arrow functions that simply return a single value immediately can drop the `{ return ...; }` parts of the construct.

mythmon:
This works, but it doesn't look quite right

shot_7710

rehandalal:
weird maybe i should just use fontawesome icons instead of standard unicode since it looks fine on my system.

kitsune/questions/static/questions/js/components/AttachmentField.jsx
rehandalal: ah cool, on it!

rehandalal:
should be fixed ^

kitsune/users/static/users/js/actions/UserAuthActions.es6.js
```diff
@@ -55,9 +55,15 @@ export function login(username, password, {inactive=false}={}) {
...
-      Dispatcher.dispatch({
-        type: actionTypes.AUTH_LOG_IN_SUCCESS,
-        username,
+      return apiFetch(`/api/2/user/${username}/`, {
+        method: 'GET',
+      })
+      .then(result => {
+        Dispatcher.dispatch({
+          type: actionTypes.AUTH_LOG_IN_SUCCESS,
+          id: result.id,
```
mythmon: The function `checkAuthState` in this file will also trigger a `AUTH_LOG_IN_SUCCESS` action if the user comes to the page already signed in. You need to update the ID there too.

rehandalal:
updated ^

mythmon:
There is a merge conflict. I'll try and fix this when I merge if it is trivial.

rehandalal:
rebased to address merge conflict ^

mythmon:
The rebase broke it because you picked up the DRF 3.3 upgrade. It should be a simple fix to change from the `action` decorator to `detail_route` (or `list_route` if it isn't for a specific object)`. Docs here: http://www.django-rest-framework.org/api-guide/routers/#extra-link-and-actions

rehandalal:
d'oh... updated ^

rehandalal:
@mythmon this g2g?

mythmon:
Hrm, it still isn't working for me. I get a 404 when it tries to post to the url `http://kitsune/en-US/upload/image/auth.User/undefined`.

kitsune/users/static/users/js/actions/UserAuthActions.es6.js
mythmon: This still needs fixed.

kitsune/users/static/users/js/actions/UserAuthActions.es6.js
```diff
@@ -15,6 +15,7 @@ export function checkAuthState() {
     if (status === 200) {
       Dispatcher.dispatch({
         type: actionTypes.AUTH_LOG_IN_SUCCESS,
+        id: data.id,
```
mythmon: `data` here doesn't have an id property. This won't work.

rehandalal:
d'oh! yes, this should be working ^

mythmon:
Ok! This works for me now, lets get it into prod.

PR #2701, opened 2015-11-04
[bug 1221244] Remove coolsearch

willkg:
This removes the frontend and backend parts for the coolsearch
prototype. We definitely want a replacement for advanced search and
this gave us some interesting ideas and was a good experiment, but
there's no future development planned and it's not currently being
used. Given that, we're going to remove it so we don't have to
maintain it. If at some future point we want to resurrect and finish
it, that's easily done because it's in git history.

r?


mythmon:
:cry:.

PR #2702, opened 2015-11-05
[Bug 1215172] Fix printer-friendly view for article pages

brittanystoroz:
This fixes the print styles for article pages (removing the left sidebar, hiding video embeds), and cleans up some width/padding and placement issues with the 'mozilla support' header.

Associated bug: https://bugzilla.mozilla.org/show_bug.cgi?id=1215172

### DIFF
--- a/kitsune/sumo/static/sumo/css/print.css
+++ b/kitsune/sumo/static/sumo/css/print.css
@@ -5,14 +5,52 @@ body {
     margin: 0;
     padding: 0;
 }
+
+body.scroll-header {
+    padding-top: 0;
+}
+
+body > header {
+    padding-bottom: 5px;
+    position: relative;
+}
+
+body > header a {
+    border-bottom: 0;
+}
+
+body.html-ltr .container_12 {
+    width: auto;
+    margin: 0;
+    padding: 0;
+}
+
+div#main-breadcrumbs {
+    border: 0;
+}
+
+ol#breadcrumbs {
+    margin: 10px;
+}
+
+body.document #main-content > div.grid_9 {
+    width: 940px;
+}
+
+article.wiki-doc {
+    padding: 0;
+}
+
 a.permalink,
 aside,
+aside[class*="grid_"],
 button,
 div.announcements,
 div.document-vote,
 div.mobile-banner,
 div.post-actions,
 div.thread-actions,
+div.youtube-embed,
 div#filter,
 div#header,
 div#more-help,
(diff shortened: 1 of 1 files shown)


willkg:
I'm surprised that this is a .css file and not a .less file. I thought all our .css files were generated. I could be horribly wrong--I haven't worked on SUMO in a while.

We should have some way to test this and make sure it's good going forward. Is there a way to test this automatically? Via Travis CI? Does it have to be manual testing? Do we have manual tests documented somewhere?

@mythmon ^^^ Thoughts?

brittanystoroz:
Yea I didn't see a print.less stylesheet in the codebase, but I could make one and move everything over there if we wanted.

mythmon:
Most of our stylesheets are .less files, but we have a few outliers like this. It probably makes sense to move it over to be a .less file at some point.

I'm not exactly sure how to test style sheets. We could create a smoketest that uses Selenium and makes sure certain elements are/aren't visible? @brittanystoroz, do you have any good ideas for how to prevent regressions via testing here?

The resulting page looks good to me, and the CSS looks like CSS. @rehandalal are you happy with this?

kitsune/sumo/static/sumo/css/print.css
```diff
@@ -5,14 +5,52 @@ body {
...
+    padding-bottom: 5px;
+    position: relative;
+}
+
+body > header a {
+    border-bottom: 0;
+}
+
+body.html-ltr .container_12 {
+    width: auto;
```
rehandalal: nit: alphabetize

rehandalal:
looks good to me, except for the small nit

kitsune/sumo/static/sumo/css/print.css
brittanystoroz: @rehandalal can you help clarify how you would want this ordered? I'm guessing:

body.html-ltr .container_12
body.scroll-header
body > header
body > header a

Not sure if classes on a tag would come before or after a tag with a child selector?  I'm not a huge fan of alphabetizing CSS...sometimes I think an alphabetic chronology obscures semantics, but I will do my best :)

mythmon: Oops. This isn't about alphabetizing the selectors. This is about alphabetizing the rules in a selector. So this should be

```css
body.html-ltr .container_12 {
  margin: 0;
  padding: 0;
  width: auto;
}
```

brittanystoroz: Oh I see, phew.

mythmon:
All of Rehan's concerns were fixed. r+

PR #2697, opened 2015-10-30
[Bug 1139114] Update DRF to 3.2.4

mythmon:
I haven't tried to run the site or do any manual testing with this, but it is probably worth looking at now. I updated DRF, then ran the tests, fixed what broke, repeat.

Most of the changes are pretty easy. Some things changed names (`from_native` became `to_internal_value`, for example). I'll add comments inline for some of the other more interesting things.

@willkg f?

PR #2703, opened 2015-11-05
[Bug 1139114] Update DRF to 3.3.0

mythmon:
See also #2697.

I updated DRF, then ran the tests, fixed what broke, repeat.

Most of the changes are pretty easy. Some things changed names (from_native became to_internal_value, for example). I'll add comments inline for some of the other more interesting things.

r?


kitsune/dashboards/api.py
```diff
@@ -4,10 +4,13 @@
...
 
 from kitsune.dashboards.models import WikiMetric
+from kitsune.products.models import Product
 
 
 class WikiMetricSerializer(ModelSerializer):
-    product = SlugRelatedField(slug_field='slug')
+    product = SlugRelatedField(
+        slug_field='slug',
+        queryset=Product.objects.all())
```
mythmon: Related fields that aren't readonly now require a queryset parameter.

kitsune/notifications/api.py
```diff
@@ -27,7 +29,7 @@ def has_object_permission(self, request, view, obj):
 
 
 class NotificationSerializer(serializers.ModelSerializer):
-    is_read = serializers.Field(source='is_read')
+    is_read = serializers.ReadOnlyField()
```
mythmon: In DRF 2.x, serializers.Field was read-only, and serializers.WriteField was the writable one. In DRF 3.x, serializers.ReadOnlyField is read-only, and serializers.Field tries to be a writable field by default.

kitsune/notifications/tests/test_api.py
```diff
@@ -80,7 +81,7 @@ def test_correct_fields(self):
         eq_(serializer.data['action_object']['type'], 'question')
         eq_(serializer.data['action_object']['id'], q.id)
         eq_(serializer.data['target'], None)
-        eq_(type(serializer.data['timestamp']), datetime)
+        datetime.strptime(serializer.data['timestamp'], '%Y-%m-%dT%H:%M:%SZ')
```
mythmon: In DRF 2.x, serialized datetimes were still datetimes. In DRF 3.x, datetimes get serialized to string encoded versions. This shouldn't affect the JSON rendered output, just the intermediary step.

kitsune/products/api.py
```diff
@@ -1,94 +1,18 @@
...
 from rest_framework import generics, serializers
 from tower import ugettext_lazy as _lazy
 
-from kitsune.products.models import Product, Topic
+from kitsune.products.models import Product, Topic, Platform
 from kitsune.sumo.api_utils import LocaleNegotiationMixin, LocalizedCharField, ImageUrlField
 from kitsune.wiki.api import DocumentShortSerializer
 
 
-class TopicField(serializers.SlugRelatedField):
```
mythmon: I moved this to kitsune/products/api_utils.py.

kitsune/products/api.py
```diff
@@ -103,8 +27,7 @@ class ProductList(generics.ListAPIView):
 
 
 class TopicShortSerializer(serializers.ModelSerializer):
-    title = LocalizedCharField(source='title',
-                               l10n_context='DB: products.Topic.title')
+    title = LocalizedCharField(l10n_context='DB: products.Topic.title')
```
mythmon: In DRF 3.x, redundant sources are an error. Here it is redundant because DRF infers the source from the name of the assignment on the serializer.

kitsune/products/api_utils.py
```diff
@@ -0,0 +1,22 @@
+from django.core.exceptions import ValidationError
+
+from rest_framework import serializers
+
+
+class TopicField(serializers.SlugRelatedField):
```
mythmon: This was moved from kitsune/products/api.py. As you can see, it also got a lot simpler thanks to changes in DRF's serializers.

kitsune/questions/api.py
```diff
@@ -22,60 +23,42 @@
...
     created = DateTimeUTCField(read_only=True)
-    creator = serializers.SerializerMethodField('get_creator')
-    involved = serializers.SerializerMethodField('get_involved_users')
-    is_solved = serializers.Field(source='is_solved')
-    is_taken = serializers.Field(source='is_taken')
-    metadata = QuestionMetaDataSerializer(source='metadata_set', required=False)
-    num_votes = serializers.Field(source='num_votes')
-    product = serializers.SlugRelatedField(required=True, slug_field='slug')
-    tags = QuestionTagSerializer(source='tags', read_only=True)
+    creator = serializers.SerializerMethodField()
```
mythmon: SerializerMethodField now have a sane default for the name of the method. Specifying a value that is equal to the sane default is an error. The default is get_{variable_name}.

mythmon:
Oh, PS: I also verified this a little by poking around with BuddyUp. That's where the extra test case comes from.

willkg:
I'll look through this later today.

kitsune/notifications/tests/test_api.py
willkg: I'm not sure what this line is doing. Is it verifying that the data can be parsed with that template? If so, maybe add a comment like:

```
# If this data isn't in the right form, then this will throw an exception.
```

kitsune/questions/api.py
```diff
@@ -108,30 +92,37 @@ class Meta:
             'updated',
         )
 
-    def get_involved_users(self, obj):
-        involved = set([Profile.objects.get(user=obj.creator)])
-        involved.update(Profile.objects.get(user=a.creator) for a in obj.answers.all())
+    def get_involved(self, obj):
+        involved = set([obj.creator.profile])
+        involved.update(a.creator.profile for a in obj.answers.all())
```
willkg: Wow--that old code is kind of weird.

kitsune/search/tests/test_api.py
```diff
@@ -122,15 +106,15 @@ def test_invalid_numbers(self):
             'q': 'search',
         })
         eq_(res.status_code, 400)
-        eq_(res.data['detail'], {
-            'max_questions': [u'Enter a whole number.'],
-            'max_documents': [u'Enter a whole number.'],
+        eq_(res.data, {
```
willkg: This suggests there's a change in the API results. It looks like it's entirely with the "something is wrong with the request" case and that the "everything is fine--here's your search results" case is ok.

Is that true?

kitsune/notifications/tests/test_api.py
mythmon: The main goal is to assert that the string returned looks like a date. I'll add a comment to make this clearer.

kitsune/sumo/api_utils.py
```diff
@@ -298,51 +253,35 @@ def has_object_permission(self, request, view, obj):
         return user == creator
 
 
-def PermissionMod(cls, permissions):
+PermissionListSerializer = None
+
+
+def PermissionMod(field, permissions=[]):
```
willkg: Generally it's bad to use mutable values for default args. Are we sure this never gets modified in any way by this code or downstream code?

kitsune/search/tests/test_api.py
mythmon: Yes, that is true. DRF changed the way it rendered errors normally. The "everything is fine" case is the same, but they unpacked the error responses a bit, moving from

```json
{
    "detail": {
        "field": ["list of errors"],
    }
}
```

to the simple

```json
{
    "field": ["list of errors"],
}
```

Considering where this API is used, I think that is an ok change to make, although technically it represents a breaking change. I'm not sure how we could force DRF to use the previous style of error handling (and I'm a bit grumpy that they changed it).

What do you think? Should we try and bend DRF into submission and make it use the old way?

kitsune/sumo/api_utils.py
```diff
@@ -298,51 +253,35 @@ def has_object_permission(self, request, view, obj):
...
+            request = self.context.get('request')
+            for Perm in permissions:
+                perm = Perm()
+                if not perm.has_permission(request, self):
+                    return False
+                if not perm.has_object_permission(request, self, obj):
+                    return False
+            return True
 
     return Modded
```
willkg: I think this messes up introspection of this thing. With functions you have `@wraps`. Classes are a little harder. This might be useful, but I only skimmed it:

http://stackoverflow.com/questions/6394511/python-functools-wraps-equivalent-for-classes

I don't know if it matters a ton, but it might if you were debugging something that involved this.

kitsune/sumo/api_utils.py
mythmon: On the one hand, I'm pretty sure this doesn't get mutated. On the other hand, we have no way to guarantee that, so this makes it a potential problem in the future. I'll change it.

kitsune/users/api.py
```diff
@@ -206,60 +205,54 @@ def get_last_answer_date(self, profile):
...
+        if 'user' in validated_data:
+            user_data = validated_data.pop('user')
+            for key, val in user_data.items():
+                setattr(instance.user, key, val)
+            instance.user.save()
+        return super(ProfileSerializer, self).update(instance, validated_data)
+
+    def validate_username(self, username):
+        if re.match(r'^[\w.-]{4,30}$', username) is None:
+            raise ValidationError('Usernames may only be letters, numbers, "." and "-".')
```
willkg: I'm totally taking the username `...`.

willkg: Bah! `....` and `.....` are taken, but `......` is not!

kitsune/users/api.py
```diff
@@ -206,60 +205,54 @@ def get_last_answer_date(self, profile):
...
-        email = attrs.get('user.email')
-        if email and User.objects.filter(email=email).exists():
-            raise ValidationError('A user with that email address '
-                                  'already exists.')
-        return attrs
+    def validate_email(self, email):
+        if not self.instance:
+            # create
+            if User.objects.filter(email=email).exists():
+                raise ValidationError({'email': 'A user with that email address already exists.'})
```
willkg: Where this is specifically validating the email address, does it need to take a dict arg or is a string arg ok?

willkg: The example here suggests this should just be a string--not a dict:

http://www.django-rest-framework.org/api-guide/serializers/#field-level-validation

The docs aren't particularly clear on whether the code handles both dict of `key` -> `error string` and `error string` fine or whether this needs to be a string.

mythmon: A string arg would be fine. I'm not sure why I was extra specific here.

kitsune/search/tests/test_api.py
willkg: I don't know. If you tested buddyup, then it's probably fine, but if it isn't, then we have to deal with a lot more work.

I'll check Fjord's sumo suggester, but I don't think I do anything with errors other than ignore them, so I think that code will be fine.

I'm not sure what else uses SUMO APIs.

mythmon: None of the consumers of the API that I know of actually check these errors.

kitsune/sumo/api_utils.py
mythmon: I think that the Stack post you linked is talking about something else. They seem to be talking about the way of writing decorators where you have a class with an `__init__` and a `__call__` method that returns a function, instead of having triply nested functions.

I'm not sure this matters. When debugging decorated functions, there isn't any standard way to get at the original function, so you try really hard to make sure not to erase information. With a class decorator that works by subclassing like this, you can get back the original class (`decorated_class.__class__.__mro__[0]`). It isn't really easy, but it does exist.

We've had this decorator-a-like for a while now, and it hasn't been a problem that we aren't doing any metadata copying on it. Further, I did a bunch of debugging around it while working on this patch, and I didn't ever think "damn, I wish I had just preserved the metadata".

Given those things, I think I'm going to leave this like it, unless you have any clever ideas on how to copy the metadata we might want.

mythmon:
^ Feedbacked.

I think that covers everything we talked about.

willkg:
Looks gouda to me. r+!

mythmon:
Thanks for reviewing this!

willkg:
You're welcome! Thank you for the comments--they made it a ton easier.

PR #2704, opened 2015-11-07
[Bug 1169416] added funtionality to show full locale name in show_tra…

skiee-mahir:
…nslations.html

### DIFF
--- a/kitsune/wiki/templates/wiki/show_translations.html
+++ b/kitsune/wiki/templates/wiki/show_translations.html
@@ -11,13 +11,13 @@
       <h3>{{ _('This document has been translated to the following locales:') }}</h3>
       <ul class="locales">
         {% for locale in translated_locales %}
-          <li><a class="translated_locale" href="{{ url("wiki.document", locale=locale, document_slug=document.slug) }}">{{ locale }}</a></li>
+          <li><a class="translated_locale" href="{{ url("wiki.document", locale=locale[0], document_slug=document.slug) }}">{{ locale[1] }} ({{ locale[0] }})</a></li>
         {% endfor %}
       </ul>
       <h3>{{ _('This document is missing translations for the following locales:') }}</h3>
       <ul class="locales">
         {% for locale in untranslated_locales %}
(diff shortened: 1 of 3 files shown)


skiee-mahir:
@mythmon can you check?

kitsune/wiki/templates/wiki/show_translations.html
```diff
@@ -1,4 +1,4 @@
-{% extends "wiki/base.html" %}
+k{% extends "wiki/base.html" %}
```
safwanrahman: k? why?

kitsune/wiki/views.py
```diff
@@ -1382,13 +1382,17 @@ def show_translations(request, document_slug):
     translated_locales = []
     untranslated_locales = []
 
-    translated_locales.append(document.locale)
-    translated_locales.extend(document.translations.all().values_list(
-        'locale', flat=True))
+    # Makes Sure "English (en-US)" is always on top
+    translated_locales.append((document.locale, LOCALES[document.locale].native))
+    # Gets all the translated locale codes
+    translated_locales_code = document.translations.all().values_list(
```
safwanrahman: hmm! Don't call '''all()''' unless you don't neeed to make database call everytime.

safwanrahman:
Just some simple neat and clean up.
Otherwise it seems ok. 
@mythmon I think this should get r+

skiee-mahir:
@safwanrahman do you have any idea why the test failed? When it was OK with the mistaken k?

safwanrahman:
@skiee-mahir Seems to be a problem with Travis. Can yo fix the above mentioned issue and squash into one commit and push again?

kitsune/wiki/views.py
mythmon: `all()` here doesn't actually do anything. It is only useful when there are no other method calls after `.objects`. Please remove it.

mythmon: Oh, I see that this code already exists. I'll fix it later then.

mythmon:
r+

PR #2706, opened 2015-11-10
[Bug 1155698] change ’ to '

pollti:
This was causing problems in https://pontoon.mozilla.org/de/sumo/search/?keyword=Do+not+disclose+personal+user+information&sources=on&translations=on.


willkg:
These strings come from the database, so you need to change them there. This file is a generated file.

Closing this out.

PR #2707, opened 2015-11-10
[bug 1223370] Clarify db_strings.py header

willkg:
People keep submitting pull requests that make changes to the
kitsune/sumo/db_strings.py file. This is a generated file, so any
changes made directly to it will get stomped on the next time it's
generated. Changes should be made to the relevant content in the db.

These changes to the header hopefully clarifies the situation to
future editors of this file.

r?


rehandalal:
good call! r+

PR #2708, opened 2015-11-10
[Bug 1204515] Noisy email

mythmon:
1. Add a lot of logging to sending email at the lowest level we can.
2. Change our logging to expose that information (I've talked with webops and we will be able to get these logs).
3. Revert the last attempt at this, since it didn't help.

r?

### DIFF
--- a/kitsune/lib/email.py
+++ b/kitsune/lib/email.py
@@ -1,66 +1,99 @@
 import logging
+import time
 
 from django.conf import settings
 from django.utils.module_loading import import_string
 
 
 log = logging.getLogger('k.lib.email')
-RealBackend = import_string(settings.EMAIL_LOGGING_REAL_BACKEND)
 
 
-class LoggingEmailBackend(RealBackend):
+class LoggingEmailBackend(object):
     """
     Wraps a email backend defined in Django's settings and logs everything it does.
     """
 
+    def __init__(self):
+        self._batch_id = None
+        # Set up the real backend
+        RealBackend = import_string(settings.EMAIL_LOGGING_REAL_BACKEND)
+        self.real_backend = RealBackend()
+
+    @property
+    def batch_id(self):
+        """Get a batch_id, or lazily make a new one."""
+        if self._batch_id is None:
+            # This is the number of milliseconds since the last midnight. It
+            # doesn't need to be globally unique, just unique enough to
+            # distinguish between email tasks that happen at a similar time.
+            self._batch_id = int(time.time() * 1000) % (1000 * 60 * 60 * 24)
+        return self._batch_id
+
+    def log(self, level, msg):
+        """Write a log message, prepending the current batch id."""
+        log.log(level, 'Batch %s - %s' % (self.batch_id, msg))
+
     def open(self):
         """Open a network connection."""
-        new_conn = super(LoggingEmailBackend, self).open()
+        new_conn = self.real_backend.open()
+
         # Checking for True or False directly instead of relying on trutheyness
         # to avoid catching non-booleans
         if new_conn is True:
-            log.debug('Succesfully opened new connection.')
+            self.log(logging.DEBUG, 'Succesfully opened new connection.')
         elif new_conn is False:
-            log.debug('Did not open a new connection. (Either cached or failed)')
+            self.log(logging.DEBUG, 'Did not open a new connection. (Either cached or failed)')
+        elif new_conn is None:
+            self.log(logging.DEBUG, 'Opened a new connection. (Unknown status)')
         else:
-            assert (
-                new_conn is None,
-                'Unexpected return from email backend open method: %r.' % (new_conn, )
-            )
-            log.debug('Opened a new connection. (Unknown status)')
+            # If new_conn is not True, False, or None, the backend is
+            # not behaving as expected. Blow up.
+            raise AssertionError(
+                'Unexpected return from email backend open method: %r.' % (new_conn, ))
+
         return new_conn
 
     def close(self):
         """Close a network connection."""
-        ret = super(LoggingEmailBackend, self).close()
-        log.debug('Closed connection.')
+        ret = self.real_backend.close()
+        self.log(logging.DEBUG, 'Closed connection.')
+        # Clear the batch ID, in case this backend object is re-used.
+        self._batch_id = None
         return ret
 
-    def send_messages(self, email_messages):
+    def send_messages(self, messages):
         """
         Sends one or more EmailMessage objects and returns the number of email messages sent.
         """
-        num_to_send = len(email_messages)
+        num_to_send = len(messages)
 
-        msg_lines = ['Attempting to send %(count)s emails:' % {'count': num_to_send}]
-        for message in email_messages:
-            msg_lines.append('\t%(subject)s - %(recipients)r' % {
+        # Build a big, multiline log "line" that lists all the emails we are trying to send.
+        first_line = 'Attempting to send %(count)s emails: ' % {'count': num_to_send}
+        msg_lines = [
+            '%(subject)s - %(recipients)r' % {
                 'subject': message.subject,
-                'recipients': message.recipients(),
-            })
-        log.debug('\n'.join(msg_lines))
+                'recipients': ', '.join(str(r) for r in message.recipients()),
+            }
+            for message in messages
+        ]
+        self.log(logging.DEBUG, first_line + '\n\t'.join(msg_lines))
 
-        num_sent = super(LoggingEmailBackend, self).send_messages(email_messages)
-        if num_sent is None:
-            num_sent = 0
-        assert num_sent <= num_to_send, 'Sent more emails than requested'
+        # Some backend sometimes return None when they don't attempt to send messsages.
+        num_sent = self.real_backend.send_messages(messages) or 0
 
         if num_sent == num_to_send:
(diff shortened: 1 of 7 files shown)


kitsune/lib/email.py
```diff
@@ -0,0 +1,66 @@
...
+
+from django.conf import settings
+from django.utils.module_loading import import_string
+
+
+log = logging.getLogger('k.lib.email')
+RealBackend = import_string(settings.EMAIL_LOGGING_REAL_BACKEND)
+
+
+class LoggingEmailBackend(RealBackend):
```
willkg: Why do the `RealBackend` thing and pull it from settings?

willkg: Nevermind--I figured it out. It's because the dev backend is different than what we have in production. That probably means that to land this, we need to change the `settings_local.py` on the servers, right?

kitsune/lib/email.py
```diff
@@ -0,0 +1,66 @@
...
+        # to avoid catching non-booleans
+        if new_conn is True:
+            log.debug('Succesfully opened new connection.')
+        elif new_conn is False:
+            log.debug('Did not open a new connection. (Either cached or failed)')
+        else:
+            assert (
+                new_conn is None,
+                'Unexpected return from email backend open method: %r.' % (new_conn, )
+            )
```
willkg: This seems odd. Why not just handle the `None` case in an `elif` that also checks to see if the backend is the console?

kitsune/lib/email.py
```diff
@@ -0,0 +1,66 @@
...
+
+        if num_sent == num_to_send:
+            log.debug('Succesfully sent %(sent)s messages' % {'sent': num_sent})
+        else:
+            log.error('Failed to send all emails. Sent %(sent)s out of %(to_send)s' % {
+                'sent': num_sent,
+                'to_send': num_to_send,
+            })
+
+        return num_sent
```
willkg: This logs a bunch of stuff over a period of time. You might want to create a batch id and prepend that to all the lines so it's easier to look at logs and match entire batches.

kitsune/lib/email.py
```diff
@@ -0,0 +1,66 @@
...
+            num_sent = 0
+        assert num_sent <= num_to_send, 'Sent more emails than requested'
+
+        if num_sent == num_to_send:
+            log.debug('Succesfully sent %(sent)s messages' % {'sent': num_sent})
+        else:
+            log.error('Failed to send all emails. Sent %(sent)s out of %(to_send)s' % {
+                'sent': num_sent,
+                'to_send': num_to_send,
+            })
```
willkg: Why handle the `<=` case with an assert, but the `==` and `>` case with this `if` block?

willkg:
What's the backend we use on the servers? How can we test this code on all the relevant backends?

kitsune/lib/email.py
mythmon: The assert line is sanity checking. The two branches in the if statement are normal operation, where the functions are doing what they are expected to.

The assert statement is an exceptional case where the backend sent more emails that it was asked to. This is something that should never ever happen. The assertion is there to make it blow up very loudly if that supposedly-impossible case ever happens.

mythmon:
On the server we use the Django standard SMTP backend: `django.core.mail.backends.console.EmailBackend`. We could test that by setting up SMTP on a dev instance. I think the docs have instructions for how to do that with Gmail.

Alternatively, we could test that by putting this on stage and trying to make some emails happen.

willkg:
This doesn't add or augment any tests. Is the expectation that we should test this by hand before landing? If so, can you walk me through the STR you used to test?

kitsune/lib/email.py
mythmon: If possible, I'd like to avoid explicitly checking what the backend is. Like the other assertion, this is about validating expectations. If this assert fails, then the backend didn't act like I expected it to, so things should blow up.

I tried to use assertions to validate my assumptions, and non-assertion code to deal with what I expected the backends to do.

willkg: This code is mixing two different ways of checking things and there aren't any comments so I find it confusing and think it just looks awkward and possibly wrong.

It'd be easier if there were comments about what's going on and the context of the checks.

kitsune/lib/email.py
willkg: I'm not wild about mixing two different ways of checking things. You could easily add an `elif` and throw an exception.

Comments would be good here.

mythmon:
My thinking while writing this is that the goal is to transparently proxy the email backend, so if the tests pass then the transparent proxy worked. However, I realize now that tests use a test-specific email backend, so none of the tests will exercise this code. Further, I don't know how to test logging output. I guess I could manually instantiate the class and use mock to assert it made calls to logs.

Do you think writing tests to test this in isolation is valuable? Or will the manual testing be enough?

To test this by hand, I used the site to do something that would send an email (I used private messages). I watched the console where `runserver` was running to make sure that both the normal email logging and the extra verbose email logging worked. I didn't test the SMTP version of this, but the process would be similar.

kitsune/lib/email.py
mythmon: That's a good idea. Do you have any clever ways of generating batch ids? My first instinct is to use a random number, or possibly something based on the when the function started.

willkg:
nose has a logcapture thing if that helps:

http://nose.readthedocs.org/en/latest/plugins/logcapture.html

Would it help to create a django command that sends a test email? That seems like it's an easy thing to implement and it'd probably help us debug additional email-sending infrastructure.

It's definitely worth testing the smtp side of this before merging this since that's the side that runs on the servers and is the thing you want to make sure works. Maybe push this branch to -stage and test it there?

kitsune/lib/email.py
willkg: The only thing you need out of it is that you can connect logging statements that happened in the same run. I don't think it needs anything more than that. Using a timestamp from when it started or something derived from that is probably fine.

kitsune/lib/email.py
mythmon: Exactly. I filed [bug 1223471](https://bugzilla.mozilla.org/show_bug.cgi?id=1223471) for the `settings_local.py` change.

mythmon:
I don't think logcapture actually makes logs available in tests. I think it only deals with printing them out when they fail. I could adapt the code from logcapture though, and add another root log handler to look at during a test.

I'll take another pass at this to try and make it easier to understand and try to add some tests.

One wrinkle about testing smtp on stage is that we don't have easy access to the logs on the servers. We'd have to get webops to help with this. We'll have to do that eventually anyways, so it isn't that big a deal.

willkg:
My bad. testfixtures can capture logging:

https://www.pythonhosted.org/testfixtures/logging.html

The thing that does it in testfixtures is also called logcapture hence my mistake.

I think I'd want a few things:

1. (optional) batch id to make reading logs easier
2. some comments around the asserts to clarify the context and why asserts vs. ifs
3. testing the SMTP backend on -stage which verifies that this does what we need it to do in the server environments
4. some promise that this is a temporary change and not a permanent one; if it's permanent, then we really want some tests

After that, I'm game for this.

mythmon:
I added the batch id idea, added tests, and did some refactoring to help with that.

The biggest change is that this no longer subclasses the real backend. Instead it just proxies to an instance of it. I couldn't get it so that I could easily swap out the backends in tests when it was subclassing.

I also removed the assert statements in favor of `raise AssertionError` when something is weird. I hope this reads better.

The tests are pretty gnarly. I hope they read clearly enough, let me know if any part of them need improvements. They could probably be factored to be fewer lines of code, but I suspect they would be a bit harder to follow.

mythmon:
Oh, and I do expect this to be pretty short lived. Once the problem goes away I think we can nix this. I would still like to have passable test coverage for it though.

requirements/dev.txt
```diff
@@ -66,5 +67,8 @@ sphinxcontrib-httpdomain==1.4.0
 # sha256: kHGqy9l6mpFQlsGq8NxoSsJnKQTNh221kECF1trJgQ4
 tabulate==0.7.5
 
+# sha256: lM7ehUx4i_w1KArTvkn85vNiWTgp_YuOPKwe0RL-yFc
+testfixtures==4.5.0
```
willkg: I think this suggests you added tests, but I don't see any tests in this PR.

mythmon: Oops.

mythmon:
Added in the missing file with tests ^

kitsune/lib/tests/test_email.py
```diff
@@ -0,0 +1,227 @@
...
+    """
+    A mock email backend that pretends to send exactly half-rounded-down of the
+    emails it is asked to send per batch.
+
+    It does not mimic the SMTP backend.
+    """
+
+    def send_messages(self, messages):
+        super(FlakyMockEmailBackend, self).send_messages(messages)
+        return len(messages) / 2
```
willkg: Interesting!

willkg:
I think this looks good. Definitely worth testing with the SMTP backend on -stage and making sure it works there. r+

mythmon:
I just realized: landing this code doesn't effect anything on stage since we haven't updated `settings_local.py` on stage. I'm going to merge this, and deploy it to stage and prod. We can change the setting on stage first and verify again.

PR #2710, opened 2015-11-10
[bug 1125866] Nix l10n completion code

willkg:
This removes the script that was figuring out l10n completion status
every day and updating the relevant file. That code now runs on
people.m.o and is maintained in another repository.

r?


mythmon:
So how does the new system work? Is there a process on people that has it's own copy of the SVN repo, and doesn't depend on Sumo at all for this data?

willkg:
The new system code is here:

https://github.com/willkg/fjord-l10n

That runs via a cronjob on people.m.o that does what this script was doing. Then I changed postatus to point to the results of that script.

It doesn't depend on SUMO servers for translation history.

However, to clarify, the linting errors is still generated by SUMO during deploys and that file is still on SUMO. We could move that, too, if you like. Might make things easier since postatus wouldn't need to parse dennis/compile-script output.

mythmon:
Ok, sounds good. I just wanted to make sure I understood what was going on.

Isn't it important that we do l10n linting during deploy to avoid compiling in bad translations?

willkg:
It is! But I'm questioning whether postatus should look at the output of that linting step in the deploy which is combined with compiling or if it should be another daily cron job that runs on people.m.o and only generates linting output.

mythmon:
Ah! I understand now. So far I don't think we've had many problems with the lint output, so I think it is fine to leave as is. If you'd like to change it, I'd be game though.

Either way, I think this PR is good, so I'll merge it.

PR #2711, opened 2015-11-10
Add missing migrations

willkg:
These migrations update the various LocaleField and TimezoneField
using models.

Quick r?


willkg:
Pretty sure this doesn't affect the data in the database or the code. It only affects the migration system and moving forward and backward in time with correct choices on these fields.

But... it's what you get when you do `./manage.py makemigrations`.

mythmon:
I'm grumpy that these need to exist, but for now they do. r+ once Travis approves of it.

willkg:
Travis is green!

PR #2713, opened 2015-11-11
Revert nunjucks changes

rehandalal:
@mythmon r?


rehandalal:
hm... oh we probably don't want this...

mythmon:
Yeah... I think we want to keep the DRF upgrade.

PR #2715, opened 2015-11-11
Downgrade nunjucks to v1.3.4

rehandalal:
phew!

r?


mythmon:
I `git bisected` until git told me which commit was the problem. It was 4dcd9d1ab00456c12ebd3ca8d572c16dff4c7676 which includes the Nunjucks upgrades, so that matches. Then I grabbed this commit, and tested it, and it works locally (using the same tests that I used for the bisect).

I'm pretty confident this will work.

mythmon:
r+

/me waits for Travis.

PR #2717, opened 2015-11-11
917754 indexing

willkg:
r?

### DIFF
--- a/kitsune/search/admin.py
+++ b/kitsune/search/admin.py
@@ -1,4 +1,5 @@
 import logging
+import time
 from datetime import datetime
 from difflib import SequenceMatcher
 
@@ -17,10 +18,8 @@ from kitsune.search.es_utils import (
     all_read_indexes, all_write_indexes)
 from kitsune.search.models import Record, get_mapping_types, Synonym
 from kitsune.search.tasks import (
-    OUTSTANDING_INDEX_CHUNKS, index_chunk_task, reconcile_task,
-    update_synonyms_task)
-from kitsune.search.utils import chunked, create_batch_id
-from kitsune.sumo.redis_utils import redis_client, RedisError
+    index_chunk_task, reconcile_task, update_synonyms_task)
+from kitsune.search.utils import chunked, to_class_path
 from kitsune.wiki.models import Document, DocumentMappingType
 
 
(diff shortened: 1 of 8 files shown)


willkg:
This makes a bunch of changes to the search maintenance admin page.

1. puts outstanding records on top for easy viewing
2. makes it clearer what's going on
3. reformats the action section so that text is on the left and action forms are on the right for easier usage
4. adds a new reconcile action
5. makes the recent records at the bottom clearer

Further it does:

1. fleshes out Record so it's way more useful and thinks about things in batches
2. nixes use of Redis
3. because of no-more-Redis, this is less flaky
4. cleans up some of the indexing code so it works better with celery and serialization

To test:

1. run the tests (or let Travis run them for you!)
2. change `CELERY_ALWAYS_EAGER = False` and go through the actions on the indexing admin page

The latter is a pain in the ass and easily takes a half day plus we're going to want to do them again on -stage. I went through them all this morning and feel good about them. I'm ok with taking the responsibility for running it on -stage when we push this out and fixing any outstanding issues there.

kitsune/search/admin.py
```diff
@@ -48,6 +38,33 @@ class DeleteError(Exception):
     pass
 
 
+def create_batch_id():
+    """Returns a batch_id"""
+    # TODO: This is silly, but it's a good enough way to distinguish
+    # between batches by looking at a Record. This is just over the
+    # number of seconds in a day.
+    return str(int(time.time()))[-6:]
```
willkg: I moved this over from another module. This is the only place in kitsune it gets used so might as well have it here.

The actual value doesn't matter a whole lot. If the way we generate it is an issue, we can change it--I don't really care one way or the other.

kitsune/search/admin.py
```diff
@@ -48,6 +38,33 @@ class DeleteError(Exception):
...
+
+    for cls, indexable in get_indexable():
+        index = cls.get_index()
+        doc_type = cls.get_mapping_type_name()
+        chunk_name = 'Reconciling: %s' % doc_type
+        rec = Record(batch_id=batch_id, name=chunk_name)
+        rec.save()
+        reconcile_task.delay(index, batch_id, rec.id, doc_type)
+
+    return HttpResponseRedirect(request.path)
```
willkg: This is new. I figure it might be helpful to be able to reconcile the db vs. the index all by itself. Having said that, it timed out on my machine when I tried. I suspect it's because it's pulling the ids for too many things from ES and thus it won't ever work. If that's true, we might want to either change it to chunkify things or nix it if we don't care.

kitsune/search/migrations/0002_delete_record.py
```diff
@@ -0,0 +1,17 @@
...
+
+    dependencies = [
+        ('search', '0001_initial'),
+    ]
+
+    operations = [
+        migrations.DeleteModel(
+            name='Record',
+        ),
+    ]
```
willkg: We don't care about previous records. It's way easier to delete the table and then create the new one rather than try to migrate between them.

willkg:
This has conflicts. I'll resolve them monday.

willkg:
^^^ Conflicts resolved.

kitsune/search/admin.py
```diff
@@ -48,6 +38,33 @@ class DeleteError(Exception):
...
+        raise ReindexError('There are %s outstanding chunks.' % outstanding)
+
+    batch_id = create_batch_id()
+
+    for cls, indexable in get_indexable():
+        index = cls.get_index()
+        doc_type = cls.get_mapping_type_name()
+        chunk_name = 'Reconciling: %s' % doc_type
+        rec = Record(batch_id=batch_id, name=chunk_name)
+        rec.save()
```
mythmon: Any reason not to use `Record.objects.create(batch_id=batch_id, name=chunk_name)` instead of the two line version?

willkg: Mmm... Nope. I'll change it.

willkg:
^^^ Switches to use `Record.objects.create`.

kitsune/search/tasks.py
```diff
@@ -53,97 +53,88 @@ def __init__(self):
...
-            rec.text, sys.exc_type, sys.exc_value)[:255]  # Truncate at 255 chars.
+        if rec is not None:
+            rec.mark_fail(u'Errored out %s %s' % (sys.exc_type, sys.exc_value))
 
         log.exception('Error while indexing a chunk')
-
-        # Some exceptions aren't pickleable and we need this to throw
-        # things that are pickleable.
-        raise IndexingTaskError()
+        raise
```
mythmon: Is there a reason we don't need to worry about pickle-ability of exceptions anymore?

mythmon:
Sorry it took so long to get to this. I think this looks good, but I haven't done the whole song and dance to verify it. I think doing that on stage makes more sense.

Can you rebase this on master one more time to pick up the email logging changes I just landed, so we can test both together (in case WebOps is extra quick today). After that we can get this on stage to do final testing there.

kitsune/search/tasks.py
willkg: I don't have the `IndexingTaskError` in Fjord. I'm not sure why I don't, but I've never had problems from it. I don't remember offhand which exceptions weren't pickleable, but I think they were elasticsearch-related ones.

I'll re-add this because I can't guarantee that removing it is a good idea.

Good catch. Sorry about that.

willkg:
^^^ Fixes pickleable things and rebases against master.

mythmon:
Alright, lets do this.

PR #2719, opened 2015-11-11
[bug 1136188] Tower extracts strings from nunjucks

rehandalal:
r?

### DIFF
--- a/kitsune/settings.py
+++ b/kitsune/settings.py
@@ -631,10 +631,11 @@ DOMAIN_METHODS = {
     ],
     'javascript': [
         # We can't say **.js because that would dive into any libraries.
-        ('kitsune/**/static/js/*-all.js', 'ignore'),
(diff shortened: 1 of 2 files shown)


mythmon:
I don't see any new strings in javascript.pot (or any other .pot file for that matter) when I run `./manage.py extract` with this patch. Did I do something wrong?

rehandalal:
weird... i see them... 
https://pastebin.mozilla.org/8851954

mythmon:
Doh. I was using the wrong branch when I was testing this. \*fail\*.

This is awesome, thanks! I think this also fixes [bug 1180783](https://bugzilla.mozilla.org/show_bug.cgi?id=1180783), at least from what I saw in the svn diff. Do you agree?

rehandalal:
yeah it also fixes that bug!

PR #2721, opened 2015-11-12
[Bug 1210490] Run smoke tests in CI

safwanrahman:
The selenium tests are running in CI. But the selenium tests are failing. I think this bug does not cover fixing the selenium tests also.
@mythmon r?

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,7 +8,7 @@ env:
   - PIP_DOWNLOAD_CACHE="pip_cache"
   matrix:
   - TEST_SUITE=lint
-  - TEST_SUITE=selenium ELASTICSEARCH_VERSION=1.2.4
+  - TEST_SUITE=smoke ELASTICSEARCH_VERSION=1.2.4
   - TEST_SUITE=django ELASTICSEARCH_VERSION=1.2.4
 matrix:
   fast_finish: true
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,6 +8,7 @@ env:
   - PIP_DOWNLOAD_CACHE="pip_cache"
   matrix:
   - TEST_SUITE=lint
+  - TEST_SUITE=selenium ELASTICSEARCH_VERSION=1.2.4
   - TEST_SUITE=django ELASTICSEARCH_VERSION=1.2.4
 matrix:
   fast_finish: true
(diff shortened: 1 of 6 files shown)


.travis.yml
```diff
@@ -8,6 +8,7 @@ env:
   - PIP_DOWNLOAD_CACHE="pip_cache"
   matrix:
   - TEST_SUITE=lint
+  - TEST_SUITE=selenium ELASTICSEARCH_VERSION=1.2.4
```
mythmon: This should be `TEST_SUITE=smoke ...`.

mythmon:
I don't know why it is failing that way.

You might want to look at how we had selenium set up in the past. Maybe there are some good ideas there. You can look at the history of `scripts/travis/test.sh` to find a version that still used selenium.

safwanrahman:
@mythmon I have fixed the screen width which was cause the failing. for low resulation screen, it was redirecting to auth page with https. As the built in server does not suppport SSL, it was showing that type of Error
Fixing screen width let the test run. But still the test is failing. I dont know why its failing. Either there is problem in the test or the test is failing because the feature is not working correctly.

mythmon:
Haha! That's a pretty hilarious set of problems, actually. Awesome job figuring it out, it explains all the problems we have been having. I think it is a bug in Sumo that it is redirecting you to an HTTPS URL when the site isn't on URL, and that it matters that the screen is very small when it does that. But that's not important for now.

Looking at the problem you're hitting now, I'm not sure why that would happen. These tests work in other places, so I think the feature works, and the test works. There must be something specific to Travis that is causing the locale not to change. Maybe the test is written in a way that running on Travis makes it hit an edge case that breaks it?

mythmon:
The tests have passed 5 times now. I'm going to guess that the one failure we saw was a fluke. We should look into why it fails more, but for now, this looks good.

One last thing, can we rename `selenium.sh` to `smoketests.sh`, and do the same to any reference in the scripts and environment variables? These are not specifically selenium tests, they just happen to be implemented with selenium. We could change them to use another technology in the future and they'd still be smoke tests.

safwanrahman:
@mythmon Fixed as per mentioned
r?

mythmon:
Thanks!

PR #2722, opened 2015-11-12
[bug 1224351] Fix language default in advanced search

willkg:
This fixes the form so that English is the default in the language
dropdown. The problem was that we weren't setting the language in
the request.GET before creating the form.

quick r?


mythmon:
Well, that's easy. r+ once Travis gets around to approving th is.

willkg:
Travis is green!

PR #2723, opened 2015-11-12
[Bug 1223970] Fix XSS in instant search

mythmon:
The actual fix was pretty easy. Just need to escape some things that go into urls.

The bulk of this work is testing the feature though. Instant search had no tests, and nothing using Nunjucks has been tested in Mocha yet. So this sets up a Nunjucks testing environment, and tests instant search.

One of the new tests is for the bug fixed here. Yay!

r?

### DIFF
--- a/kitsune/sumo/static/sumo/js/tests/instant_search_tests.js
+++ b/kitsune/sumo/static/sumo/js/tests/instant_search_tests.js
@@ -1,0 +1,118 @@
+import {default as mochaJsdom, rerequire} from 'mocha-jsdom';
+import {default as chai, expect} from 'chai';
+import React from 'react';
+import chaiLint from 'chai-lint';
+import sinon from 'sinon';
+
+import mochaK from './fixtures/mochaK.js';
+import mochaJquery from './fixtures/mochaJquery.js';
+import mochaGoogleAnalytics from './fixtures/mochaGoogleAnalytics.js';
+import mochaNunjucks from './fixtures/mochaNunjucks.js';
+import mochaGettext from './fixtures/mochaGettext.js';
+
+chai.use(chaiLint);
+
+describe('instant search', () => {
+  mochaJsdom({useEach: true});
+  mochaJquery();
+  mochaK();
+  mochaGoogleAnalytics();
+  mochaGettext();
+  mochaNunjucks();
+  /* globals window, document, $ */
+
+  describe('', () => {
+    let $sandbox;
+    let clock;
+    let cxhrMock;
+
+    beforeEach(() => {
+      clock = sinon.useFakeTimers();
+      cxhrMock = sinon.mock({request: () => {}});
+      window.k.CachedXHR = () => cxhrMock.object;
+
+      rerequire('../i18n.js');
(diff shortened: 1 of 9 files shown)


rehandalal:
travis no likey

mythmon:
Oops. I forgot to repin the npm requirements.

mythmon:
Ugh. I'm not sure what's wrong here. I can sort of reproduce it locally, but only after I `rm  node_modules; npm install`. So there must have been something in my environment that made this work.

mythmon:
Haha, I figured it out. The problem was that even though I was overriding the template loader in tests, the original template loader still got loaded. Because these tests technically run in Node, Nunjucks loaded the FileSystemLoader. Over and over and over again. Every time it loaded, it registered a thing that would watch for changes on the file system. Over and over and over again. Linux imposes a limit on how many watches on the file system you can have, and we were hitting it.

Preventing FileSystemLoaders from trying to watch the FS makes this go away.

mythmon:
@brittanystoroz, I hear you dug into the tests a bit lately. Can you look through the tests here?

kitsune/sumo/static/sumo/js/tests/instant_search_tests.js
```diff
@@ -0,0 +1,140 @@
...
+describe('instant search', () => {
+  mochaJsdom({useEach: true});
+  mochaJquery();
+  mochaK();
+  mochaGoogleAnalytics();
+  mochaGettext();
+  mochaNunjucks();
+  /* globals window, document, $ */
+
+  describe('', () => {
```
brittanystoroz: Is there something that makes sense to pass in here for what's being described? I'm a little confused by the need for the nested `describe` blocks? Is this setup required because of a library we're using?

mythmon: Oh, that could be better, sorry. This is mainly to group the things below it into one logical block that can share beforeEach and afterEach hooks. This is mainly because all of the `mocha*` functions above set up beforeEach and afterEach hooks.

mythmon: The describe blocks usually go something like 

```js
describe('library', () => {
  describe('method in library', () => {
    it('should do the things');
  });
});
```

There isn't really a library here, so that second layer doesn't make a lot of sense, but it is still needed for the `mocha*` helpers. At least, I think it is. I remember going through a lot of debugging here, so maybe this is just a spurious change.

mythmon:
I'm going to merge this now, since I want to get the underlying issue fixed ASAP. Please continue to review the tests if you've got anything else to say, and we can fix them up in another PR.

kitsune/sumo/static/sumo/js/tests/instant_search_tests.js
mythmon: Um, scratch all that. I had a lot of trouble getting Nunjucks working, so I did a lot of stuff that didn't always end up helping. I guess this was one of them. I can rearrange the code a bit to remove the describe block, since it is confusing.

PR #2724, opened 2015-11-13
[Bug 1198961] Fix post reply links

brittanystoroz:
This [bug](https://bugzilla.mozilla.org/show_bug.cgi?id=1198961) was caused by the `selectionStart` property being represented as an integer rather than a string. I originally changed the strict equals to compare against an integer, but I think it might be better to just do a `==` as I wasn't really able to determine when that property started being represented as an integer...I imagine there are scenarios where it may still be represented as a string.

I'm writing some tests for the Marky tool this code is a part of, but figured this was a harmless enough of a change to not be blocked by those. 

Thoughts?

### DIFF
--- a/kitsune/sumo/static/sumo/js/markup.js
+++ b/kitsune/sumo/static/sumo/js/markup.js
@@ -156,7 +156,7 @@
         // IE/Opera
         selText = document.selection.createRange().text;
       } else if (this.textarea.selectionStart ||
-        this.textarea.selectionStart == '0') {
+        parseInt(this.textarea.selectionStart) === 0) {
         // Firefox/Safari/Chrome/etc.
         selText = this.textarea.value.substring(
           this.textarea.selectionStart, this.textarea.selectionEnd);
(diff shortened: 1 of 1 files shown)


rehandalal:
wfm. but maybe add a comment about why we're using the `==` instead of `===` so it's not "fixed" by someone by mistake in the future.

mythmon:
The linters might have a problem with this, actually. Normally they would run in CI, but, well, they're broken right now. Merging this will probably make turning the linters back on harder.

We could do a few things:

1. Use `parseInt` to convert the value into a number. I think that would be safe on both strings and numbers, and then we can use === to compare it with a number.
2. Tell the linter to ignore the triple-equals rule on this line only, by adding a comment like `// eslint-disable-line eqeqeq` to the end of the line.
3. Decide we don't like the triple-equals rule, and disable it globally.

brittanystoroz:
I'll go the parseInt route

rehandalal:
+1 on `parseInt`

mythmon:
This looks good to me, the linters like it, and it fixes the problem. Thanks!

PS: You should remember to make comments if you want people to notice that you pushed new code ;)

PR #2726, opened 2015-11-17
Remove a level of nesting in the instant search tests.

mythmon:
@brittanystoroz pointed out that I had a weird amount of nesting in this test suite. This removes it and does a little bit of reorganization to make it work. See the comment in the change for the reasoning.

@brittanystoroz r?


brittanystoroz:
Closing for now as this doesn't make as much of an improvement as I'd expected. Will think about this more.

PR #2728, opened 2015-11-17
1225557 reconcile

willkg:
r?


willkg:
There's no test coverage for any of this code. (AWESOME!)

Best way to test it is to run the unit tests for kicks (which Travis will do) and then do the `./manage.py runserver` thing and make sure the admin doesn't blow up (a 5-second project).

mythmon:
Nothing exploded when I poked around the admin pages, and search still works. r+

PR #2729, opened 2015-11-17
Upgrade dev requirements to match requires.io.

mythmon:
Eventually I'd like to get to a point where everything on requires.io is up to date. That's a big project though. Upgrading the dev requirements should be pretty easy though. I updated anything that requires.io complained about, except Sphinx, and added `functools32` because precommit wouldn't run without it.

I didn't upgrade Sphinx because it would require us to upgrade Babel, which is not a dev dependency. Maybe it would be easy to do, but I didn't want to touch that file right now.

Things that should work:

* tests locally
* tests in travis
* making the docs
* pre-commit hooks
* ./manage.py runserver
* any normal debugging workflow you care to test

These all work for me locally

r?

### DIFF
--- a/requirements/dev.txt
+++ b/requirements/dev.txt
@@ -1,53 +1,56 @@
 # These requirements are used for the development and testing of
 # Kitsune, but they are not required for running it on servers.
 
-# sha256: s6eaI9N7WgL6pVC5LLu-vrSqHXfmScPrOcGav1Ji2gQ
-argparse==1.3.0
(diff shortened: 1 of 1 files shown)


willkg:
Looks fine to me. Generally, for dev requirements I just eyeball it and move on. If there are issues, someone will catch them and they're usually easy to fix.

mythmon:
That was kind of my thinking too. We touch this stuff a lot, presumably, so we should find any bugs quick.

PR #2730, opened 2015-11-18
[Bug 1225672] Fix list of fields for ImageAttachmentSerializer.

mythmon:
It's a dumb test, but it would have caught the problem this bug exhibits.

r?


willkg:
Looks good to me. r+

PR #2732, opened 2015-11-18
[Bug 1199555] RTL Locales out of view

brittanystoroz:
Adding some padding to the textarea prevents the RTL locales from being cut off when you type (as noted in bugzilla [here](https://bugzilla.mozilla.org/show_bug.cgi?id=1199555). I could add some specificity to this by wrapping it in the `body.html-rtl` selector as well, but it seems a minimal enough change that applying it at all times shouldn't hurt. 

Screenshot attached. 
rtl-locales

### DIFF
--- a/kitsune/sumo/static/sumo/less/questions.less
+++ b/kitsune/sumo/static/sumo/less/questions.less
@@ -409,7 +409,8 @@ h2 {
 
     textarea {
       margin-top: 10px;
-      width: 510px;
+      width: 500px;
+      padding: 0 5px;
     }
 
     .editor-tools {
--- a/kitsune/sumo/static/sumo/less/questions.less
+++ b/kitsune/sumo/static/sumo/less/questions.less
@@ -409,8 +409,8 @@ h2 {
 
     textarea {
       margin-top: 10px;
-      width: 500px;
       padding: 0 5px;
+      width: 500px;
     }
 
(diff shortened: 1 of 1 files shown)


kitsune/sumo/static/sumo/less/questions.less
```diff
@@ -409,7 +409,8 @@ h2 {
 
     textarea {
       margin-top: 10px;
-      width: 510px;
+      width: 500px;
```
rehandalal: nit: alphabetize properties

rehandalal:
lgtm, just needs that one little nit addressed

kitsune/sumo/static/sumo/less/questions.less
mythmon: We really should get a linter for this. Well, that and fix the linters.

rehandalal: +1

willkg: We should write a linter! It'd be easy. It'd take us like an afternoon to do.

brittanystoroz:
Eeeeeek FIXED. Linter would be a good idea, I'm probably going to forget this a lot for a long time. :+1: @rehandalal @mythmon

rehandalal:
r+

PR #2734, opened 2015-11-20
[Bug 1064335] Added space between left and right columns in questions.

anushbmx:
This adds space between the text 
"xx questions in the last 24 hours have no reply. Help solve them!"  which is little longer in other language like french that it goes to next line that is fine but the problem is text "Total questions: xx" ( Right side ) which is floated to right and has no padding to its left side it makes appear like a single line if the text on right is long enough.

Here is how it looks in English:  https://support.mozilla.org/en-US/questions/thunderbird

xx questions in the last 24 hours have no reply. Help solve them!  ^^^^^^^^^  Total questions: xx
( Have enough space )
                                                                                                       
In French  https://support.mozilla.org/fr/questions/thunderbird
screenshot from 2015-11-20 12 40 59
Without space between the two words

screenshot from 2015-11-20 12 40 16
With space.

### DIFF
--- a/kitsune/sumo/static/sumo/less/questions.less
+++ b/kitsune/sumo/static/sumo/less/questions.less
@@ -1117,6 +1117,7 @@ h2 {
   .total {
     float: right;
     margin: 0 6px;
+    padding-left: 50px;
   }
 }
 


safwanrahman:
Looks good! This should get r+
LGTM!

anushbmx:
@safwanrahman haha it is just a simple fix, I flagged you to use this bug for the sprint you were organizing but It was kept idle :) so thought to close it.

safwanrahman:
@anushbmx Yap. I know. Actually, I dont have much Idea about CSS or front end! :disappointed:. So I could not mentor in the sprint about the front end.

rehandalal:
cool, thanks!

PR #2737, opened 2015-11-27
[Bug 1225507] Disable signout during AAQ

brittanystoroz:
This hides the signout button which wasn't working during the AAQ flow (https://bugzilla.mozilla.org/show_bug.cgi?id=1225507) 

I also renamed the `hide_aaq_link` variable to something a bit more generic, as it's responsible for a little more now.

### DIFF
--- a/kitsune/questions/templates/questions/new_question_react.html
+++ b/kitsune/questions/templates/questions/new_question_react.html
@@ -6,6 +6,8 @@
 {% set title = _('Ask a Question') %}
 {% block breadcrumbs %}{% endblock %}
 
+{% set hide_signout_link = True %}
+
 {% block content %}
   {{ csrf() }}{# CSRF token for AJAX actions. #}
   <script type="application/json" class="data" name="products">{{ products_json|safe }}</script>
--- a/kitsune/questions/templates/questions/new_question_react.html
+++ b/kitsune/questions/templates/questions/new_question_react.html
@@ -6,6 +6,7 @@
 {% set title = _('Ask a Question') %}
 {% block breadcrumbs %}{% endblock %}
 
+{% set hide_aaq_link = True %}
 {% set hide_signout_link = True %}
 
 {% block content %}
--- a/kitsune/sumo/templates/includes/common_macros.html
+++ b/kitsune/sumo/templates/includes/common_macros.html
@@ -11,9 +11,9 @@
   </form>
 {% endmacro %}
 
-{% macro aux_nav(user, hide_aaq_link=False) %}
+{% macro aux_nav(user, within_aaq_flow=False) %}
   {% if not settings.READ_ONLY %}
-    {% if not hide_aaq_link %}
+    {% if not within_aaq_flow %}
       {% if request.LANGUAGE_CODE in AAQ_LANGUAGES %}
         {% set ask_url = url('questions.aaq_step1') %}
       {% else %}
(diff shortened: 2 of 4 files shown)


kitsune/sumo/templates/includes/common_macros.html
```diff
@@ -34,12 +34,14 @@
           <li><a href="{{ url('search')|urlparams(asked_by=user.username, a=1, w=2) }}">{{ _('My Questions') }}</a></li>
           <li><a href="{{ profile_url(user, edit=True) }}">{{ _('Edit Profile') }}</a></li>
           <li><a href="{{ url('users.edit_settings') }}">{{ _('Settings') }}</a></li>
+          {% if not within_aaq_flow %}
           <li>
```
rehandalal: we would usually indent all of this stuff in the `if` block one more level

rehandalal:
This works, but I'm not sure it addresses the problem in the bug.

We want to hide the sign in link from the new react-based AAQ not the current AAQ. I believe the template you want to edit is `questions/new_question_react.html`. 

And since we don't want to hide the link for all instances of being within the AAQ flow I would probably change that variable black and modify the macro to accept a second argument which specifies whether to hide the sign in link or not.

brittanystoroz:
Ahh, yea I see that now. Sounds good I'll update this today thanks @rehandalal

brittanystoroz:
Updated to target the new AAQ flow rather than old. Thoughts @rehandalal ?

One thing I noticed is that it looks like the pre-existing `hide_aaq_link` logic effects the old flow rather than the new. (i.e. the 'Ask A Question' link still appears throughout the new AAQ flow) Is this intended or an oversight? I double-checked on master to make sure it was not this PR that introduced that behavior.

kitsune/questions/templates/questions/new_question_react.html
```diff
@@ -6,6 +6,8 @@
 {% set title = _('Ask a Question') %}
 {% block breadcrumbs %}{% endblock %}
 
+{% set hide_signout_link = True %}
```
rehandalal: you're right about the "Ask A Question" link on this page. we should probably also add:
`{% set hide_aaq_link = True %}`

brittanystoroz:
@rehandalal Added.

safwanrahman:
As far as I see, [Bug 1225507](bugzil.la/1225507) covers correcting the sign out link. The **sign out** link is not working perfectly.
In [comment 1](https://bugzilla.mozilla.org/show_bug.cgi?id=1225507#c1), its mentioned that the **sign in** link should be hidden. Which seems to be [bug 1225511](bugzil.la/1225511).

This patch hides the **Sign out** link from the AAQ. Which is not a good UX I think. Hiding the **Sign in** is a better choice.

brittanystoroz:
@safwanrahman So in that case, I don't think we can do just one or the other. (Hide the sign in link OR the sign out link.) Hiding the sign in link won't fix the fact that once they are signed in, the sign out link still doesn't work.

I think we should:

1) definitely remove the sign in link if the user is already authenticated and 
2) either fix whatever is wrong with the sign out link, or remove it (in the last check-in meeting we had decided on just removing it because it was unlikely someone would bother to sign out mid-AAQ. Obviously it would be better to fix the issue than just hide it, but if the bug is more trouble than it's worth, I think hiding it is a fine option and better than leaving it on there broken.)

rehandalal:
@safwanrahman the bug is really the place to have this discussion not this pull request. however this pull does fix the bug in question as per the discussion in that bug.

rehandalal:
test failures looked unrelated. i restarted the tests.

rehandalal:
travis is happy.

rehan is happy.

r+

safwanrahman:
@rehandalal Sorry my bad. 
But, **sign out** link should not be be hidden as per bug. In [comment 1](https://bugzilla.mozilla.org/show_bug.cgi?id=1225507#c1), its mentioned that the **Sign In** link should be hidden. not the **Sign Out** link

PR #2738, opened 2015-11-29
[Bug 1189361] Starting search from one of the products sections, can't switch to 'All Products' at search page

rsajdok:


mythmon:
Hi @rsajdok, thanks for this PR. It's a little unclear what is happening here. Can you update the commit message to explain what the change was, instead of explaining the problem? If you look at the other commits in the git history, we try to make it clear what the commit contains, and the reason for the change, instead of describing the state of the code before the commit.

Also, can you change the commit message to start with "[Bug 1189361] Starting..." instead of "Bug 1189361 - Starting..."? We have some tooling that looks for the former.

rsajdok:
@mythmon 
"The possibility of changing to "All products" filter at any time."
Will it be ok?

mythmon:
@rsajdok I'd rather an explanation of what the actual code change was, and why it fixed the problem. You are describing very high level user visible version of the effect of this change. My problem with this is that I don't understand your code change. Why is this a useful change? How do we know it is correct?

mythmon:
This PR has gone almost a month without any activity, so I'm going to close it to keep our PR queue tidy. If you would like to fix the issues I've outlined above, please re-open this pull request with the changes.

PR #2739, opened 2015-11-30
[Bug 1226771] Update Django to 1.7.11

safwanrahman:
@mythmon r?

### DIFF
--- a/requirements/default.txt
+++ b/requirements/default.txt
@@ -48,8 +48,8 @@ cryptography==0.7.2
 # sha256: _eJnrrA6jUnjNB5_caIoxnSNzkO1wA9lWBeA2Ztwstk
 dennis==0.7
 
-# sha256: uTV9LOvmGZcFXUF9YH-cZQ6BfNGjg7mhuIvx7a15fHU
-Django==1.7.10
+# sha256: IDkUT86PG2A9A_paVkNXjfGtAHxO1BphfwKjlD9wWaE
+Django==1.7.11
 
 # sha256: M7YuOXN4svxB-YixuXUuT1Y5rfNYJN5wxGqKDmv1pVs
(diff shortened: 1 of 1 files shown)


requirements/default.txt
```diff
@@ -48,8 +48,9 @@ cryptography==0.7.2
 # sha256: _eJnrrA6jUnjNB5_caIoxnSNzkO1wA9lWBeA2Ztwstk
 dennis==0.7
 
-# sha256: uTV9LOvmGZcFXUF9YH-cZQ6BfNGjg7mhuIvx7a15fHU
-Django==1.7.10
+# sha256: EAFkVWiXwSGfM3BuY6ZWuISNM9CbAWHi3u_MUJeM9i0
+# sha256: IDkUT86PG2A9A_paVkNXjfGtAHxO1BphfwKjlD9wWaE
```
mythmon: We usually only need one hash here. What is the other one?

safwanrahman:
I dont know. The hash maybe depends on platform. I have got 2 hashes for Django

requirements/default.txt
safwanrahman: I dont know. The hash maybe depends on platform. I have got 2 hashes for Django

mythmon: Usually when this happens is because one of the hashes is a wheel. Django shouldn't be platform dependent in the tarball format that we should be using.

safwanrahman: So having 2 hashes are Ok or need to have one only?

mythmon: It needs to be one only.

safwanrahman:
Updated!

mythmon:
I verified the hash and version number. This looks good, thank you!

PR #2740, opened 2015-11-30
[Bug 743360] Added function to retrive the last searched term

anushbmx:
This is fill the search field with previous searched team if exist instead of the placeholder.

### DIFF
--- a/kitsune/search/views.py
+++ b/kitsune/search/views.py
@@ -266,6 +266,10 @@ def advanced_search(request, template=None):
     # 2. Build form.
     search_form = AdvancedSearchForm(r, auto_id=False)
     search_form.set_allowed_forums(request.user)
+    # get value for search input from last search term.
+    last_search = request.COOKIES.get(settings.LAST_SEARCH_COOKIE)
+    if last_search:
+    	r['q'] = urlquote(last_search)
 
     # 3. Validate request.
     # Note: a == 2 means "show the form"--that's all we use it for now.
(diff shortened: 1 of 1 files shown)


kitsune/search/views.py
```diff
@@ -266,6 +266,10 @@ def advanced_search(request, template=None):
     # 2. Build form.
     search_form = AdvancedSearchForm(r, auto_id=False)
     search_form.set_allowed_forums(request.user)
+    # get value for search input from last search term.
+    last_search = request.COOKIES.get(settings.LAST_SEARCH_COOKIE)
+    if last_search:
+    	r['q'] = urlquote(last_search)
```
mythmon: Shouldn't you only set `r['q']` in the case that it is empty, to avoid overriding other searches?

anushbmx:
@mythmon I have added a check for that too.

kitsune/search/views.py
```diff
@@ -266,6 +266,10 @@ def advanced_search(request, template=None):
     # 2. Build form.
     search_form = AdvancedSearchForm(r, auto_id=False)
     search_form.set_allowed_forums(request.user)
+    # get value for search input from last search term.
+    last_search = request.COOKIES.get(settings.LAST_SEARCH_COOKIE)
+    if last_search and not 'q' in r:
```
mythmon: Instead of `not 'q' in r` this is usually written as `'q' not in r`.

anushbmx:
@mythmon I have updated it.

PR #2741, opened 2015-12-01
[Bug 1224267] Redirect warnings to logging infra.

mythmon:
This should do two things

* Make collect_tweets stop spamming my inbox
* In the future we can more easily deal with these kind of warnings via logging instead of ad-hoc dumping of them to stdout.

r?

### DIFF
--- a/kitsune/log_settings.py
+++ b/kitsune/log_settings.py
@@ -84,3 +84,4 @@ else:
     )
 
 dictConfig(config)
+logging.captureWarnings(True)
--- a/scripts/crontab/gen-crons.py
+++ b/scripts/crontab/gen-crons.py
@@ -14,7 +14,7 @@ def main():
                       help="Location of kitsune (required)")
     parser.add_option("-u", "--user",
                       help=("Prefix cron with this user. "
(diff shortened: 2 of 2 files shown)


kitsune/log_settings.py
```diff
@@ -84,3 +84,4 @@
     )
 
 dictConfig(config)
+logging.captureWarnings(True)
```
willkg: What effect does this happen in celery and wsgi and `./manage.py runserver`? Seems like it'd be fine, but I'm curious if you tested it out or not and if the other log settings send the warnings anywhere useful.

willkg:
I had a question but this looks fine to me. r+

kitsune/log_settings.py
mythmon: Anything with `DEBUG=True` gets most logging sent to stdout, including warnings. This should cover the runserver case. I don't have a great way to generate warnings in local dev right now, but I assume this would look slightly different but still have the same info.

Anything with `DEBUG=False` gets most logging sent to syslog, where webops can slice and dice it as they see fit, and maybe someday we'll be able to look at it. This should cover the celery and wsgi cases.

PR #2742, opened 2015-12-05
[Bug 1173526] Replace all model makers with factoryboy classes.

mythmon:
Wooh boy. This is a big-un. +2125/-2598 lines. Touched 146 files. I edited almost every single test. GitHub won't show me the full diff.

A lot of this was trivial. A lot of this I did with mass find/replace. Some of those find/replaces might be a little uglier than they need to be.

Some of this is not trivial. Some of the factories have subtly different semantics than the model makers. Most of the tests are logically the same, with a different spelling. A couple tests I changed the logic of, but I think they are still testing the same thing.

This only affects the test infrastructure (in theory). If Travis passes, we know that didn't break. Besides that, check out the tests, make sure I didn't make any of them invalid. Check out the new factories, see if they are terrible or awesome.

I probably owe a drink to whoever reviews this. Luckily we'll all be together next week.

r?

PS: I have a version of this git history on my laptop that is module-by-module. It is messier, but if you'd prefer to review that, I can put it up.

### DIFF
--- a/kitsune/access/tests/__init__.py
+++ b/kitsune/access/tests/__init__.py
@@ -1,20 +1,21 @@
 from django.contrib.contenttypes.models import ContentType
 from django.test.client import RequestFactory
 
+import factory
 from authority.models import Permission
 from nose.tools import eq_
 
 from kitsune import access
-from kitsune.sumo.tests import TestCase, with_save
+from kitsune.sumo.tests import TestCase
 from kitsune.sumo.urlresolvers import reverse
-from kitsune.users.tests import user
+from kitsune.users.tests import UserFactory
 
 
-@with_save
-def permission(**kwargs):
-    if 'approved' not in kwargs:
-        kwargs['approved'] = True
-    return Permission(**kwargs)
+class PermissionFactory(factory.DjangoModelFactory):
+    class Meta:
+        model = Permission
+
+    approved = True
 
 
 class AccessTests(TestCase):
@@ -25,11 +26,11 @@ class AccessTests(TestCase):
 
     def test_admin_perm_thread(self):
         """Super user can do anything on any forum."""
-        from kitsune.forums.tests import restricted_forum
-        f1 = restricted_forum()
-        f2 = restricted_forum()
+        from kitsune.forums.tests import RestrictedForumFactory
+        f1 = RestrictedForumFactory()
+        f2 = RestrictedForumFactory()
 
-        admin = user(is_staff=True, is_superuser=True, save=True)
+        admin = UserFactory(is_staff=True, is_superuser=True)
 
         # Loop over all forums perms and both forums
         perms = ('thread_edit_forum', 'thread_delete_forum', 'post_edit_forum',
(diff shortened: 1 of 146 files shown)


willkg:
So... I'm game for reviewing this, but when I click on the "Files changed" tab, Github says "I suck. Too big. Cannot show." or something like that.

Couple of possible fixes off the top of my head:

1. break it up into a couple of PRs
2. use a different review tool
3. land it because review rhymes with shoe and when the shoe fits, wear it

I think these all kind of suck. Can you think of better options? If not, maybe we should break it up.

mythmon:
2 and 3 both sound lame. Lets not do that.

I see another option: review it in person next week. That's probably not a great idea, since it is really probably really time consuming and leaves no paper trail.

I think I'll work on making this into two PRs. It won't be easy, but I think it is possible.

willkg:
Mmm... So, it'd be great if this didn't take us ages to work through. I want to get the ES stuff done this quarter plus we have all the work for Django 1.8. So the faster we get through this, the better. I think moving faster on this is more important than a comprehensive paper trail. This is test related so I claim it's "ok" to be a little loose with review. I do recognize the ramifications of having bugs in the tests, but I also claim there are probably already bugs with the tests and maybe this is a zero sum change in that regard. Anyhow, I'd urge speed over comprehensiveness.

Given that, I could probably do this in person. Maybe a better approach is that I pull the PR down to my computer and look at the diffs locally and then write up a commit fixing anything **bad** and pushing that as a PR against your branch. Would that work for you? Seems like it'd get us further faster.

mythmon:
That all sounds reasonable, and a lot easier (at least for me) than redoing this PR. Lets go with that. Speed sounds especially good when the sheer number of lines is taken into account. The chance for merge conflicts if this sits much longer is really high.

willkg:
Awesome. I'll start working on it in the morning. WE WILL CRUSH THIS PR!

willkg:
The diff is too big for github to handle, so I pulled the branch to my local system and read through the diff. My tweaks are in this commit:

https://github.com/willkg/kitsune/commit/e23a9280a47693b0b9c14ad3067f1af57e90b8b5

Generally, I:

1. skimmed most of the changes in the tests since they seem to be "change this to that" kinds of things
2. read through the new factory classes more carefully
3. read through non-test code more carefully

It's a lot of changes to a huge swatch of codebase, so it's hard/impossible to read through the changes and notice subtle issues. I did read through all the changes and they do pass. I think that's the best we can do without spending gobs more time and essentially evaluating every test on a test-by-test basis which seems like a poor use of time.

I had a couple of changes in my commit that should happen and a bunch of comments about converting the code to do things in a batch. Other than that, r+.

mythmon:
I cherry picked willkg@e23a928 into this branch, and made another small tweak to the comments.

I want to go through the batch optimizations at another point in the future. I think this PR is good to merge, as soon as Travis runs on the latest commit.

:fireworks: :fireworks: :fireworks:

willkg:
Woo hoo!!!

PR #2743, opened 2015-12-08
1217076 signal

willkg:


mythmon:
I think we all like this. Travis failures were spurious. Merging.

PR #2744, opened 2015-12-09
[Bug 1217084] Fix Focus and IOS sprites on mobile.

mythmon:
The Focus changes were based on the iOS changes, which also missed the mobile icons. This should fix it.

r?


mythmon:
I'm going to self r+ this since Travis likes it and we are going to test it on stage.

PR #2745, opened 2015-12-10
[Bug 1154833] Add searchresult timestamps

brittanystoroz:
This adds a last_updated timestamp to question search results as per [bug 1154833](https://bugzilla.mozilla.org/show_bug.cgi?id=1154833)

Had to make some styling decisions to avoid awkward line-wraps with the additional data.

timestamps-styling

### DIFF
--- a/kitsune/sumo/static/sumo/less/includes/search-result.less
+++ b/kitsune/sumo/static/sumo/less/includes/search-result.less
@@ -46,9 +46,23 @@
     }
   }
 
+  .question-status {
+    font-size: 13px;
+    margin: 12px 0 0 0;
+
+    .solved {
+      background: transparent @iconsSprite no-repeat scroll -49px -862px;
+      color: rgb(129, 188, 46);
+      display: inline-block;
+      font-family: "Open Sans Light",Arial,Helvetica,sans-serif;
+      line-height: 18px;
+      text-indent: 22px;
+    }
+  }
+
   .thread-meta {
     list-style: none;
-    margin: 12px 0 0 0;
+    margin: 3px 0 0 0;
     padding: 0;
 
     li {
--- a/kitsune/sumo/static/sumo/less/includes/search-result.less
+++ b/kitsune/sumo/static/sumo/less/includes/search-result.less
@@ -54,6 +54,7 @@
     li {
       border-left: 1px solid @borderLightGrey;
       float: left;
+      font-size: 13px;
       margin-left: 10px;
       padding-left: 10px;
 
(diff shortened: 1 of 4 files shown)


kitsune/sumo/static/sumo/less/includes/search-result.less
```diff
@@ -46,14 +46,29 @@
...
 
+  .question-status {
+    font-size: 13px;
+    margin: 12px 0 0 0;
+
+    .solved {
+      background: transparent @iconsSprite no-repeat scroll -49px -862px;
+      color: rgb(129, 188, 46);
+      display: inline-block;
+      font-family: "Open Sans Light",Arial,Helvetica,sans-serif;
```
rehandalal: nit: spaces after `,`

brittanystoroz: I literally copy and pasted that from elsewhere in the CSS!! Haha but I will fix. And whatever other instance I copied from. At least they are all alphabetized this time ;)

brittanystoroz: jk maybe I copied it from devtools :) <3

brittanystoroz:
Travis no likey. Something with python timestamps in the AdvancedSearchResults tests:

ERROR: kitsune/search/tests/test_search_advanced.py:AdvancedSearchTests
TypeError: datetime.datetime(2015, 12, 9, 21, 9, 11) is not JSON serializable

I had to convert the `updated` timestamp in [search/views.py#L97](https://github.com/brittanystoroz/kitsune/blob/add-searchresult-timestamps/kitsune/search/views.py#L97) because otherwise the `timesince` helper in the [result.html](https://github.com/brittanystoroz/kitsune/blob/add-searchresult-timestamps/kitsune/search/templates/search/includes/result.html#L11) template complains about not having a `tzinfo` attribute. Am I doing this in the right place/the right way? 

I thought dates where only hard in javascript =P

brittanystoroz:
@mythmon travis seems happy, timestamps seem happy. any thoughts on that last commit d40def007c1b8bbe4cf9588b526f88615b55435b?

safwanrahman:
@brittanystoroz I think there should be some test for the API Changes you have made. 
Also, it will be great if there is a a test for the feature you have implemented. So we can avoid any kind of regression in future

brittanystoroz:
@safwanrahman So there's not actually an API change here...I'm simply adding the `last_updated` field so I can surface it in the template. Checking whether or not something renders in a template doesn't seem to be something heavily tested in this repo, but do point me in the right direction if I'm wrong here.

mythmon:
The new styles look good to me. I don't have a strong opinion either way about tests here.

rehandalal:
r+

PR #2746, opened 2015-12-10
Revert "Merge pull request #2740 from anushbmx/master"

mythmon:
#2740 tried to improve advanced search by remembering query parameters in the text box by reading a cookie. Unfortunatly, this search interface was used by other parts of the site, and remember the last search query broke those parts. This PR reverts That PR.

This reverts commit 5e14c43318bc6b07072c80ebb49dd21f09cd4e4f, reversing
changes made to 0f0d7dabceb811619c7f996304aff264fe294b38.

r?

PR #2747, opened 2015-12-10
Rewriting and fixing the Patch #2740 of Bug 743360

safwanrahman:
#2740 Added feature for remembering the input of advance search. But, as the view is used for many other purposes, it raises issue from the user end.
#2746 is filled for reverting the changes that @anushbmx made. **But, in the meantime I have rewrite his patch and fix the issue.**
So I think the revert of #2740 will not be necessary if my patch is correct and being landed.

@mythmon @rehandalal r?

### DIFF
--- a/kitsune/search/views.py
+++ b/kitsune/search/views.py
@@ -266,10 +266,6 @@ def advanced_search(request, template=None):
     # 2. Build form.
     search_form = AdvancedSearchForm(r, auto_id=False)
     search_form.set_allowed_forums(request.user)
-    # get value for search input from last search term.
-    last_search = request.COOKIES.get(settings.LAST_SEARCH_COOKIE)
-    if last_search and 'q' not in r:
-    	r['q'] = urlquote(last_search)
 
     # 3. Validate request.
     # Note: a == 2 means "show the form"--that's all we use it for now.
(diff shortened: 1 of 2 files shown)


anushbmx:
That was a bad patch from me :-1:  , forgot to look other issue. thanks @safwanrahman for fix the issues.

kitsune/search/views.py
```diff
@@ -281,11 +277,18 @@ def advanced_search(request, template=None):
...
+        data = {'advanced': True,
                 'request': request,
-                'search_form': search_form}),
+                'search_form': search_form}
+        # get value for search input from last search term.
+        last_search = request.COOKIES.get(settings.LAST_SEARCH_COOKIE)
+        # If there is any cached input from last search, pass it to template
+        if last_search and 'q' not in r:
+            cached_field = urlquote(last_search)
+            data.update({'cached_field': cached_field})
```
mythmon: Nit: I'd probably write this as `data['cached_field'] = urlquote(last_search)`.

mythmon:
This looks good, thanks for the fix!

kitsune/search/views.py
safwanrahman: Oh! I just did not think about this! 
Out of my mind! sorry!

PR #2748, opened 2015-12-15
[Bug 1225506] (WIP) Show product & topic in AAQ urls

brittanystoroz:
Stopping here for some review before this PR gets too unruly...

This adds the product & topic values to the path of the URL. If the question store doesn't have this data readily available (i.e. if you use the back button or refresh the page), we can now check for this information in the URL.

There's some cleanup/refactoring I can still do here, but more immediately I'm curious if we think `step` as a query parameter is still necessary? Does the new URL structure with a more specific path give us all the information we need to imply the step? (i.e. `/questions/new/product` implies `step=topic`, `questions/new/product/topic implies `step=title`) My first impression is that it should be pretty safe to remove but perhaps someone else knows better off the top of their head ¯\_(ツ)_/¯

### DIFF
--- a/kitsune/questions/static/questions/js/components/TitleContentEditor.jsx
+++ b/kitsune/questions/static/questions/js/components/TitleContentEditor.jsx
@@ -19,14 +19,13 @@ export default class TitleContentEditor extends AAQStep {
   }
 
   setPropsFromUrl() {
-      UrlActions.getPropsFromPath("/questions/new/(.*)/(.*)", ['product', 'topic']);
-      let urlData = UrlStore.get('pathProps');
-      if (urlData.product) {
-        AAQActions.setProduct(urlData.product);
-      }
-      if (urlData.topic) {
-        AAQActions.setTopic(urlData.topic);
-      }
+    let urlData = UrlStore.get('pathProps');
+    if (urlData.product) {
+      AAQActions.setProduct(urlData.product);
+    }
+    if (urlData.topic) {
+      AAQActions.setTopic(urlData.topic);
+    }
   }
 
   componentWillMount() {
--- a/kitsune/questions/static/questions/js/components/TitleContentEditor.jsx
+++ b/kitsune/questions/static/questions/js/components/TitleContentEditor.jsx
@@ -2,6 +2,7 @@
 import AAQStep from './AAQStep.jsx';
 import AAQActions from '../actions/AAQActions.es6.js';
 import aaqGa from '../utils/aaqGa.es6.js';
+import UrlStore from '../../../sumo/js/stores/UrlStore.es6.js';
 
 export default class TitleContentEditor extends AAQStep {
   handleChange(ev) {
@@ -18,6 +19,11 @@ export default class TitleContentEditor extends AAQStep {
 
   componentDidMount() {
     AAQActions.checkTroubleshootingAvailable();
+    if (this.shouldExpand() === false) {
+      let urlData = UrlStore.get('pathData');
+      AAQActions.setProduct(urlData.product);
+      AAQActions.setTopic(urlData.topic);
+    }
   }
 
   shouldExpand() {
--- a/kitsune/questions/static/questions/js/components/TitleContentEditor.jsx
+++ b/kitsune/questions/static/questions/js/components/TitleContentEditor.jsx
@@ -30,8 +30,7 @@ export default class TitleContentEditor extends AAQStep {
   }
 
   componentWillMount() {
-    if (!this.props.question.topic &&
-            !this.props.question.product) {
+    if (!this.props.question.topic && !this.props.question.product) {
       this.setPropsFromUrl();
     }
   }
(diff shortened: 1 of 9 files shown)


mythmon:
The step parameter was the easiest way at the time that I could see to make the forward and back buttons work within the AAQ. If that still works, I think you're right: we can derive the step from all the other information, and we can probably get rid of it. Otherwise we might get in weird states like having step=topic, but not having a product set.

I'll do an actual review here tomorrow.

kitsune/questions/static/questions/js/components/TitleContentEditor.jsx
```diff
@@ -16,6 +18,24 @@ export default class TitleContentEditor extends AAQStep {
...
+        AAQActions.setProduct(urlData.product);
+      }
+      if (urlData.topic) {
+        AAQActions.setTopic(urlData.topic);
+      }
+  }
+
+  componentWillMount() {
+    if (!this.props.question.topic &&
+            !this.props.question.product) {
```
mythmon: Nit: In Python we have a 99 character line limit. In JS we don't have an explicit line length limit. Either way, I think this could be wrapped to one line, if you think that reads better.

kitsune/sumo/static/sumo/js/constants/UrlConstants.es6.js
```diff
@@ -3,6 +3,7 @@ import constantMap from '../../../sumo/js/utils/constantMap.es6.js';
 export const actionTypes = constantMap([
   'UPDATE_QUERY_STRING',
   'UPDATE_QUERY_STRING_DEFAULTS',
+  'UPDATE_URL_PATH',
```
mythmon: I think you also need to add `GET_PROPS_FROM_PATH` here.

kitsune/sumo/static/sumo/js/stores/UrlStore.es6.js
```diff
@@ -22,17 +28,37 @@ class _UrlStore extends BaseStore {
 const UrlStore = new _UrlStore();
 
 UrlStore.dispatchToken = Dispatcher.register((action) => {
-  let params, qs;
+  let params, pathProps, qs, urlPath;
   switch (action.type) {
+    case actionTypes.GET_PROPS_FROM_PATH:
+      pathProps = createDictFromPath(action.pathRegex, action.propNames);
+      urlData.pathProps = pathProps;
```
mythmon: Nit: This can probably just be `urlData.pathProps = createDictFromPath(action.pathRegex, action.propNames);`, unless there is a good reason to have it split like this.

brittanystoroz:
@mythmon thanks for the review here, fixed the nits, I want to try to pull out the `step` query param as we discussed but is that something you'd want in a different PR or would it be okay to tack onto this one?

mythmon:
Adding it on to this PR seems fine, if you'd like to do that. Right now the code is in a weird middle state that does both ways.

brittanystoroz:
@mythmon this has been updated to take the `step` out of the query param, and determine everything based on the URL path instead. Kind of a beast of a pull request, lmk if you have trouble following

kitsune/sumo/static/sumo/js/main.js
```diff
@@ -5,6 +5,29 @@ window.k = window.k || {};
 (function () {
   k.LAZY_DELAY = 500;  // delay to lazy loading scripts, in ms
   k.STATIC_URL = $('body').data('static-url');
+  k.getPathAsDict = function(propertyNames) {
+    var pathDict = {},
+      pathArray = window.location.pathname.split('/').filter(Boolean);
+
+    propertyNames.map(function(value, index) {
+      pathDict[value] = pathArray[index];
```
mythmon: `.map` is for transforming an array into a new array. Since this is just looping over the array it would be better to use `propertyNames.forEach()`.

kitsune/sumo/static/sumo/js/main.js
```diff
@@ -5,6 +5,29 @@ window.k = window.k || {};
...
+    });
+
+    return pathDict;
+  };
+
+  k.pathStringFromDict = function(pathDict) {
```
mythmon: Since this and `getPathAsDict` are only used by new ES6 code, how do you feel about defining them directly in `kitsune/sumo/static/sumo/js/utils/urls.es6.js` instead of here and then letting `urls.es6.js` get them in the shim?

brittanystoroz:
@mythmon put in 2 extra commits.  One switches out `.map()` with `.forEach()`. Not sure what my thought process was there at the time, but good catch. Also agreed with moving the new URL helper functions to `urls.es6.js` and cleaned that up.

Travis is complaining about a 502 Gateway error...something about django-badger? I'm assuming this is not my fault/out of my control?

mythmon:
I restarted the build that failed in Travis.

mythmon:
Sorry this took so long to get to. I had some issues that I solved in #2762.

This works well in the browser, and the code looks fine. The more I read this code the more I regret not using a router package instead of trying to do it all like this, but that's in the past. I think it is best to land this, and possibly improve it in the future.

r+

PR #2749, opened 2015-12-15
[bug 1232637] Fix extract_db re: empty strings

willkg:
Empty string msgids are super bad. This reduces the likelihood we have
those.

r?


willkg:
There are no tests for `extract_db.py`. That's probably something we might want to fix at some point. I did some hand-testing and I think this is correct.

The idea is that if `item[i]` is empty (can't strip it--it has to literally be an empty string), then we don't create a string and put it in the `db_strings.py` file.

mythmon:
+1 to having more tests.

I eyeballed the code, and it looks fine. r+

PR #2750, opened 2015-12-15
Fix pre commit

mythmon:
Like I expected, the Pre-commit problems went away after we waited a bit and then upgraded everything. This

1. Upgrades pre-commit and it's dependencies
2. Fixes a bunch of lint errors that eslint found. Not sure how these snuck in. Almost all of them were fixed with `eslint ${files} --fix`.
3. Re-enable linting in Travis.

If Travis likes this, there should be no problem, since Travis was our primary problem.

r?

### DIFF
--- a/kitsune/sumo/static/sumo/js/kbox.js
+++ b/kitsune/sumo/static/sumo/js/kbox.js
@@ -79,7 +79,7 @@
   '<div class="kbox-title"></div>' +
   '<div class="kbox-wrap"><div class="kbox-placeholder"/></div>' +
   '</div>',
-  OVERLAY =
+    OVERLAY =
   '<div id="kbox-overlay"></div>';
 
   // The KBox type
@@ -220,7 +220,7 @@
     },
     setPosition: function(position) {
       var self = this,
-      toX, toY, $parent, parentOffset, minX, minY, scrollL, scrollT;
+        toX, toY, $parent, parentOffset, minX, minY, scrollL, scrollT;
       if (!position) {
         position = self.options.position;
       }
--- a/kitsune/sumo/static/sumo/js/kbox.js
+++ b/kitsune/sumo/static/sumo/js/kbox.js
@@ -73,14 +73,14 @@
 
   'use strict'; // Giving this a shot!
 
-  var TEMPLATE =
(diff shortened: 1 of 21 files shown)


kitsune/sumo/static/sumo/js/kbox.js
```diff
@@ -79,7 +79,7 @@
   '<div class="kbox-title"></div>' +
   '<div class="kbox-wrap"><div class="kbox-placeholder"/></div>' +
   '</div>',
-  OVERLAY =
+    OVERLAY =
```
willkg: Why did this need to get indented?

willkg:
It failed with this:

```
FAIL: kitsune/kpi/tests/test_api.py:KpiApiTests.test_vote
  File "/home/travis/build/mozilla/kitsune/kitsune/kpi/tests/test_api.py", line 180, in test_vote
    eq_(r['objects'][0]['ans_helpful'], 2)
  File "/home/travis/virtualenv/python2.7.9/lib/python2.7/site-packages/nose/tools.py", line 31, in eq_
AssertionError: 3 != 2
```

That test fails for me locally periodically, but not all the time. So I restarted Travis.

mythmon:
That error is fixed in #2751.

willkg:
YAY! Though I thought it had been flakey for a while. There are a couple of others that fail on me periodically, but not all the time, too. Something to do with topics. Next time it happens, I'll be a responsible adult and write up a bug report.

kitsune/sumo/static/sumo/js/kbox.js
mythmon: It is part of a multi-variable `var` statement (which I don't like, but whatever). The linter says that variables in a multi-variable `var` statement that are on not-the-first-line need indented one step. If I had fixed this by hand (instead of with `eslint --fix`, I would have also indented the markup above this. I'll do that now.

willkg:
Travis failed on the same test. So ... maybe land #2751 first, then freebase this, then we can see where things are at.

Or just land it. I'm r+ either way.

mythmon:
I'm just going to merge it. Mainly because I don't want to bust out the cli to get this PR closed. `#lazy`.

PR #2751, opened 2015-12-16
Fix flaky test in KPI.

mythmon:
I noticed this test was failing randomly. It was because of one of the fuzzy model makers. Stating more assumptions in the test fixed it.

r?


willkg:
Yay!

willkg:
Passed! r+!

PR #2752, opened 2015-12-16
[bug 1232637] Add check for production db

willkg:
I can't figure out a way to distinguish an anonymized production db
dump from a db dump generated by sample data without doing hand-wavey
"are there at least x users?" or other things which are fragile. So
instead, I added a reminder prompt.

r?

### DIFF
--- a/kitsune/sumo/management/commands/extract_db.py
+++ b/kitsune/sumo/management/commands/extract_db.py
@@ -66,6 +66,12 @@ class Command(BaseCommand):
         except AttributeError:
             raise CommandError('DB_LOCALIZE setting is not defined!')
 
+        areyousure = raw_input(
+            'Are you sure you have a recent production db? y/n: '
+        )
+        if areyousure.lower() != 'y':
+            raise CommandError('Not sure if it is a recent production db.')
+
         strings = []
         for app, models in apps.items():
             for model, params in models.items():


willkg:
I'm not a huge fan of "are you really sure you're a responsible adult?" type questions, but I figure this is an easy change to make and might be good enough.

Another thing we could do is add some kind of check like "The last question asked in this db was on YYYY-MM-DD HH:MM and there are x questions in the db." And then ask the question, "Are you sure this is a recent db?"

That gives the user some information to gauge the answer and/or know when it's probably horribly wrong. Would that be better?

mythmon:
To be honest, if this got merged I'd probably just add `yes y | ./manage.py extract_db` to my script because I'm terrible.

Maybe a better approach would be to look for things that exist in the auto generated database. For example, the sample data leaves the descriptions to all topics blank. That was the actual problem. Could we key off of details like that?

willkg:
I don't think we can. Because we can't know if someone ran generatedata after getting a production db which I do to "catch up" so i don't have to update it. (At least, I did this with fjord. The kitsune generatedata isn't great.)

willkg:
Also, I think if you're just going to do a "yes | ..." we should just drop this and you can deal with issues.

willkg:
To clarify, we specifically need a "fresh db dump". So a db with generatedata generated data isn't great. A db based on a dump from a few days ago, but with some additional data that came from testing or feature implementation isn't great, either. I don't think we can distinguish between those cases to do anything automatically.

Having said all that, I suspect this issue really only affects you. Maybe this is something you should deal with in your scripts rather than us dealing with this in the codebase.

mythmon:
Thinking about this a bit, I figured out a way I can make this pretty bullet proof: In my script, I always import a DB and then do extract_db. I can assert that the place the database gets imported to and the place that extract_db is pulling data from are the same.

The root cause here is that I changed my Django DB settings to test the generatedata command at the work week, and then forgot to change it back. Checking for that is pretty easy, but not something we should probably do in extract_db.

willkg:
That sounds perfect! I'll close this out.

PR #2753, opened 2015-12-16
[Bug 1150592] Make font CORS .htaccess rules to apply to all static files.

mythmon:
Some fonts aren't loading on Sumo prod, because fonts require CORS rules, and we aren't setting them. We have fonts outside of `static/fonts`, where these rules used to apply. In particular, we have fonts in `static/fontawesome/...`.

If I understand this correctly, it will have the result of setting CORS headers to allow cross origin use of all of our static resources on any Mozilla site. I think that is OK.

r?


mythmon:
As far as I know, the only way to test this is to actually get it running in Apache. That means either setting up Apache locally, or getting this to dev/stage/prod and testing it there.

safwanrahman:
As the Static Files of Production are loading from CDN, the only way to test it is by deploying into production. Last time while we test the bug, the font was loading in Stage but not in production. Because Stage does not use CDN.

static/.htaccess
```diff
@@ -30,3 +30,13 @@ FileETag MTime
 <FilesMatch "\.txt$">
     FileETag None
 </FilesMatch>
+
+# CORS only for mozilla domains
```
Osmose: This comment led me to believe it was allowing _any_ Mozilla domain to access our static resources, but I think it's actually allowing only SUMO itself if it's on a Mozilla domain.

Osmose:
r+, nice work!

PR #2754, opened 2015-12-17
[bug 1171413] Fix highlighting for <code>.

Osmose:
The parsers for CodeMirror's syntax highlighting of `<code>`, `<pre>`, and `<nowiki>`  now consume input until their closing tags are found, and no longer have the odd behavior around `</nowiki>` tags.

I'm not sure whether I should add tests for this; if I did, I'd be testing if the `<div class="CodeMirror-code">` tag within the CodeMirror editor has `display: none` applied to it. I'm also unsure of where I'd add that test: Would it be with the in-browser tests in smoketests? Somewhere else? I'm happy to add them if I can get guidance on where to add them. :D
r?

### DIFF
--- a/kitsune/sumo/static/sumo/js/codemirror.sumo-mode.js
+++ b/kitsune/sumo/static/sumo/js/codemirror.sumo-mode.js
@@ -68,18 +68,18 @@
         regex: /(<del>)(.*?)(<\/del>)/
       },
       {
-        token: ['variable.language', 'comment'],
-        regex: /(<nowiki>)(.*?)/,
+        token: 'variable.language',
+        regex: /<nowiki>/,
         next: 'nowiki'
       },
       {
-        token: ['variable.language', 'comment'],
-        regex: /(<code>)(.*?)/,
+        token: 'variable.language',
+        regex: /<code>/,
         next: 'code'
       },
       {
-        token: ['variable.language', 'comment'],
-        regex: /(<pre>)(.*?)/,
+        token: 'variable.language',
+        regex: /<pre>/,
         next: 'pre'
       },
       {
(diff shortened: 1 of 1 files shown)


mythmon:
Thinking about this more, I wonder if the `display: none` behavior is intentional, and there is a source of parser errors we should be displaying?

The tests for this would like live in something like [`kitsune/sumo/static/sumo/js/tests`](https://github.com/mozilla/kitsune/tree/master/kitsune/sumo/static/sumo/js/tests). These could also exist in `kitsune/questions`, but there aren't any there yet.

Osmose:
> Thinking about this more, I wonder if the display: none behavior is intentional, and there is a source of parser errors we should be displaying?

I don't think so. I'm pretty sure this is where it's being set: https://github.com/codemirror/CodeMirror/blob/9951761dc3e690cb958b4f11f7a84092466c9c56/lib/codemirror.js#L722

Looks like it's set to `none` for big updates to speed up DOM and style changes. And because we're throwing an exception in the middle of highlighting, it's stuck in that state.

I'll see about adding tests in the place you mentioned, thanks!

mythmon:
> Looks like it's set to none for big updates to speed up DOM and style changes. And because we're throwing an exception in the middle of highlighting, it's stuck in that state.

Well, damn. I was hoping it wasn't that. Oh well, that makes sense.

Osmose:
After some digging, I've hit a blocker in getting CodeMirror running in the mocha tests: It relies on `document.createRange`, which [hasn't been implemented in jsdom yet](https://github.com/tmpvar/jsdom/issues/317).

At this point I'm of the opinion that a regression test for this isn't going to be worth the effort required to pull it off. Unless maybe I should write a smoketest for it?

mythmon:
I consider writing Selenium tests a special kind of hell. If you'd like to do it, go for it, but I'd be fine without it.

@rdalal, I don't know much about the code mirror parser stuff. Can you take a look at this?

rehandalal:
code looks fine to me

PR #2755, opened 2015-12-17
[Bug 1233485] Add server requirements to deploy.

mythmon:
r?


willkg:
Looks good to me. You could also do it in one big peep line:

```
peep.sh install -r requirements/default.txt -r requirements/server.txt
```

I don't know if you get anything out of that, though.

PR #2756, opened 2015-12-22
[bug 1225595] Fix the wiki share link

rehandalal:
r?


mythmon:
We've done the opposite of this before: https://github.com/mozilla/kitsune/commit/ccd86c4637846bef68dd0ae68e26678ec946c927

That's weird, but it does definitely fix the problem.

mythmon:
Either way, this fixes the issue, so lets go with it.

PR #2757, opened 2015-12-23
[bug 1234789] Fix submit button on wiki translate page

rehandalal:
r?


willkg:
Wow--that template is pretty complicated.

I verified that the `<form>` gets closed below `if documents`, so this looks ok to me. I'm not set up to test this, so I think we'll have to get the fix verified by someone else or on -stage. r+ with caveats.

PR #2758, opened 2015-12-27
[bug 1225595] fix tooltip strings in revision history page

yfdyh000:
https://github.com/mozilla/kitsune/commits/master/kitsune/wiki/templates/wiki/includes/revision_list.html

e.g. https://support.mozilla.org/en-US/kb/add-ons-signing-firefox/history


safwanrahman:
Awesome! According to [django doc](https://docs.djangoproject.com/en/1.7/topics/i18n/translation/#contextual-markers) it seems right! Thanks a lot!
This should get  **r+**

rehandalal:
good catch! thanks!

PR #2759, opened 2016-01-01
[Bug 1235806] Review Revision page shows <label> tag

safwanrahman:
Changing jija2 filter "f" to "fe" should work. As it will escape the <label> tag.
Also modified the previous test so this issue will not be raise in future.
@mythmon @rehandalal r?


rehandalal:
This is great. Thanks for updating the test as well! r+

PR #2760, opened 2016-01-03
[bug 1219056] Fix API routing for usernames with periods

rehandalal:
r?

### DIFF
--- a/kitsune/users/tests/test_api.py
+++ b/kitsune/users/tests/test_api.py
@@ -188,11 +188,11 @@ class TestUserView(TestCase):
         self.client = APIClient()
 
     def test_usernames_with_periods(self):
-        p = ProfileFactory(user__username='something.something')
-        url = reverse('user-detail', args=[p.user.username])
+        u = UserFactory(username='something.something')
+        url = reverse('user-detail', args=[u.username])
         res = self.client.get(url)
         eq_(res.status_code, 200)
-        eq_(res.data['username'], p.user.username)
+        eq_(res.data['username'], u.username)
 
     def test_only_self_edits(self):
         p1 = ProfileFactory()
--- a/kitsune/users/tests/test_api.py
+++ b/kitsune/users/tests/test_api.py
@@ -187,6 +187,12 @@ class TestUserView(TestCase):
     def setUp(self):
         self.client = APIClient()
 
+    def test_usernames_with_periods(self):
+        p = ProfileFactory(user__username='something.something')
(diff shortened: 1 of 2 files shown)


kitsune/users/tests/test_api.py
```diff
@@ -187,6 +187,12 @@ class TestUserView(TestCase):
     def setUp(self):
         self.client = APIClient()
 
+    def test_usernames_with_periods(self):
+        p = ProfileFactory(user__username='something.something')
+        url = reverse('user-detail', args=[p.user.username])
+        res = self.client.get(url)
+        eq_(res.status_code, 200)
```
willkg: You might want to also verify that the username is in the data that comes back. Given that this is an API, that's possibly JSON. You might need to also send an Accept header. Verifying that the data that comes back contains the username guarantees you that this actually worked rather than returning something else.

willkg:
Other than that, this looks ok to me. However, I haven't used those DRF mixins before, so I don't know whether this makes sense in the grand scheme of things and/or whether there are better ways to do it.

@mythmon: Thoughts?

kitsune/users/tests/test_api.py
rehandalal: done!

willkg: bd

kitsune/users/tests/test_api.py
```diff
@@ -187,6 +187,13 @@ class TestUserView(TestCase):
     def setUp(self):
         self.client = APIClient()
 
+    def test_usernames_with_periods(self):
+        p = ProfileFactory(user__username='something.something')
+        url = reverse('user-detail', args=[p.user.username])
```
mythmon: The `UserFactory` is better than the `user` model maker, in that it will create a profile for all users. Given that, you should be able to do this instead, which saves some typing:

```python
u = UserFactory(username='something.something')
url = reverse('user-detail', args=[u.username])
```

mythmon:
I didn't know about this setting in DRF. After thinking about it for a bit and reading the docs, it makes sense why it is this way, and I think it makes sense for us to change it.

I imagine the default is the way it is because some apis look like `/v2/user/bob.yml`, and it would be weird to look for a user named `bob.yml` instead of returning YAML formatted info about the user named `bob`. That doesn't apply to us though.

We may have other places where we want to use this setting. It might be worth auditing the code for models that use slugs as url keys that allow periods.

rehandalal:
updated ^

mythmon:
Awesome, thanks!

PR #2763, opened 2016-01-06
Upgrade peep to support pip 7.

mythmon:
Peep finally supports pip 7! This should make things a little easier for setting up a new environment.

We'll probably want to do something like this on Input too, since it has a similar peep.sh.

r?

### DIFF
--- a/bin/vagrant_provision.sh
+++ b/bin/vagrant_provision.sh
@@ -69,7 +69,6 @@ VENV=/home/vagrant/virtualenv
 sudo -H -u vagrant -s -- <<EOF
 virtualenv $VENV
 source $VENV/bin/activate
-pip install -U "pip<7"
 cd ~/kitsune
 ./peep.sh install -r requirements/default.txt
 ./peep.sh install -r requirements/dev.txt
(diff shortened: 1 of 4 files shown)


willkg:
I wouldn't bother editing Input. pip 8 should be out soon (possibly this month) and then we ditch peep altogether.

bin/vagrant_provision.sh
```diff
@@ -69,7 +69,6 @@ VENV=/home/vagrant/virtualenv
 sudo -H -u vagrant -s -- <<EOF
 virtualenv $VENV
 source $VENV/bin/activate
-pip install -U "pip<7"
```
safwanrahman: I think a specific version of pip should be mentioned here. So in future it wil not be broken if peep does not work with the future versions.
``pip install -U pip==7.1.2`` should be a better choice

safwanrahman:
I think this should get merged. What do you think?

bin/vagrant_provision.sh
mythmon: The vagrant box comes with a fixed version of Pip that doesn't change. If we change the base box for virtualbox, it may come with a newer version of Pip. If someone is changing the base box, they should hit a problem here.

safwanrahman: Hmm. The vagrant box maybe Ubuntu 14.04. It has pip 1.5.x. So I think this should be upgrade to a fixed version of ``7.1.2`` which peep supports.

PR #2764, opened 2016-01-08
[Bug 1236631] en-us missing in fallback loacales

pollti:


mythmon:
@pollti Can you please add a test case that demonstrates the issue this is fixing?

kitsune/wiki/views.py
```diff
@@ -1527,7 +1528,10 @@ def get_fallback_locale(doc, request):
     #   * wiki fallbacks for that locale
 
     for locale in all_accepted_locales:
-        if locale in translated_locales:
+        if locale == doc.locale:
```
safwanrahman: The ``doc.locale`` can be any locale of the requested doc. I think this should be ``settings.WIKI_DEFAULT_LANGUAGE``

kitsune/wiki/tests/test_views.py
```diff
@@ -177,6 +177,22 @@ def test_custom_event_while_fallback_locale(self):
         assert '"Fallback Locale"' in doc('body').attr('data-ga-push')
         assert '"Not Localized"' not in doc('body').attr('data-ga-push')
 
+    def test_en_in_accept_language_header(self):
```
safwanrahman: I dont see any ``accpeted_language_header`` is passed through the client for fallback. can you please check other tests where the header is passed and tested?

safwanrahman: Moreover, I think this test should be in the ``FallbackLocale`` test class. check that

safwanrahman:
I believe the test should be in [FallbacSystem class](https://github.com/mozilla/kitsune/blob/master/kitsune/wiki/tests/test_views.py#L801). @pollti please check the tests and make a test with the en-US accepted header.

kitsune/wiki/views.py
safwanrahman: I think its better to add the ``en-US`` in the translation list rather than escaping it from here.

pollti: @safwanrahman I think adding en-US to the translations list would cause a different note in the article:
>This page does not exist in German. You have been redirected to the English version instead. If you would like to localize it into German, click here. You can also see the English version of this page.

That would not be very logical. Do you agree?

safwanrahman: Yeah. I accept it would not be logical. moreover, it will raise issue about GA tracking as we are considering the other non en-US locale. Most of the accepted language header have en-US, so showing en-US content thorough the fallback system will not be the thing we expected.
So therefore, better exclude it from the list. And your approach is correct

safwanrahman:
The test looks good and it passed. I have reviewed the code and it seems that it would not break anything. Also it did not fail any of the test.
Thanks a lot @pollti for the awesome work
@mythmon I think this should get r+. What do you think?

mythmon:
These changes make sense and the test is robust. Thanks for the help!

pollti:
Thanks @safwan and @mythmon!

PR #2766, opened 2016-01-10
[Bug 619284] Create a "draft" status for unfinished translations

safwanrahman:
A long waiting patch. :smiley: :sunglasses: Saving a revision as draft is badly needed from localization end. The saving and restoring system can be better by using javascript, but I dont have much Idea about javascript. so leaving it for others to work in future. The javascript in this patch may not be as good because of that, so please consider reviewing. Though most of the work is done in the backend.
@mythmon r?

### DIFF
--- a/kitsune/sumo/static/sumo/js/wiki.js
+++ b/kitsune/sumo/static/sumo/js/wiki.js
@@ -1,5 +1,5 @@
 /* globals k:false, jQuery:false, ShowFor:false, Marky:false, VERSIONS:false,
-           BrowserDetect:false, gettext:false, KBox:false, CodeMirror:false */
+           BrowserDetect:false, gettext:false, KBox:false, CodeMirror:false, interpolate:false */
 /*
  * wiki.js
  * Scripts for the wiki app.
@@ -73,6 +73,7 @@
 
     if ($body.is('.translate')) {  // Translate page
       initToggleDiff();
+      initTranslationDraft();
     }
 
     initEditingTools();
@@ -636,6 +637,33 @@
                     $contentOrDiff.toggleClass('content diff');
                   }));
     }
+  }
+
+  function initTranslationDraft() {
+    var $draftButton = $('.btn-draft'),
+      url = $('.btn-draft').data('draft-url'),
+      $draftMessage = $('#draft-message');
+
+    $draftButton.click( function() {
+      var message = gettext('<strong>Draft is saving...</strong>'),
+        image = '<img src="/static/sumo/img/customercare/spinner.gif">',
+        bothData = $('#both_form').serializeArray(),
+        docData = $('#doc_form').serializeArray(),
+        revData = $('#rev_form').serializeArray(),
+        totalData = $.extend(bothData, docData, revData);
+
+      $draftMessage.html(image + message).removeClass('success error').addClass('info').show()
+      $.post(url, totalData)
+        .done(function() {
+          var time = new Date(),
+            message = interpolate(gettext('<strong>Draft has been saved on:</strong> %s'), [time]);
+          $draftMessage.html(message).toggleClass('info success').show();
+        })
+        .fail(function() {
+          var message = gettext('<strong>Draft is saving...</strong>');
+          $draftMessage.html(message).toggleClass('info error').show();
+        });
+    });
   }
 
   function initRevisionList() {
--- a/kitsune/sumo/static/sumo/js/wiki.js
+++ b/kitsune/sumo/static/sumo/js/wiki.js
@@ -660,7 +660,7 @@
           $draftMessage.html(message).toggleClass('info success').show();
         })
         .fail(function() {
-          var message = gettext('<strong>Draft is saving...</strong>');
+          var message = gettext('<strong>Error saving draft</strong>');
           $draftMessage.html(message).toggleClass('info error').show();
         });
     });
--- a/kitsune/wiki/forms.py
+++ b/kitsune/wiki/forms.py
@@ -9,7 +9,7 @@ from kitsune.products.models import Product, Topic
 from kitsune.sumo.form_fields import MultiUsernameField, StrippedCharField
 from kitsune.wiki.config import SIGNIFICANCES, CATEGORIES
 from kitsune.wiki.models import (
-    Document, Revision, MAX_REVISION_COMMENT_LENGTH)
+    Document, Revision, DraftRevision, MAX_REVISION_COMMENT_LENGTH)
 from kitsune.wiki.tasks import add_short_links
 from kitsune.wiki.widgets import (
     RadioFieldRendererWithHelpText, ProductTopicsAndSubtopicsWidget,
@@ -274,6 +274,22 @@ class RevisionForm(forms.ModelForm):
 
         new_rev.save()
         return new_rev
+
+
+class DraftRevisionForm(forms.ModelForm):
+    class Meta(object):
+        model = DraftRevision
+        fields = ('keywords', 'summary', 'content', 'slug', 'title', 'based_on')
+
+    def save(self, request):
+        """save the draft revision and return the draft revision"""
+        creator = request.user
+        doc_data = self.cleaned_data
+        parent_doc = doc_data['based_on'].document
+        locale = request.LANGUAGE_CODE
+        draft, created = DraftRevision.objects.update_or_create(
+            creator=creator, document=parent_doc, locale=locale, defaults=doc_data)
+        return draft
 
 
 class ReviewForm(forms.Form):
--- a/kitsune/wiki/models.py
+++ b/kitsune/wiki/models.py
@@ -828,18 +828,27 @@ register_for_indexing(
 MAX_REVISION_COMMENT_LENGTH = 255
 
 
-class Revision(ModelBase, SearchMixin):
-    """A revision of a localized knowledgebase document"""
-    document = models.ForeignKey(Document, related_name='revisions')
-    summary = models.TextField()  # wiki markup
-    content = models.TextField()  # wiki markup
-
+class AbstractRevision(models.Model):
+    # **%(class)s** is being used because it will allow  a unique reverse name for the field
+    # like created_revisions and created_draftrevisions
+    creator = models.ForeignKey(User, related_name='created_%(class)ss')
+    created = models.DateTimeField(default=datetime.now)
+    # The reverse name should be revisions and draftrevisions
+    document = models.ForeignKey(Document, related_name='%(class)ss')
     # Keywords are used mostly to affect search rankings. Moderators may not
     # have the language expertise to translate keywords, so we put them in the
     # Revision so the translators can handle them:
     keywords = models.CharField(max_length=255, blank=True)
 
-    created = models.DateTimeField(default=datetime.now)
+    class Meta:
+        abstract = True
+
+
+class Revision(ModelBase, SearchMixin, AbstractRevision):
+    """A revision of a localized knowledgebase document"""
+    summary = models.TextField()  # wiki markup
+    content = models.TextField()  # wiki markup
+
     reviewed = models.DateTimeField(null=True)
     expires = models.DateTimeField(null=True, blank=True)
 
@@ -849,7 +858,6 @@ class Revision(ModelBase, SearchMixin):
     comment = models.CharField(max_length=MAX_REVISION_COMMENT_LENGTH)
     reviewer = models.ForeignKey(User, related_name='reviewed_revisions',
                                  null=True)
-    creator = models.ForeignKey(User, related_name='created_revisions')
     is_approved = models.BooleanField(default=False, db_index=True)
 
     # The default locale's rev that was the latest ready-for-l10n one when the
@@ -1053,6 +1061,15 @@ class Revision(ModelBase, SearchMixin):
     @classmethod
     def get_mapping_type(cls):
         return RevisionMetricsMappingType
+
+
+class DraftRevision(ModelBase, SearchMixin, AbstractRevision):
+    based_on = models.ForeignKey(Revision)
+    content = models.TextField(blank=True)
+    locale = LocaleField(blank=False, db_index=True)
+    slug = models.CharField(max_length=255, blank=True)
+    summary = models.TextField(blank=True)
+    title = models.CharField(max_length=255, blank=True)
 
 
 @register_mapping_type
--- a/kitsune/wiki/templates/wiki/translate.html
+++ b/kitsune/wiki/templates/wiki/translate.html
@@ -46,6 +46,20 @@
           </li>
         {% endif %}
 
+        {% if draft_revision %}
+          <li class="info">
+            <p><strong>{{ _('You have a draft revision for this article saved on {date_time}')|f(date_time=draft_revision.created) }}</strong></p>
+            <form action="" method="get">
+              <input class="btn" name="restore" value={{ _('Restore') }} type="submit">
+              <input class="btn" name="discard" value={{ _('Discard') }} type="submit">
+            </form>
+          </li>
+        {% endif %}
+        {% if more_updated_rev %}
+          <li class="warning">
+            <p><strong>{{ _('This version is outdated, but there is a new version <a href="{link}">{revision}</a> available.')|fe(link=url('wiki.revision', document.slug, more_updated_rev.id, locale=locale), revision=more_updated_rev.id) }}</strong></p>
+          </li>
+        {% endif %}
         {{ edit_messages(document, show_revision_warning) }}
         {{ document_lock_warning() }}
       </ul>
@@ -58,7 +72,7 @@
       </div>
       {% if not document %}
         {# If this is the first translation to this locale, we use 1 big form. #}
-        <form action="" method="post" data-json-url="{{ url('wiki.json') }}">
+        <form id="both_form" action="" method="post" data-json-url="{{ url('wiki.json') }}">
           {{ csrf() }}
           <input type="hidden" name="form" value="both" />
           <input type="hidden" name="slug" value="{{ parent.slug }}" />
@@ -70,7 +84,7 @@
           {{ errorlist(document_form) }}
           {% if document %}
             {# If there are existing translations in this locale, we use 2 separate forms. #}
-            <form action="" method="post" data-json-url="{{ url('wiki.json') }}" data-document-id="{{ document.id }}">
+            <form id="doc_form" action="" method="post" data-json-url="{{ url('wiki.json') }}" data-document-id="{{ document.id }}">
               {{ csrf() }}
               <input type="hidden" name="form" value="doc" />
               <input type="hidden" name="slug" value="{{ document.slug }}" />
@@ -116,7 +130,7 @@
           {{ errorlist(revision_form) }}
           {% if document %}
             {# If there are existing translations in this locale, we use 2 separate forms. #}
-            <form action="" method="post">
+            <form id="rev_form" action="" method="post">
               {{ csrf() }}
               <input type="hidden" name="form" value="rev" />
               <input type="hidden" name="slug" value="{{ document.slug }}" />
@@ -186,7 +200,7 @@
             {# If the document has been created and has a current revision,
                we allow the localizer to keep the translation out of date
                with this new revision. #}
-            {{ submit_revision(revision_form, show_no_update_checkbox=(document and document.current_revision), include_diff=True) }}
+            {{ submit_revision(revision_form, show_no_update_checkbox=(document and document.current_revision), include_diff=True, translate=True) }}
             <div id="preview" class="cf"></div>
             <div id="preview-diff">
                 <div class="from">{{ revision_form.content.value() }}</div>
@@ -195,6 +209,8 @@
               </div>
             <div class="submit" id="preview-bottom">
               {{ submit_revision(revision_form, buttons_only=True, include_diff=True) }}
+            </div>
+            <div id="draft-message" class="alert" hidden>
             </div>
           </div>
           {% if document %}
--- a/kitsune/wiki/tests/test_templates.py
+++ b/kitsune/wiki/tests/test_templates.py
@@ -25,7 +25,7 @@ from kitsune.wiki.config import (
     CATEGORIES, CANNED_RESPONSES_CATEGORY, TEMPLATES_CATEGORY, TEMPLATE_TITLE_PREFIX)
 from kitsune.wiki.tasks import send_reviewed_notification
 from kitsune.wiki.tests import (
-    TestCaseBase, DocumentFactory, RevisionFactory, ApprovedRevisionFactory,
+    TestCaseBase, DocumentFactory, DraftRevisionFactory, RevisionFactory, ApprovedRevisionFactory,
     RedirectRevisionFactory, TranslatedRevisionFactory, LocaleFactory, new_document_data)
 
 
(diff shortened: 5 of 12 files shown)


kitsune/sumo/static/sumo/js/wiki.js
```diff
@@ -787,4 +787,14 @@
...
   }
 
+  var draftbutton = document.getElementsByClassName('btn-draft')[0];
+  var url = $( '.btn-draft' ).data( 'draft-url' );
+  draftbutton.addEventListener('click', function() {
+    var both_data = $('#both_form').serializeArray();
+    var doc_data = $('#doc_form').serializeArray();
+    var rev_data = $('#rev_form').serializeArray();
+    var total_data = both_data.concat(doc_data, rev_data);
+    $.post( url, total_data );
```
mythmon: This should give feedback to users, reporting whether the action was a success or failure.

kitsune/sumo/static/sumo/js/wiki.js
```diff
@@ -787,4 +787,14 @@
     });
   }
 
+  var draftbutton = document.getElementsByClassName('btn-draft')[0];
```
mythmon: Vanilla-JS DOM access is better written as `var draftButton = document.querySelector('.btn-draft');`

kitsune/sumo/static/sumo/js/wiki.js
```diff
@@ -787,4 +787,14 @@
...
+  var url = $( '.btn-draft' ).data( 'draft-url' );
+  draftbutton.addEventListener('click', function() {
+    var both_data = $('#both_form').serializeArray();
+    var doc_data = $('#doc_form').serializeArray();
+    var rev_data = $('#rev_form').serializeArray();
+    var total_data = both_data.concat(doc_data, rev_data);
```
mythmon: Mixing Vanilla-JS DOM access and jQuery DOM access is confusing. You should pick one and use that for all the operations.

kitsune/wiki/forms.py
```diff
@@ -276,6 +276,23 @@ def save(self, creator, document, based_on_id=None, base_rev=None,
         return new_rev
 
 
+class DraftRevisionForm(forms.ModelForm):
+    class Meta(object):
+        model = DraftRevision
+        fields = ('keywords', 'summary', 'content', 'slug', 'title',
+                  'based_on')
```
mythmon: Nit: This line doesn't need to wrap.

kitsune/wiki/forms.py
```diff
@@ -276,6 +276,23 @@ def save(self, creator, document, based_on_id=None, base_rev=None,
...
+
+    def save(self, user, parent_doc, doc_data):
+        """save the draft revision and return the draft revision"""
+        locale = doc_data["locale"]
+        doc_data.pop("locale", None)
```
mythmon: `.pop` is used to remove *and return* a value from a dict. This is better written as `locale = doc_data.pop('locale')`.

kitsune/wiki/views.py
```diff
@@ -537,6 +537,22 @@ def edit_document(request, document_slug, revision_id=None):
...
+@require_POST
+def draft_revision(request):
+    """Create a Draft Revision.
+
+    User can have only one draft revision for a translated document. Store the draft with
+    parent document, user and locale. Get the parent document from the based on revision"""
+    draft_form = DraftRevisionForm(request.POST)
+    if draft_form.is_valid():
+        _create_draft_revision(request, draft_form)
+        return HttpResponse('')
```
mythmon: Better would be `return HttpResponse(status=204)`. [`HTTP 204`](http://http.cat/204) is "No Content" and is appropriate to use in this context.

kitsune/wiki/views.py
```diff
@@ -1428,6 +1485,15 @@ def _save_rev_and_notify(rev_form, creator, document, based_on_id=None,
     EditDocumentEvent(new_rev).fire(exclude=new_rev.creator)
 
 
+def _create_draft_revision(request, draft_form):
```
mythmon: This seems like it should be integrated into `DraftForm` in some way.

kitsune/wiki/views.py
```diff
@@ -813,23 +829,50 @@ def translate(request, document_slug, revision_id=None):
...
+    draft = DraftRevision.objects.filter(creator=user, parent_doc=parent_doc,
+                                         locale=request.LANGUAGE_CODE)
+
     doc_form = rev_form = None
     base_rev = None
+    restore_draft = None
+    # Check if the user has requested to restore or discard draft
+    draft_request = request.GET.get('form') == 'draft'
+    draft_revision = None
+    # If Draft is available and the user has requested with draft paramaeters
```
mythmon: Typo: paramaeters should be parameters.

kitsune/wiki/views.py
```diff
@@ -1417,6 +1462,18 @@ def _document_form_initial(document):
             'needs_change_comment': document.needs_change_comment}
 
 
+def _draft_initial(type, draft_revision):
```
mythmon: This seems like it should be two methods, `_draft_initial_doc` and `_draft_initial_rev`.

kitsune/sumo/static/sumo/js/wiki.js
safwanrahman: @mythmon Actually, I dont have much knowledge about javascript. If you can show me any code which do this type of work, I think I can understand how to show a feedback to the users.

kitsune/wiki/views.py
safwanrahman: What do you think about returning http 201? as its creating a draft revision, HTTP 201 maybe better status code.

kitsune/sumo/static/sumo/js/wiki.js
safwanrahman: Thanks. I did not have idea about it

safwanrahman:
@mythmon Updated. r?

brittanystoroz:
@safwanrahman I'll take a look through your javascript on this today and let you know how it can be improved.

safwanrahman:
Thanks @brittanystoroz. Please let me know so that I can fix

kitsune/sumo/static/sumo/js/wiki.js
```diff
@@ -787,4 +787,20 @@
     });
   }
 
+  var draftbutton = $('.btn-draft');
+  var url = $('.btn-draft' ).data( 'draft-url' );
+  var draftmessage = $('.draft-message').html();
```
brittanystoroz: When you're declaring a lot of variables all at once, you can just specify `var` once in the beginning, and put each new variable on a new line separated by commas. You can see an example of this [here](https://github.com/mozilla/kitsune/blob/master/kitsune/sumo/static/sumo/js/wiki.js#L125-L131)

It's also a good convention to prefix any variable names with `$` if they represent a jQuery-selected DOM element. So in this block of code you might want to change `draftbutton` to `$draftButton`.

kitsune/sumo/static/sumo/js/wiki.js
```diff
@@ -787,4 +787,20 @@
     });
   }
 
+  var draftbutton = $('.btn-draft');
+  var url = $('.btn-draft' ).data( 'draft-url' );
```
brittanystoroz: Nit: remove spaces inside the parenthesis in `( 'draft-url'  )`

kitsune/sumo/static/sumo/js/wiki.js
```diff
@@ -787,4 +787,20 @@
...
+  var draftmessage = $('.draft-message').html();
+  draftbutton[0].addEventListener('click', function() {
```
brittanystoroz: jQuery has a built-in `.click` [method we can use](https://api.jquery.com/click/) instead of `addEventListener`. This will remove the need for the `[0]` and clean up the syntax a little bit. So this could be re-written like:

```javascript
$draftButton.click(function() {
   // do things
});
```

kitsune/sumo/static/sumo/js/wiki.js
```diff
@@ -787,4 +787,20 @@
...
+    var both_data = $('#both_form').serializeArray();
+    var doc_data = $('#doc_form').serializeArray();
+    var rev_data = $('#rev_form').serializeArray();
```
brittanystoroz: Nit: camelCase variable names rather than snake_casing. 

Also, can we think of some more meaningful ID names for these form elements? At first glance, it's really hard for me to decipher what `both_form` `doc_form` and `rev_form` mean. If you can help clarify their differences for me a bit I can try to think of some more informative IDs.

kitsune/sumo/static/sumo/js/wiki.js
```diff
@@ -787,4 +787,20 @@
...
+    var total_data = $.extend(both_data, doc_data, rev_data);
+    var posting = $.post( url, total_data );
+    posting.done(function() {
+      var time = new Date();
+      var message = draftmessage + time
+      $('.draft-message').css('display', 'block').empty().append(message)
```
brittanystoroz: Instead of emptying and appending a new message each time, we can simplify this with `.html()`:
`$('.draft-message').html(message);` would replace the entire inner html of the `div` each time. 

For toggling the display of the element, we can use `.show()` instead of `.css()`. We would probably also want this to come *after* we replace the inner content so we're sure the content is updated before we display the element:

`$('.draft-message').html(message).show();`

brittanystoroz:
@safwanrahman added some line notes with small improvements to start, couple more things we might want to do here but let's get these in first and take a look again after

kitsune/sumo/static/sumo/js/wiki.js
safwanrahman: There are 3 type of forms a localizers can see in time by time.
``doc_form`` refers to a form, which contains only document content like title and slug
``rev_form`` refers to a form, which contains revision content like revision content, search summery, and others
``both_form`` refers to a big form, which has both document content and revision content. This form is being showed to the localizers when they localize any article for first time. 
I think the ID name is meaningful as per their use. As we define like the same in the backend views.

safwanrahman:
@brittanystoroz updated the javacript. Travis fail seems not related. can you re run the test?

safwanrahman:
@brittanystoroz I have updated the patch by adding Error message if saving draft revision fail. also added some CSS for styling the message. What do you think about the fontend  now?

kitsune/sumo/static/sumo/js/wiki.js
```diff
@@ -787,4 +787,30 @@
     });
   }
 
+  var $draftButton = $('.btn-draft'),
```
brittanystoroz: Let's take allllll of this javascript you wrote (Lines 790 - 814) and wrap it in a function called `initTranslationDraft` and move it up to line 641 - in between the `initToggleDiff` and `initRevisionList` functions.  Right now it's kind of just hanging out at the bottom of the file, unattached. This will make it a little more consistent with the way the rest of the file is written.

Then let's invoke this new `initTranslationDraft` function on [line 75](https://github.com/safwanrahman/kitsune/blob/draft/kitsune/sumo/static/sumo/js/wiki.js#L75) right beneath `initToggleDiff();`  This will ensure that all of your code is only being used when we're actually on the translation page where it's needed.

safwanrahman:
@mythmon r?

brittanystoroz:
Travis is finally happy :tada: :tada: :tada:  @mythmon I am good with the front-end on this, wanna take a look at the back-end when you have a chance?

rehandalal:
@mythmon and I had a quick chat about this and we think it might be a good idea to refactor this slightly so that you create an `AbstractRevision` class for the common elements of `Revision` and `DraftRevision` and have both those classes inherit from `AbstractRevision`. I can walk you through this if you need help with it.

mythmon:
The docs about abstract models are [here](https://docs.djangoproject.com/en/1.7/topics/db/models/#model-inheritance).

kitsune/sumo/static/sumo/js/wiki.js
```diff
@@ -638,6 +639,34 @@
...
+        totalData = $.extend(bothData, docData, revData);
+      $.ajax({
+        type: 'POST',
+        url: url,
+        data: totalData,
+        timeout: 3000
+      })
+        .done(function() {
+          var time = new Date(),
+            message = draftMessage + time;
```
mythmon: This way of building a message is bad for localization. You should move the text out of the HTML and into the JS and make this line something like `message = k.safeInterpolate(_('Draft has been saved on: {time}'), {time: time});`.

mythmon: This is because not all languages will want to put `{time}` at the end of the string. Some might want to put it in the middle or at the beginning.

kitsune/wiki/forms.py
```diff
@@ -276,6 +276,24 @@ def save(self, creator, document, based_on_id=None, base_rev=None,
...
+    def save(self, request):
+        """save the draft revision and return the draft revision"""
+        creator = request.user
+        doc_data = self.cleaned_data
+        parent_doc = doc_data['based_on'].document
+        locale = request.LANGUAGE_CODE
+        draft, created = DraftRevision.objects.update_or_create(creator=creator,
+                                                                parent_doc=parent_doc,
+                                                                locale=locale,
+                                                                defaults=doc_data)
```
mythmon: This matches our style guide, but personally I think it is more readable like this:

```python
draft, created = DraftRevision.objects.update_or_create(
    creator=creator, parent_doc=parent_doc, locale=locale, defaults=doc_data)
```

safwanrahman:
@mythmon @rehandalal Updated with the Inherited class. r?
I would like to squash it into one before merging. :+1:

kitsune/wiki/models.py
```diff
@@ -857,8 +865,6 @@ class Revision(ModelBase, SearchMixin):
     # this is simply the latest of the default locale's revs as of that time.
     # Used to determine whether localizations are out of date.
     based_on = models.ForeignKey('self', null=True, blank=True)
-    # TODO: limit_choices_to={'document__locale':
```
mythmon: I don't think you have handled this TODO item in this PR. Then again, I don't really know what it is talking about. Do you?

safwanrahman: I think its about having the ``based_on`` into ``settings.WIKI_DEFAULT_LANGUAGE`` only. Lets keep this as it is

kitsune/sumo/static/sumo/js/wiki.js
```diff
@@ -787,4 +809,5 @@
     });
   }
 
+
```
rehandalal: nit: this seems like an unnecessary extra line

safwanrahman:
updated @rehandalal. Is everything Ok?

kitsune/wiki/tests/test_templates.py
```diff
@@ -1013,6 +1013,143 @@ def test_keywords_require_permission(self):
         new_rev = Revision.objects.filter(document=doc).order_by('-id')[0]
         eq_('keyword1 keyword2', new_rev.keywords)
 
+    def test_restore_draft_revision(self):
+        rev = ApprovedRevisionFactory(is_ready_for_localization=True)
+        doc = rev.document
+        doc.is_localizable = True
+        doc.save()
```
mythmon: You could do this more easily with `rev = ApprovedRevisionFactory(is_ready_for_localization=True, document__is_localizable=True)`.

kitsune/wiki/tests/test_templates.py
```diff
@@ -1013,6 +1013,143 @@ def test_keywords_require_permission(self):
...
+        trans_content = pq(trans_resp.content)
+        eq_(0, len(trans_content('.user-messages li')))
+        eq_(1, len(trans_content('.submit .btn-draft')))
+        data = {
+            'content': 'Test Content bla bla bla',
+            'keywords': 'something',
+            'summary': 'some summary',
+            'title': 'test title',
+            'slug': 'test-slug',
+            'based_on': int(rev.id)}
```
mythmon: Should `rev.id` already be an int?

safwanrahman: I strongly believed it would be int. but sometime the test fail because it return ``45L`` or something like that.

kitsune/wiki/tests/test_templates.py
```diff
@@ -1013,6 +1013,143 @@ def test_keywords_require_permission(self):
...
+        cache.clear()
+        trans_resp = self.client.get(trans_url)
+        trans_content = pq(trans_resp.content)
+        # Check user message is shown there
+        eq_(1, len(trans_content('.user-messages li')))
+        # Check there are two buttons for restore and discard
+        eq_(2, len(trans_content('.user-messages .info form .btn')))
+        # Restore with the draft data
+        draft_request = {'restore': 'Restore'}
+        cache.clear()
```
mythmon: What's in the cache that you keep clearing?

safwanrahman: ``trans_resp = self.client.get(trans_url)``

safwanrahman: at line ``1041``

mythmon: That shouldn't be cached. These tests pass fine for me without clearing the cache. What specifically (what key and value) do you expect to be in the cache?

kitsune/wiki/tests/test_templates.py
```diff
@@ -1013,6 +1013,143 @@ def test_keywords_require_permission(self):
...
+        trans_content = pq(trans_resp.content)
+        eq_(rev2.content, trans_content('.content textarea').text())
+        # Now Restore the draft which is based on the past revision
+        draft_request = {'restore': 'Restore'}
+        cache.clear()
+        trans_resp = self.client.get(trans_url, draft_request)
+        trans_content = pq(trans_resp.content)
+        eq_(rev1.content, trans_content('.content textarea').text())
+        # As the old based on draft is restored, the latest revision content should not be there
+        assert not rev2.content == trans_content('.content textarea').text()
```
mythmon: you probably want `assert rev2.content != trans_content('.content textarea').text()`. That's a bit clearer.

kitsune/wiki/tests/test_templates.py
safwanrahman: I dont know. While requesting the same page in same test, it raised error for me because it was cached. The translation page get cached and when I tried to get the page again, it returned the cached page. 
If test passes without clearing the cache, I am happy to remove it

kitsune/wiki/tests/test_templates.py
safwanrahman: Oh! Yeah.

mythmon:
I have some trouble with the tests for this. I think are functional, but they are very muddled. They involve a lot of parts and each test relies on many things. I think the tests should be split up into smaller units, and we need to work on the way you set up the tests.

In particular, you often are using the http interfaces to create models for testing. The only time you should be calling the views (with something like `self.client.get`) is when you are testing the view. Otherwise, you should be creating the models through a factory. You'll need to create a DraftRevisionFactory in `kitsune/wiki/tests/__init__.py`.

I'd go so far as to say that each test should probably only have one http request in it. This is not an absolute rule, it is possible to write a good test that has more than one, but it is probably a good goal to strive for.

Does that make sense? I can give more focused feedback if that would be useful. We can also chat on IRC about this, if it would be easier.

kitsune/sumo/static/sumo/less/main.less
```diff
@@ -307,6 +307,27 @@ input[type="submit"]{
   }
 }
 
+/* Taken from Bootstrap V3 */
+.alert {
+  border: 1px solid transparent;
+  border-radius: 4px;
+  background: #FFE url("../img/warning-stripes.png") repeat-x scroll left top;
```
rehandalal: nit: alphabetized - ba < bo

kitsune/sumo/static/sumo/less/main.less
```diff
@@ -307,6 +307,27 @@ input[type="submit"]{
...
+  padding: 15px;
+  margin-bottom: 20px;
+}
+
+.success {
```
rehandalal: same with everything here. alphabetize.

kitsune/sumo/static/sumo/less/main.less
```diff
@@ -307,6 +307,27 @@ input[type="submit"]{
...
+  color: #3c763d;
+  background-color: #dff0d8;
+  border-color: #d6e9c6;
+}
+
+.info {
```
rehandalal: more alphabetizing required.

kitsune/wiki/models.py
```diff
@@ -1055,6 +1063,15 @@ def get_mapping_type(cls):
         return RevisionMetricsMappingType
 
 
+class DraftRevision(ModelBase, SearchMixin, AbstractRevision):
+    based_on = models.ForeignKey(Revision, blank=True, null=True)
+    content = models.TextField(blank=True)
+    locale = LocaleField(default=settings.WIKI_DEFAULT_LANGUAGE)
```
rehandalal: do we need this locale field? don't we just use the locale of the document?

kitsune/wiki/tests/test_views.py
```diff
@@ -243,6 +243,21 @@ def setUp(self):
...
+    def create_a_draft_revision(self, user, locale, based_on):
+        self.client.login(username=user.username, password='testpass')
+        url = reverse('wiki.draft_revision', locale='bn-BD')
+        data = {
+            'content': 'Test Content bla bla bla',
+            'keyword': 'something',
+            'summary': 'some summary',
+            'title': 'test title',
+            'slug': 'test-slug',
+            'based_on': int(based_on.id)
```
rehandalal: should this be an int?

rehandalal: oh @mythmon  already addressed this. it appears all over the tests.

kitsune/wiki/models.py
safwanrahman: As the document will always be a ``en-US`` one, we can not. We identify with parent document, and locale

kitsune/sumo/static/sumo/less/main.less
mythmon: To clarify: Putting CSS rules in alphabetical order is good style, and actually has a small impact on the compressed file size. Smaller files means faster pages. So it seems silly, but it helps.

safwanrahman: Actually, this is copy from bootstrap. so it was like just be there.

safwanrahman:
@mythmon updated the tests. r?

kitsune/wiki/tests/test_views.py
```diff
@@ -470,6 +471,98 @@ def test_needs_change(self):
...
+            'summary': 'some summary',
+            'title': 'test title',
+            'slug': 'test-slug',
+            'based_on': rev.id
+        }
+        resp = self.client.post(url, data)
+        eq_(201, resp.status_code)
+        obj = DraftRevision.objects.get(creator=self.u, document=rev.document, locale='bn-BD')
+        eq_(obj.content, data['content'])
+        eq_(obj.based_on, rev)
```
mythmon: This test is great. It tests one thing simply and is easy to follow. Nice job!

kitsune/wiki/tests/test_views.py
```diff
@@ -470,6 +471,98 @@ def test_needs_change(self):
...
+
+    def test_draft_revision_restore_in_translation_page(self):
+        """Check Draft Revision is restored when a user click on the Restore Button"""
+        # Create a draft revision
+        draft = DraftRevisionFactory(creator=self.u)
+        doc = draft.document
+        # Get the Translation page of the document
+        trans_url = reverse('wiki.translate', locale=draft.locale, args=[doc.slug])
+        trans_resp = self.client.get(trans_url)
+        eq_(200, trans_resp.status_code)
```
mythmon: I thinks getting the Translation page of the document is unneeded. The API to store the draft page should be stateless and not depend on the translation page being loaded first. Lines 519 through 521 should probably be removed.

kitsune/sumo/static/sumo/less/main.less
```diff
@@ -307,6 +307,27 @@ input[type="submit"]{
...
+}
+
+.success {
+  background-color: #dff0d8;
+  border-color: #d6e9c6;
+  color: #3c763d;
+}
+
+.info {
+  background-color: #fcf8e3;
```
mythmon: This CSS rule is too broad, and is affecting other parts of the site. Specifically, this: 
shot_14917

@rehandalal any ideas for better CSS rules here? Should they just be scoped to `#localize-document`?

kitsune/wiki/templates/wiki/translate.html
```diff
@@ -196,6 +210,12 @@
             <div class="submit" id="preview-bottom">
               {{ submit_revision(revision_form, buttons_only=True, include_diff=True) }}
             </div>
+            <div id="draft-message-success" class="alert success" hidden>
+            </div>
+            <div id="draft-message-saving" class="alert info" hidden>
+              <img src="/static/sumo/img/customercare/spinner.gif">
+              <strong>{{ _('Draft is saving...........') }} </strong>
```
mythmon: You don't need 11 `.` here. 3 would do just fine.

kitsune/sumo/static/sumo/js/wiki.js
```diff
@@ -638,6 +639,27 @@
...
+        docData = $('#doc_form').serializeArray(),
+        revData = $('#rev_form').serializeArray(),
+        totalData = $.extend(bothData, docData, revData);
+      $.post(url, totalData)
+        .done(function() {
+          var time = new Date(),
+            message = interpolate(gettext('<strong>Draft has been saved on:</strong> %s'), [time]);
+          $('#draft-message-success').html(message).show();
+          $('#draft-message-saving').hide();
+        });
```
mythmon: You should add a `.fail(function() { ... })` handler to this to show an error message in case the server can't be contacted.

mythmon: The docs here are useful: http://api.jquery.com/jQuery.post/

mythmon:
Most of this looks good, and the back end should work well. I'd like to see a couple more improvements to the frontend, and then this should be ready to go.

mythmon:
The lint error is fixed on master, so the next time you add commits to this, it should go away.

kitsune/sumo/static/sumo/js/wiki.js
safwanrahman: added

safwanrahman:
@mythmon Updated the front end. r?<
After getting r+, I would squash it into one.

safwanrahman:
@mythmon Yap. I like the Idea.
I have made the correction and submitted another commit. can you please check again and let me know?

mythmon:
Ok, I think this is ready to go, thank you!

Go ahead and squash this, and then once Travis has passed we can merge it.

safwanrahman:
:D Squashed and test passed! :green_apple: 
Its finally ready for merge. :tada: :+1: :100: :zzz:

safwanrahman:
@mythmon can merge?

mythmon:
Merged. We can deploy this next week.

Osmose:
:tada:

PR #2767, opened 2016-01-10
Upgrade Jingo

safwanrahman:
We were using jingo 8.1 with a fix of @rehandalal [jingo #76](https://github.com/jbalogh/jingo/pull/76)
The problem got fixed in jingo with [jingo #77](https://github.com/jbalogh/jingo/pull/77) and a version 8.2 released. So now we can download it from Pypi
r?


rehandalal:
Thanks Safwan! I was going to upgrade this but @mythmon has already worked on ditching jingo completely for django-jinja, so I'm going to close this out.

safwanrahman:
@rehandalal I understand. But I can not understand why the Django badger has problem with it. I think the Django badger is not working with jingo 0.8.2
what do you think?

PR #2770, opened 2016-01-12
[bug 1229417] Remove double search box on mobile

rehandalal:
r?


mythmon:
Looks like the main content was absolutely positioned to make room for this element, so you'll need to update those css rules too.
screen shot 2016-01-12 at 11 01 41

rehandalal:
oops! had removed the styles but forgot to add the .less file before committing.

rehandalal:
rebased to include lint changes ^

PR #2771, opened 2016-01-13
[Bug 1238252] Properly encode search URL in mobile instant search.

mythmon:
This is very similar to [bug 1223970](https://bugzilla.mozilla.org/show_bug.cgi?id=1223970).

Nunjucks tries to escape a lot of things, but it only does HTML encoding, not URI encoding.

small r?


mythmon:
@rehandalal r?

There is a problem with the linter that I'll fix.

rehandalal:
makes sense... does this need to be updated for desktop search as well?

mythmon:
We did this already on desktop.

mythmon:
Rebased to fix lint problems ^

rehandalal:
cool r+

PR #2772, opened 2016-01-13
Fix new lint errors.

mythmon:
Pre-commit is lame and doesn't pin things. Recently it caused a new kind of lint rule. This fixes the code to keep us up to date with lint errors, and will fix the lint problems in #2771.

r?


rehandalal:
lgtm r+

Guada174 (commented):
> ``795411

PR #2773, opened 2016-01-13
[Bug 1225511] Hide signin menu link during AAQ

brittanystoroz:
This removes the sign in link/drop-down during the AAQ flow. We previously just hid the sign out button because it was having problems, but this will hide the entire thing in order to solve https://bugzilla.mozilla.org/show_bug.cgi?id=1225511

### DIFF
--- a/kitsune/questions/templates/questions/new_question_react.html
+++ b/kitsune/questions/templates/questions/new_question_react.html
@@ -7,7 +7,7 @@
 {% block breadcrumbs %}{% endblock %}
 
 {% set hide_aaq_link = True %}
-{% set hide_signout_link = True %}
+{% set hide_signin = True %}
 
 {% block content %}
   {{ csrf() }}{# CSRF token for AJAX actions. #}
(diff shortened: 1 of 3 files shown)


rehandalal:
lint issues were fixed in #2772, rebasing should get travis passing.

kitsune/questions/templates/questions/new_question_react.html
```diff
@@ -7,7 +7,7 @@
 {% block breadcrumbs %}{% endblock %}
 
 {% set hide_aaq_link = True %}
-{% set hide_signout_link = True %}
+{% set hide_signin = True %}
```
rehandalal: nit: i think we should name this variable something else since it hides the sign in, sign out and contributor tools.

rehandalal:
after looking at this again. it occurs to me that nothing is being output by the `aux_nav` macro on the AAQ page.

i propose we get rid of the hide_signin param altogether and just have something like this instead:
```
{% set hide_aux_nav = True %}

...

{% if not hide_aux_nav %}
  {{ aux_nav(user, hide_aaq_link) }}
{% endif %}
```

brittanystoroz:
@rehandalal updated

rehandalal:
hawt! r+

PR #2774, opened 2016-01-14
[bug 1225370] Calculate retention metrics for KPI dashboard

rehandalal:
r?

### DIFF
--- a/kitsune/kpi/cron.py
+++ b/kitsune/kpi/cron.py
@@ -542,9 +542,9 @@ def cohort_analysis():
             for retention_range in ranges[i:]:
                 retained_user_count = _count_contributors_in_range(querysets, cohort_users,
                                                                    retention_range)
-                retention_metric, _ = RetentionMetric.objects.update_or_create(
+                RetentionMetric.objects.update_or_create(
                     cohort=cohort, start=retention_range[0], end=retention_range[1],
-                    defaults=retained_user_count)
+                    defaults={'size': retained_user_count})
 
 
 def _count_contributors_in_range(querysets, users, date_range):
--- a/kitsune/kpi/cron.py
+++ b/kitsune/kpi/cron.py
@@ -1,4 +1,6 @@
+import operator
 from datetime import datetime, date, timedelta
+from functools import reduce
 
 from django.conf import settings
 from django.db.models import Count, F, Q
@@ -14,7 +16,7 @@ from kitsune.kpi.models import (
     KB_ENUS_CONTRIBUTORS_METRIC_CODE, KB_L10N_CONTRIBUTORS_METRIC_CODE, L10N_METRIC_CODE,
     SUPPORT_FORUM_CONTRIBUTORS_METRIC_CODE, VISITORS_METRIC_CODE, SEARCH_SEARCHES_METRIC_CODE,
     SEARCH_CLICKS_METRIC_CODE, EXIT_SURVEY_YES_CODE, EXIT_SURVEY_NO_CODE,
-    EXIT_SURVEY_DONT_KNOW_CODE, CONTRIBUTOR_COHORT_CODE, KB_CONTRIBUTOR_COHORT_CODE,
+    EXIT_SURVEY_DONT_KNOW_CODE, CONTRIBUTOR_COHORT_CODE, KB_ENUS_CONTRIBUTOR_COHORT_CODE,
     KB_L10N_CONTRIBUTOR_COHORT_CODE, SUPPORT_FORUM_HELPER_COHORT_CODE, AOA_CONTRIBUTOR_COHORT_CODE)
 from kitsune.kpi.surveygizmo_utils import (
     get_email_addresses, add_email_to_campaign, get_exit_survey_results,
@@ -513,7 +515,7 @@ def cohort_analysis():
             (Revision.objects.all(), ('creator', 'reviewer',)),
             (Answer.objects.not_by_asker(), ('creator',)),
             (Reply.objects.all(), ('user',))]),
-        (KB_CONTRIBUTOR_COHORT_CODE, [
+        (KB_ENUS_CONTRIBUTOR_COHORT_CODE, [
             (Revision.objects.filter(document__locale='en-US'), ('creator', 'reviewer',))]),
         (KB_L10N_CONTRIBUTOR_COHORT_CODE, [
             (Revision.objects.exclude(document__locale='en-US'), ('creator', 'reviewer',))]),
@@ -524,7 +526,7 @@ def cohort_analysis():
     ]
 
     for kind, querysets in reports:
-        cohort_kind = CohortKind.objects.get_or_create(code=kind)[0]
+        cohort_kind, _ = CohortKind.objects.get_or_create(code=kind)
 
         for i, cohort_range in enumerate(ranges):
             cohort_users = _get_cohort(querysets, cohort_range)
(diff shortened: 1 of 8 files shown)


rehandalal:
updated to include AoA ^

kitsune/kpi/cron.py
```diff
@@ -496,3 +496,92 @@ def survey_recent_askers():
...
+            filters = {'%s__in' % field: users, 'created__gte': start, 'created__lt': end}
+            retained_users = retained_users.union(
+                set(getattr(o, field) for o in queryset.filter(**filters)))
+
+    return len(retained_users)
+
+
+def _get_cohort(querysets, date_range):
+    start, end = date_range
+    cohort = []
```
mythmon: Should this be a `set()` to filter out duplicate users?

kitsune/kpi/cron.py
```diff
@@ -496,3 +496,92 @@ def survey_recent_askers():
...
+
+    for queryset, fields in querysets:
+        contributions_in_range = queryset.filter(created__gte=start, created__lt=end)
+        potential_users = set()
+
+        for field in fields:
+            potential_users = potential_users.union(
+                set(getattr(cont, field) for cont in contributions_in_range))
```
mythmon: If you wanted, this could be `potential_users |= set(getattr(cont, field) for cont in contributions_in_range)`. The `|` operator is equivalent to `.union()`. `&` is `.intersection()`.

kitsune/kpi/cron.py
```diff
@@ -496,3 +496,92 @@ def survey_recent_askers():
...
+        (KB_L10N_CONTRIBUTOR_COHORT_CODE, [
+            (Revision.objects.exclude(document__locale='en-US'), ('creator', 'reviewer',))]),
+        (SUPPORT_FORUM_HELPER_COHORT_CODE, [
+            (Answer.objects.not_by_asker(), ('creator',))]),
+        (AOA_CONTRIBUTOR_COHORT_CODE, [
+            (Reply.objects.all(), ('user',))])
+    ]
+
+    for kind, querysets in reports:
+        cohort_kind = CohortKind.objects.get_or_create(code=kind)[0]
```
mythmon: I've usually seen this written as `cohort_kind, _ = CohortKind.objects.get_or_create(code=kind)`. The second value is whether the object was created or not.

kitsune/kpi/cron.py
```diff
@@ -496,3 +496,92 @@ def survey_recent_askers():
...
+            cohort_users = _get_cohort(querysets, cohort_range)
+
+            # Sometimes None will be added to the cohort_users list, so remove it
+            if None in cohort_users:
+                cohort_users.remove(None)
+
+            cohort = Cohort.objects.get_or_create(kind=cohort_kind, start=cohort_range[0],
+                                                  end=cohort_range[1])[0]
+            cohort.size = len(cohort_users)
+            cohort.save()
```
mythmon: You may be interested in [`Cohort.objects.update_or_create`](https://docs.djangoproject.com/en/1.7/ref/models/querysets/#update-or-create).

kitsune/kpi/cron.py
```diff
@@ -496,3 +496,92 @@ def survey_recent_askers():
...
+
+        for field in fields:
+            potential_users = potential_users.union(
+                set(getattr(cont, field) for cont in contributions_in_range))
+
+        def is_in_cohort(u):
+            qs = [Q(**{field: u}) for field in fields]
+            filters = qs.pop()
+            for q in qs:
+                filters |= q
```
mythmon: So here you are `or`ing together all the filters, right? I think this deserves a comment, it is a bit dense.

What do you think of this version?

```python
import operators
from functools import reduce

qs = [Q(**{field: u}) for field in fields]
filters = reduce(operators.or, qs)
```

kitsune/kpi/cron.py
mythmon: PS: I really l like how this avoids needing data migrations for this.

kitsune/kpi/tests/test_cron.py
```diff
@@ -1,24 +1,153 @@
...
 
 
+class CohortAnalysisTests(TestCase):
+    def setUp(self):
+        today = datetime.today()
+        self.start_of_first_week = today - timedelta(days=today.weekday(), weeks=12)
+
+        revisions = ApprovedRevisionFactory.create_batch(3, created=self.start_of_first_week)
+
+        reviewer = ProfileFactory()
```
mythmon: You can safely use `UserFactory()` here, and it will create a profile too.

kitsune/kpi/tests/test_cron.py
```diff
@@ -1,24 +1,153 @@
...
+
+        c1r1 = c1.retention_metrics.get(start=self.start_of_first_week + timedelta(weeks=1))
+        eq_(c1r1.size, 0)
+
+        c1r2 = c1.retention_metrics.get(start=self.start_of_first_week + timedelta(weeks=2))
+        eq_(c1r2.size, 2)
+
+        c2 = Cohort.objects.get(kind__code=AOA_CONTRIBUTOR_COHORT_CODE,
+                                start=self.start_of_first_week + timedelta(weeks=1))
+        eq_(c2.size, 0)
```
mythmon: Wow. Those tests are tough to read. I think they are all good though.

kitsune/questions/managers.py
```diff
@@ -62,3 +62,9 @@ def escalated(self):
 class QuestionLocaleManager(Manager):
     def locales_list(self):
         return self.values_list('locale', flat=True)
+
+
+class AnswerManager(Manager):
+    def not_by_asker(self):
+        """Answers by anyone except the user who asked the question"""
+        return self.exclude(creator=F('question__creator'))
```
mythmon: :+1:

kitsune/kpi/models.py
```diff
@@ -16,6 +16,12 @@
 EXIT_SURVEY_NO_CODE = 'exit-survey:no'
 EXIT_SURVEY_DONT_KNOW_CODE = 'exit-survey:dont-know'
 
+CONTRIBUTOR_COHORT_CODE = 'contributor'
+KB_CONTRIBUTOR_COHORT_CODE = 'contributor:kb'
+KB_L10N_CONTRIBUTOR_COHORT_CODE = 'contributor:kb:l10n'
+SUPPORT_FORUM_HELPER_COHORT_CODE = 'contributor:support_forum:helper'
+AOA_CONTRIBUTOR_COHORT_CODE = 'contributor:aoa'
```
mythmon: I think I would make `contributor:kb` into `contributor:kb:english`. The way it is now makes me think that `contributor:kb:l10n` is a subset of `contributor:kb` (It isn't, right?). This is a minor issue though.

kitsune/kpi/cron.py
```diff
@@ -496,3 +496,92 @@ def survey_recent_askers():
...
+            # Sometimes None will be added to the cohort_users list, so remove it
+            if None in cohort_users:
+                cohort_users.remove(None)
+
+            cohort = Cohort.objects.get_or_create(kind=cohort_kind, start=cohort_range[0],
+                                                  end=cohort_range[1])[0]
+            cohort.size = len(cohort_users)
+            cohort.save()
+
+            for retention_range in ranges[i + 1:]:
```
mythmon: This creates `RetentionMetric` objects starting with the week after the cohort forms, right? So the cohort from 2015-10-19 through 2015-10-26 would get its first RetentionMetric starting 2015-10-26 through 2015-11-02?

I think this might make the API/charting this a little harder. I think making the first RetentionMetric the same week the cohort forms will make creating the charts easier, since then the chart only has to care about RetentionMetric objects, and not have to special case the first week for any given cohort.

Thoughts?

kitsune/kpi/cron.py
rehandalal: oh yeah, good call on the `cohort_kind, _`. I'll change that up.

rehandalal: And yeah I started off with thinking data migrations but then figured this was waaaaaay better

kitsune/kpi/cron.py
rehandalal: oh! definitely.

kitsune/kpi/cron.py
rehandalal: Makes sense. I will add that in!

kitsune/kpi/cron.py
rehandalal: ah yeah, probably a good call.

kitsune/kpi/cron.py
rehandalal: neat!

kitsune/kpi/cron.py
rehandalal: i love that! way simpler.

kitsune/kpi/models.py
rehandalal: yeah true, I should probably also rename it to `KB_ENUS_CONTRIBUTOR_COHORT_CODE`

kitsune/kpi/tests/test_cron.py
rehandalal: Oh yeah! I'm so used to the old create a profile, then use the user thing we used to have to do.

kitsune/kpi/tests/test_cron.py
rehandalal: Yeah writing them hurt my brain a little.

rehandalal:
updated ^

mythmon:
This looks good to me. I'd like @brittanystoroz to sign off on this data model since she will be the one building UIs with it, but r+ from me.

brittanystoroz:
r+ :tada:

PR #2777, opened 2016-01-20
[bug 1240780] Create API for retention data

rehandalal:
r?


brittanystoroz:
r+ :tada: assuming travis is happy

rehandalal:
@mythmon could you do a quick pass and make sure there's nothing wonky with my DRF stuff?

kitsune/kpi/urls_api.py
```diff
@@ -1,8 +1,13 @@
 from django.conf.urls import patterns, url
+from rest_framework import routers
 
 from kitsune.kpi import api
 
-urlpatterns = patterns(
+router = routers.SimpleRouter()
+router.register(r'api/2/cohort', api.CohortViewSet)
```
mythmon: I think having `api/2/cohort` here will do strange things. The User/Profile API in [`kitsune/users/urls_api.py`](https://github.com/mozilla/kitsune/blob/7e1d343a6d582c4df4f207f94ee17d296e23843c/kitsune/users/urls_api.py#L19) does this by `include()`ing the router in another url route, so it can mount it in the right place. I think that is a better way to do it.

rehandalal:
updated ^

did the rest of it look okay?

mythmon:
The rest looks good. r+

PR #2778, opened 2016-01-22
[Bug 1241970] Fix survey logic. Again.

mythmon:
*grumble*.

This logic gives us trouble historically. Please read [the bug](https://bugzilla.mozilla.org/show_bug.cgi?id=1241970) and [the previous bug](https://bugzilla.mozilla.org/show_bug.cgi?id=1215607) and compare the logic there with the logic here to make sure I got it right.

r?


rehandalal:
cool, lgtm. r+

PR #2780, opened 2016-01-29
Updated requirements to omit crashes

Glorf:
Updated to cryptography-1.1.2 (and its dependencies) that:
"Fixed a runtime error undefined symbol EC_GFp_nistp224_method that occurred with some OpenSSL installations."
This bug occurred to me while building on fedora-23

### DIFF
--- a/requirements/default.txt
+++ b/requirements/default.txt
@@ -32,8 +32,8 @@ celery==3.1.17
 # sha256: 95yMBNfrUAcbytZ_0j88EPq2xy1WhXrfhINngGhF1uU
 click==3.3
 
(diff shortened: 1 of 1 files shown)


safwanrahman:
@mythmon What do you think about it? 
Change log seems to be hare https://cryptography.io/en/latest/changelog/

mythmon:
When I deployed this to dev to test it, dev crashed. I think that we have an incompatible version of some library. Our server environments are pretty difficult to change, so unless this is a high priority (read: security issue), I don't think we should do this righ tnow.

PR #2781, opened 2016-02-01
[Bug 1238055] Copy django-authority into this repo.

mythmon:
1. Copy django-authority's Python package into the Kitsune repo
2. Make its tests pass
3. Make it pass our linter
4. Remove a couple instances of Django < 1.7 specific code
5. Add migrations for it

Step 5 is the most important one, as it should allow us to upgrade to Django 1.8 more easily. On my system having this package here results in Python picking up the correct version of django-authority, even if it is installed in the virtualenv. We'll need to verify this on the servers too, but we can deploy this without checking. If the servers don't use the correct version will have to get webop's help to remove the old library from the virtualenv.

r?

### DIFF
--- a/authority/AUTHORS
+++ b/authority/AUTHORS
@@ -1,0 +1,11 @@
+Jannis Leidel <jannis@leidel.info>
+Martin Mahner <martin@mahner.org>
+Diego Búrigo Zacarão <diegobz@gmail.com>
+James Pic
+Wes Winham <winhamwr@gmail.com>
+Kyle Gibson <kyle.gibson@policystat.com>
+Jason Ward <jason.ward@policystat.com>
+Travis Chase <http://www.supercodepoet.com/>
+Remigiusz Dymecki <remigiusz@gmail.com>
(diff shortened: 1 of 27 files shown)


safwanrahman:
I strongly believe that, we need to move into the latest version of pip from peep. So therefore we can destroy the old virtualenv in the deploy and create a new virtualenv and install the dependencies. As pip store the dependencies in the cache, it will not take much time to deploy.

As we don't recreate the virtualenv, it may raise an issue with version conflict.

mythmon:
This isn't really the place to discuss that. And anyways, the reason we can't delete the virtualenv is not only about the speed of it. Since our cron jobs run on the same machine that does the deploys, deleting the virtualenv will cause cron jobs to fail during deploys, which isn't ok. Having cached downloads would help, but not enough.

mythmon:
Travis failures were unrelated. This is passing now.

rehandalal:
yay! if travis is happy, so am i! r+

PR #2782, opened 2016-02-02
[Bug 1225371] Kpi retention chart

brittanystoroz:
Adds retention chart to the kpi dashboard page.


mythmon:
The linter errors here are my fault. If you rebase on master, that problem will go away.

kitsune/kpi/static/kpi/js/components/Chart.es6.js
```diff
@@ -0,0 +1,182 @@
...
+      axes: {
+        xAxis: {
+          labels: [],
+          labelOffsets: { x: 15, y: 23 }
+        },
+        yAxis: {
+          labels: [],
+          labelOffsets: { x: 5, y: 23 }
+        },
+        getPosition: (position, axis, index) => {
```
mythmon: This style for functions works fine. You can also use another new syntax for functions in objects which feels more natural to me, like this:

```javascript
let adder = {
  value: 5,
  add(x) {
    return x + this.value;
  }
}
```

Use whichever you like better.

kitsune/kpi/static/kpi/js/components/Chart.es6.js
```diff
@@ -0,0 +1,182 @@
...
+  }
+
+  // Render whatever pieces of the chart we can while waiting for data
+  preRender() {
+    // draw the container svg for the chart
+    this.dom.svg = d3.select(this.dom.graphContainer).append('svg')
+      .attr('width', this.width + this.margin.left + this.margin.right)
+      .attr('height', this.height + this.margin.top + this.margin.bottom)
+      .append('g')
+        .attr('transform', 'translate(' + this.margin.left + ',' + this.margin.top + ')');
```
mythmon: Template strings are cool:

```javascript
.attr('transform', `translate(${this.margin.left}px, ${this.margin.top}px)`);
```

kitsune/kpi/static/kpi/js/components/Chart.es6.js
```diff
@@ -0,0 +1,182 @@
...
+        .attr('x', function(d, i) {
+          return self.axes.getPosition('x', axis, i) + self.axes[axis].labelOffsets.x;
+        })
+        .attr('y', function(d, i) {
+          return self.axes.getPosition('y', axis, i) + self.axes[axis].labelOffsets.y;
+        })
+  }
+
+  setupLegend() {
+    let self = this;
```
mythmon: You shouldn't need this trick anymore, since fat-arrow functions (`x => x * 2`) keep the `this` value of their context.

kitsune/kpi/static/kpi/js/components/Chart.es6.js
```diff
@@ -0,0 +1,182 @@
...
+    self.dom.cohorts
+      .attr('width', self.width)
+      .attr('height', self.gridSize/2)
+      .attr('x', 0)
+      .attr('y', function(d, i) {
+        return i * self.gridSize/2;
+      })
+      .attr('class', function(d, i) {
+        return 'cohort-group ' + kindFilter;
+      });
```
mythmon: This can just be a constant, right? `.attr('class', 'cohort-group ' + kindfilter)`?

kitsune/kpi/static/kpi/js/components/Chart.es6.js
```diff
@@ -0,0 +1,182 @@
...
+      .data(filteredData);
+
+    self.dom.cohorts.enter().append('g');
+
+    self.dom.cohorts
+      .attr('width', self.width)
+      .attr('height', self.gridSize/2)
+      .attr('x', 0)
+      .attr('y', function(d, i) {
+        return i * self.gridSize/2;
```
mythmon: Arrow functions can make this a it nicer

```javascript
.attr('y', (d, i) => i * self.gridSize / 2)
```

kitsune/kpi/static/kpi/js/components/Chart.es6.js
brittanystoroz: So I don't need it everywhere in the methods, but when you get into the d3 stuff like setting x and y attributes, the `this` context changes to the d3 selection.

kitsune/kpi/static/kpi/js/components/Chart.es6.js
brittanystoroz: So this one I wrote as an arrow function because otherwise the `this` context gets messed up. The way it's written now, `this` represents the Chart (so I have access to `this.gridSize`). If I rewrite it the other way, the `this` context becomes the `axes` object rather than the Chart (which doesn't have a reference to the gridSize).

kitsune/kpi/static/kpi/js/components/Chart.es6.js
brittanystoroz: Would you prefer I save anything I need within the d3 methods to variables at the top of each function so I can avoid doing this trick?

kitsune/kpi/static/kpi/js/components/Chart.es6.js
mythmon: Ah, that makes sense. Sorry I missed that.

kitsune/kpi/static/kpi/js/components/Chart.es6.js
mythmon: Are you talking about things like `.style('fill', function(d, i) { return self.chartColors[i]; });`? If you write that as `.style('fill', (d, i) => this.chartColors[i])`, I'm pretty sure it works fine without self, based on [this jsbin](http://jsbin.com/vacinigowo/edit?html,js,output).

If you do need to get around `this` getting overridden, `self` is a fine way to do that. I just don't think it is needed in this case. I could be mistaken though.

brittanystoroz: Ah my mistake, you're right. Be gone `self`!

kitsune/kpi/templates/kpi/dashboard.html
```diff
@@ -9,6 +9,25 @@
 {% block content %}
   <article id="kpi-dash-rickshaw">
 
+    <section id="kpi-cohort-analysis"
+      data-url="/api/2/cohort">
+
+      <h2>{{_('Retention') }}</h2>
```
safwanrahman: small nit: should have a space at the lift side ``{{ _('Retention') }}

brittanystoroz:
@mythmon @rdalal this should be all fixed up. There will be a bit more refactoring to the `kpi.browserify.js` file once I add the retention chart to the new `/community/metrics` page but I'd like to save that for a new PR corresponding to that bug.

rehandalal:
This looks good to me... I assume you'll be dropping the dropdown to select the type of graph in the next PR once those other charts move to the community metrics page?

rehandalal:
r+

PR #2783, opened 2016-02-04
Make showfor more resilient to missing data.

mythmon:
Someone pointed out that the minimal KB view wasn't loading images. It turns out this is because showfor was trying to load non-existent data and crashing, taking down everything else with it (including lazy loading images).

This doesn't fix the issue of showfor data not being on the minimal page, but it does make showfor not crash and burn the page to the ground, which helps this particular issue.

@rehandalal r?


rehandalal:
wfm. r+

PR #2785, opened 2016-02-11
[Bug 1238055][Bug 1225606] Upgrade Django to 1.8.

mythmon:
I'm sorry about the giant diff. I'm sorry. You can see a lot of it here, but technically there are some files missing.

If you look at this on the cli, you want to do something like `git show --find-copies-harder --patience`. That produces a much more interesting diff.

The general changes:

* Update libraries
* Move all the templates from `templates/` to `/jinja2/`
* Move all the helpers from `helpers.py` to `templatetags/jinja_helpers.py`
* Rename some template stuff (like `{{ csrf() }}` to `{% csrf_token %}`)
* Adjust whitespace handling in `*.ltxt` files

There are some other smaller changes to deal with differences in the libraries that got updated. And some miscellaneous stuff.

r? and I'm sorry.


mythmon:
The tests now take over an hour to run. I don't really know why, and also that's terribly bad. I'm going to dig into it next week, but any insight anyone else has would be great.

safwanrahman:
From my local, the admin page can not be accessible from ``http://127.0.0.1:8000/admin``

safwanrahman:
The login page has this CSRF >> << HERE
firefox_screenshot_2016-02-14t13-07-19 565z

mythmon:
@safwanrahman Nice catch with the CSRF thing. I forgot about that.

As far as the admin not being available, it is working fine for me. Do you get any errors?

safwanrahman:
If you logout and login to the admin page, its redirecting you to ``http://127.0.0.1:8000/en-US/users/auth?next=%2Fadmin%2Flogin%2F%3Fnext%3D%2Fadmin%2F``

safwanrahman:
Seems like this issue get solved if anyone enter wrong username and password.
STR:

- run ``./manage.py flush`` to delete all the data
- then, create a superuser
- Try to go ``http://127.0.0.1:8000/admin``

mythmon:
Both old and new superusers work fine for me logging in to the admin. We won't be flushing the databases on the servers, so that's not really a worrisome issue. Did you run the migrations?

kitsune/kpi/api.py
```diff
@@ -335,23 +335,33 @@ def _start_date():
...
     """
     for label, qs in kwargs.iteritems():
-        yield dict((date(x['year'], x['month'], x.get('day', 1)),
-                   {label: x['count']}) for x in qs)
+        res = {}
+        # For each date mentioned in qs, sum up the counts for that day
+        # Note: days may be duplicated
+        for x in qs:
+            key = date(x['year'], x['month'], x.get('day', 1))
+            res.setdefault(key, {label: 0})[label] += x['count']
```
Osmose: Maybe a `defaultdict` instead?

kitsune/questions/cron.py
```diff
@@ -72,7 +72,8 @@ def auto_archive_old_questions():
 
         cursor = connection.cursor()
         cursor.execute(sql)
-        transaction.commit_unless_managed()
+        if not transaction.get_connection().in_atomic_block:
```
Osmose: I think what you want is to surround the cursor code with `with transaction.atomic()` rather than this, since this doesn't actually start a transaction.

kitsune/questions/tasks.py
```diff
@@ -66,7 +66,8 @@ def update_question_vote_chunk(data):
         """ % ids
     cursor = connection.cursor()
     cursor.execute(sql)
-    transaction.commit_unless_managed()
+    if not transaction.get_connection().in_atomic_block:
```
Osmose: Same here about using `transaction.atomic()`

kitsune/questions/cron.py
mythmon: I think that would change the semantics of what the code did.

kitsune/settings.py
```diff
@@ -586,15 +599,13 @@
 
 
 def JINJA_CONFIG():
```
Osmose: Is this even still used?

kitsune/sumo/static/sumo/css/print.css
```diff
@@ -24,7 +24,7 @@ body.html-ltr .container_12 {
     padding: 0;
     width: auto;
 }
-
+tap here to go to
```
Osmose: Whoops

kitsune/settings.py
willkg: Puente can extract the list of Jinja2 extensions to use from the old jingo style (i.e. this ^^^).

https://github.com/mozilla/puente/blob/master/puente/commands.py#L14

However, I'd recommend ditching it and adding the appropriate things to the PUENTE config.

Osmose:
`settings_local.py-dist` needs to be updated to remove some old settings that are unnecessary or misconfigured.

Other than that and the stuff I've mentioned, this looks code from a review perspective. I'm now trying to replicate the issues that @safwanrahman ran into.

willkg:
One thing to do if you haven't done it already is do a strings extract and merge and make sure that it doesn't change any of the msgids.

mythmon:
I haven't done that.

Progress on the "tests take 90 minutes" problem: They are nice and quick for me on my laptop. They are slow on @Osmose's laptop. They are slow on Travis. Osmose and Travis used MySQL and I'm using MariaDB. I tried switching Travis to MariaDB. The tests didn't go faster.

Osmose:
I was able to replicate issues logging in, but it was due to a misconfigured local cache. Once I got that sorted out (by just removing the `CACHES` setting from my local setting config to use the default in-memory cache) login to the admin worked fine.

willkg:
I spent some time looking at the "OMG! TESTS TAKE 90 MINUTES!". I can reproduce it on my machine. @mythmon couldn't. So we compared pastebins. I noticed he's doing `REUSE_STATIC=1`.

So then I wrote this script which simulates what travis does:

```
#!/bin/bash

# This fails if it's not there, but that's ok because all we want to do
# is guarantee there's no db.
mysql -e "drop database test_kitsune;"

# Set environment variables here.
# export FORCE_DB=1      # Force creation of a new db.
# export REUSE_DB=0      # Don't reuse the db.
# export REUSE_STATIC=1  # Reuse static assets.

# This is what travis runs with the minor addition of limiting it to a
# specific set of 8 tests so we don't wait all fucking day for an iteration.
python manage.py test \
  --noinput --logging-clear-handlers \
  --no-skip \
  --with-nicedots \
  kitsune/forums/tests/test_urls.py:BelongsTestCase
echo 'Booyahkasha!'
```

That limits it to just one test case which has 8 tests in it. Things are good if it takes 1.6s and bad if it takes 33s. Note that this is explicitly the time it takes to run the tests and doesn't include the time to destroy/create the db, process static files, walk to the pharmacy for nyquil, etc.

So... 1.6s good :smile:. 33s bad :angry:.

Then I fiddled with combinations of the environment variables and if I add `export REUSE_STATIC=1`, it always runs in 1.6s. :smile:

So I looked at that code:

https://github.com/mozilla/kitsune/blob/master/kitsune/sumo/tests/__init__.py#L55

It only kicks off when setting up the test environment. It doesn't kick off for every test. I verified that.

I noticed we upgraded django-pipeline from 1.4.7 to 1.6.4. Maybe because we're doing this with `call_command`, it's leaving behind "state" causing shit to take FOREVER to run?

With that theory, I changed the `call_command` to a subprocess call and now it takes 1.6s. :smile:

```
        if not getenv('REUSE_STATIC', 'false').lower() in ('true', '1', ''):
            # Collect static files for pipeline to work correctly--do this with
            # subprocess because collectstatic somehow retains emotional baggage
            # which causes all the tests to take FOREVER to run.
            cmdline = [sys.executable, 'manage.py', 'collectstatic', '--noinput']
            print 'Running %r' % cmdline
            subprocess.call(cmdline)
```

Now it takes 1.6s with or without the `export REUSE_STATIC=1` :smile:.

Someone should verify all that just in case I screwed something up somewhere and also to make sure it's not magic pixie dust on my machine.

willkg:
As a side note, I also upgraded to django-pipeline 1.6.5 which was released last week and includes the `PIPELINE_COLLECTOR_ENABLED = DONT FUCKING COLLECT FOR EVERY DAMN USE OF THE TAG` fix when `DEBUG=True`. It's nice. Fixes the EXCRUCIATING PAIN that is using django-pipeline for development in a vagrant environment. I highly recommend it.

mythmon:
Yes! @willkg you rock, that fixed the problem. Tests are still a little slow (17 minutes), but they are fast enough for me. There are still a few more things to do here.

- [ ] Put this on stage for a while.
- [ ] Inform QA what's up, and any particularly problematic areas. (Right now the only one I know of is the review flow for KB articles).
- [ ] Test out string extraction.

I think that's it. If I missed anything, let me know. I'm going to work on getting this deployed to stage now.

Osmose:
Oh yeah, I should probably be explicit. r+ :D

mythmon:
# LEEEEROOOOOYYYYY JENNNKINS!

PR #2788, opened 2016-02-22
[bug 1241661] Calculate CSAT metrics

rehandalal:
I'm still trying to figure out this campaign thing. But I figured I should push this so at least the front-end work can be unblocked.

r?

### DIFF
--- a/kitsune/kpi/cron.py
+++ b/kitsune/kpi/cron.py
@@ -1,3 +1,4 @@
+import json
 import operator
 from datetime import datetime, date, timedelta
 from functools import reduce
@@ -6,6 +7,7 @@ from django.conf import settings
 from django.db.models import Count, F, Q
 
 import cronjobs
+import requests
 from statsd import statsd
 
 from kitsune.customercare.models import Reply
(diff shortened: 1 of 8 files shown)


kitsune/kpi/cron.py
```diff
@@ -581,3 +586,85 @@ def is_in_cohort(u):
...
+        KB_L10N_CONTRIBUTORS_CSAT_METRIC_CODE: 0,
+    }
+
+    while more_pages:
+        response = requests.get(
+            'https://restapi.surveygizmo.com/v2/survey/{survey}'
+            '/surveyresponse?'
+            'filter[field][0]=datesubmitted'
+            '&filter[operator][0]=>=&filter[value][0]={start}+0:0:0'
+            'filter[field][1]=datesubmitted'
```
mythmon: Should this have an '&' at the beginning here?

kitsune/kpi/cron.py
```diff
@@ -17,7 +19,10 @@
     SUPPORT_FORUM_CONTRIBUTORS_METRIC_CODE, VISITORS_METRIC_CODE, SEARCH_SEARCHES_METRIC_CODE,
     SEARCH_CLICKS_METRIC_CODE, EXIT_SURVEY_YES_CODE, EXIT_SURVEY_NO_CODE,
     EXIT_SURVEY_DONT_KNOW_CODE, CONTRIBUTOR_COHORT_CODE, KB_ENUS_CONTRIBUTOR_COHORT_CODE,
-    KB_L10N_CONTRIBUTOR_COHORT_CODE, SUPPORT_FORUM_HELPER_COHORT_CODE, AOA_CONTRIBUTOR_COHORT_CODE)
+    KB_L10N_CONTRIBUTOR_COHORT_CODE, SUPPORT_FORUM_HELPER_COHORT_CODE, AOA_CONTRIBUTOR_COHORT_CODE,
+    CONTRIBUTORS_CSAT_METRIC_CODE, AOA_CONTRIBUTORS_CSAT_METRIC_CODE,
+    KB_ENUS_CONTRIBUTORS_CSAT_METRIC_CODE, KB_L10N_CONTRIBUTORS_CSAT_METRIC_CODE,
+    SUPPORT_FORUM_CONTRIBUTORS_CSAT_METRIC_CODE)
```
mythmon: Maybe all these codes should be in some sort of shared object? Something like this? Or do you think it is preferable to import them all like this?

```python
from kitsune.kpi import config
config.SUPPORT_FORUM_CONTRIBUTORS_CSAT_METRIC_CODE
```

mythmon:
This looks fine. That URL is sort of nasty, but I don't think there is any reasonable way to improve it. My above issue about a missing `&` stands, but besides that, I think this is good to go.

Does this need a new API for the front end? Or do existing APIs cover this already?

@brittanystoroz Does this look good to you?

kitsune/kpi/cron.py
rehandalal: oops, yes!

rehandalal: funny that it doesn't complain about this when i run the cronjob

rehandalal:
I was going to put the API stuff in another PR once this landed, but it probably makes sense to just include it here, so that's what I did. ^

mythmon:
This looks good to me. Lets do it. r+

PR #2790, opened 2016-02-23
[Bug 1246937] Fix AAQ event tracking

brittanystoroz:
This fixes the event tracking for the new AAQ flow as outlined [here](https://bugzilla.mozilla.org/show_bug.cgi?id=1246937). 

The submission failure tracking isn't implemented exactly as described:
* Authentication errors are being tracked with a simple 'registration failed' or 'login failed' message, under the `SPA User Auth` scope rather than the AAQ scope.
* Submission failures as a result of missing or incorrect data are being tracked, but not specific to the actual problem (i.e. 'no product selected'). The UI doesn't render the submit button if there is missing information, which makes submission failures of this nature much less likely. (That said, the UI could be improved to render helpful error messages rather than just hiding the submit button. But that is a separate issue.)

### DIFF
--- a/kitsune/questions/static/questions/js/components/TitleContentEditor.jsx
+++ b/kitsune/questions/static/questions/js/components/TitleContentEditor.jsx
@@ -107,7 +107,7 @@ class SuggestionItem extends React.Component {
       <li className={classString}>
         <a
           href={suggestion.url}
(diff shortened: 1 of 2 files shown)


mythmon:
This looks great, thanks!

PR #2791, opened 2016-02-23
[Bug 1241629] Use correct domain for jsi18n urls.

Osmose:
Two problems are being solved:

1. We were using the `javascript` domain instead of `djangojs` in a bunch of places for Javascript translations, due to missing these uses when we switched to Puente.
2. We were relying on Pipeline's `BUILD_ID_JS`, which no longer works, to cache-bust the JS translations. I've used the staticfiles URL function in an attempt to make this cache-busting work again, but testing it out locally is proving to be very difficult. We'll need to verify on a server that a cache-busting hash or whatever is added to the URL after it hits a server.

   If it ends up not working, this will still improve our current situations of translations being completely broken, and instead will simply freeze our translations at the the point we deploy.

### DIFF
--- a/kitsune/sumo/helpers.py
+++ b/kitsune/sumo/helpers.py
@@ -4,6 +4,7 @@ import re
 import urlparse
 
 from django.conf import settings
+from django.contrib.staticfiles.templatetags.staticfiles import static
 from django.core.urlresolvers import reverse as django_reverse
 from django.http import QueryDict
 from django.utils.encoding import smart_str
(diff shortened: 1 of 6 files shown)


mythmon:
So much :boom: in Travis.

mythmon:
From Travis: `Unknown command: 'compilejs18n'`

Osmose:
> From Travis: Unknown command: 'compilejs18n'

Whoops, missed the i.

PR #2792, opened 2016-02-23
Collect jsi18n files via staticfiles to get hashed filenames.

Osmose:
`STATICI18N_ROOT` causes the `compilejsi18n` command to write the jsi18n files to `/jsi18n` in the repo root. Then, adding that to `STATICFILES_DIRS` causes them to be collected during `collectstatic`, which saves them with the hashes we expect in their filenames.

I had to update the deploy script to collect the jsi18n files before running `collectstatic`, as well.

### DIFF
--- a/jsi18n/.gitignore
+++ b/jsi18n/.gitignore
@@ -1,0 +1,2 @@
+*
+!.gitignore
--- a/kitsune/settings.py
+++ b/kitsune/settings.py
@@ -362,6 +362,7 @@ STATIC_ROOT = path('static')
 STATIC_URL = '/static/'
 STATICFILES_DIRS = (
     path('bower_components'),
+    path('jsi18n'),  # Collect jsi18n so that it is cache-busted
 )
 STATICFILES_FINDERS = (
     'django.contrib.staticfiles.finders.FileSystemFinder',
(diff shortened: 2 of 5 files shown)


Osmose:
I deployed this on staging and JavaScript localization seems to be working: https://support.allizom.org/af/kpi/dashboard

Because this updates the deploy script itself, it _might_ break on the first push; I'm not entirely sure if the update script will pick up deploy script changes immediately. Break here is defined as an exception on every page that uses the base templates loading this file, which is bad. The issue is that the `static` helper fails when it can't find a static file that has been requested.

A second deploy should fix the issue. This might be acceptable. If it's not, then I think we we can do is add an extra commit that keeps the deploy and collection changes, but removes the change to generating the URLs in the template. Then, add another commit reverting back to the new URL generation via the `static` helper. That way, we can deploy only the collection changes but avoid calling the static helper. Then we deploy the latest commit, and things should work!

I can add those changes if we want them. Do we want them?

mythmon:
Changes to the deploy script definitely require two pushes. We've run into this before. That would mean about 10 minutes of effective downtime, which I'm not thrilled about.

It sounds like it is pretty easy to split this into two commits that we can deploy in two steps. Lets do that.

Osmose:
Alternate fix: Do not blow up when generating static URLs that don't exist.

I have implemented that.

mythmon:
This works for me locally. Are you happy with your testing on stage/dev? Is this ready to go out?

Osmose:
I'd like to test it on stage one more time after merging before we hit prod, but I'm otherwise happy with this.

mythmon:
Ok. The code here is good. Merge when you're happy with the deployment situation.

r+

PR #2793, opened 2016-02-24
[Bug 1206139] Reviews Pending e-mails subject is in English for all locales

MikkCZ:
I hope, this will fix it.


mythmon:
Nice catch!

PR #2794, opened 2016-02-25
Docs: Update multiple typo fixes

staranjeet:
There were many typos in the docs, which I found out while going through them. Let me know if I need to commit per file.


mythmon:
Thanks for the fixes!

PR #2795, opened 2016-02-26
Migrate functional UI tests from sumo-tests to kitsune

davehunt:
This copies the functional UI tests from the https://github.com/mozilla/sumo-tests repository, which will be deactivated once this is merged. This means these tests will be run for every commit, and allow the tests to be maintained alongside the application code base. This is part of an effort by Web QA to improve the visibility of these automated tests, and to bring them into the developer workflows.

At this time only three of the functional UI tests are passing in Travis CI. When running a local instance of kitsune I get many more tests passing, so I suspect that the instance running in Travis is not entirely representative of a fully running application. I've marked the passing tests as 'smoke', so we'll get notifications in GitHub if any of these tests regress. The full suite is also running but is allowed to fail - this should allow someone to investigate the cause of the failures and hopefully once whatever issues are addressed, we can add the 'smoke' marker to more and more tests.

I've renamed this suite from 'smoketests' as there are likely to be more than just smoke tests here. The full suite will continue to run as appropriate against dev/stage/prod but will run from this repository once this patch lands. I've also included a documentation update, which should help developers to run these tests locally.

/cc @rbillings @stephendonner @Osmose


docs/tests.rst
```diff
@@ -167,33 +167,91 @@ Here are a few tips for writing tests:
...
-They're located in the ``smoketests/`` directory. See the ``README.rst``
-file in that directory for how to set them up as well as other details.
+We can do more comprehensive front-end testing with the functional UI tests.
+They're located in the ``tests/functional`` directory.
 
+Installing dependencies
+-----------------------
 
-.. _tests-chapter-qa-test-suite:
+Followed the steps in :ref:`the installation docs <hacking-howto-chapter>`,
```
rbillings: Nit: s/b "Follow the steps"

scripts/travis/install.sh
```diff
@@ -12,12 +12,9 @@ echo "Install Python dependencies"
...
 
-# Installing dependencies for smoke tests
-if [[ $TEST_SUITE == "smoke" ]]; then
-  virtualenv venv_smoketests
-  source venv_smoketests/bin/activate
-  pip install -r smoketests/requirements.txt
-  deactivate
+# Installing dependencies for UI tests
+if [[ $TEST_SUITE == "ui" ]]; then
+  ./peep.sh install -r requirements/test.txt
```
safwanrahman: There was a discussion or something like that to install the smoketests dependencies in a separate virtualenv. I strongly believe that that would be better

mythmon:
@davehunt Awesome! I'm really glad to see this.

Looking at the errors you ran into, I think I know what is going on in a couple places. @safwanrahman discovered in #2721 that running the tests with a very small (virtual) screen causes strange errors, including ones like you're seeing. In particular, the site is redirecting to SSL, and for some reason changing the screen size prevents that. Crazy, but it was pretty easy to verify in #2721.

It may also be worth running the tests entirely with SSL using `runsslserver`, [described here](http://kitsune.readthedocs.org/en/latest/notes.html?highlight=runsslserver#about-support-api). This would produce a more prod-like environment, and might fix some problems.

scripts/travis/install.sh
mythmon: The main reason we had a different virtualenv before was that the Input smoke tests, which we copied everything from, were incompatible with with Input's normal requirements. If we can make the set of requirements all work together, that's fine. And CI will make sure it stays that way.

requirements/default.txt
```diff
@@ -209,6 +209,9 @@ MySQL-python==1.2.3c1
 # sha256: sboSwcRTI_2Glb5oznIoRWuG5RI08WcCfrvEzNqFLGE
 nose==1.1.2
 
+# sha256: REZqm8tW0uVodQ-RUE0SeMdOq7JZowWwbpdbh7UWNdo
```
safwanrahman: Is it possible to include in the test.txt?

docs/tests.rst
```diff
@@ -167,33 +167,91 @@ Here are a few tips for writing tests:
...
-=================
+Create test users
+-----------------
+
+Some of the tests require logging in as a administrator, and others require
+logging in as a user. To run these tests you will need to create accounts in
+the target environment. If you're running against a local instance of the
+application you can create these users by running the following script::
+
+  $ ./manage.py shell < ./scripts/create_user_and_superuser.py
```
mythmon: Django has a scripts functionality built in that we should use instead of piping to `shell`. It would be `./manage.py runscript create_user_and_superuser`, and `scripts/create_user_and_superuser.py` would need some minor changes. You can use [`scripts/get_media.py`](https://github.com/mozilla/kitsune/blob/80990a688ae66681922d2182dd429b1336d324a9/scripts/get_media.py) as an example.

requirements/default.txt
```diff
@@ -209,6 +209,9 @@ MySQL-python==1.2.3c1
...
+nose-exclude==0.4.1
```
mythmon: I suspect this should either got in `dev.txt` or `test.txt`.

scripts/travis/install.sh
safwanrahman: In future the requirements may be incompatible and that may raise issues. I strongly believe that it will be better for long run if this requirements are installed in different virtualenv and so we can debug it easily in future if any failure.

mythmon: That being said, if we don't want to do the work of making the ui tests stay compatible with the main requirements, making a separate virtualenv is a great way to do that.

requirements/default.txt
mythmon: I see now why this can't go in `test.txt`, since it is used to exclude tests during normal test runs. It should go in `dev.txt`.

requirements/default.txt
mythmon: See my comment on the next line. This is here so that we can configure normal test runs to ignore the pytest tests. It should be in `dev.txt`.

scripts/travis/install.sh
davehunt: This will potentially prevent us from using `pytest-django` for the UI tests in the future. I suspect this is worth looking into, as it is likely to allow much more control over test setup, including the creation of users and test data.

requirements/default.txt
davehunt: Sure, I originally put it alongside nose, as if nose is present and in use, we probably want to control the exclusion of the pytests.

docs/tests.rst
davehunt: This was taken directly from @safwanrahman's work on the previous attempt to migrate the tests. I'll take a stab at making the suggested changes, but if I hit issues I suspect it may be better to follow up with an improvement.

davehunt:
> Looking at the errors you ran into, I think I know what is going on in a couple places. @safwanrahman discovered in #2721 that running the tests with a very small (virtual) screen causes strange errors, including ones like you're seeing. In particular, the site is redirecting to SSL, and for some reason changing the screen size prevents that. Crazy, but it was pretty easy to verify in #2721.

I haven't changed the screen size, it's still set to 1280x1024x16 in https://github.com/mozilla/kitsune/pull/2795/files#diff-0f775ad83429f23374705da3dafdbd63R7

> It may also be worth running the tests entirely with SSL using runsslserver, described here. This would produce a more prod-like environment, and might fix some problems.

I'd like to suggest pushing this investigation to a separate issue. This is already a reasonably large patch, and I'd like to encourage follow ups that address the failing tests and increase the size of the smoke suite.

mythmon:
I'm ok with pushing this off to future work. Even this is an improvement over what we have now.

docs/tests.rst
davehunt: When attempting to change this I hit the following exception, which I can only guess is safely ignored in the current version:

```
$ ./manage.py runscript create_user_and_superuser
Exception while running run() in 'scripts.create_user_and_superuser'
  File "./manage.py", line 32, in <module>
    execute_from_command_line(sys.argv)
  File "/home/dhunt/kitsune/venv/local/lib/python2.7/site-packages/django/core/management/__init__.py", line 385, in execute_from_command_line
    utility.execute()
  File "/home/dhunt/kitsune/venv/local/lib/python2.7/site-packages/django/core/management/__init__.py", line 377, in execute
    self.fetch_command(subcommand).run_from_argv(self.argv)
  File "/home/dhunt/kitsune/venv/local/lib/python2.7/site-packages/django/core/management/base.py", line 288, in run_from_argv
    self.execute(*args, **options.__dict__)
  File "/home/dhunt/kitsune/venv/local/lib/python2.7/site-packages/django/core/management/base.py", line 338, in execute
    output = self.handle(*args, **options)
  File "/home/dhunt/kitsune/venv/local/lib/python2.7/site-packages/django_extensions/management/commands/runscript.py", line 147, in handle
    run_script(mod, *script_args)
  File "/home/dhunt/kitsune/venv/local/lib/python2.7/site-packages/django_extensions/management/commands/runscript.py", line 74, in run_script
    mod.run(*script_args)
  File "/home/dhunt/kitsune/scripts/create_user_and_superuser.py", line 33, in run
    enter_data(data)
  File "/home/dhunt/kitsune/scripts/create_user_and_superuser.py", line 26, in enter_data
    create_user(username=user['username'], password=user['password'], email=user['email'])
  File "/home/dhunt/kitsune/scripts/create_user_and_superuser.py", line 18, in create_user
    User.objects.create_user(username=username, password=password, email=email)
  File "/home/dhunt/kitsune/venv/local/lib/python2.7/site-packages/django/contrib/auth/models.py", line 183, in create_user
    **extra_fields)
  File "/home/dhunt/kitsune/venv/local/lib/python2.7/site-packages/django/contrib/auth/models.py", line 178, in _create_user
    user.save(using=self._db)
  File "/home/dhunt/kitsune/venv/local/lib/python2.7/site-packages/django/db/models/base.py", line 589, in save
    force_update=force_update, update_fields=update_fields)
  File "/home/dhunt/kitsune/venv/local/lib/python2.7/site-packages/django/db/models/base.py", line 626, in save_base
    update_fields=update_fields, raw=raw, using=using)
  File "/home/dhunt/kitsune/venv/local/lib/python2.7/site-packages/django/dispatch/dispatcher.py", line 198, in send
    response = receiver(signal=self, sender=sender, **named)
  File "/home/dhunt/kitsune/kitsune/search/models.py", line 232, in update
    maybe_call_method(instance, kw.get('raw'), 'index_later')
  File "/home/dhunt/kitsune/kitsune/search/models.py", line 226, in maybe_call_method
    obj = instance_to_indexee(instance)
  File "/home/dhunt/kitsune/kitsune/users/models.py", line 312, in get_profile
    return Profile.objects.get(user=u)
  File "/home/dhunt/kitsune/venv/local/lib/python2.7/site-packages/django/db/models/manager.py", line 92, in manager_method
    return getattr(self.get_queryset(), name)(*args, **kwargs)
  File "/home/dhunt/kitsune/venv/local/lib/python2.7/site-packages/django/db/models/query.py", line 351, in get
    num = len(clone)
  File "/home/dhunt/kitsune/venv/local/lib/python2.7/site-packages/django/db/models/query.py", line 122, in __len__
    self._fetch_all()
  File "/home/dhunt/kitsune/venv/local/lib/python2.7/site-packages/django/db/models/query.py", line 966, in _fetch_all
    self._result_cache = list(self.iterator())
  File "/home/dhunt/kitsune/venv/local/lib/python2.7/site-packages/django/db/models/query.py", line 265, in iterator
    for row in compiler.results_iter():
  File "/home/dhunt/kitsune/venv/local/lib/python2.7/site-packages/django/db/models/sql/compiler.py", line 701, in results_iter
    for rows in self.execute_sql(MULTI):
  File "/home/dhunt/kitsune/venv/local/lib/python2.7/site-packages/django/db/models/sql/compiler.py", line 787, in execute_sql
    cursor.execute(sql, params)
  File "/home/dhunt/kitsune/venv/local/lib/python2.7/site-packages/django/db/backends/utils.py", line 81, in execute
    return super(CursorDebugWrapper, self).execute(sql, params)
  File "/home/dhunt/kitsune/venv/local/lib/python2.7/site-packages/django/db/backends/utils.py", line 65, in execute
    return self.cursor.execute(sql, params)
  File "/home/dhunt/kitsune/venv/local/lib/python2.7/site-packages/django/db/utils.py", line 94, in __exit__
    six.reraise(dj_exc_type, dj_exc_value, traceback)
  File "/home/dhunt/kitsune/venv/local/lib/python2.7/site-packages/django/db/backends/utils.py", line 65, in execute
    return self.cursor.execute(sql, params)
  File "/home/dhunt/kitsune/venv/local/lib/python2.7/site-packages/django/db/backends/mysql/base.py", line 129, in execute
    return self.cursor.execute(query, args)
  File "/home/dhunt/kitsune/venv/local/lib/python2.7/site-packages/MySQLdb/cursors.py", line 173, in execute
    self.errorhandler(self, exc, value)
  File "/home/dhunt/kitsune/venv/local/lib/python2.7/site-packages/MySQLdb/connections.py", line 36, in defaulterrorhandler
    raise errorclass, errorvalue
django.db.utils.OperationalError: (1054, "Unknown column 'users_profile.csat_email_sent' in 'field list'")
```

davehunt:
I've pushed the two easy tweaks from the review comments. Thanks @safwanrahman and @mythmon for the feedback!

mythmon:
Ok. Lets push the harder stuff off to the future.

@davehunt Can you make sure the remaining tasks get recorded somewhere? My preference would be bugs in the Sumo product.

docs/tests.rst
davehunt: Raised as [bug 1251694](https://bugzilla.mozilla.org/show_bug.cgi?id=1251694).

davehunt:
I've raised [bug 1251698](https://bugzilla.mozilla.org/show_bug.cgi?id=1251698) for investigating the failures and increasing the smoke test suite.

PR #2797, opened 2016-02-29
[Bug 1225374] Create retention chart for community metrics page

brittanystoroz:
There's kind of a lot in here, but if you take it commit-by-commit the changes should be pretty readable. This adds a community metrics page as outlined [here](https://bugzilla.mozilla.org/show_bug.cgi?id=1225374) with the retention chart.

A couple of notes:

* I took the drop-down menu out of the retention chart on the KPI dashboard, but left it in the community metrics page. We're going to be adding the CSAT charts to this page as well, so I think it makes sense to use a drop-down to filter the charts rather than breaking each section out into their own graph. If anyone has a problem with this when we go over it during the check-in meeting I can adjust. 
* Right now there's some extra unnecessary dependencies being loaded on the community metrics page (mainly the rickshaw stuff), but those libs will be needed anyway when we add the CSAT charts so they'll be put to use.
* I wasn't sure where to add a reference to the community metrics page on the Community Hub index. Right now it's literally just a link under 'Top Contributors' and looks pretty lame. I played around with adding a heading and a chart thumbnail (just an image) that would look something like this: (Thoughts?)
community-hub


kitsune/community/templates/community/metrics.html
```diff
@@ -0,0 +1,30 @@
...
+{% block content %}
+  <article id="kpi-dash">
+    <h1>{{ _('Contributor Metrics') }}</h1>
+      <section id="kpi-cohort-analysis" data-url="/api/2/cohort">
+      <h2>{{ _('Retention') }}</h2>
+      <div id="retention-chart" class="graph" data-chart-type="d3" data-slug="retention">
+        <div class="inline-controls">
+          <div>
+            <select id="toggle-cohort-type">
+              <option value="contributor">{{ _('Contributors') }}</option>
```
rehandalal: I think we can ditch this option

rehandalal:
With the exception of the one thing I mentioned this lgtm.

brittanystoroz:
@rehandalal updated

rehandalal:
hawt. r+

PR #2804, opened 2016-03-04
Fix share link escaping.

Osmose:
No bug number because I didn't bother filing after someone reported it on IRC. 

r?


safwanrahman:
https://bugzilla.mozilla.org/show_bug.cgi?id=1253563 Bug!

kitsune/wiki/tests/test_templates.py
```diff
@@ -415,6 +415,16 @@ def test_document_with_fallback_locale(self):
         # Check that content is available in bn-BD
         eq_(pq(trans_doc.html)('div').text(), doc('#doc-content div').text())
 
+    def test_document_share_link_escape(self):
```
safwanrahman: Great that a test has been added. It was bothering many times before.

PR #2805, opened 2016-03-04
Include non-supported locales in robots.txt

mythmon:
Webops noticed that robots are hitting urls for locales that get redirected. This should stop that.

r?


Osmose:
r+, will merge when Travis is happy

PR #2806, opened 2016-03-05
[Bug 1253743] Fix Weekly Solution API

safwanrahman:
I have changed the logic to optimize the code. I strongly believe the code is optimized now than before.
Also, re-factor the test so hope that this issue will not raise again.
**Note: I dont have any Idea why the previous code was raising issues**
@mythmon @Osmose r?

### DIFF
--- a/kitsune/users/tests/test_api.py
+++ b/kitsune/users/tests/test_api.py
@@ -243,9 +243,9 @@ class TestUserView(TestCase):
         top_answer_number = 15
         for i in range(12):
             user = UserFactory()
-            [SolutionAnswerFactory(creator=user) for i in range(top_answer_number)]
+            SolutionAnswerFactory.create_batch(top_answer_number, creator=user)
             user_info_list.append((user.username, top_answer_number))
-            top_answer_number = top_answer_number - 1
+            top_answer_number -= 1
 
         res = self.client.get(reverse('user-weekly-solutions'))
         eq_(res.status_code, 200)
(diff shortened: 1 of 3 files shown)


kitsune/users/tests/test_api.py
```diff
@@ -229,21 +230,33 @@ def test_generated_users_tagged(self):
...
+        res = self.client.get(reverse('user-weekly-solutions'))
+        eq_(res.status_code, 200)
+        eq_(len(res.data), 0)
+
+        # Check that the data about the contributors is showing currectly
+        user_info_list = []  # Info list with username and their number of solutions
+        top_answer_number = 15
+        for i in range(12):
+            user = UserFactory()
+            [SolutionAnswerFactory(creator=user) for i in range(top_answer_number)]
```
mythmon: Factories have a method that makes this simpler:

```python
SolutionAnswerFactory.create_bulk(top_answer_number, creator=user)
```

kitsune/users/tests/test_api.py
```diff
@@ -229,21 +230,33 @@ def test_generated_users_tagged(self):
...
+            user_info_list.append((user.username, top_answer_number))
+            top_answer_number = top_answer_number - 1
```
mythmon: This is probably better written as `top_ansewr_number -= 1`.

mythmon:
This looks a lot faster, thanks! I'm not sure why the code was written the way it was before, but I think this should work the same way. I have a couple nits, but then this looks good to go.

safwanrahman:
@mythmon updated. :+1:

mythmon:
The tests seem sound, and an improvement that should catch the issue, and this should certainly be quicker.

r+

PR #2811, opened 2016-03-08
Rename "xx" test locale to "xx_testing" to match locales repo

mythmon:
White trying to figure out [bug 1254199](https://bugzilla.mozilla.org/show_bug.cgi?id=1254199), I noticed that our testing locale got renamed in SVN. This makes all our systems correctly refer to the locale as `xx_testing` instead of `xx`.

I suspect this is actually causing problems in prod, since the `xx` locale is suddenly accessible on the site. It seems to be very error prone. I hope this will fix that too.

### DIFF
--- a/docs/localization.rst
+++ b/docs/localization.rst
@@ -381,10 +381,10 @@ Run::
 
 It'll extract all the strings, create a ``.pot`` file, then create a
 Pirate translation of all strings. The Pirate strings are available in
-the xx locale. After running the ``test_locales.sh`` script, you can
-access the xx locale with:
+the xx_testing locale. After running the ``test_locales.sh`` script, you can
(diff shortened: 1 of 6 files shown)


safwanrahman:
If SVN repo is the issue, I strongly believe the SVN repo name should be changed rather than this locale name in our code. the locale name seems too long.
We also need to change a lot of changes in order to change the test locale name

safwanrahman:
and according to [kitsune.sumo.models.LocaleField](https://github.com/mozilla/kitsune/blob/master/kitsune/sumo/models.py#L42) it should be maximum 7 character. So having this big locale name should fail surely!

mythmon:
That's a good point. I'll explore another way to get this fixed.

PR #2812, opened 2016-03-08
[Bug 1253685] Move community/metrics template to jinja2 directory

mythmon:
r?


safwanrahman:
Is it possible to also test if the correct template is being loaded?

mythmon:
It is, sort of (this got more complicatd when we switching from jingo to jinja2). The test I've added here is a style that I find very useful, where it does a very minimal amount of testing. these tests are great canaries, that will fail when we get blatant errors like this.

The test you describe sounds like a decent test to have, but it would not have caught the error this pr is fixing, and it is more work.

mythmon:
Actually, on more consideration, I mispoke: that test would have also caught the error, but only because the view would have crashed anyways. So it wouldn't function any differently than the test I wrote.

safwanrahman:
I understand.I understand.
I believe this should get r+

PR #2814, opened 2016-03-10
[bug 1253686] Handle no total_pages from SurveyGizmo API

rehandalal:
small r?

### DIFF
--- a/kitsune/kpi/cron.py
+++ b/kitsune/kpi/cron.py
@@ -632,7 +632,7 @@ def calculate_csat_metrics():
             timeout=300)
 
         results = json.loads(response.content)
-        total_pages = results['total_pages']
+        total_pages = results['total_pages'] if 'total_pages' in results else 1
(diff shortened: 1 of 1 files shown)


kitsune/kpi/cron.py
```diff
@@ -632,7 +632,7 @@ def calculate_csat_metrics():
             timeout=300)
 
         results = json.loads(response.content)
-        total_pages = results['total_pages']
+        total_pages = results['total_pages'] if 'total_pages' in results else 1
```
safwanrahman: I think its better written as
``results.get('total_pages', 1)``
What do you think?

mythmon:
I think @safwanrahman is right. `.get(key, default_val)` is more idiomatic and easier to read. Besides that, this should fix the problem, so r+wc.

rehandalal:
fixed ^

safwanrahman:
r+

PR #2816, opened 2016-03-11
[Bug 1254690] Fix escaping of share link on mobile articles

mythmon:
This blatantly steals the same solution used in #2804.

@Osmose r?

### DIFF
--- a/kitsune/wiki/tests/test_templates.py
+++ b/kitsune/wiki/tests/test_templates.py
@@ -438,6 +438,16 @@ class MobileArticleTemplate(MobileTestCase):
         eq_(200, response.status_code)
         assert template_used(response, 'wiki/mobile/document.html')
 
+    def test_document_share_link_escape(self):
+        """Ensure that the share link isn't escaped."""
+        r = ApprovedRevisionFactory(
+            content='Test',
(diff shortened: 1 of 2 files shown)


kitsune/wiki/tests/test_templates.py
```diff
@@ -438,6 +438,16 @@ def test_document_view(self):
         eq_(200, response.status_code)
         assert template_used(response, 'wiki/mobile/document.html')
 
+    def test_document_share_link_escape(self):
+        """Ensure that the share link isn't escaped."""
+        r = ApprovedRevisionFactory(
+            content='Test',
+            document__share_link='https://www.example.org',
+        )
+        response = self.client.get(r.document.get_absolute_url())
```
safwanrahman: is it going fetching the mobile page template? how?

safwanrahman: **"going to fetch"

mythmon: Yes, it will, since this is a MobileTestCase, which makes sure that the mobile templates are used.

safwanrahman:
strongly believe this should get r+

Osmose:
Late r+

PR #2817, opened 2016-03-16
[Bug 1246660] Localization content saved automatically

safwanrahman:
I have thought and found that, its a better way to save the auto saving content into ``localstorage``.  If we do save it at our server with the draft API, some contributors may have problem as it will consume much data.
small javascript patch. As I am not that expert in javascript, please let me know where can be improved.
@mythmon r?

### DIFF
--- a/kitsune/sumo/static/sumo/js/wiki.js
+++ b/kitsune/sumo/static/sumo/js/wiki.js
@@ -682,6 +682,7 @@
   function initAutoSave() {
     var key = getAutoSaveKey(),
       autoSavedContent = localStorage.getItem(key),
+      time = localStorage.getItem('time' + key),
       contentField = document.querySelector('#id_content');
 
     // Check if the syntex highlighting in enabled
@@ -699,7 +700,10 @@
     }
 
     function save(text) {
+      var time = new Date();
+
       localStorage.setItem(key, text);
+      localStorage.setItem('time' + key, time);
     }
 
     // As the normal editor does not depend on full page load,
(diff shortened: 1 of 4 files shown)


brittanystoroz:
Right now the UI of this isn't super intuitive -- it seems to conflict a bit with the 'save draft' button. For example, if I have a saved draft (from clicking the 'Save Draft' button), I get a purpley notification bar at the very top of the page asking me to restore or discard my draft.  If I also happened to have a more recent, unsaved version in localstorage, I'm completely unaware of that unless I scroll all the way down to where the textbox of content is, where I can see the yellow bar that says I have 'unsaved content'. If I had chosen to restore my saved draft in the purple box, I wouldn't expect to see another yellow box telling me I still had unsaved content that I could restore/discard.

I understand why this works the way it does right now, but a typical user isn't going to know the difference between a draft saved to the server versus a draft stored in localstorage. I think the UI needs some improvements before this can be landed. Some initial ideas:

- Put both notification boxes for an unsaved draft and unsaved localstorage copy near each other and have them styled in the same way. This will ensure users know they may have conflicting, unsaved versions of their draft before they choose to restore or discard them.
- Put a date/timestamp on the notification for an unsaved localstorage copy.
- Allow users to see a diff of their draft & the localstorage version so they can decide which one they want to restore.

brittanystoroz:
Additionally, if a user chooses to discard their saved draft, they should not have the option to restore the copy in localstorage. Restoring the copy in localstorage would re-introduce any article changes from the draft they just chose to discard.

safwanrahman:
@brittanystoroz Updated :cool:

Osmose:
I agree with @brittanystoroz 's feedback and I think that, as designed, this feature is too confusing for users and doesn't provide a huge benefit over things like [Lazarus](https://addons.mozilla.org/en-US/firefox/addon/lazarus-form-recovery/). See [my comment on the bug](https://bugzilla.mozilla.org/show_bug.cgi?id=1246660#c5).

Thanks for the PR, @safwanrahman, but I don't think we'll accept this right now.

safwanrahman:
@Osmose This issue is on discussion and we are trying to improve the feature. Does it a better way to close the Pull Request?

Osmose:
I'm suggesting that the feature is not something we want at all, which is why I closed the PR (I talked about this with @brittanystoroz over IRC before doing so and she agreed). The issues I have around having two different backup locations isn't really something that can be improved by updating this patch, and I don't want to keep PRs open on the repo that we're not going to accept, so I closed it. Sorry. :(

safwanrahman:
@Osmose I strongly cant accept to close the PR while the discussion has not finished in the bug. I dont know whats the convention, but as the bug is not closed yet, I dont think the PR should be closed anyway.
Have the decision has been made in the bug that this feature can not be implemented?

Osmose:
Sorry, I worded my last comment a little wrong.

Generally, bugs are for discussion about a feature, while PRs are for discussion of a specific implementation of a feature. In this case, the bug is for the feature of automatically saving localizations while localizers are working on it, and this PR is your specific implementation of that feature.

I propose that the feature itself is not something we want, and that discussion is still happening in the bug. I could easily be wrong on that. But regardless of that, the dev team has decided that the specific implementation of automatic saving that uses local client storage alongside the existing draft-saving feature is not one that we'll accept. And since this PR is about that specific implementation, we've closed it. 

If the bug discussion ends up with the decision that the feature should exist, then a different implementation will have to be written and submitted as a separate PR. It could even reuse some code from this PR. But leaving it open only serves to clutter up the list of pending PRs that need review and work that we will eventually merge.

So, to be perfectly clear, this PR closing doesn't mean the bug itself has been rejected. But it does mean the implementation in this PR has been. I appreciate the work you put in to it: Thank you! But it's fundamentally not an implementation that we're willing to merge.

safwanrahman:
Thanks @Osmose for clearing. I have understand the fact.

PR #2818, opened 2016-03-17
Updating pytest

stephendonner:


requirements/test.txt
```diff
@@ -6,9 +6,8 @@ execnet==1.4.1
 # sha256: plAZY8cl_CVU2r_s6K6aj7XhScCsCkL9KwLFwcV_wRQ
 py==1.4.31
 
-# sha256: f60wfCePxbGNJ9V6zO0x6euZ90zMyBccMJhi7eu7orc
-# sha256: 1anD2zUVq5WQK6CMCdHpQaIaRA3C1l8xusAmZ1CKnvw
-pytest==2.7.3
+# sha256: b61TzL8JA8aduT1nuD31IIGLBsdZftioQHvF_f_V5A4
```
Osmose: The rest of the requirements have two hashes: one for the source package and one for the wheel. You can use a tool like [hashin](https://github.com/peterbe/hashin) to easily add both hashes to the requirements file.

Osmose:
Test failure is odd and doesn't look related; I'm rerunning them.

requirements/test.txt
safwanrahman: I think we does not use wheel.

Osmose:
Hooray! Tests work!

Add that second hash and this is good to merge!

requirements/test.txt
Osmose: Actually pip prefers wheels over source distributions, I'm pretty sure. It will work without the hash because pip will fall back to the source distribution, but if there's a wheel available pip will use that first.

safwanrahman: Yap. But we had a decission to not use wheel in kitsune.

safwanrahman: check this
https://github.com/mozilla/kitsune/blob/master/peep.sh

Osmose: Huh, you're right. I wonder why all our packages still have two hashes then?

Oh well!

PR #2819, opened 2016-03-18
[Bug 1257606] Show always current copyright year in the site footer

MikkCZ:


safwanrahman:
I have used ``{% now "Y" %}`` in the django template loader. dont know if it also works in jinja2 template.

MikkCZ:
Does not for me. Anyway I decided to use a little more generic function, so it can be better reused. If the previous [single-use](https://github.com/MikkCZ/kitsune/commit/3984286c97c733d4813be5ed4efc410201b9e3d9) was better, np to revert back.

MikkCZ:
> FAIL: kitsune/wiki/tests/test_templates.py:NewRevisionTests.test_warning_showing_while_new_revision

@safwanrahman Tests failed on something, which seems unrelated to me. For the previous commit version with a single-use function it [passed](https://travis-ci.org/mozilla/kitsune/builds/116942525).

kitsune/sumo/templatetags/jinja_helpers.py
```diff
@@ -456,6 +456,11 @@ def static(path):
         return ''
 
 
+@library.global_function
+def datetime_now_strftime(format):
```
safwanrahman: The function name looks better as ``datetime_now(format)``

safwanrahman:
The Test fail seems unrelated. Lets someone from the sumodev team restart it!

MikkCZ:
The test is passing now again. ;)

r?

safwanrahman:
Good Job :+1:
r+
This should get merged.

kitsune/sumo/templatetags/jinja_helpers.py
```diff
@@ -456,6 +456,11 @@ def static(path):
         return ''
 
 
+@library.global_function
+def datetime_now(format):
```
Osmose: There's already a `datetimeformat` template function earlier in this file. You could rename this to just `now` and have it return the datetime, and then use the existing template function to format it.

Osmose:
This looks good besides the format logic duplication. Get that sorted and I think this will be good. Nice work!

MikkCZ:
Using `datetimeformat` there are a lot of test failures cased by

> AttributeError: 'WSGIRequest' object has no attribute 'session'

or

> AttributeError: 'RequestFactory' object has no attribute 'session'

Is there something special about how these work in the tests? Running the server and looking at the footer, it works.

mythmon:
There is a problem here that is probably left over from the Django 1.8 upgrade that prevents formatting dates in templates. That's why the tests are failing.

MikkCZ:
Hmm, so wait until the Django update thing is repaired, or fix this in the most naive way replacing `datetimeformat` functionality with `strftime`?

mythmon:
If you're feeling adventurous, you could probably fix the problem with the helper yourself. The problem is in [this function](https://github.com/mozilla/kitsune/blob/master/kitsune/sumo/templatetags/jinja_helpers.py#L207-L272), and the traceback is 

```

  File "/home/travis/build/mozilla/kitsune/kitsune/sumo/tests/test_views.py", line 45, in test_deprecated_redirect

    resp = deprecated_redirect(req, url='home')

  File "/home/travis/build/mozilla/kitsune/kitsune/sumo/views.py", line 113, in deprecated_redirect

    'dest': dest, 'proto': proto, 'host': host})

  File "/home/travis/build/mozilla/kitsune/kitsune/sumo/monkeypatch.py", line 128, in _more_info

    resp = fun(request, template_name, *args, **kwargs)

  File "/home/travis/virtualenv/python2.7.9/lib/python2.7/site-packages/django/shortcuts.py", line 67, in render

    template_name, context, request=request, using=using)

  File "/home/travis/virtualenv/python2.7.9/lib/python2.7/site-packages/django/template/loader.py", line 99, in render_to_string

    return template.render(context, request)

  File "/home/travis/virtualenv/python2.7.9/lib/python2.7/site-packages/django_jinja/backend.py", line 105, in render

    return self.template.render(context)

  File "/home/travis/virtualenv/python2.7.9/lib/python2.7/site-packages/jinja2/environment.py", line 891, in render

    return self.environment.handle_exception(exc_info, True)

  File "/home/travis/build/mozilla/kitsune/kitsune/sumo/jinja2/sumo/deprecated.html", line 3, in top-level template code

    {% set refresh = (dest, 10) %}

  File "/home/travis/build/mozilla/kitsune/kitsune/sumo/jinja2/base.html", line 198, in top-level template code

    {% trans year=datetimeformat(now(), 'year') %}

  File "/home/travis/build/mozilla/kitsune/kitsune/sumo/templatetags/jinja_helpers.py", line 229, in datetimeformat

    if 'timezone' not in request.session:

AttributeError: 'WSGIRequest' object has no attribute 'session'
```

Making that helper safe when no session exists will be helpful to Sumo in general, and it might be a while until anyone else gets to it.

On the other hand, all you need is the year, so just printing that out with `strftime` would be fine. Your choice.

safwanrahman:
I think the problem is caused because the test depends on ``RequestFactory()``. As its a mock request and isolated, it does not go with the middleware class ``session``. So no ``session`` object is inserted there. Therefore the issue is raised issue.

I have investigate this issue and it seems that, the ``base.html`` is a independent template that is used in every template. So while testing, the template is loaded with ``render_to_string`` with a mock request.(``RequestFactory()``).
As the request does not go with any middleware class, some information must be missing in the request like ``session``, ``user`` etc. So for fixing the issue, there can be three approach:

- Upgrade the test so the ``user`` and ``session`` middleware content is passed there

- Make the base template like before approach. Like with ``strftime``

- Make the helpers to work without the ``session`` and ``user``

MikkCZ:
Nice. I will wait until your PR is merged and check if it fixes this.

safwanrahman:
@mythmon I think it can be merged now.

mythmon:
Looks good, thanks @MikkCZ and @safwanrahman for this!

PR #2820, opened 2016-03-21
[Bug 1258268] Minimal theme footer includes double-escaped share link

MikkCZ:


MikkCZ:
r?

safwanrahman:
This should have a test so in futurevit will not broken!

MikkCZ:
Part of some UI test? Or are there any template specific tests?

safwanrahman:
This should be in template test

Osmose:
I added a test for this same issue on the non-minimal template: https://github.com/mozilla/kitsune/commit/76719c695abdb640e78077c1fa9734dbb7a654a7#diff-f96d1b40b0f6f4785dcc421f2c0e35d6R418

Basically add another test just like that but using the minimal template. Otherwise, this fix looks fine to me. Nice work!

(Make sure to leave a comment once you've added the test so I know to check it out.)

MikkCZ:
I've added a template test case. Btw. is there any reason, why is minimal treated as view instead of template, and the normal template uses class `wiki-doc`, when mobile and minimal id `wiki-doc`?

MikkCZ:
r?

safwanrahman:
I am going to review this tomorrow if anyone dont do it in the mean time.

kitsune/sumo/tests/__init__.py
```diff
@@ -126,6 +126,15 @@ def setUp(self):
         self.client.cookies[settings.MOBILE_COOKIE] = 'on'
 
 
+class MinimalViewTestCase(TestCase):
+    def setUp(self):
+        super(MinimalViewTestCase, self).setUp()
+
+    def get_minimal_url(self, document):
+        minimalurl = '?minimal=1&mobile=1'
+        return document.get_absolute_url()+minimalurl
```
Osmose: Add spaces between the operator and operands please (IE `document.get_absolute_url() + minimalurl`).

@rehandalal I don't know enough about triggering the minimal view to judge whether this is a fine way of doing so. Thoughts?

Osmose:
I'm r+ on this pending @rehandalal 's comments and the spacing fix around the `+`. I'll merge once that change is made and once rdalal shares his thoughts.

MikkCZ:
The spacing is fixed (sorry for that).

kitsune/sumo/tests/__init__.py
rehandalal: yep that should work

Osmose:
Nice work!

PR #2822, opened 2016-03-23
Remove Verbatim mentions

MikkCZ:
As a follow up to [bug 1230580](https://bugzilla.mozilla.org/show_bug.cgi?id=1230580), I suggest to remove/replace all Verbatim mentions left. Here is the doc localization page fixed. What about https://github.com/mozilla/kitsune/blob/master/kitsune/customercare/tests/test_templates.py#L97 and https://github.com/mozilla/kitsune/blob/master/docs/sop.rst?

### DIFF
--- a/docs/localization.rst
+++ b/docs/localization.rst
@@ -10,7 +10,7 @@ Kitsune is localized with `gettext <http://www.gnu.org/software/gettext/>`_.
 User-facing strings in the code or templates need to be marked for gettext
 localization.
 
-We use `Verbatim <http://localize.mozilla.org/>`_ to provide an easy interface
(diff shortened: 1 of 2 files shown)


MikkCZ:
@mythmon r? or what do you think about the last occurrences mentioned above?

mythmon:
Lets change the code comment (the first occurence you mentioned). For the SOP document, leave it as is for now. It isn't your responsibility to update that, and it is terribly out of date anyways.

MikkCZ:
Updated.

mythmon:
Thanks. r+, will merge once CI passes.

PR #2823, opened 2016-03-24
[Bug 1259342] Fix datetimeformat jinja helpers

safwanrahman:
The issue should fix the jinja helper. Issue raised in #2819 

@Osmose @mythmon r?

### DIFF
--- a/kitsune/sumo/templatetags/jinja_helpers.py
+++ b/kitsune/sumo/templatetags/jinja_helpers.py
@@ -211,14 +211,6 @@ def datetimeformat(context, value, format='shortdatetime'):
     Returns a formatted date/time using Babel's locale settings. Uses the
     timezone from settings.py, if the user has not been authenticated.
     """
-
-    def get_user_timezone(user):
-        try:
-            timezone = Profile.objects.get(user=user).timezone
-            return timezone
-        except (Profile.DoesNotExist, AttributeError):
-            return None
-
     if not isinstance(value, datetime.datetime):
         # Expecting date value
         raise ValueError(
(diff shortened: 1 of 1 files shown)


kitsune/sumo/templatetags/jinja_helpers.py
```diff
@@ -211,6 +211,14 @@ def datetimeformat(context, value, format='shortdatetime'):
     Returns a formatted date/time using Babel's locale settings. Uses the
     timezone from settings.py, if the user has not been authenticated.
     """
+
+    def get_user_timezone(user):
```
mythmon: Why did you move this code into an inner function? It seems a bit oddly placed, and doesn't really help with the readability.

kitsune/sumo/templatetags/jinja_helpers.py
```diff
@@ -226,16 +234,18 @@ def datetimeformat(context, value, format='shortdatetime'):
...
+        if 'timezone' not in request.session:
+            if request.user.is_authenticated():
+                user_timezone = get_user_timezone(request.user)
+                if user_timezone is not None:
+                    convert_tzinfo = user_timezone
+            request.session['timezone'] = convert_tzinfo
+        else:
+            convert_tzinfo = request.session['timezone']
+
+    except AttributeError:
```
mythmon: I worry that this try/except is too broad. I'd rather see something like `if hasattr(request, 'session')`, which would be very specific. Catching generic `AttributeError`s, in particular, can have unwanted side effects as it can hide programmer errors, like misspelling a variable name.

mythmon:
This should fix the problem, thanks! I think we can make this a little more robust before we merge it, though.

kitsune/sumo/templatetags/jinja_helpers.py
safwanrahman: If not split into another inner function, the code looks messy and the code become hard to read. I think, getting the User timezone should be in another function.

safwanrahman:
@mythmon updated. r?

MikkCZ:
According to http://stackoverflow.com/questions/610883/how-to-know-if-an-object-has-an-attribute-in-python, try/catch will be faster, but this looks less as an _exception driven programming_.

PR #2826, opened 2016-03-30
[Bug 1259377] Extract strings from jinja2 templates

mythmon:
@rehandalal r?

### DIFF
--- a/kitsune/settings.py
+++ b/kitsune/settings.py
@@ -625,6 +625,7 @@ PUENTE = {
             ('**/templates/**.ltxt', 'jinja2'),
             ('kitsune/**.py', 'python'),
             ('kitsune/**/templates/**.html', 'jinja2'),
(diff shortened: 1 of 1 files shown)


safwanrahman:
Need to restart the test! :disappointed:

kitsune/settings.py
```diff
@@ -625,6 +625,7 @@ def JINJA_CONFIG():
             ('**/templates/**.ltxt', 'jinja2'),
             ('kitsune/**.py', 'python'),
             ('kitsune/**/templates/**.html', 'jinja2'),
```
rehandalal: we can probably remove this line

mythmon: I think we have a couple templates that override Django templates that this would catch. I'm not sure if they use localization though?

rehandalal: ah yeah, then we should leave that in.

rehandalal:
r+

PR #2828, opened 2016-04-02
Simplify contributor metrics headlines l10n

MikkCZ:


safwanrahman:
It is not a good way to have this change.
The localizer should know what they are localizing. They can never know the concept 
like ``Contributor Satisfaction:  %(area)s'``, they will never know whats the area. They should know whats the Area and the localization is depend on it.

PR #2830, opened 2016-04-02
Set UTF-8 as default encoding for peep

MikkCZ:
This was necessary for me to make the build work. Maybe will help someone too, and hopefully not break anything.


safwanrahman:
Peep is deprecated and we will move to 'pip 8' which has built in hashing support.
And this file should not be changed as its a library file. you can try to make a PR to the peep repository.

PR #2831, opened 2016-04-03
[Bug 1255056] Make class_selected() safe

MikkCZ:
Would be great if this can be tested each time `class_selected()` is called...

Or do we want to use `class_selected()` or `if {% if %}class="selected"{% endif %}` as [here](https://github.com/mozilla/kitsune/blob/6650c53623127d60b34eedb3808e131dc646b412/kitsune/questions/jinja2/questions/question_list.html#L202)?

### DIFF
--- a/kitsune/questions/tests/test_templates.py
+++ b/kitsune/questions/tests/test_templates.py
@@ -1101,6 +1101,21 @@ class QuestionsTemplateTestCase(TestCaseBase):
         response = self.client.get(url)
         eq_(200, response.status_code)
 
+    def test_owner_tab_selected_in_list(self):
+        # Test one tab is selected for no show arg specified
+        questions_list = urlparams(reverse('questions.list', args=['all']))
+        response = self.client.get(questions_list)
+        doc = pq(response.content)
+        eq_(1, len(doc('#owner-tabs > .selected')))
+
+        # Test one tab is selected for all show args
+        show_args = ['needs-attention', 'responded', 'done', 'all']
+        for show_arg in show_args:
+            questions_list = urlparams(reverse('questions.list', args=['all']), show=show_arg)
+            response = self.client.get(questions_list)
+            doc = pq(response.content)
+            eq_(1, len(doc('#owner-tabs > .selected')))
+
     def test_product_filter(self):
         p1 = ProductFactory()
         p2 = ProductFactory()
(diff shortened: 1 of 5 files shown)


MikkCZ:
Can someone please re-run the tests? Failed on installation.

safwanrahman:
Can you please write a test so that we can avoid it in future?

MikkCZ:
Sure I can write a test for the particular occurrence reported in the bug, but not sure how to do it for all occurrences (will avoid mistakes for current places where used, but not for any future use somewhere else).

mythmon:
There is a better way of doing this. Since the output of `class_selected` is always safe, we can mark it as such in the helper function. That lives in [`kitsune/sumo/templatetags/jinja2_helpers.py`](https://github.com/mozilla/kitsune/blob/master/kitsune/sumo/templatetags/jinja_helpers.py#L468-L476). Wrapping the output of that function in a `Markdown(...)` instance should fix this.

MikkCZ:
Thank you, I will check it tomorrow.

mythmon:
This looks good to me. @safwanrahman, do you still want a test for this? What kind of test do you want?

@homu delegate=safwanrahman

safwanrahman:
So it should test, the class is there and the text is there. I have made a similar kind of test
https://github.com/mozilla/kitsune/pull/2759/files#diff-f96d1b40b0f6f4785dcc421f2c0e35d6R1699

MikkCZ:
Tests committed, by due to the GitHub outage not visible here yet. Hopefully will align itself soon.

MikkCZ:
Force pushed to run the tests. @safwanrahman r?

safwanrahman:
I was hoping to have a test of templates rather than template tags. 
is it possible for you to write the test for template?
Otherwise @mythmon should review the test as I dont have much experience with the jinja2 tags and their tests.

MikkCZ:
The fix was in changing how the filter works, so I wanted to test it. If you think I should test also the template, I will do.

MikkCZ:
Test for the question list added.

safwanrahman:
@homu retry

mythmon:
@homu r+

mythmon:
@MikkCZ it looks like you failed the linter. Just some whitespace issues to fix. You should be able to run these tests locally with `precommit run --all-files`. You can make sure this happens automatically in the future by running `precommit install`.

MikkCZ:
Cool! Fixed now. :)

safwanrahman:
@homu r+

PR #2835, opened 2016-04-04
Make warnings on translate page more consistent in tests

mythmon:
This makes the tests pass consistently for me, and I think the logic is unchanged (though the result is a bit surprising).

@safwanrahman r?

@homu delegate=safwanrahman

### DIFF
--- a/kitsune/wiki/views.py
+++ b/kitsune/wiki/views.py
@@ -1517,8 +1517,7 @@ def _get_next_url_fallback_localization(request):
 
 def _show_revision_warning(document, revision):
     if revision:
-        return document.revisions.filter(created__gt=revision.created,
-                                         reviewed=None).exists()
+        return document.revisions.filter(id__gt=revision.id, reviewed=None).exists()
     return False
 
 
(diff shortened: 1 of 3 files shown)


mythmon:
@homu priority=1

safwanrahman:
I think, Changing the way it test is a better approach.
Rather than counting the number, we should check if there is the warning message. 
So add a class [here](https://github.com/mozilla/kitsune/blob/master/kitsune/wiki/jinja2/wiki/translate.html#L59) and in test, check if the class is present. That makes more sense.

mythmon:
I made those changes ^

kitsune/wiki/views.py
```diff
@@ -1517,8 +1517,7 @@ def _get_next_url_fallback_localization(request):
 
 def _show_revision_warning(document, revision):
     if revision:
-        return document.revisions.filter(created__gt=revision.created,
-                                         reviewed=None).exists()
+        return document.revisions.filter(id__gt=revision.id, reviewed=None).exists()
```
safwanrahman: This one is Awesome for Future work. Thanks a lot!

safwanrahman:
@homu r+

mythmon:
Homu seems to have forgotten about us.

PR #2837, opened 2016-04-06
Remove vestiges of lockdown. It isn't used in prod.

mythmon:
Lockdown is only serving today to make our CI fail very often. I think the output of the NPM registries it proxies has changed. Additionally, it hasn't had an update in over a year, so I don't expect this problem to get fixed upstream.

Additionally, we aren't even verifying these hashes in prod. They only make CI more complex. I think we should remove them.

@rehandalal thoughts? r?


mythmon:
PS: if you want to merge this, you should use `@homu r+`. It will help the automation deal with the other PRs.

rehandalal:
sure if it's causing a bunch of issues we should ditch it.

rehandalal:
@homu r+

PR #2839, opened 2016-04-06
[Bug 1261896] Show community support link for all products and locales

MikkCZ:


kitsune/products/jinja2/products/includes/topic_macros.html
```diff
@@ -9,14 +9,12 @@
       </a>
     </li>
     {% if loop.last %}
-      {% if show_community_support %}
```
mythmon: This variable is no unused. It should be removed from the macro, and any places this macro is called.

safwanrahman:
Can you describe how and why the change would fix the issue?

mythmon:
The code looks good, but I'm not sure if this is the change we want. Lets hold off until we get a decision in the bug.

MikkCZ:
A link to the article is in the menu instead of AAQ if the SUMO support forum is not opened for the locale, so I do not see a reason why to hide it here. But let's discuss in the bug. (check https://support.mozilla.org/de/)

MikkCZ:
Updated to show the link for locales with AAQ enabled (the same behavior as now) or get-community-support article localized (requested behavior).

Waiting to know from Rachel, if the link should be added to the en-US fallback info, too.

MikkCZ:
r?

kitsune/products/views.py
```diff
@@ -35,6 +37,16 @@ def product_landing(request, template, slug):
         return HttpResponse(json.dumps({'topics': topic_list}),
                             content_type='application/json')
 
+    if request.LANGUAGE_CODE == settings.WIKI_DEFAULT_LANGUAGE:
+        show_community_support = True
+    else:
+        try:
+            doc = Document.objects.get(locale=settings.WIKI_DEFAULT_LANGUAGE,
+                                       slug='get-community-support')
```
safwanrahman: Hard Coded Slug is not a good choice after all! Because the slug can change anytime

kitsune/products/views.py
```diff
@@ -35,6 +37,16 @@ def product_landing(request, template, slug):
...
+            show_community_support = doc.translated_to(request.LANGUAGE_CODE)
```
safwanrahman: Is it possible to get the document with only one call?

safwanrahman:
This Should have test. Let me know if you need any help writing the test

kitsune/products/views.py
```diff
@@ -35,6 +37,16 @@ def product_landing(request, template, slug):
         return HttpResponse(json.dumps({'topics': topic_list}),
                             content_type='application/json')
 
+    if request.LANGUAGE_CODE == settings.WIKI_DEFAULT_LANGUAGE:
+        show_community_support = True
+    else:
+        try:
```
safwanrahman: This can be rewritten like this

```python
show_community_support = Document.objects.filter(locale=request.LANGUAGE_CODE, 
                                                 parent__slug='get-community-support').exclude(
                                                 current_revision=None).exists()
```

It will check whether any translated document is exists of the English document having slug ``get-community-support``. It will also exclude any document having no current revision.

kitsune/products/views.py
MikkCZ: I agree, but I am afraid there is no other option how to check, if the article has been translated. It's already done on few other places, this would be the first one outside of templates. Maybe I can rewrite how this is done to be called from the template with the slug.

MikkCZ:
I have rewritten the code basically from the scratch and added some tests.

safwanrahman:
The past approach was a better one I think rather than this. This is not a reusable thing that need to be live in template tags

MikkCZ:
So what change do you propose. To move the call of is_localized to product.html, so the help_topics() "interface" does not change? Or to move it to the views.py? I thought you do not like having the "get-community-support" slug on two distinct places (views.py and the template itself too).

safwanrahman:
> I thought you do not like having the "get-community-support" slug on two distinct places (views.py and the template itself too).

Sorry for misunderstanding. I have got that the slug is the only way to make the work done. I did not mean to not have it in views
Sorry again

safwanrahman:
I would propose have the change like that in past, just the approach i mentioned for checking whether the article is localized.

MikkCZ:
OK. Now I am failing some tests, so I will need to fix them first.

kitsune/products/views.py
```diff
@@ -44,12 +56,15 @@ def product_landing(request, template, slug):
         else:
             latest_version = 0
 
+    show_community_support = is_localized('get-community-support', request.LANGUAGE_CODE)
```
safwanrahman: The Past approach was correct. no need to query database if the requested locale is English.
Making another function is overkill

MikkCZ: I added the second query for case the article does not exist. Maybe I can change it a little and put the check for request and default locale equality outside is_localized().

MikkCZ:
The functionality itself is OK, but I have problems with the tests. @safwanrahman can you please check what I am doing wrong, probably with the test clients?

kitsune/products/views.py
```diff
@@ -44,12 +54,16 @@ def product_landing(request, template, slug):
         else:
             latest_version = 0
 
+    show_community_support = request.LANGUAGE_CODE == settings.LANGUAGE_CODE or is_localized(
```
safwanrahman: What is ``settings.LANGUAGE_CODE`` ? I think this should be ``settings.WIKI_WIKI_DEFAULT_LANGUAGE``

kitsune/products/views.py
```diff
@@ -21,6 +23,14 @@ def product_list(request, template):
         'products': products})
 
 
+def is_localized(slug, request_language_code):
```
safwanrahman: Please take the function at the end of the file

safwanrahman: Having  ``parent_slug`` and ``locale`` this two parameters makes is good readability

kitsune/products/tests/test_templates.py
```diff
@@ -158,3 +159,52 @@ def test_subtopics(self):
...
+
+    def test_community_support_shown_for_default_language(self):
+        """Verify the get community support is shown in the topics for en-US."""
+        p = ProductFactory()
+        d = DocumentFactory(slug='get-community-support', locale=settings.LANGUAGE_CODE)
+        rev = ApprovedRevisionFactory(document=d)
+        d.current_revision = rev
+        d.save()
+
+        self.refresh()
```
safwanrahman: What is done with ``self.refresh``?

kitsune/products/tests/test_templates.py
```diff
@@ -158,3 +159,52 @@ def test_subtopics(self):
         eq_(200, r.status_code)
         pqdoc = pq(r.content)
         eq_(1, len(pqdoc('li.subtopic')))
+
+    def test_community_support_shown_for_default_language(self):
+        """Verify the get community support is shown in the topics for en-US."""
+        p = ProductFactory()
+        d = DocumentFactory(slug='get-community-support', locale=settings.LANGUAGE_CODE)
```
safwanrahman: The ``Get Community Support`` Should be show all the time for default locale.

kitsune/products/views.py
safwanrahman: Oh! Seems like ``settings.LANGUAGE_CODE`` is the default locale. sorry

safwanrahman:
I have checked the code. Seems like some little improvement is needed.
I will check the test Errors later tonight or tomorrow. Will get back to you regarding the test failors

kitsune/products/jinja2/products/product.html
```diff
@@ -81,7 +81,7 @@ <h1 class="product-title cf">
   <div class="grid_12">
     <section id="help-topics">
       <h1>{{ _('Help Topics') }}</h1>
-      {{ help_topics(topics, show_community_support=product.questions_enabled(request.LANGUAGE_CODE)) }}
+      {{ help_topics(topics, show_community_support) }}
```
safwanrahman: As The default value of ``show_community_support`` is ``False`` [here](https://github.com/mozilla/kitsune/blohttps://github.com/mozilla/kitsune/blob/a8ed6f598293be6d7286cae2bf8ae3f4d9ea7d8b/kitsune/products/jinja2/products/includes/topic_macros.html#L1b/a8ed6f598293be6d7286cae2bf8ae3f4d9ea7d8b/kitsune/products/jinja2/products/includes/topic_macros.html#L1), So until you pass the value through it, it will use the default value.

kitsune/products/tests/test_templates.py
MikkCZ: It's in `kitsune/search/tests/__init__.py` - `ElasticTestCase.refresh()`. I assume it's needed to properly show everything needed on the landing page, as the above tests use it too.

kitsune/products/tests/test_templates.py
MikkCZ: Right, that was testing the previous behavior with checking, if the document even exists.

kitsune/products/jinja2/products/product.html
MikkCZ: This code works, at least when running it locally, I see it working properly. The show_community_support "variable", or how is it called in templates, is being passed to `product.html` from `kitsune/products/views.py`, so it shouldn't be needed to `show_community_support=show_community_support` the same as it's not needed to `topics=topics`.

kitsune/products/tests/test_templates.py
```diff
@@ -158,3 +159,46 @@ def test_subtopics(self):
         eq_(200, r.status_code)
         pqdoc = pq(r.content)
         eq_(1, len(pqdoc('li.subtopic')))
+
+    def test_community_support_shown_for_default_language(self):
+        """Verify the get community support is shown in the topics for en-US."""
+        p = ProductFactory()
+
```
safwanrahman: So I think, as there are no topic is created for the product, the ``topics`` become ``None``. Thats why the test is failing.
Lets create a Topic by ``TopicFactory(product=p, visible=True)``

MikkCZ:
I think I have found the trouble. It's caused by how the `help_topics()` macro actually works. It outputs everything in a loop over `topics`. So in case there is no topic for the product, no `ul` including `li#community-support` there. @safwanrahman is this behavior OK and I should workaround it in the tests by adding topics, or should I change, how the macro works for zero topics?

safwanrahman:
@MikkCZ I think its better to add a topic in the test at this moment. Otherwise the PR will be too big to get merged with limited dev resources

MikkCZ:
Huh, this is hopefully the last commit to this pull request. Let's give it ~15 minutes the finish the tests. Thank you @safwanrahman .

safwanrahman:
I neeed some time to take a closer look and review this Pull request. Maybe tomorrow or the next day!
@MikkCZ thanks a lot!

MikkCZ:
@safwanrahman , what do you think? Are those test sufficient?

glogiotatidis:
Thanks for the PR. Due to ongoing discussions about the future of the platform we'll avoid working on any pull requests not directly related to security or maintenance. I've marked this pull request as "Platform Decision Pending" to flag it for later review.

glogiotatidis:
Closing PR since development has been halted. Thanks your contribution.

PR #2843, opened 2016-04-11
Make "said" text localizable when quoting

MikkCZ:


MikkCZ:
My local venv gets broken. Have to fix it and check this.

MikkCZ:
Should work as expected.

safwanrahman:
Have you tested it? let me know, so easy for review.
Thanks

MikkCZ:
I was unable to file any question on local installation fot some reason, so I saved html file with injected JS files. Would be great if can test on real, but not sure what's missing in the local instance.

MikkCZ:
@glogiotatidis r?

MikkCZ:
Finally managed to test locally (article discussion + manually pasting https://support.cdn.mozilla.net/static/jsi18n/cs/djangojs.538269379a03.js content with manually added string into console) and it works.

@glogiotatidis r?

glogiotatidis:
Thanks @MikkCZ

PR #2844, opened 2016-04-12
Fix "./manage.py" NameError when "USE_DEBUG_TOOLBAR" is enabled"

MikkCZ:


MikkCZ:
Actually breaks almost every ./manage.py run.

safwanrahman:
we do not use jingo anymore. 
So this is not a correct way to fix it

MikkCZ:
Thanks, fixed "the best possible way" - deleting. ;)

safwanrahman:
Yap. Since we dont use jinja as a default template loader, so we dont need ``JINGO_EXCLUDE_APPS`` anymore.
:+1: for removing it
r+ from my end

MikkCZ:
@glogiotatidis r?

glogiotatidis:
r+ Thanks!

PR #2852, opened 2016-05-05
Add .idea to .gitignore

MikkCZ:
I am testing PyCharm and it looks pretty good. Let's make sure no IDE project files are committed.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -23,3 +23,4 @@ james.ini
 build.py
 .vagrant/
 venv/
+.idea/


safwanrahman:
Kitsune Should be Platform Independent and also IDE. Its not wise to add any specific IDE file into git ignore.

MikkCZ:
For the same reason I think no IDE files should get into the master ever, and .gitignore is the way to avoid it.

glogiotatidis:
SInce we have no other platform / ide specifics on .gitignore I suggest that we keep it this way. But I agree that we should protect master from accidental additions. @MikkCZ have you tried putting files to our own project specific gitignore at `$GIT_DIR/info/exclude` ? Does this work for you?

MikkCZ:
Oh, that works too. :)

PR #2854, opened 2016-05-05
[fix bug 1259953] Wrap GA in DNT conditional.

glogiotatidis:
Respect user's DNT flag.

### DIFF
--- a/kitsune/bundles.py
+++ b/kitsune/bundles.py
@@ -292,6 +292,7 @@ PIPELINE_JS = {
             'sumo/js/geoip-locale.js',
             'mailcheck/src/mailcheck.js',
             'sumo/js/ui.js',
+            'sumo/js/dnt-helper.js',
             'sumo/js/analytics.js',
             'sumo/js/surveygizmo.js',
             'sumo/js/instant_search.js',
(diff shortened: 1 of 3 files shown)


glogiotatidis:
@safwanrahman @MikkCZ r?

MikkCZ:
Pulled, tested and works. Could we add the JS test you linked in https://bugzilla.mozilla.org/show_bug.cgi?id=1259953#c4 too?

glogiotatidis:
> Pulled, tested and works. Could we add the JS test you linked in https://bugzilla.mozilla.org/show_bug.cgi?id=1259953#c4 too?

Good point. I added a note that this is direct copy. Since we're not diverging i'll use this as a library and skip adding tests.

MikkCZ:
:+1: r+

safwanrahman:
It seems like, the ``dnt-helper.js`` will not load in mobile site and minimal view. Can you please check @glogiotatidis

safwanrahman:
Also test fail seems relative. The linter style is not matched. You can test it locally. the instruction is  [here](http://kitsune.readthedocs.io/en/latest/hacking_howto.html#install-linting-tools)

glogiotatidis:
> Also test fail seems relative. The linter style is not matched. You can test it locally. the instruction is here

FYI Added proper editor config for JS https://github.com/mozilla/kitsune/commit/1db1d9ece8eda1eb6d63a362d348cb880542c74c

glogiotatidis:
it should be fine now, @safwanrahman ?

safwanrahman:
Looks Good @glogiotatidis 
r+ :+1:

safwanrahman:
Lets test it into stage first before sending to production

PR #2856, opened 2016-05-05
[Bug 1265027] More space for user-meta in the questions list

MikkCZ:
Caused by long translation (lines got wrapped). We have more space there thanks to tag-list 280px padding.

### DIFF
--- a/kitsune/sumo/static/sumo/less/questions.less
+++ b/kitsune/sumo/static/sumo/less/questions.less
@@ -785,7 +785,7 @@ h2 {
         position: absolute;
         right: 20px;
         text-align: right;
-        width: 250px;
+        width: 270px;
       }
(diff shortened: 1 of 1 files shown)


MikkCZ:
r?

safwanrahman:
I dont have much Idea about CSS. So someone with front end experience should review it.

MikkCZ:
@glogiotatidis do you have any experience with the front end and it's reviews?

glogiotatidis:
Well we can merge this as a better solution to the one we currently have but it doesn't eliminate the problem. You can still end up with question text overlapping or nearly close to the user details. IMO we should add the user details bellow the tags instead of next to the tags.

Thanks for working on this @MikkCZ!

PR #2857, opened 2016-05-05
[Bug 1150382] Add placeholders indicating valid date input format for new announcements

MikkCZ:


MikkCZ:
r?

safwanrahman:
I think a datepicker was added there? 
https://bugzilla.mozilla.org/show_bug.cgi?id=1083914

MikkCZ:
True. I thought the reported would like to see the time format in case it's pasted manually. But in case the data picker was added after the report, we might close this and the bug as well.

safwanrahman:
I think so. Because, the bug was filled before we added the date Picker.
So better to close the bug!

MikkCZ:
I should definitely go through the whole list or support.mozilla.org bugs and check, what's hidden there.

Closed the bug as duplicate and closing this PR too. Thank you for noticing a finding the original bug.

PR #2858, opened 2016-05-06
 [Bug 1205093] Show Ready for L10n in the KB dashboard based on the latest normal or major revision

MikkCZ:
Waiting for Joni info about the minor revisions correct behavior.

### DIFF
--- a/kitsune/dashboards/readouts.py
+++ b/kitsune/dashboards/readouts.py
@@ -230,10 +230,16 @@ def kb_overview_rows(mode=None, max=None, locale=None, product=None, category=No
             'trans_url': reverse('wiki.show_translations', args=[d.slug],
                                  locale=settings.WIKI_DEFAULT_LANGUAGE),
             'title': d.title,
-            'num_visits': d.num_visits,
-            'ready_for_l10n': d.revisions.filter(is_approved=True,
-                                                 is_ready_for_localization=True).exists()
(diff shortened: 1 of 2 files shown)


safwanrahman:
The Dashboard part is so critical that, without a full time commited dev, I would rather not make any changes there. Because it is very critical and raises a lot of regression in the past.
Your work is awesome, but I think its better to keep the dashboard as it is till we dont get any full time dev.

MikkCZ:
OK. Let's do not merge this until we have a full time developer or we know the future direction of the platform development.

I will anyway update info in this PR once I get the info from Jonii and commit any necessary changes if I am able to do them, so this is not stuck somewhere in the middle.

MikkCZ:
Looking at my other PR, I will write some tests for this, too...

MikkCZ:
According to https://bugzilla.mozilla.org/show_bug.cgi?id=1205093#c7 the behavior I expected is correct. So it's time to make some tests for this.

glogiotatidis:
Thanks for the PR. Due to ongoing discussions about the future of the platform we'll avoid working on any pull requests not directly related to security or maintenance. I've marked this pull request as "Platform Decision Pending" to flag it for later review.

glogiotatidis:
Closing PR since development has been halted. Thanks your contribution.

PR #2862, opened 2016-05-08
[WIP] [Bug 1271139] Fix test failures with locale cloned and compiled

MikkCZ:


MikkCZ:
I should rather set up a test suite including localization compilation, to show it failing, and fix it afterwards.

MikkCZ:
This is not the right way to go.

PR #2870, opened 2016-05-27
[bug 1268346] Redirect to lang only if lang is supported

glogiotatidis:
Fixes 1268346 by performing the lang setting only if `lang` is in `SUMO_LANGUAGES`.

You can test this code on support.allizom.org

### DIFF
--- a/kitsune/sumo/middleware.py
+++ b/kitsune/sumo/middleware.py
@@ -34,9 +34,10 @@ class LocaleURLMiddleware(object):
         set_url_prefixer(prefixer)
         full_path = prefixer.fix(prefixer.shortened_path)
 
-        if 'lang' in request.GET:
+        if request.GET.get('lang', '') in settings.SUMO_LANGUAGES:
             # Blank out the locale so that we can set a new one. Remove lang
             # from the query params so we don't have an infinite loop.
+
             prefixer.locale = ''
             new_path = prefixer.fix(prefixer.shortened_path)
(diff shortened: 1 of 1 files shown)


MikkCZ:
Hi @glogiotatidis. Can you please add me to the bug? Without access there, I do not know what to test. Want to check it too, @safwanrahman ?

safwanrahman:
I also dont have access to read the bug therefore I can not know what is the approach.
I strongly believe, this will break our fallback locale system which is hardcoded in settings. We use some specific fallback system like ``bn`` should be redirected to ``bn-BD`` not ``bn-IN``. Its hardcoded in the settings. 
Sometime, the locale maybe not in our language list, but we may have custom fallback for that locale.

glogiotatidis:
> I strongly believe, this will break our fallback locale system which is hardcoded in settings. We use some specific fallback system like bn should be redirected to bn-BD not bn-IN. Its hardcoded in the settings.

The patch doesn't break this behavior, you can test on stage.

glogiotatidis:
@jgmize r?

jgmize:
>>  some specific fallback system like bn should be redirected to bn-BD not bn-IN. Its hardcoded in the settings.

> The patch doesn't break this behavior, you can test on stage.

Code lgtm, and I verified that support.allizom.org/bn/ is redirected to support.allizom.org/bn-BD/.

r+

PR #2871, opened 2016-06-09
Pipeline perms fix

glogiotatidis:
- unit tests pass, nevermind the frontend tests. they fail due to other changes.
- had to patch django-pipeline to respect Storages permissions. See also https://github.com/jazzband/django-pipeline/pull/570
- Had to remove virtualenv and re-create it to force update the packages

@jgmize r?

### DIFF
--- a/requirements/default.txt
+++ b/requirements/default.txt
@@ -119,9 +119,8 @@ django-nose==1.4.3
 # sha256: jTddtwi8h_bfm7OulO3QG-0su2polKkneF3tmu3pg-8
 https://github.com/gintas/django-picklefield/archive/54b11bdb177fd6140c3ae7d6e9744eaa2deda261.tar.gz#egg=django-picklefield
 
-# sha256: pqnOIAOuEOcpdotteaSz2zMNh2uGIl63dDtvVpa6-lk
-# sha256: t83C7-J-R43dOOh43XteWsEhwomP2FrM8xh-GRN_5rc
-django-pipeline==1.6.8
+# sha256: gTftCmQ1YK4jAOESeOBJB_ChIIAxoAaUdHccq2gA8yU
+https://github.com/glogiotatidis/django-pipeline/archive/103bb549cd5f93b5648e10f1be6c2c6c8c5e9610.tar.gz#egg=django-pipeline
 
 # django-ratelimit: tags/v0.4.0~8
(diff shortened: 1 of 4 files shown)


safwanrahman:
We had broke the virtualenv and recreate it before and it made a lot of errors in the cronjob. So I would propose to not do it!

jgmize:
@glogiotatidis I agree with both @safwanrahman's comment here and your comments in irc that virtualenv recreation should be done manually on an as-needed basis rather than on every deploy. Other than that and the failing tests this lgtm, and as you mentioned those are due to a separate issue, and while I'd of course prefer to see those passing first I'll leave it to your judgement on whether that should be a blocker for this. r+wc

requirements/default.txt
```diff
@@ -31,19 +31,63 @@ celery==3.1.17
...
+# sha256: Kdkqca2Igv-fpUyXURLPEWqi5Zi9U90ji9chw-g2quA
+# sha256: VNOHaFdBZQnOtHkdk0fL32Gz_ZQPRr2ELP9A7VGE6QI
+# sha256: chp9gSb4ufTMJoviSAEC9zoCQM651tWcDCeT0KKE4CY
+# sha256: 2XkAjo_ptE0vyhEsxtNx5HeEAGSdVxDOaH0UaJ2qEII
+# sha256: hGQ4FZee5FqGdqoA5_dQEe3rM-VZvd0l5g6-p4uhLgo
+# sha256: I_7i45dhm8K8xIrf1rBF6Z7QxhiPqGCd01ajWgRgwZA
+# sha256: 0oL3ZesgmSbvEgjGUyYk-lCHpZjkCPincq6KZSXymHM
+# sha256: 9SRb08DUoxtK_Dl9FY7RsFG3A9lpBSdHLoyjcbINhzU
+# sha256: uxSVQO2QxLIXG_aU_mmR1jMbwUmuYjyP9BkyT0Ii0Sg
+cryptography==1.4
```
safwanrahman: We have tried to upgrade the library before and it broke our developer site. Hope to concern about it.

glogiotatidis:
I merged 
https://github.com/mozilla/kitsune/commit/25c4c1c9814882a2e9e8083dade8bbc6800fa3ff
and https://github.com/mozilla/kitsune/commit/6ae4ae24aa7814dc9af54aa711373f9feb8dfe2c and pushed to production.

We made only manual changes to the virtualenv to remove django-pipeline and to force install of the new version.

Things appear to be working.

Thanks for the feedback!

PR #2872, opened 2016-06-30
 [fix bug 1283148] Check if _gaq is defined before using it. 

glogiotatidis:


MikkCZ:
Search works again in staging, but:
1. I see code duplication of `function trackEvent` in both `googleAnalytics.es6.js` and `analytics.js`. This can be resolved later.
2. Builds are failing - smoke on some driver issue (probably unrelated), but django fails at the very beginning with complains about `TypeError: Property 'trackEvent' of object [object Object] is not a function`.

MikkCZ:
@glogiotatidis Could you please look at least at the second point? I won't be able to investigate and fix before the weekend. :(

glogiotatidis:
> Search works again in staging, but:
> 1. I see code duplication of function trackEvent in both googleAnalytics.es6.js and analytics.js. This can be resolved later.
There's code duplication in many different places. Since we're deprecating this, I don't consider this a problem for now.

> 2. Builds are failing - smoke on some driver issue (probably unrelated), but django fails at the very beginning with complains about TypeError: Property 'trackEvent' of object [object Object] is not a function.

Tests are broken anyway. Smoke indeed unrelated. Will try to fix trackEvent but if not an easy fix I'll push code to fix the broken search asap.

Thanks for checking Michal.

MikkCZ:
You are welcome. :) Or maybe just revert https://github.com/mozilla/kitsune/pull/2854 if the time is expensive?

glogiotatidis:
That's also a possibility, yes

glogiotatidis:
Tests fixed. Merging.

PR #2875, opened 2016-07-19
Upgrade to Django 1.8.14

glogiotatidis:


glogiotatidis:
@safwanrahma r?

safwanrahman:
Looks good. Test passes :+1: @glogiotatidis 
According to our convention, we do not include the wheel of hashes. Can you please remove that?
Except that **r+wc**

glogiotatidis:
> According to our convention, we do not include the wheel of hashes. Can you please remove that?

Can you please explain why we don't include hashes of wheels?

safwanrahman:
@glogiotatidis The reason of not using wheel is written [here ](https://github.com/mozilla/kitsune/blob/master/peep.sh#L14)

We always pass ``--no-use-wheel`` while downloading dependencies. So thats the reason we should not include the hash of wheel.

jayy7312:
> 

- Right

dow572 (commented):
Flare

requirements/default.txt
```diff
@@ -92,8 +92,8 @@ cryptography==1.4
 # sha256: _eJnrrA6jUnjNB5_caIoxnSNzkO1wA9lWBeA2Ztwstk
 dennis==0.7
 
-# sha256: 7BSL5zVI2gkN12wujFfJjosehPLLh1ALm-VCAYekNfs
-Django==1.8.11
+# sha256: UoLEi5D7spUHKZ9ZIhU3joSaKAj0heQURibmZxXVRk0
```
jmayes41, line 95: *
```suggestion
# sha256: UoLEi5D7spUHKZ9ZIhU3joSaKAj0heQURibmZxXVRk0
```
**

tinandanny (approved):

candaceblewett42 (commented):
image

Mikevin:
@candaceblewett42 Is this someone's personal data? Looks like a privacy concern.

@glogiotatidis Can you check if this is relevant or remove it? Reporting to GitHub itself looks like a slow process.

Chanvannak52 (commented):
requirements/default.txt

candaceblewett42 (commented):
.

candaceblewett42 (commented):
Hey partner

tran09:
Nice work man

tran09 (commented):
Nice work

tran09 (commented):
What doesn't this do?

PR #2880, opened 2017-01-07
Update to Windows 10

stephendonner:
@davehunt / @rbillings r?

I know this isn't all that useful as this project transitions over to Lithium's platform, but it does put us on a better-supported/later version of Windows.

Passing ad-hoc job here: http://webqa-ci-staging1.qa.scl3.mozilla.com:8080/job/sumo.adhoc/18/

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -5,7 +5,7 @@ def environments = [
   desktop: [
     browserName: 'Firefox',
     version: '47.0',
-    platform: 'Windows 7'
+    platform: 'Windows 10'
   ],
   mobile: [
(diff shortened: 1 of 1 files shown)


adriansanmigel:
01-06-2017
Having the master License Why was i not notified of this.

davehunt:
We're retiring these tests very soon, so I don't think we need this change.

stephendonner:
Agreed; I should've bothered to read https://github.com/mozilla/kitsune/blob/bd5b8ab17c30acb9237771dd01ec16d393a77082/README.rst :-)

PR #2881, opened 2017-04-20
[bug 1357563] Try to redirect users to the correct page.

glogiotatidis:


kitsune/middleware.py
```diff
@@ -0,0 +1,186 @@
...
+        cached_redirect = cache.get(cache_key)
+        if cached_redirect:
+            return HttpResponseRedirect(cached_redirect)
+
+        if urlparts[2] in self.product_pages:
+            return HttpResponseRedirect(self.product_pages[urlparts[2]])
+
+        path = request.path_info
+        url = 'https://secure02.lithium.com.edgekey.net' + path
+        response = requests.get(
```
pmac: Should set a reasonable timeout on this request.

glogiotatidis:
- Removed search page redirect after talking with Ben
- Hardcoded a couple more under `extra_pages`
- Added timeout

kitsune/middleware.py
```diff
@@ -0,0 +1,199 @@
...
+# build a list of topics translated for reverse lookup
+TOPICS = {}
+for topic in set(Topic.objects.values_list('title', flat=True)):
+    for lang in settings.SUMO_LANGUAGES:
+        with translation.override(lang):
+            trans_text = pgettext('DB: products.Topic.title', topic)
+        TOPICS[trans_text] = topic
+
+PRODUCTS = {}
+for topic in set(Product.objects.values_list('title', flat=True)):
```
mythmon: Nit: this probably should be `for product in set(...`.

kitsune/middleware.py
```diff
@@ -0,0 +1,199 @@
...
+        with translation.override(lang):
+            trans_text = pgettext('DB: products.Product.title', topic)
+        PRODUCTS[trans_text] = topic
+
+
+class LithiumRedirectionMiddleware():
+    """
+    Redirect requests for user profiles from /<username> to
+    /u/<username> to avoid breaking profile urls with the new url
+    schema.
```
mythmon: Is this the right docstring for this class? It doesn't seem to match anything else around here.

kitsune/middleware.py
```diff
@@ -0,0 +1,199 @@
...
+    }
+
+    extra_pages = {
+        '/t5/How-To/Cookies/ta-p/16348': '/kb/cookies-information-websites-store-on-your-computer',
+        '/t5/Fix-slowness-crashing-error/What-does-quot-Your-connection-is-not-secure-quot-mean/ta-p/30354': '/kb/what-does-your-connection-is-not-secure-mean',
+        '/t5/Firefox-for-Android/ct-p/Firefox-Android': '/products/mobile',
+    }
+
+    def process_response(self, request, response):
+        if (response.status_code != 404):
```
mythmon: nit: Extra parens aren't needed.

kitsune/middleware.py
```diff
@@ -0,0 +1,199 @@
...
+            if not urlparts[1].startswith('t5'):
+                return response
+        except KeyError:
+            return response
+
+
+        # Oh boy, Lithium URL. Let's see if we can turn things around.
+
+        # Cache first
+        cache_key = hashlib.sha1(request.path_info.encode('utf-8')).hexdigest()
```
mythmon: It's not a big deal, but i would probably be better not to use sha1 anymore, as a general rule.

kitsune/middleware.py
```diff
@@ -0,0 +1,199 @@
...
+            sentry_client.captureException()
+            return handle404(request)
+
+        try:
+            response.raise_for_status()
+        except requests.exceptions.RequestException:
+            sentry_client.captureException()
+            return handle404(request)
+        else:
+            for content, prop in self.meta_regex.findall(response.content):
```
mythmon: Maybe it would be better to use `pq` to parse the page, instead of using regex on html?

kitsune/middleware.py
```diff
@@ -0,0 +1,199 @@
...
+
+                        title = TOPICS.get(title, title)
+                        product = PRODUCTS.get(product, product)
+
+                        try:
+                            bar = Topic.objects.get(title=title, product__title=product)
+                        except (Topic.DoesNotExist, Topic.MultipleObjectsReturned):
+                            pass
+                        except:
+                            sentry_client.captureException()
```
mythmon: This is an interesting pattern. The goal is to record the error in sentry, without interrupting the view, right? That's pretty clever.

kitsune/middleware.py
```diff
@@ -0,0 +1,199 @@
...
+# 1. Check if the response is 404 and it's from a Lithium URL (/t5/*)
+# 2. Checks cache, redirects if successful
+# 3. Checks Lithium localized product_pages from LithiumRedirectionMiddleware.product_pages, redirects if successful
+# 4. Fetches the page from Lithium and tries to extract title of the page.
+#    Use og:title instead of HTML Title because Lithium strips the later down to 50 chars or something.
+#    If it fails it returns 404
+# 5a. Try to find a Document with the same title, redirect if successful
+# 5b. Try to find a Topic with the same title under a product. Product is also extracted by title. For non-English pages we need to do a reverse translation lookup from TOPICS
+# 5c. Try to find a Product. For non-English pages we need to do a reverse translation lookup from PRODUCTS
+# 5d. If everything fails redirect to the search page with title as search string.
```
mythmon: This sounds like a great idea, but I don't see it implemented in the code below. Am I missing something?

glogiotatidis: Good catch, we talked with ben and agreed to remove this to track the failures. updating the docs.

kitsune/middleware.py
glogiotatidis: copy / pasting fail

kitsune/middleware.py
glogiotatidis: i'm kinda used to always have parrens but good point

kitsune/middleware.py
glogiotatidis: I thought of beautiful soup but i didn't want to bring another dep inside given how  fragile the deployment method is. Also i love regexes (!)

kitsune/middleware.py
glogiotatidis: yap!

glogiotatidis:
Thanks for your time @mythmon. Updated with your comments

kitsune/middleware.py
mythmon: I think `pq` is already part of the dependencies, which is why I mentioned it. I recall using it in the tests, but I don't know if we ever used it in production. If it isn't available, then carry on.

PR #2882, opened 2017-05-21
Switch SESSION_ENGINE to memcached

jgmize:


jgmize:
Due to increased traffic levels the DB is not able to keep up with all of the sessions being written with the current `SESSION_ENGINE`, and @fox2mike made this change to the local settings file. So far the change appears to have alleviated the DB performance issues but it is likely that editors will need to log in more often.

fox2mike:
Pushed to prod by https://bugzilla.mozilla.org/show_bug.cgi?id=1366571

safwanrahman:
@jgmize Do you have any idea how long the session will live in memcached?
As far as I remembered, there were some issue in the past where we faced contributors complain of their session time out.

safwanrahman:
So this was added with this bug
https://bugzilla.mozilla.org/show_bug.cgi?id=649452

glogiotatidis:
@safwanrahman thanks for the feedback. We pushed this in prod to mitigate a session issue we had and we're already working on getting the sessions live as long as before.

glogiotatidis:
OK the memcached only works for now we'll continue to work with this on production. @safwanrahman please reach out if you get reports that ppl's sessions are timing out in an aggressive and/or annoying manner.

PR #2883, opened 2017-05-23
JSON Serializer

glogiotatidis:
See also https://bugzilla.mozilla.org/show_bug.cgi?id=1366558


glogiotatidis:
This cannot be used due to the type of data we store in the session.

PR #2884, opened 2017-07-05
Implement docker with docker-compose

safwanrahman:
This is very early stage of implementing docker for kitsune. Its not ready yet. Needs heavy rewrite.
Its actually based on the experience got from ``kuma``
Next stage can be:
- [x] Configure docker to build successfully
- [x] Pass All the unit tests in docker
- [ ] Pass all the smoketests in docker
- [ ] Configure the CI so docker image is successfully build there
- [ ] Make all the tests pass in docker environment
- [ ] Configure CI to run all the tests in docker environment
- [ ] Run the smoketest for the docker environment


glogiotatidis:
Great work @safwanrahman . 

I'd de-prioriotize any CI related work related to docker. We'll build docker images and push to Deis from our Jenkins instance and therefore we can keep Travis for PR testing which can be done outside the docker env, given that building docker and getting it right on Travis can be a pain. What do you think?

safwanrahman:
> We'll build docker images and push to Deis from our Jenkins instance

Hmm. Actually the build is done after merging any Pull request.

> building docker and getting it right on Travis can be a pain.
Its actually a pain, but its easier for long run. In `kuma`, we have implemented it so that we can see if any of the PR break the docker environment and if any test failed in the environment. I have worked mostly in implementing to get all the unit test pass in docker environment with travis ci. I think that experience will help here also.

Let it not remain unsaid that, for the contributors who use low speed Internet like me, Its hard to build docker image every time they make any pull request. So its easier for them to get their error if the test run in Travis.

glogiotatidis:
I understand your comments and we can agree. What I suggested is to de-priotize docker work related to CI for now so we get docker and tests working before we start fighting Travis / Jenkins. That said if it's actually easier for you to build on Travis yes we can have a working `.travis.yml`.

safwanrahman:
So most of the test pass at this moment. Some tests still fail. Need to debug more to investigate why they fail

Dockerfile
```diff
@@ -0,0 +1,49 @@
...
+    python-pip gettext build-essential \
+    libxml2-dev libxslt1-dev zlib1g-dev git\
+    libjpeg-dev libffi-dev libssl-dev libxslt1.1\
+    libmysqlclient-dev software-properties-common apt-transport-https
+
+# Install mariadb-client 5.5
+RUN apt-key adv --recv-keys --keyserver keyserver.ubuntu.com 0xcbcb082a1bb943db
+RUn add-apt-repository 'deb [arch=amd64,i386] http://sfo1.mirrors.digitalocean.com/mariadb/repo/5.5/debian wheezy main'
+RUN apt-get update
+RUN apt-get install -y --no-install-recommends  mariadb-client-5.5 libmariadbclient-dev
```
glogiotatidis: Hmm not sure if that's needed here. We probably are going to run on mariadb 10+ so we can just use the default packages. Feel free to leave this here for now and we can revise later

Dockerfile
```diff
@@ -0,0 +1,49 @@
...
+
+# Install nodejs older version 0.10.x
+RUN curl -sL https://deb.nodesource.com/setup_0.10 | bash -
```
glogiotatidis: OK for now but we need to integrate what this script does into this dockerfile.

Dockerfile
```diff
@@ -0,0 +1,49 @@
...
+COPY ./bower.json /app/bower.json
+
+RUN pip install pip==7.0
+RUN ./peep.sh install -r requirements/default.txt --no-binary :all:
+
+# Add a non root user and use it further
+RUN adduser -u 1001 --gecos 'kitsune' --no-create-home kitsune
+# Give permission to the created user to write in directory
+RUN chown -R kitsune /app
+# npm and bower tries to write in /home directory. So this permission is needed
```
glogiotatidis: Hmm since there's no `home` for kitsune user what are you trying to achieve here?

docker-compose.yml
```diff
@@ -0,0 +1,80 @@
...
+    image: redis
+
+  rabbitmq:
+    image: rabbitmq
+    environment:
+      - RABBITMQ_DEFAULT_USER=kitsune
+      - RABBITMQ_DEFAULT_PASS=kitsune
+      - RABBITMQ_DEFAULT_VHOST=kitsune
+
+volumes:
```
glogiotatidis: Why you want volumes for node_modules and bower_components and not let them in `/app`?

kitsune/settings.py
```diff
@@ -45,13 +49,13 @@ def path(*parts):
         # Or path to database file if sqlite3.
         'NAME': 'kitsune',
         # Not used with sqlite3.
-        'USER': '',
+        'USER': config('DATABASE_USER'),
```
safwanrahman: I have done this in the commit https://github.com/mozilla/kitsune/pull/2884/commits/a6c529a222e47f41b216d53f81b5df1fa84aa299
But it was failing some of the tests. so needed to revert it back!

Dockerfile
```diff
@@ -0,0 +1,49 @@
+FROM quay.io/mozmar/base:latest
+WORKDIR /app
+
+ENV PYTHONDONTWRITEBYTECODE=1 \
+    PIP_DISABLE_PIP_VERSION_CHECK=1
+
+RUN apt-get update
+RUN apt-get upgrade -y
```
safwanrahman: Yeah. its needed. I dont know why I added this.

Dockerfile
```diff
@@ -0,0 +1,49 @@
...
+RUN apt-get install -y --no-install-recommends python2.7 libpython2.7 python-dev \
```
safwanrahman: Yeah. Its also needed.

Dockerfile
safwanrahman: As far as I know, our production runs with mariadb 5.5 and in the documentation, 5.5 is also mentioned. 
I do not want to raise new issue by upgrading to 10+

Dockerfile
safwanrahman: Yeah. we can keep the script in our project and just run it!

docker-compose.yml
```diff
@@ -0,0 +1,80 @@
...
+      - ALLOWED_HOSTS=*
+      - BROKER_HOST=rabbitmq
+      - BROKER_PORT=5672
+      - BROKER_USER=kitsune
+      - BROKER_PASSWORD=kitsune
+      - BROKER_VHOST=kitsune
+      - REDIS_URL=redis://redis:6379/0
+      - CELERY_ALWAYS_EAGER=False
+      - CSRF_COOKIE_SECURE=False
+      - DATABASE_URL=mysql://${DATABASE_USER:-root}:${DATABASE_PASSWORD:-kitsune}@mariadb:3306/${DATABASE_NAME:-kitsune}
```
safwanrahman: Why?

docker-compose.yml
safwanrahman: Because, while building docker image, ``node_modules`` and ``bower components`` directory are created. We are mounting the local directory to ``./app``. In the local machine directory, there are no ``node_modules`` and ``bower components``. So need to add this volume to keep the files in `./app`

Dockerfile
glogiotatidis: I hear your concern. We can leave this now as is but we'll need upgrade to 10.x to move to AWS b/c that's the version RDS provides.

docker-compose.yml
glogiotatidis: I don't see a use for that and it complicates things. We define a database to use and we already know the variables. There's no need to allow override those using env variables. If for some reason one wants to change database they can override `DATABASE_URL` entirely.

docker-compose.yml
glogiotatidis: Hmm, I see your point but I'm not sure I like this solution b/c I don't like volumes. I'll investigate other solutions as well.

glogiotatidis:
@safwanrahman let's keep the docker development in the `docker` branch. I changed the branch of this PR

safwanrahman:
So some tests are failing because of permission issue. One test fail is critical to debug, because it could not create unicode files in the system. Do not know why its happening.

safwanrahman:
So now only 1 test failing!
```python
ERROR: test_status (kitsune.search.tests.test_cmds.ESCommandTests)
  File "/app/kitsune/search/tests/test_cmds.py", line 48, in test_status
    call_command('esstatus')
  File "/usr/local/lib/python2.7/dist-packages/django/core/management/__init__.py", line 120, in call_command
    return command.execute(*args, **defaults)
  File "/usr/local/lib/python2.7/dist-packages/django/core/management/base.py", line 445, in execute
    output = self.handle(*args, **options)
  File "/app/kitsune/search/management/commands/esstatus.py", line 16, in handle
    es_status_cmd(options['checkindex'], log=FakeLogger(self.stdout))
  File "/app/kitsune/search/es_utils.py", line 640, in es_status_cmd
    log.info('  Version                 : %s', es_deets['version']['number'])
UnboundLocalError: local variable 'es_deets' referenced before assignment
```

Dockerfile
```diff
@@ -0,0 +1,45 @@
...
+RUN curl -sL https://deb.nodesource.com/setup_0.10 | bash -
+RUN apt-get install -y nodejs=0.10.48-1nodesource1~xenial1
+
+# https://github.com/pypa/setuptools/issues/544
+RUN pip install --upgrade setuptools
+COPY ./requirements /app/requirements
+COPY ./package.json /app/package.json
+COPY ./bower.json /app/bower.json
+
+RUN pip install -r requirements/default.txt --no-binary :all: --require-hashes
```
glogiotatidis: Please remove the no-binary flag here and in other places. We can use wheels in this setup

docker-compose.yml
```diff
@@ -0,0 +1,80 @@
+version: '2.1'
+services:
+  web:
+    image: quay.io/mozmar/kitsune_base
```
glogiotatidis: You don't need to define `image` when using `build`. Also what's ` quay.io/mozmar/kitsune_base`?

docker-compose.yml
```diff
@@ -0,0 +1,80 @@
...
+      - PYTHONUNBUFFERED=True
+
+  memcached:
+    image: memcached
+
+  mariadb:
+    image: mariadb:5.5
+    environment:
+      - MYSQL_ROOT_PASSWORD=kitsune
+      - MYSQL_USER=${DATABASE_USER:-kitsune}
```
glogiotatidis: Also in this sections not needed to have env variables.

docker-compose.yml
```diff
@@ -0,0 +1,80 @@
...
+      - CELERY_ALWAYS_EAGER=True
+      - CSRF_COOKIE_SECURE=False
+      - DATABASE_URL=mysql://root:kitsune@mariadb:3306/kitsune
+      - DATABASE_USER=root
+      - DATABASE_PASSWORD=kitsune
+      - DATABASE_HOST=mariadb
+      - DATABASE_PORT=3306
+      - DEBUG=${DEBUG:-True}
+      - DOMAIN=localhost
+      - ES_URLS=elasticsearch:9200
```
glogiotatidis: this should be `ES_URLS=http://elasticsearch:9200`

safwanrahman:
@glogiotatidis Fixed the issue you have mentioned.
Now I hope all the unittest pass.
r?

glogiotatidis:
Unit tests verified locally to work. Nice work @safwanrahman  🐘 

This is OK to allow us to start working with docker. Merging this PR as a base and we'll continue to work on docker

PR #2885, opened 2017-07-13
Upgrade to pip8

glogiotatidis:
Kept the same version of packages. Switched to pip8 hashing, removed peep.


glogiotatidis:
@safwan @pmac r?

safwanrahman:
Remembering the past, I have made a PR #2813 for moving to `pip8` and we have discussed that its not right because we have shortage of developers. We have discussed over IRC that as `peep` was working fine, do not change it untill it breaks.
I think, we can keep peep as it is and move forward without changing this. We can upgrade to pip8 once we verified all other things.

glogiotatidis:
> Remembering the past, I have made a PR #2813 for moving to pip8 and we have discussed that its not right because we have shortage of developers. We have discussed over IRC that as peep was working fine, do not change it untill it breaks.
> I think, we can keep peep as it is and move forward without changing this. We can upgrade to pip8 once we verified all other things.

I understand how this made sense in the past, given that `peep` is also part of the Chief deployment method. Since we're building a new deployment pipeline it's better that we modernise some things to make our life easier. We're not changing any python packages or versions so I we're going to be just fine with this change. When we merge this you can rebase your docker PR on top and remove all the `peep` steps from there as well.

glogiotatidis:
@pmac updated. Some references to peep did not make sense to update (e.g. vagrant) so I just removed them. This has been tested locally and installs correctly.

Travis will fail b/c it's broken anyway. We'll fully update Travis and fix the failing tests in upcoming commits, especially after we merge the first iteration of dockerfile in #2884

safwanrahman:
Can you also remove the vagrant part from the docs? As you are removing vagrant also

glogiotatidis:
> Can you also remove the vagrant part from the docs? As you are removing vagrant also

I did https://github.com/mozilla/kitsune/pull/2885/commits/fe289a1c0db0c8f30754a2eecf2f666167ec2971#diff-a3f43adba2793d92ca3ed1cffb76cd02

glogiotatidis:
Thanks both!

safwanrahman:
@glogiotatidis ``factory-boy`` is missing in the requirements! 😞

glogiotatidis:
> @glogiotatidis factory-boy is missing in the requirements! 😞

that was a rebase mistake, nice catch @safwanrahman . You can re-add this in  #2884

PR #2886, opened 2017-07-17
[fix bug 1381436] Replace ES facets with DB queries.

glogiotatidis:
ElasticSearch v2 has deprecated facets in favor of aggregations. To upgrade to
aggregations we need to update elastic search python libraries. This is a
combination of ES and Database queries to serve as a workaround until we move
codebase to v2.

Since the DB is quite slower compared to ES, the results get cached for 15
minutes.

This has been tested on staging on ESv2 and works fine.

### DIFF
--- a/kitsune/community/utils.py
+++ b/kitsune/community/utils.py
@@ -1,11 +1,14 @@
 from datetime import datetime, date, timedelta
+from django.core.cache import cache
 from django.conf import settings
+from django.db.models import Count
 
 from kitsune.customercare.models import ReplyMetricsMappingType
 from kitsune.products.models import Product
 from kitsune.questions.models import AnswerMetricsMappingType
 from kitsune.search.es_utils import F
-from kitsune.users.models import UserMappingType
+from kitsune.users.models import User
+from kitsune.users.templatetags.jinja_helpers import profile_avatar
 from kitsune.wiki.models import RevisionMetricsMappingType
 
 
@@ -13,38 +16,57 @@ from kitsune.wiki.models import RevisionMetricsMappingType
 # section.  There isn't a way to tell ES to just return everything.
 BIG_NUMBER = 3000
 
+# This should be higher than the max number of questions, revisions or tweets
+HUGE_NUMBER = 1500000
+
 
 def top_contributors_questions(start=None, end=None, locale=None, product=None,
-                               count=10, page=1):
+                               count=10, page=1, use_cache=True):
     """Get the top Support Forum contributors."""
     # Get the user ids and contribution count of the top contributors.
-    query = (
-        AnswerMetricsMappingType
-        .search()
-        .facet('creator_id', filtered=True, size=BIG_NUMBER))
+
+    cache_key = 'top_contributors_questions_{}_{}_{}_{}_{}_{}'.format(start, end, locale,
+                                                                      product, count, page)
+    cached = cache.get(cache_key, None)
+    if use_cache and cached:
+        return cached
+
+    query = AnswerMetricsMappingType.search()
 
     # Adding answer to your own question, isn't a contribution.
     query = query.filter(by_asker=False)
-
     query = _apply_filters(query, start, end, locale, product)
 
-    return _get_creator_counts(query, count, page)
+    answers = [q.id for q in query.all()[:HUGE_NUMBER]]
+    users = (User.objects
+             .filter(answers__in=answers)
+             .annotate(query_count=Count('answers'))
+             .order_by('-query_count'))
+
+    counts = _get_creator_counts(users, count, page)
+    if use_cache:
+        cache.set(cache_key, counts, 60*15)  # 15 minutes
+    return counts
 
 
-def top_contributors_kb(start=None, end=None, product=None, count=10, page=1):
+def top_contributors_kb(start=None, end=None, product=None, count=10, page=1, use_cache=True):
     """Get the top KB editors (locale='en-US')."""
     return top_contributors_l10n(
-        start, end, settings.WIKI_DEFAULT_LANGUAGE, product, count)
+        start, end, settings.WIKI_DEFAULT_LANGUAGE, product, count, use_cache)
 
 
 def top_contributors_l10n(start=None, end=None, locale=None, product=None,
-                          count=10, page=1):
+                          count=10, page=1, use_cache=True):
     """Get the top l10n contributors for the KB."""
+
+    cache_key = 'top_contributors_l10n_{}_{}_{}_{}_{}_{}'.format(start, end, locale,
+                                                                 product, count, page)
+    cached = cache.get(cache_key, None)
+    if use_cache and cached:
+        return cached
+
     # Get the user ids and contribution count of the top contributors.
-    query = (
-        RevisionMetricsMappingType
-        .search()
-        .facet('creator_id', filtered=True, size=BIG_NUMBER))
+    query = RevisionMetricsMappingType.search()
 
     if locale is None:
         # If there is no locale specified, exclude en-US only. The rest are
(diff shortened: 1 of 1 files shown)


kitsune/community/utils.py
```diff
@@ -1,75 +1,116 @@
...
     # Get the user ids and contribution count of the top contributors.
-    query = (
-        AnswerMetricsMappingType
-        .search()
-        .facet('creator_id', filtered=True, size=BIG_NUMBER))
+
+    cache_key = 'top_contributors_questions_{}_{}_{}_{}_{}_{}'.format(start, end, locale,
+                                                                      product, count, page)
+    cached = cache.get(cache_key, None)
+    if use_cache and cached:
```
pmac: Probably a micro-optimization, but you could avoid generating the key and checking the cache when `use_cache` is `False`.

kitsune/community/utils.py
```diff
@@ -1,75 +1,116 @@
...
+
+    query = AnswerMetricsMappingType.search()
 
     # Adding answer to your own question, isn't a contribution.
     query = query.filter(by_asker=False)
-
     query = _apply_filters(query, start, end, locale, product)
 
-    return _get_creator_counts(query, count, page)
+    answers = [q.id for q in query.all()[:HUGE_NUMBER]]
```
safwanrahman: There are a `BIG_NUMBER`, why introduce another one?

safwanrahman: Oh. got the point actually

safwanrahman:
[Test Fail](https://travis-ci.org/mozilla/kitsune/jobs/254392754) seems related!

kitsune/community/utils.py
safwanrahman: I think it can be like this. More faster to query I believe!
```python
users_id = [q.creator_id for q in query.all()[:HUGE_NUMBER]]
user = User.objects.filter(id__in=set(users_id)).annotate(query_count=Count('answers')).order_by('-query_count')
```

kitsune/community/utils.py
```diff
@@ -95,40 +136,27 @@ def _apply_filters(query, start, end, locale=None, product=None):
...
-
-        user_lookup[r['id']] = r
-
-    # Add the user to each dict in the creator_counts array.
-    for item in creator_counts:
-        item['user'] = user_lookup.get(item['term'], None)
-
-    return ([item for item in creator_counts if item['user'] is not None],
-            total)
+    total = query.count()
```
safwanrahman: As we must need to load the `user` into memory, lets call ``len()``. ``count()`` will make another sql query

kitsune/community/utils.py
```diff
@@ -95,40 +136,27 @@ def _apply_filters(query, start, end, locale=None, product=None):
...
+    results = []
+    now = datetime.now()
```
safwanrahman: More optimization can be done here by using `only` and `select_related` like
query = query.select_related('profile').only('id', 'username')

kitsune/community/utils.py
```diff
@@ -1,75 +1,116 @@
...
-        .facet('creator_id', filtered=True, size=BIG_NUMBER))
+    query = RevisionMetricsMappingType.search()
 
     if locale is None:
         # If there is no locale specified, exclude en-US only. The rest are
         # l10n.
         query = query.filter(~F(locale=settings.WIKI_DEFAULT_LANGUAGE))
 
     query = _apply_filters(query, start, end, locale, product)
+    revisions = [q.id for q in query.all()[:HUGE_NUMBER]]
```
safwanrahman: Here also can be implemented like above, I think

kitsune/community/utils.py
```diff
@@ -95,40 +136,27 @@ def _apply_filters(query, start, end, locale=None, product=None):
...
+        last_contribution_date = user.profile.last_contribution_date
+        days_since_last_activity = None
+        if last_contribution_date:
+            days_since_last_activity = now - last_contribution_date
+
+        data = {
+            'count': user.query_count,
+            'term': user.id,
+            'user': {
+                'username': user.username,
```
safwanrahman: here should be `user.id` also
https://github.com/mozilla/kitsune/blob/master/kitsune/community/jinja2/community/index.html#L163

kitsune/community/utils.py
```diff
@@ -1,75 +1,116 @@
...
-                               count=10, page=1):
+                               count=10, page=1, use_cache=True):
     """Get the top Support Forum contributors."""
     # Get the user ids and contribution count of the top contributors.
-    query = (
-        AnswerMetricsMappingType
-        .search()
-        .facet('creator_id', filtered=True, size=BIG_NUMBER))
+
+    cache_key = 'top_contributors_questions_{}_{}_{}_{}_{}_{}'.format(start, end, locale,
```
safwanrahman: There should be unicode data. There is a test fail because of that!
```python
ERROR: test_product_filter (kitsune.questions.tests.test_templates.QuestionsTemplateTestCase)
  File "/app/kitsune/questions/tests/test_templates.py", line 1146, in test_product_filter
    check(p1.slug, [q2])
  File "/app/kitsune/questions/tests/test_templates.py", line 1132, in check
    response = self.client.get(url)
  File "/usr/local/lib/python2.7/dist-packages/django/test/client.py", line 500, in get
    **extra)
  File "/usr/local/lib/python2.7/dist-packages/django/test/client.py", line 303, in get
    return self.generic('GET', path, secure=secure, **r)
  File "/usr/local/lib/python2.7/dist-packages/django/test/client.py", line 379, in generic
    return self.request(**r)
  File "/app/kitsune/sumo/tests/__init__.py", line 116, in request
    return super(LocalizingClient, self).request(**request)
  File "/usr/local/lib/python2.7/dist-packages/django/test/client.py", line 466, in request
    six.reraise(*exc_info)
  File "/usr/local/lib/python2.7/dist-packages/django/core/handlers/base.py", line 132, in get_response
    response = wrapped_callback(request, *callback_args, **callback_kwargs)
  File "/usr/local/lib/python2.7/dist-packages/mobility/decorators.py", line 23, in wrapper
    return f(request, *args, **kw)
  File "/app/kitsune/questions/views.py", line 301, in question_list
    locale=forum_locale, product=product)
  File "/app/kitsune/community/utils.py", line 30, in top_contributors_questions
    cached = cache.get(cache_key, None)
  File "/usr/local/lib/python2.7/dist-packages/django/core/cache/backends/memcached.py", line 83, in get
    key = self.make_key(key, version=version)
  File "/usr/local/lib/python2.7/dist-packages/django/core/cache/backends/memcached.py", line 76, in make_key
    return force_str(super(BaseMemcachedCache, self).make_key(key, version))
  File "/usr/local/lib/python2.7/dist-packages/django/core/cache/backends/base.py", line 102, in make_key
    new_key = self.key_func(key, self.key_prefix, version)
  File "/usr/local/lib/python2.7/dist-packages/django/core/cache/backends/base.py", line 35, in default_key_func
    return '%s:%s:%s' % (key_prefix, version, key)
UnicodeDecodeError: 'ascii' codec can't decode byte 0xc4 in position 43: ordinal not in range(128)
```

safwanrahman: adding `u'top_contributors_questions_{}_{}_{}_{}_{}_{}' would fix this!

kitsune/community/utils.py
glogiotatidis: Unfortunately this won't work as expected as it won't respect the `start`, `end`, `locale` and `product` filters. Very good thought though.

kitsune/community/utils.py
glogiotatidis: Same here as above

kitsune/community/utils.py
safwanrahman: @glogiotatidis Why it wont respect the filters? Does not it suppose to do so?

kitsune/community/utils.py
glogiotatidis: Hmm not really. We will need the relations to `create_revisions` or `tweet_replies` or other things counted. So the elephant in the room in the loading of ES results into memory, I wouldn't worry too much about db optimization at this point as it can complicate code.

kitsune/community/utils.py
glogiotatidis: Your query will operate on user ids and not on answers. Therefore it will count all answers from a specific user and not the exact answers we selected from ES

glogiotatidis:
Thanks both!

PR #2888, opened 2017-07-18
Adding Malay to available SUMO locales

vesper-mozilla:
Having completed the UI translation, the Malay community requested re-enabling Malay content on SUMO.

PR #2904, opened 2017-08-01
Fixing unittest based on the changes of configuration

safwanrahman:
The unittest were failing for some reason. This PR will fix that
@glogiotatidis r?

### DIFF
--- a/kitsune/sumo/tests/test_googleanalytics.py
+++ b/kitsune/sumo/tests/test_googleanalytics.py
@@ -32,7 +32,7 @@ class GoogleAnalyticsTests(TestCase):
         visits = googleanalytics.visitors_by_locale(
             date(2013, 01, 16), date(2013, 01, 16))
 
-        eq_(58, len(visits))
+        eq_(59, len(visits))
         eq_(221447, visits['en-US'])
         eq_(24432, visits['es'])
 
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,3 +1,4 @@
+dist: precise
 sudo: false
 language: python
 python: 2.7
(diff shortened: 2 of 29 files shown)


safwanrahman:
Now it pass `flake8`
r? @glogiotatidis

glogiotatidis:
This is in my todo list, I'll review next monday.

kitsune/sumo/tests/test_googleanalytics.py
```diff
@@ -32,7 +32,7 @@ def test_visitors_by_locale(self, _build_request):
         visits = googleanalytics.visitors_by_locale(
             date(2013, 01, 16), date(2013, 01, 16))
 
-        eq_(58, len(visits))
+        eq_(59, len(visits))
```
glogiotatidis: Do you have more info why did this changed?

safwanrahman: `Malay` locale was added in kitsune with #2888.
Therefore, there are total 59 locale.

glogiotatidis: Hmmm OK but the test is limited to Jan 2013 which is before this change. Also this test doesn't seem to fail anymore

glogiotatidis: I updated the test in #2965 with this change. Thanks!

glogiotatidis:
Thank you for your time @safwanrahman. A lot of changes in this PR will be deprecated once #2965 gets merged. It also contains a lot of pep8 changes which are not required. 

Besides those it still has a number of useful improvements related to the search tests. I'd welcome a future PR -after #2965 gets merged- with those updates.

PR #2905, opened 2017-08-02
Update and serve contribute.json

glogiotatidis:
Ops asked to update this for [bug 1383785](https://bugzilla.mozilla.org/show_bug.cgi?id=1383785)

pmac please r?

### DIFF
--- a/kitsune/urls.py
+++ b/kitsune/urls.py
@@ -3,6 +3,7 @@ from django.conf import settings
 from django.views.i18n import javascript_catalog
 from django.views.decorators.cache import cache_page
 from django.views.generic.base import RedirectView
+from django.views.static import serve as servestatic
 
 import authority
 import badger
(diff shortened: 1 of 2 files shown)


kitsune/urls.py
```diff
@@ -84,6 +85,9 @@
 
     # These API urls include both v1 and v2 urls.
     (r'^api/', include('kitsune.users.urls_api')),
+
+    # contribute.json url
+    url(r'^(?P<path>contribute\.json)$', servestatic, kwargs={'document_root': settings.ROOT}),
```
safwanrahman: We used to route this kind of url to `kitsune.sumo.urls`
[here](https://github.com/mozilla/kitsune/blob/ae1f5ff5b160d460f7c45005c1f6ec50b18ad241/kitsune/sumo/urls.py#L15)
Like we server `robots.txt` from the route. I think this url rule should live there

glogiotatidis: Thanks for the tip @safwanrahman. I like to have things that reply on `/` on the very top `urls.py` namespace, otherwise it's harder to find like in this case.  It seems that we have 404/500 handlers here too.  I'd probably move `robots.txt` here in the future.

PR #2910, opened 2017-08-14
Changing locale name for Malay as requested by the community

vesper-mozilla:
The change should update the existing Arabic script to a Latin one, as
per a community request from the localizers.


glogiotatidis:
ack!

PR #2911, opened 2017-08-15
[Bug 1385087] Add firefox download button for non fx user

safwanrahman:
So it will show Firefox Download Button to non fx users in all pages except Firefox Landing page because it already have a download firefox button.
@pmac @bensternthal r?


kitsune/sumo/jinja2/base.html
```diff
@@ -244,6 +244,11 @@
   {% javascript script %}
 {% endfor %}
 
+{# Show firefox download button to unauthenticated users only #}
+{% if not user.is_authenticated() %}
```
pmac: Should be the same logic from line 141 I think

```jinja
{% if not (hide_fx_download or user.is_authenticated()) %}
```

safwanrahman:
@pmac Updated. r?

PR #2914, opened 2017-08-31
[Bug 1219716] Expand a string shown above articles with l10n in progress

MikkCZ:
The change is basically a copy&past what's above for completely untranslated articles.

@vesper-mozilla does the text look good, or do you want to chime it a little bit?


vesper-mozilla:
The text of the message LGTM - thank you, @MikkCZ!

babybg69:
is this a virus?

glogiotatidis:
Thanks!

PR #2918, opened 2017-09-20
Update SHA hashes after GitHub Change.

glogiotatidis:
GitHub changed the way it produces tarballs which results to different hashes
for same packages.

Other projects are experiencing the same issue too:

  https://github.com/kubernetes/kubernetes/issues/52307

pushed to stage just fine.


glogiotatidis:
Thanks!

PR #2919, opened 2017-09-22
[bug 1388942] Add snowball analyzer in document mapping 

glogiotatidis:
In ESv2 document mapping must include the 'snowball' analyzer to work with searches that use it. This change brings Wiki Document mappings in part with Question mappings which already define the analyzer. 

This also brings back support support synonims

### DIFF
--- a/kitsune/search/config.py
+++ b/kitsune/search/config.py
@@ -1,3 +1,3 @@
 ES_SYNONYM_LOCALES = [
-
+    'en-US',
 ]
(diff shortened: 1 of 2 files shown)


glogiotatidis:
@pmac I already re-indexed using this code and appears to be working fine. Waiting for confirmation that this is the case from the SUMO folks as well.

Meanwhile could you please r?

glogiotatidis:
Thanks!

safwanrahman:
`snowball` analyzer was included in document mapping but later removed by #1527 in order to support analyzer for other locale.
I have discovered the issue while working in #2904. [2 tests are failing in Elasticsearch V1](https://travis-ci.org/mozilla/kitsune/jobs/280906677#L3108). Looking over the failing, it seems that the analyzer for other language is not working.

PR #2801, opened 2016-03-04
[bug 1253528] Fix CSRF on mobile login.

Osmose:
r?

PR #2924, opened 2017-10-05
[Bug 1405844] User Registration on mobile fails

safwanrahman:
It should fix the issue. Similar to #2801
@glogiotatidis @pmac r?

### DIFF
--- a/kitsune/questions/jinja2/questions/mobile/new_question_login.html
+++ b/kitsune/questions/jinja2/questions/mobile/new_question_login.html
@@ -6,7 +6,7 @@
 {% set hide_extra_content = True %}
 {% set loginform = login_form %}
 {% from "layout/errorlist.html" import errorlist %}
-{% from "questions/includes/aaq_macros.html" import mobile_register_forms %}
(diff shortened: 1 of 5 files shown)


glogiotatidis:
Thanks @safwanrahman 

I tried a similar fix yesterday and it didn't work. I now pushed yours to staging (you're using `with context`) which I didn't and it still doesn't work. `csrf` is not included in the POSTed data.

Does this work for you locally?

safwanrahman:
@glogiotatidis Its working locally for me. Can you test this in your local machine? have you restart the server in stage while deploying? template change may not detected without a restart.

safwanrahman:
@glogiotatidis Sorry, its my bad. I thought the AAQ registration flow and normal registration flow were using the same template. Therefore I was testing the registration process only locally.
Looking at the code and workflow, it seems that both are using different template. I have fixed that in the latest commit and tested the AAQ flow locally in mobile view. it seems to be working now.
Moreover, I have added another test that would test the functionality.
r?

glogiotatidis:
Verified working. Thanks @safwanrahman !

PR #2951, opened 2017-11-08
Update Header & Footer

bensternthal:
@glogiotatidis here is an updated header/mini-header and footer.

Please do some testing.. my html/css skillz are quite rusty. And if we need to pull in a front end engineer let me know.


glogiotatidis:
Header renders with pixels for me and the scroll view seems a bit off? I know it's SVG and that shouldn't be the case so probably it has something to do with the SVG processing. I wonder what our frontend wizards have say about this?

https://screenshots.firefox.com/H6yuP7M2PwvkJ3cG/support.allizom.org
https://screenshots.firefox.com/br2yxi5afQzmegjh/support.allizom.org

Also I see you removed the 'download firefox' button from the scroll header, I guess to make some space. Will that impacts downloads originating from SUMO?

I pushed our PR to stage so we can preview easily. Thoughts?

bensternthal:
OK let me loop in some experts. Perhaps it is how I exported the SVG.

bensternthal:
@glogiotatidis can you view the logo directly and let me know if it is still jaggy:

https://support.allizom.org/static/sumo/img/mozilla-support.1dcea6dea6d7.svg

glogiotatidis:
> @glogiotatidis can you view the logo directly and let me know if it is still jaggy:

yap, still jaggy

bensternthal:
Bah! Ok let me see if I can get this to look right with the PNG creative provided.. will also request a SVG from them. Could be that I am creating the SVG from the PNG and not the source.

bensternthal:
@glogiotatidis ok converted to PNG. Let me know how it looks on your end.

glogiotatidis:
Looks good now, thanks!

PR #2952, opened 2017-11-08
Fix #2899: Use s3 for user media

### LINKED ISSUE #2899: Switch user media to s3 + CDN
Uploaded media needs to go somewhere external to the servers in AWS. S3 is good for this with a CDN (probably CloudFront) in front.

pmac:
Will require changes to server config

### DIFF
--- a/kitsune/settings.py
+++ b/kitsune/settings.py
@@ -550,6 +550,7 @@ INSTALLED_APPS = (
     'authority',
     'timezones',
     'waffle',
+    'storages',
     'kitsune.access',
     'kitsune.sumo',
     'kitsune.search',
@@ -756,6 +757,16 @@ MAX_FILENAME_LENGTH = 200
 MAX_FILEPATH_LENGTH = 250
 # Default storage engine - ours does not preserve filenames
 DEFAULT_FILE_STORAGE = 'kitsune.upload.storage.RenameFileStorage'
+
+# AWS S3 Storage Settings
+# local_settings must include the following to use s3 storage
+# AWS_ACCESS_KEY_ID
+# AWS_SECRET_ACCESS_KEY
+# AWS_STORAGE_BUCKET_NAME
+AWS_S3_OBJECT_PARAMETERS = {
+    'CacheControl': 'max-age=86400',
+}
+AWS_S3_CUSTOM_DOMAIN = 'support.cdn.mozilla.net'
 
 # Auth and permissions related constants
 LOGIN_URL = '/users/login'
--- a/kitsune/upload/storage.py
+++ b/kitsune/upload/storage.py
@@ -3,7 +3,10 @@ import itertools
 import os
 import time
 
-from django.core.files.storage import FileSystemStorage as DjangoStorage
+try:
+    from storages.backends.s3boto3 import S3Boto3Storage as DjangoStorage
+except ImportError:
+    from django.core.files.storage import FileSystemStorage as DjangoStorage
 
 
 class RenameFileStorage(DjangoStorage):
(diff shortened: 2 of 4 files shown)


pmac:
I've not tested this fully, but it should work, or is at least a good start. It'd be nice to have a good trial run of this. Do we have an instance in scl3 we can deploy this to for testing @glogiotatidis ?

kitsune/upload/storage.py
```diff
@@ -3,7 +3,10 @@
 import os
 import time
 
-from django.core.files.storage import FileSystemStorage as DjangoStorage
+try:
+    from storages.backends.s3boto3 import S3Boto3Storage as DjangoStorage
```
safwanrahman: It will use S3 in local instance?

safwanrahman:
As its going to work for AWS, I think its better to first push the `docker` branch things to master, than rebase according to it. We have decided to use `pip` instead of `peep`. So the requirement file change will not be relavant.

pmac:
@safwanrahman that's true, but we're going to push this change to the current datacenter first to make the transition to AWS more smooth. I'll port this change to master once it's been successful there.

kitsune/upload/storage.py
pmac: Only if you have django-storages and boto3 installed. We could also add a setting when this moves to the docker/master branch.

kitsune/settings.py
```diff
@@ -757,6 +758,16 @@ def JINJA_CONFIG():
 # Default storage engine - ours does not preserve filenames
 DEFAULT_FILE_STORAGE = 'kitsune.upload.storage.RenameFileStorage'
 
+# AWS S3 Storage Settings
+# local_settings must include the following to use s3 storage
+# AWS_ACCESS_KEY_ID
+# AWS_SECRET_ACCESS_KEY
+# AWS_STORAGE_BUCKET_NAME
+AWS_S3_OBJECT_PARAMETERS = {
+    'CacheControl': 'max-age=86400',
```
safwanrahman: Current max age is `2592000`

glogiotatidis:
I did some preliminary testing on this and seems to be OK. I'll continue extensivelly testing and try to deploy on staging on monday.

kitsune/settings.py
```diff
@@ -757,6 +758,16 @@ def JINJA_CONFIG():
...
+}
+AWS_S3_CUSTOM_DOMAIN = 'support.cdn.mozilla.net'
```
glogiotatidis: Let's change this to the production cdn for user media `prod-cdn.sumo.mozilla.net`.

If we decide to push static to s3 we will need to create 2 different storage classes to set two different custom domains.

kitsune/settings.py
glogiotatidis: Please keep this the same value unless there's a reason to change it.

glogiotatidis:
I tested by creating with questions but as @safwanrahman pointed out uploading images in the gallery does not work.

There 're two related Sentry Errors:

 1. https://sentry.prod.mozaws.net/operations/sumo-stage/issues/688144/
 2. https://sentry.prod.mozaws.net/operations/sumo-stage/issues/688123/

which I made some progress fixing in https://github.com/mozilla/kitsune/commit/18a2009f5d9ef024e6465f99c6c5820ceb921b96  and we now hit

https://sentry.prod.mozaws.net/operations/sumo-stage/issues/688209/

Also `compress_image` task needs some rework to function with s3 objs https://github.com/mozilla/kitsune/blob/18a2009f5d9ef024e6465f99c6c5820ceb921b96/kitsune/upload/tasks.py#L107

pmac:
This one is outdated and replaced by #2983. I'll open a new one for `legacy` branch after we merge the one for master.

PR #2953, opened 2017-11-10
Mobile UI Tweaks Requested By FF UX In Support of Quantom IOS - Color…

bensternthal:
… Changes & Filter Bar Changes

@glogiotatidis this is the PR I mentioned via email, if you can check things out and get on staging tomorrow morning we can let have some folks test.


glogiotatidis:
Thanks Ben

PR #2955, opened 2017-11-10
Fixed Link Color On Mobile Search Result

bensternthal:
@glogiotatidis 

Fixed one issue where on mobile search result sub-header font was white and impossible to read.


glogiotatidis:
Thanks Ben!

PR #2959, opened 2017-11-18
update to latest newrelic

metadave:
https://github.com/mozmeao/infra/issues/640


safwanrahman:
@metadave We use peep instead of pip in SCL3.
You can create the hash with `peepin`

metadave:
note: peep appears to be [deprecated](https://pypi.python.org/pypi/peep)

glogiotatidis:
> note: peep appears to be deprecated

indeed. We'll be switching to pip9 in the master branch

PR #2960, opened 2017-11-18
add latest new relic dep w/ peepin

metadave:
https://github.com/mozmeao/infra/issues/640


glogiotatidis:
Thanks Dave

glogiotatidis:
For some reason this didn't go through in production, despite staging being ok. Given the fragile nature of the current prod deployment I reverted the change. If there's a big benefit from updating the NR client we can discuss and reconsider.

PR #2965, opened 2017-11-20
12Factorize + docker(-compose) and Travis updates

glogiotatidis:
This PR completes the 12Factorization of kitsune, updates Dockerfiles and makes Travis run the tests within docker.

Apologies this ended up being huge, it has a lot of moving bits.

Some tests where skipped to fix the test suite and I created issues to track those in the future


package.json
```diff
@@ -21,7 +21,7 @@
     "gulp-watch": "0.6.8",
     "jquery": "2.1.4",
     "jsdom": "3.1.2",
-    "less": "2.4.0",
+    "less": "1.7.0",
```
glogiotatidis: This is the current version on servers

glogiotatidis:
@safwanrahman if you have the time please take a look too

safwanrahman:
@glogiotatidis Sure, I will check it at tonight.

Dockerfile
```diff
@@ -1,44 +1,35 @@
...
 RUN chown -R kitsune /app
-# npm and bower tries to write in /home directory. So this permission is needed
-RUN chown -R kitsune /home
 USER kitsune
-
-RUN npm install
-RUN ./node_modules/.bin/bower install
-
-ENV WEB_CONCURRENCY=4
-CMD ["gunicorn", "--bind", "0.0.0.0:8000", "--timeout=120", "kuma.wsgi:application"]
```
safwanrahman: gunicorn is needed for starting a webserver

.travis.yml
```diff
@@ -1,26 +1,42 @@
...
+    - AXES_BEHIND_REVERSE_PROXY=False
+    - PIPELINE_ENABLED=True
+    - ES_LIVE_INDEXING=False
+    - ALLOWED_HOSTS=*
+    - BROKER_URL=redis://redis:643
+    - REDIS_DEFAULT_URL=redis://redis:6379/1
+    - REDIS_HELPFULVOTES_URL=redis://redis:6379/2
+    - CACHE_URL=redis://redis:6379/3
+    - CSRF_COOKIE_SECURE=False
+    - DATABASE_URL=mysql://root:kitsune@mariadb:3306/kitsune
```
safwanrahman: If not set, it would take from the `.env-dist`

.travis.yml
```diff
@@ -1,26 +1,42 @@
...
+    - DOMAIN=localhost
+    - ES_URLS=elasticsearch:9200
+    - SITE_URL=http://localhost:8000
+    - SECRET_KEY=secret
+services:
+  - docker
+install:
+  - pip install -U docker-compose
+  - env > .env
```
safwanrahman: rather than doing this, you can use `env_file` configuration option in `docker-compose`

.travis.yml
```diff
@@ -1,26 +1,42 @@
...
+  - docker-compose up -d mariadb
```
safwanrahman: It can be simplified like this
`docker-compose up -d mariadb elasticsearch redis`

.travis.yml
```diff
@@ -1,26 +1,42 @@
...
+  - docker-compose up -d elasticsearch
+  - docker-compose up -d redis
+  - docker-compose build
+  - docker-compose run --user root web ./manage.py nunjucks_precompile
```
safwanrahman: rathar than running as root you can export `UID` into environment variable and it will run with the host uid. So there will be no problem in creating files.
like
```
export UID
docker-compose run web ./manage.py nunjucks_precompile
```

kitsune/settings.py
```diff
@@ -67,22 +59,16 @@ def path(*parts):
...
 }
 
 # Setting this to the Waffle version.
 WAFFLE_CACHE_PREFIX = 'w0.7.7a:'
 
 # Addresses email comes from
-DEFAULT_FROM_EMAIL = 'notifications@support.mozilla.org'
-DEFAULT_REPLY_TO_EMAIL = 'no-reply@mozilla.org'
-SERVER_EMAIL = 'server-error@support.mozilla.org'
-EMAIL_SUBJECT_PREFIX = '[support] '
```
safwanrahman: Any particular reason to remove this?

kitsune/settings.py
```diff
@@ -407,7 +389,7 @@ def path(*parts):
 )
 
 # Make this unique, and don't share it with anybody.
-SECRET_KEY = '#%tc(zja8j01!r#h_y)=hy!^k)9az74k+-ib&ij&+**s3-e^_z'
+SECRET_KEY = config('SECRET_KEY')
```
safwanrahman: you should add some default value

kitsune/settings.py
```diff
@@ -475,15 +457,9 @@ def path(*parts):
...
 
-    # This should come before TokenLoginMiddleware, because
-    # TokenLoginMiddleware uses this to tell users they have been
-    # automatically logged. It also has to come after
-    # NoVarySessionMiddleware.
+    # This has to come after NoVarySessionMiddleware.
     'django.contrib.messages.middleware.MessageMiddleware',
 
-    # This middleware should come after AuthenticationMiddleware.
-    'kitsune.users.middleware.TokenLoginMiddleware',
```
safwanrahman: Whats the reason to remove this?

kitsune/settings.py
```diff
@@ -518,7 +493,6 @@ def path(*parts):
 # Auth
 AUTHENTICATION_BACKENDS = (
     'django.contrib.auth.backends.ModelBackend',
-    'kitsune.users.auth.TokenLoginBackend',
```
safwanrahman: this one also

kitsune/settings.py
```diff
@@ -706,25 +681,26 @@ def JINJA_CONFIG():
     PIPELINE['BROWSERIFY_ARGUMENTS'] += ' -d'
 
 NUNJUCKS_PRECOMPILE_BIN = 'nunjucks-precompile'
```
safwanrahman: repetative

kitsune/settings.py
pmac: Nope. Not adding a default value makes it error on startup, which we want if no unique secret key is set. It's a security risk to check in a value that may accidentally be used in prod.

kitsune/settings.py
pmac: It was just moved later in the file.

Dockerfile
glogiotatidis: Yes, but this is not a production ready image. gunicorn with configuration will come in a future PR.

kitsune/settings.py
glogiotatidis: This was a mistake, restored. Nice catch!

.travis.yml
glogiotatidis: I want to have a Travis specific environment, so i don't want to hardcode this env into docker-compose. For the time being I believe I'll stick with different `.env` files. We will iterate on this and see what fits best.

.travis.yml
glogiotatidis: tried that without luck in https://travis-ci.org/mozilla/kitsune/builds/305240903

I 'll welcome a PR to fix this.

glogiotatidis:
Thanks both!

PR #2969, opened 2017-11-22
Upgrade to latest newrelic.

glogiotatidis:


glogiotatidis:
Removed the `server.txt` requirements file. @pmac how do you feel about merging all requirements into one file?

pmac:
I'm of two minds on this.

1. I really like simplifying things like this and having a single file is pretty great.
2. It does seem silly to have a dev user or docker image install newrelic and other "prod" specific things that are of no use in other contexts. I'd be okay with two files, one for "prod" and one for everything else.

So I could be convinced either way. If we don't have many "prod only" things then let's just have the one file. If it is significantly more than one prod-only thing then maybe we should have two. From the look of the current `server.txt` file I'm 👍  on removing it.

glogiotatidis:
> I'm of two minds on this.
> 
>     I really like simplifying things like this and having a single file is pretty great.
>     It does seem silly to have a dev user or docker image install newrelic and other "prod" specific things that are of no use in other contexts. I'd be okay with two files, one for "prod" and one for everything else.
> 
> So I could be convinced either way. If we don't have many "prod only" things then let's just have the one file. If it is significantly more than one prod-only thing then maybe we should have two. From the look of the current server.txt file I'm 👍 on removing it.

I hear you. What's interesting is that we currently have `dev.txt` and `test.txt` which seems to over-complicate things. Let's make things better, i.e. simple overall, better for dev, more secure for prod.

PR #2970, opened 2017-11-22
Python docker caching

glogiotatidis:


pmac:
Nice! Looks like good stuff. More cache plz 💯

.travis.yml
```diff
@@ -3,13 +3,28 @@ language: python
 python: 2.7
 services:
   - docker
+before_install:
+  - echo "ENV GIT_SHA ${TRAVIS_COMMIT}" >> Dockerfile
```
safwanrahman: Is it possible to do it with build arguments?

safwanrahman:
So the approach can be something like building a docker image with git hash tag, and use the tag only to start the container with that hash.
I will try with another approach today.

glogiotatidis:
Thanks!

PR #2971, opened 2017-11-22
Add jenkins pipeline for docker image building 

glogiotatidis:
Adding a Jenkinsfile to build images and push to docker hub.

@davehunt @stephendonner I moved your Jenkinsfile under `tests/` along with the rest of WebQA stuff. Is this ok?

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -1,175 +1,62 @@
-import groovy.json.JsonOutput
+@Library('github.com/mozmeao/jenkins-pipeline@20171123.1')
+def config
+def docker_image
 
-/** Map of Tox environments and associated capabilities */
-def environments = [
-  desktop: [
-    browserName: 'Firefox',
-    version: '47.0',
-    platform: 'Windows 7'
-  ],
-  mobile: [
-    browserName: 'Browser',
-    platformName: 'Android',
-    platformVersion: '5.1',
-    deviceName: 'Android Emulator',
-    appiumVersion: '1.5.3'
-  ]
-]
-
-/** Transform a map into a list of maps
- *
- * @param aMap map to transform
- * @return a list of maps
-*/
-@NonCPS
-def entrySet(aMap) {
-  aMap.collect {
-    k, v -> [key: k, value: v]
-  }
-}
-
-/** Write capabilities to JSON file
- *
- * @param desiredCapabilities capabilities to include in the file
-*/
-def writeCapabilities(desiredCapabilities) {
-    def defaultCapabilities = [
-        build: env.BUILD_TAG,
-        public: 'public restricted'
-    ]
-    def capabilities = defaultCapabilities.clone()
-    capabilities.putAll(desiredCapabilities)
-    def json = JsonOutput.toJson([capabilities: capabilities])
-    writeFile file: 'capabilities.json', text: json
-}
-
-/** Run Tox
- *
- * @param environment test environment to run
-*/
-def runTox(environment) {
-  def processes = env.PYTEST_PROCESSES ?: 'auto'
-  try {
-    wrap([$class: 'AnsiColorBuildWrapper']) {
-      withCredentials([[
-        $class: 'StringBinding',
-        credentialsId: 'SAUCELABS_API_KEY',
-        variable: 'SAUCELABS_API_KEY']]) {
-        withEnv(["PYTEST_ADDOPTS=${PYTEST_ADDOPTS} " +
-          "-n=${processes} " +
-          "--base-url=${PYTEST_BASE_URL} " +
-          "--driver=SauceLabs " +
-          "--variables=capabilities.json " +
-          "--color=yes"]) {
-          sh "tox -e ${environment}"
-        }
-      }
-    }
-  } catch(err) {
-    currentBuild.result = 'FAILURE'
-    throw err
-  } finally {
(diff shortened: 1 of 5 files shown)


stephendonner:
@glogiotatidis yep; we don't actively use this, and regardless, that's where it lives in nearly all of our other in-development-repo projects.  Thanks!

Jenkinsfile
```diff
@@ -1,175 +1,60 @@
...
+
+    stage("Build") {
+      if (!dockerImageExists(docker_image)) {
+            sh "echo 'ENV GIT_SHA ${GIT_COMMIT}' >> Dockerfile"
+
+            // Docker always copies data as root. Chowning files to root before
+            // the build command will prevent docker cache invalidation due to
+            // different file metadata. When we upgrade to a newer docker
+            // version we should probably use --chown flag for COPY. See
+            // https://github.com/moby/moby/pull/34263
```
pmac: This is fascinating. Thanks! I've been wondering why this didn't seem to work quite right sometimes. But it does seem to use cache for me locally. Why is that?

jenkins.yml
```diff
@@ -0,0 +1,10 @@
+project:
+  name: kitsune
+  docker_name: mozorg/kitsune
```
pmac: We have a `mozmeao` docker org as well that I think is the place for things in future. The "mozorg" org is primarly for bedrock and I think bedrock should move as well. Not a blocker, just something I think we should eventually do.

Jenkinsfile
pmac: Ah! I get it. It works fine from the same system, but when building from another system it'll not use cache because it was built with a different user on a different machine. Interesting.

Jenkinsfile
```diff
@@ -1,175 +1,60 @@
...
-            withEnv(["PYTEST_ADDOPTS=--variables=${VARIABLES}"]) {
-              runTox(environment)
-            }
-          }
-        }
+    docker_image = "${config.project.docker_name}:${GIT_COMMIT_SHORT}"
+
+    stage("Build") {
+      if (!dockerImageExists(docker_image)) {
+            sh "echo 'ENV GIT_SHA ${GIT_COMMIT}' >> Dockerfile"
```
safwanrahman: I think its kinda hacky way to accomplish the task. We could use build argument to pass git commit do dockerfile.

pmac:
We will do that. For now the docker version on our Jenkins box is too old I
think.
-- 
- pmac

safwanrahman:
@pmac I think its updated enough to support build argument. in kuma, we are build argument.

glogiotatidis:
> @glogiotatidis yep; we don't actively use this

@stephendonner does that mean that you don't test SUMO at all?

Jenkinsfile
```diff
@@ -1,175 +1,60 @@
...
+
+            // Docker always copies data as root. Chowning files to root before
+            // the build command will prevent docker cache invalidation due to
+            // different file metadata. When we upgrade to a newer docker
+            // version we should probably use --chown flag for COPY. See
+            // https://github.com/moby/moby/pull/34263
+            dockerRun('debian', [
+                    "docker_args": "-v `pwd`:/app",
+                    "cmd": "cd /app && chown -c root.root -R requirements bower.json package.json && chmod a+rw -R requirements bower.json package.json",
```
glogiotatidis: @pmac fiy i updated this command to 
1. first chown to root to fix the docker bug
2. chmod to a+rw to make jenkins able to alter the files on next build

jenkins.yml
glogiotatidis: Updated

Jenkinsfile
glogiotatidis: Yeah it's... complicated. I had this question too and  it cost me a fair amount to time to get this working.

glogiotatidis:
This builds and pushes images to docker hub. Merging!

davehunt:
@glogiotatidis the Jenkinsfile now in `/tests` has not been used for some time (as far as I know) and hasn't been kept up to date. If I recall correctly we agreed to stop running these tests when SUMO first moved off kitsune. Should we discuss the situation and see if there's a need to restore this coverage? If the tests aren't being run then we should probably remove them so we're not giving a misleading impression of quality/coverage.

glogiotatidis:
> @glogiotatidis the Jenkinsfile now in /tests has not been used for some time (as far as I know) and hasn't been kept up to date. If I recall correctly we agreed to stop running these tests when SUMO first moved off kitsune. Should we discuss the situation and see if there's a need to restore this coverage? If the tests aren't being run then we should probably remove them so we're not giving a misleading impression of quality/coverage.

Since we're keeping Kitsune I suggest that we update those tests and get them running. We're currently working on migrating Kitsune in AWS and we're not developing any features. This work will continue in 2018Q1. Maybe it's a good opportunity to get those fixed during this time as well. Is this something you and your team can take care of right away or do we need to bring it to Patrick to request resources from your team?

davehunt:
I haven't touched these tests for almost 2 years, so it's hard to say how much work is involved in restoring them. If you have a new deployment pipeline, it might make sense to get the tests running in there, rather than the old approach of a separate Jenkins job running daily against a deployed instance? Either way, I suspect requesting resources might be the best approach, allowing us to prioritise it appropriately.

PR #2973, opened 2017-11-23
Move from crontabs to APScheduler. Fixes #2946.

### LINKED ISSUE #2946: Migrate crontabs to APScheduler

glogiotatidis:


kitsune/settings.py
```diff
@@ -1021,3 +1021,40 @@ def read_only_mode(env):
         'environment': config('SENTRY_ENVIRONMENT', default=''),
     }
 }
+
+# Dead Man Snitches
```
pmac: That is a whole lotta snitches... We need to find another way to monitor things like this. We just have too many.

pmac:
Also needs a manual rebase.

PR #2974, opened 2017-11-23
Add gunicorn and meinheld worker. Fixes #2939.

### LINKED ISSUE #2939: Add gunicorn + Meinheld worker

glogiotatidis:


wsgi/config.py
```diff
@@ -0,0 +1,11 @@
+# see http://docs.gunicorn.org/en/latest/configure.html#configuration-file
+
+from os import getenv
+
+
+bind = '0.0.0.0:{}'.format(getenv('PORT', 8000))
+workers = getenv('WSGI_NUM_WORKERS', 2)
+errorlog = '-'
+loglevel = getenv('WSGI_LOG_LEVEL', 'info')
+keepalive = getenv('WSGI_KEEP_ALIVE', 2)
```
pmac: Would it be better here to use decouple.config so you can cast the env var as an int?

wsgi/config.py
```diff
@@ -0,0 +1,11 @@
...
+worker_class = getenv('GUNICORN_WORKER_CLASS', "meinheld.gmeinheld.MeinheldWorker")
```
pmac: NIT: double quotes

wsgi/app.py
```diff
@@ -0,0 +1,29 @@
+"""
```
pmac: Suggestion: If we're moving the wsgi file, the typical Django pattern (these days at least) is to add it to the app package. e.g. `kitsune/wsgi.py`. It's not what all our sites do though... so either way.

pmac:
This is looking good. I don't see a script or command anywhere that will use gunicorn and this wsgi setup though. Could be a good time to add that as well.

wsgi/app.py
glogiotatidis: Since we also have `config.py` probably better to keep them both under `wsgi/`

wsgi/config.py
```diff
@@ -0,0 +1,11 @@
+# see http://docs.gunicorn.org/en/latest/configure.html#configuration-file
+
+from os import getenv
```
glogiotatidis: from decouple import config


bind = '0.0.0.0:{}'.format(config('PORT', default=8000, cast=int))
workers = config('WSGI_NUM_WORKERS', default=2, cast=int)
errorlog = '-'
loglevel = config('WSGI_LOG_LEVEL', default='info')
keepalive = config('WSGI_KEEP_ALIVE', default=2, cast=int)
worker_class = config('GUNICORN_WORKER_CLASS', default='meinheld.gmeinheld.MeinheldWorker')

glogiotatidis: I switched to decouple and then for some strange reason the django app wouldn't read the settings from `.env`. Since this is a small file which I don't expect to change very often I just added `int` to cast the values

glogiotatidis:
> This is looking good. I don't see a script or command anywhere that will use gunicorn and this wsgi setup though. Could be a good time to add that as well.

I intentionally left those out at first b/c we're not even copying the code into the docker image yet. But since you mentioned that I added `./run-common.sh` and `./run-prod.sh` to execute gunicorn.

I also added `Procfile` which is useful with Deis and will be useful with k8s and `honcho` to run the different commands (gunicorn, celery, nginx?) in the future.

pmac:
Great! +1 for `honcho`. Love that project. Sounds like we're not allowed to use Deis though, so you can probably just go ahead and make the Procfile honcho specific whenever you're ready.

wsgi/config.py
pmac: That is really odd... No idea why that would be, but this works fine :)

PR #2975, opened 2017-11-23
Delete settings_test.py and allow configuration of PIPELINE_ENABLED.

glogiotatidis:


kitsune/settings.py
```diff
@@ -1021,3 +1021,5 @@ def read_only_mode(env):
         'environment': config('SENTRY_ENVIRONMENT', default=''),
     }
 }
+
+PIPELINE_ENABLED = config('PIPELINE_ENABLED', default=False)
```
pmac: Need a `cast=bool` kwarg in this one.

glogiotatidis: good catch, updated

Issue #2966, opened 2017-11-21
Improve Travis Build Time


glogiotatidis:
Travis needs 15-20 minutes to complete testing a pull request. We need to improve this time to iterate faster.

Half of the time is spend re-building the image. Current docker version in Travis (17.03) does not support the `cache_from` directive in `docker-compose.yml` so we need to find another way to use pre-built docker images

metadave:
https://cloud.google.com/container-registry/

metadave:
We'll use Jenkins to build and deploy, and Travis to run PR tests.

akatsoulas:
This issue is too old and probably resolved by now. Please open a new issue if this is not the case.

PR #2977, opened 2017-11-24
[Fix #2964  #2966] Run test in travis as non root and fix cache

### LINKED ISSUE #2964: Update travis config to run tests without root user

safwanrahman:
This will add functionality to run test using non root user.
It will also fix the cache invalidation issue. Docker thinks its directory if trailing slash `/` is present in destination path, so it copy the file content. Otherwise it copy the file with metadata. Which cause the cache invalidation without any reason.

I have fixed the things and remove the fixup done in jenkins.

Moreover, Its better to use `mozmeao/kitsune:latest` image in docker compose. otherwise if anyone open a repo with `kitsune`, docker compose will try to fetch from there.

The build cache affect cant be seen in the build of this PR because image with this commits is not pushed yet. Its something like [Chicken and Egg Problem](https://en.wikipedia.org/wiki/Chicken_or_the_egg). So this PR needs te be merged first then Jenkins will make a build and push to docker hub and from next build, we can see the affect.

@glogiotatidis @pmac r?

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,29 +10,20 @@ before_install:
 install:
   - pip install -U docker-compose
   - cp .env-travis .env
+  - export UID
+  - docker build -t mozmeao/kitsune:latest --cache-from mozmeao/kitsune .
   - docker-compose up -d mariadb
   - docker-compose up -d elasticsearch
   - docker-compose up -d redis
+  - docker-compose up -d
 
-  # Docker always copies data as root. Chowning files to root before the build
-  # command will prevent docker cache invalidation due to different file
-  # metadata. Once Travis updates to the latest docker version we should
-  # probably use --chown flag for COPY. See https://github.com/moby/moby/pull/34263
-  - sudo chown -c 0:0 -R requirements package.json bower.json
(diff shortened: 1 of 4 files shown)


glogiotatidis:
@safwanrahman please limit your pull requests to the minimum when possible. This PR contains lots of useful work some of which requires more discussion and from other we can immediately benefit from. 

For example you can PR just the cache fix, which is a two line change, and the all the other dockerfile changes for another PR and all the docker-compose changes in a third. It's not easy to make decisions or discussions on PRs that change multiple things at once.

glogiotatidis:
@safwanrahman I took your suggestions and created another approach using:
 - multi level dockerfiles in #2985 
 - circleci
 - and got npm building to a separate docker image which still installs as root to avoid all the perm issues but it's never deployed to production.

Docker caching of python requirements doesn't work neither with your or mine trick, so if you have any insights please PR. As I mentioned in a previous comment, small focused PRs helps us move faster and code getting merged instead of being blocked due to the many open for discussion subjects.

Thanks for your input!

Issue #2968, opened 2017-11-21
Include locales + build messages in Docker image.


glogiotatidis:

safwanrahman:
Going to work on this, hoping to finish it within this weekend!

glogiotatidis:
Fixed in https://github.com/mozilla/kitsune/pull/2985

PR #2978, opened 2017-11-26
[#2968] Include locales + build messages in Docker image

safwanrahman:
@glogiotatidis r?

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -31,5 +31,12 @@ COPY ./bower.json /app/bower.json
 RUN npm install
 
 RUN ./node_modules/.bin/bower install --allow-root
+
+COPY . /app
+
+# Add locale files
+ADD https://github.com/mozilla-l10n/sumo-l10n/archive/master.tar.gz /app/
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -31,5 +31,12 @@ COPY ./bower.json /app/bower.json
 RUN npm install
 
 RUN ./node_modules/.bin/bower install --allow-root
+
+COPY . /app
+
+# Add locale files
+ADD https://github.com/mozilla-l10n/sumo-l10n/archive/master.tar.gz /app/
+RUN tar --extract --file=master.tar.gz sumo-l10n-master && mv sumo-l10n-master locale
```
glogiotatidis: --strip-components is a very neat tar flag you may want to check out.

glogiotatidis:
I figured out that having locales installed caused the tests to fail. Interestingly this PR has passing tests and this is because it's using a docker-compose file which mounts the code from the host filesystem, which doesn't have the locales. 

This along with other concerns and improvements were tackled in #2985 and make this one obsolete. 

Thanks for your input @safwanrahman

PR #2983, opened 2017-11-29
Fix #2899: Use s3 for user media

### LINKED ISSUE #2899: Switch user media to s3 + CDN

pmac:
This is the same as the PR for legacy, but for master so that we can use docker for testing. Once this is merged I'll back-port to legacy branch for use in scl3.

### DIFF
--- a/docker/dockerfiles/base
+++ b/docker/dockerfiles/base
@@ -21,7 +21,7 @@ RUN apt-get update && apt-get install apt-transport-https && \
             libmariadbclient-dev mariadb-client && \
     rm -rf /var/lib/apt/lists/*
 
-COPY ./requirements/ /app/requirements/
+COPY ./requirements/*.txt /app/requirements/
 
 RUN bash -c '\
     virtualenv /venv && \
--- a/kitsune/settings.py
+++ b/kitsune/settings.py
@@ -536,6 +536,7 @@ INSTALLED_APPS = (
     'authority',
     'timezones',
     'waffle',
+    'storages',
     'kitsune.access',
     'kitsune.sumo',
     'kitsune.search',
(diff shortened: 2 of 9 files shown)


pmac:
Well, it's the same except the issues with the compress and optipng tasks should be fixed.

pmac:
cc @safwanrahman

safwanrahman:
@pmac is it deployed to anywhere? without deployment and aws s3 access key, its quite hard to test this thing.

safwanrahman:
[Test fail](https://travis-ci.org/mozilla/kitsune/builds/308733103#L7860) seems related

pmac:
It's not deployed yet. Once we have the deis app setup sometime this week we'll have a public place to test it. Until then I'm just using some other AWS credentials locally. I'll check out the test.

pmac:
Tests should now be fixed.

kitsune/settings.py
```diff
@@ -744,6 +745,15 @@ def JINJA_CONFIG():
 # Default storage engine - ours does not preserve filenames
 DEFAULT_FILE_STORAGE = 'kitsune.upload.storage.RenameFileStorage'
 
+# AWS S3 Storage Settings
+AWS_ACCESS_KEY_ID = config('AWS_ACCESS_KEY_ID', default='')
+AWS_SECRET_ACCESS_KEY = config('AWS_SECRET_ACCESS_KEY', default='')
+AWS_STORAGE_BUCKET_NAME = config('AWS_STORAGE_BUCKET_NAME', default='')
+AWS_S3_CUSTOM_DOMAIN = config('AWS_S3_CUSTOM_DOMAIN', default='prod-cdn.sumo.mozilla.net')
+AWS_S3_OBJECT_PARAMETERS = {
+    'CacheControl': 'max-age=2592000',
```
safwanrahman: The `max-age` is not same as now

glogiotatidis:
@pmac I setup sumo-dev.frankfurt.moz.works as a deis app with an empty db. You can copy conf to another app and deploy this code. Unfortunately we won't be able to test yet b/c we're blocked by #2984

pmac:
@glogiotatidis I filed #2986 to fix it the hacky way. We can decide to do that or not, but it's probably the way to get it running this week.

kitsune/settings.py
pmac: Looks the same to me. 

```bash
$ curl -I https://support.cdn.mozilla.net/media/uploads/gallery/images/2017-10-11-15-35-09-6b82a5.png | grep Cache-Control
Cache-Control: max-age=2592000
```

pmac:
IT WORKS!!!

https://sumo-s3-media.frankfurt.moz.works/en-US/gallery/images

I uploaded that PNG just now and it seems to have made the thumbnail and optimized the image. Let me know what you think.

docker/dockerfiles/base
```diff
@@ -21,7 +21,7 @@ RUN apt-get update && apt-get install apt-transport-https && \
             libmariadbclient-dev mariadb-client && \
     rm -rf /var/lib/apt/lists/*
 
-COPY ./requirements/ /app/requirements/
+COPY ./requirements/*.txt /app/requirements/
```
glogiotatidis: Is this what we need to get proper requirements caching?

pmac: I'm hoping so. It's kindof a guess as the docs aren't clear on this, but I think it could help since it's the documented way of copying multiple files in a dir.

PR #2985, opened 2017-11-30
Split Dockerfiles + Image building + CircleCI

glogiotatidis:
In this PR:
 - We create split Dockerfiles
   - Smaller final image
   - Node packages not installed in the final image to reduce attach surface
 - Added locales
   - Note that unit tests fail with locales installed. Thus we produce an image without no locales and test against it.
 - Added image build scripts
 - Moved to CircleCI which features the latest and greatest docker engine
   - Builds new images and pushes to docker hub
   - Runs lint, mocha and unit tests.
 - Updated docker-compose configuration:
   - The base `docker-compose.yml` contains minimum conf.
   - Created docker-compose.test.yml with good configuration for running unit tests, i.e. with mounting `.` directory 
   - Created `docker-compose.dev.yml` with good configuration for development. Removed `node_modules` and `bower_components` from volumes. In development you want to have those in host dir. Run with `docker-compose -f docker-compose.yml -f docker/composefiles/docker-compose.dev.yml`
 
Also:
 - Upgraded to Pillow 4.3.0: previous version was throwing `import Image` errors
 - Fix gunicorn path typo
 - Switch to latest stable Whitenoise (v3.3.0)

An image produced by those scripts answers at `https://sumo-dev.frankfurt.moz.works`

I wish this could be a smaller PR but too many interconnected pieces. :|

Oh and docker caching when it comes to `requirements/` is non-sense.


.circleci/config.yml
```diff
@@ -0,0 +1,41 @@
...
+      - image: python:2-stretch
+    steps:
+      - checkout
+      - setup_remote_docker:
+          version: 17.10.0-ce
+      - run:
+          name: Install Docker client
+          command: |
+              set -x
+              curl -L https://get.docker.com/ | sh
```
pmac: Why not use [their python-2 image](https://hub.docker.com/r/circleci/python/) so we get the latest docker and whatnot by default? `circleci/python:2`.

docker/dockerfiles/base
```diff
@@ -15,21 +18,17 @@ RUN apt-get update && apt-get install apt-transport-https && \
...
-RUN pip install --no-cache-dir -r requirements/dev.txt --require-hashes
-RUN pip install --no-cache-dir -r requirements/test.txt --require-hashes
-
-COPY ./package.json /app/package.json
-COPY ./bower.json /app/bower.json
+COPY ./requirements/ /app/requirements/
 
-RUN npm install
+RUN bash -c '\
+    virtualenv /venv && \
```
pmac: Why do we need a virtualenv? I've never run into issues when using the system python in the python images.

pmac: I see. You're copying the venv to other images and this one is not inherited from for the prod image. Interesting.

.circleci/config.yml
glogiotatidis: Yes! Thank you

docker/dockerfiles/base
glogiotatidis: I used a virtualenv to contain everything so I can easily copy them other to the final image. The final docker image is based on top of python-slim and does not include all the compiler / dev libs required to install the python packages. 

This reduces attack surface and limits the size of the final image. win-win!

pmac: Great point! Didn't think of that. I'll look at that technique for bedrock as well. 🎉

PR #2986, opened 2017-11-30
Hack around no longer getting x-forwarded-proto headers

pmac:
In some prod clusters we can assume that it's always HTTPS, so we get that from the environment.

Also rearrange some code and imports.


pmac:
Hmm... tests failed because Circle ran the tests even though there's no circle config in this branch. And Travis it seems didn't run. Odd. This change shouldn't affect the tests though.

pmac:
Agreed. Thanks.

PR #2987, opened 2017-11-30
Fix Bug 1416886 - Update Download Firefox Icon

bensternthal:
Due to some issues with my local docker install I could not test this locally. Can you deploy to stage so I can test? After that we should be good to go.


glogiotatidis:
Almost there! https://screenshots.firefox.com/v4xHHs8Cmr4sIY0k/support.allizom.org

bensternthal:
Ugh. OK let me tweak.

Also let me see if I can get the docker install running again locally so I can avoid having you manually check.

bensternthal:
Was able to get docker running so I could tweak. This version should be OK but please double check.

glogiotatidis:
Yap, looks good. Thanks! 🍺

Issue #2991, opened 2017-12-04
Update development documentation


glogiotatidis:
Development documentation under `docs/` needs to explain how to setup things with Docker.

PR #2992, opened 2017-12-04
Docker development image and instructions

glogiotatidis:
Added quick docker development instructions.  Also filed a ticket #2991 to update the rtfd docs.

Also created a `base-dev` image which install `node`, needed to run `npm install` during development. In the future probably we will want to remove `dev.txt` and `test.txt` requirements from the `base` image and only install them in the `base-dev` image. I thought that there're enough changes already and we can probably test if this dev flow works before making more.


pmac:
Looks like a good start. Also looks like you figured out the ENTRYPOINT issue? 👍

glogiotatidis:
> Also looks like you figured out the ENTRYPOINT issue? 👍

Nope, Deis still appears to ignore it (?!)

pmac:
Damn. Strange. Like I said on IRC I do think we have another option. We just add the `venv`'s `bin` dir to the `PATH` in the image and then all of the python stuff should work. Just using that binary activates the venv enough to import the dependencies.

PR #2994, opened 2017-12-07
Use PATH instead of ENTRYPOINT for virtualenv activation

pmac:


pmac:
Decided to play with this and it seems to be working well. Interested in your thoughts.

glogiotatidis:
Created a build and push to dev and verified that this is working. Nice work @pmac ! 🍻

PR #2995, opened 2017-12-07
Add cache to app volume to improve performance on OSX. 

bensternthal:
See:

https://docs.docker.com/docker-for-mac/osxfs-caching/

and

https://docs.docker.com/compose/compose-file/#caching-options-for-volume-mounts-docker-for-mac


glogiotatidis:
Thanks! 🍄

PR #3005, opened 2018-01-09
Ref #2899: Use s3 for user media on legacy infra

pmac:
This is a port of the changes from #2983 to the legacy branch for use on our legacy infra in SCL3.

### DIFF
--- a/kitsune/settings.py
+++ b/kitsune/settings.py
@@ -550,6 +550,7 @@ INSTALLED_APPS = (
     'authority',
     'timezones',
     'waffle',
+    'storages',
     'kitsune.access',
     'kitsune.sumo',
     'kitsune.search',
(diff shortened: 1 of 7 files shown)


glogiotatidis:
I pushed to stage. Previous user media appear to be working and uploading to gallery also works. \o/

@pmac @safwanrahman can you please test that everything looks OK?

kitsune/settings.py
```diff
@@ -757,6 +758,15 @@ def JINJA_CONFIG():
 # Default storage engine - ours does not preserve filenames
 DEFAULT_FILE_STORAGE = 'kitsune.upload.storage.RenameFileStorage'
 
+# AWS S3 Storage Settings
+AWS_ACCESS_KEY_ID = config('AWS_ACCESS_KEY_ID', default='')
```
safwanrahman: This should not work in SCL3

safwanrahman:
I dont know if its related, but it seems like some image is missing!
like this one
https://stage-cdn.sumo.mozilla.net/uploads/gallery/images/thumbnails/2016-06-19-03-30-16-25fe03.png

screen shot 2018-01-12 at 3 50 59 am

safwanrahman:
This portion need to be rewritten.
https://github.com/mozilla/kitsune/blob/ae1f5ff5b160d460f7c45005c1f6ec50b18ad241/kitsune/sumo/views.py#L208-L230

safwanrahman:
It will not be possible to use the group avatar edit page as the view check the file path exists.
https://github.com/mozilla/kitsune/blob/ae1f5ff5b160d460f7c45005c1f6ec50b18ad241/kitsune/groups/views.py#L84-L87

Need to fix this. I can not test as I do not have admin access in stage. But according to code, it should must fail

kitsune/settings.py
glogiotatidis: correct we already tackled that

glogiotatidis:
> I dont know if its related, but it seems like some image is missing!
> like this one
> https://stage-cdn.sumo.mozilla.net/uploads/gallery/images/thumbnails/2016-06-19-03-30-16-25fe03.png

some images are missing from SCL3 and thus were not transferred to S3.

glogiotatidis:
> It will not be possible to use the group avatar edit page as the view check the file path exists.
> 
> kitsune/kitsune/groups/views.py
> 
> Lines 84 to 87 in ae1f5ff
>  if prof.avatar and os.path.isfile(prof.avatar.path): 
>      # Need to store the path, or else django's 
>      # form.is_valid() messes with it. 
>      old_avatar_path = prof.avatar.path 
> 
> Need to fix this. I can not test as I do not have admin access in stage. But according to code, it should must fail

Good catch. @pmac this raises https://sentry.prod.mozaws.net/operations/sumo-stage/issues/1210673/ when trying to edit photo in https://support.allizom.org/en-US/groups/italian-localization/avatar

PR #3019, opened 2018-01-18
Update READ_ONLY configuration for AWS.

glogiotatidis:


kitsune/settings.py
```diff
@@ -503,13 +517,22 @@ def path(*parts):
     'commonware.middleware.XSSProtectionHeader',
     'commonware.middleware.RobotsTagHeader',
     # 'axes.middleware.FailedLoginMiddleware'
-)
+]
+if READ_ONLY:
```
pmac: You could also just include the middleware in the list and [mark it as unused at runtime](https://docs.djangoproject.com/en/1.8/topics/http/middleware/#marking-middleware-as-unused). Not a requirement, but an option if you're not happy with this. I'm happy either way, and this is perhaps a bit more explicit.

glogiotatidis: I like this, thanks!

glogiotatidis:
@pmac updated with your suggestion

pmac:
Looks great! 💯

glogiotatidis:
many thanks!

Issue #2916, opened 2017-09-15
[Bug 1386808] Shutdown Army of Awesome


bensternthal:
Tracking the shutdown of Army of Awesome.

https://bugzilla.mozilla.org/show_bug.cgi?id=1386808

PR #3020, opened 2018-01-18
Remove links and Cron jobs related to Customer Care (Army of Awesome)

glogiotatidis:
I started removing the related cron jobs to reduce the amount of things we have to babysit in AWS and while I's at it I removed the links to AoA (Army of Awesome).

Removing the code entirely is bit more involved are there're references all over the place.

See also #2916


glogiotatidis:
Thanks!

PR #3032, opened 2018-01-29
Fix #3023: Add liveness and readiness URLs

### LINKED ISSUE #3023: Add liveness and readiness endpoints
Add `/healthz/` and `/readiness/` to check the health of the app and it's ability to serve traffic. 

Needed for the k8s deployment.

See also mdn's example https://github.com/mozilla/kuma/blob/master/kuma/health/views.py

pmac:


kitsune/sumo/views.py
```diff
@@ -26,11 +27,54 @@
...
 
 
+# copied from
+# https://github.com/mozilla/kuma/blob/master/kuma/health/views.py
+@require_safe
+def liveness(request):
+    """
+    A successful response from this endpoint simply proves
+    that Django is up and running. It doesn't mean that its
+    supporting services (like MySQL, memcached, Celery) can
```
metadave: This comment can be updated for SUMO

kitsune/sumo/views.py
```diff
@@ -26,11 +27,54 @@
...
+    supporting services (like MySQL, Redis, Celery) can
+    be successfully used from within this service.
+    """
+    return HttpResponse(status=204)
+
+
+# copied from
+# https://github.com/mozilla/kuma/blob/master/kuma/health/views.py
+@require_safe
+@cache_page(10)
```
safwanrahman: I believe this endpoint should not be cached. If cached for 10 seconds, the endpoint will not hit the DB for 10 seconds.

pmac: The health checks will be more than 10s apart, and this is an extra check to make sure the cache is working (as mentioned in the doc string).

glogiotatidis:
Thanks!

PR #3033, opened 2018-01-30
Fix rtfd requirements.

glogiotatidis:


glogiotatidis:
Tested and works https://kitsune.readthedocs.io/en/fix-rtfd/

PR #3036, opened 2018-01-30
Fix #2998: Add host enforcement middleware

### LINKED ISSUE #2998: Redirect support.mozilla.com to support.mozilla.org

pmac:
This will ensure that other domains we point at the site (e.g. support.mozilla.com) will redirect to the canonical domain.

### DIFF
--- a/kitsune/settings.py
+++ b/kitsune/settings.py
@@ -475,6 +475,7 @@ MIDDLEWARE_CLASSES = (
     'multidb.middleware.PinningRouterMiddleware',
     'django_statsd.middleware.GraphiteMiddleware',
     'commonware.request.middleware.SetRemoteAddrFromForwardedFor',
+    'enforce_host.EnforceHostMiddleware',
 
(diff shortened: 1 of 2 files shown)


kitsune/settings.py
```diff
@@ -1037,6 +1038,10 @@ def JINJA_CONFIG():
 ]
 
 ALLOWED_HOSTS = config('ALLOWED_HOSTS', default='', cast=Csv())
+# in production set this to 'support.mozilla.org' and all other domains will redirect.
+# can be a comma separated list of allowed domains.
+# the first in the list will be the target of redirects.
+ENFORCE_HOST = config('ENFORCE_HOST', default='', cast=Csv())
```
glogiotatidis: I believe this should default to `None` https://github.com/dabapps/django-enforce-host/blob/master/enforce_host/__init__.py#L24

pmac: Hmm... great catch. I'll update shortly.

pmac:
Updated. Thanks!

Issue #3027, opened 2018-01-22
[Bug 1431412] Google Analytics Enhancements


bensternthal:
Google analytics enhancements requested by Peter G & Analytics Pros.

https://bugzilla.mozilla.org/show_bug.cgi?id=1431412

jpetto:
As [Wes mentioned in the bug](https://bugzilla.mozilla.org/show_bug.cgi?id=1431412#c4), this change should replace all existing GA on SUMO. What does this mean in terms of [googleanalytics.py](https://github.com/mozilla/kitsune/blob/master/kitsune/sumo/googleanalytics.py), which appears to feed an admin dashboard with GA reports? Will this file continue to function after replacing existing GA? I'm guessing no (but I have nothing concrete to base that on).

If the answer is indeed "no", what's the plan for this file and the reports it generates? Easiest would just be a follow-up task, but we should be prepared before it happens. (I have no idea how important these existing GA reports are.)

I'm not sure who this is a question for - Analytics Pros, @glogiotatidis?

bensternthal:
From what I understand googleanalytics.py grabs a few data points via the ga API. Because the new GA js will continue to send data to the same GA account my thinking is this should behave normally. I looked through and most of the API calls grab basic data such as pageviews and number of visitors.

As far as I can tell I do not think any of the underlying data that is used to generate the reports will change. 

There is one [funky call for search conversion,](https://github.com/mozilla/kitsune/blob/master/kitsune/sumo/googleanalytics.py#L253). But the way this is configured in ga (see attached) to use URL patterns.....I do not think anything we are doing as part of your work would impact this.

<img width="754" alt="screen shot 2018-01-31 at 9 00 44 am" src="https://user-images.githubusercontent.com/198507/35636229-558b7658-0665-11e8-8231-626710009d57.png">

glogiotatidis:
> (I have no idea how important these existing GA reports are.)

Emailed Madalina and invited her to comment here.

madasan:
There are multiple dashboards out there. We investigated this as a team and we couldn't identify which are the ones that might be affected. Can you give us more details as in which are the dashboards that are powered up by this code?

PR #3037, opened 2018-01-31
[fix bug 1431412] Update GA script.

jpetto:
In the interest of keeping review on this PR sane, I'm going to split it into (at least) two commits.

1. Adding the new GA script and product tagging.
2. Adding `data-*` attributes all over the place.

@glogiotatidis - Can you give this first commit a review? Not being familiar with SUMO, I want to make sure I've got the new GA script on all necessary templates. It would also be nice if we had a public comment somewhere detailing our data collection practices (as we have for [mozilla.org](https://bugzilla.mozilla.org/show_bug.cgi?id=1122305#c8)).

### DIFF
--- a/kitsune/settings.py
+++ b/kitsune/settings.py
@@ -911,6 +911,7 @@ GA_KEY = config('GA_KEY', default='longkey')  # Google API client key
 GA_ACCOUNT = config('GA_ACCOUNT', 'something@developer.gserviceaccount.com')  # Google API Service Account email address
 GA_PROFILE_ID = config('GA_PROFILE_ID', default='12345678')  # Google Analytics profile id for SUMO prod
 GA_START_DATE = date(2012, 11, 10)
+GTM_CONTAINER_ID = config('GTM_CONTAINER_ID', default='') # Google container ID
 
 MOBILE_COOKIE = config('MOBILE_COOKIE', default='msumo')
 
--- a/kitsune/bundles.py
+++ b/kitsune/bundles.py
@@ -293,8 +293,6 @@ PIPELINE_JS = {
             'sumo/js/geoip-locale.js',
             'mailcheck/src/mailcheck.js',
             'sumo/js/ui.js',
-            'sumo/js/dnt-helper.js',
-            'sumo/js/analytics.js',
             'sumo/js/surveygizmo.js',
             'sumo/js/instant_search.js',
         ),
@@ -351,7 +349,6 @@ PIPELINE_JS = {
     },
     'mobile-common': {
         'source_filenames': (
-            'sumo/js/dnt-helper.js',
             'sumo/js/templates/mobile-search-results.js',
             'moment/moment.js',
             'sumo/js/i18n.js',
@@ -366,7 +363,6 @@ PIPELINE_JS = {
             'sumo/js/search_utils.js',
             'sumo/js/aaq.js',
             'sumo/js/mobile/ui.js',
-            'sumo/js/analytics.js',
             'sumo/js/instant_search.js',
             'sumo/js/mobile/instant_search.js',
         ),
@@ -546,8 +542,6 @@ PIPELINE_JS = {
             'modernizr/modernizr.js',
             'sumo/js/browserdetect.js',
             'sumo/js/mobile/ui.js',
-            'sumo/js/dnt-helper.js',
-            'sumo/js/analytics.js',
             'sumo/js/libs/jquery.cookie.js',
             'sumo/js/libs/jquery.lazyload.js',
             'sumo/js/showfor.js',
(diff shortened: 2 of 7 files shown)


kitsune/sumo/static/sumo/js/analytics.js
```diff
@@ -1,119 +0,0 @@
...
-        _gaq.push(['_trackEvent',
-                   'Landing Page Read - 5 seconds',
-                   $('body').data('product-slug')]);
-      }
-    }, 5000);
-  }
-}
-
-
-function trackEvent(category, action, value) {
```
glogiotatidis: What's the plan with pages calling this and `trackPageview` functions?

kitsune/settings.py
```diff
@@ -911,6 +911,7 @@ def JINJA_CONFIG():
 GA_ACCOUNT = config('GA_ACCOUNT', 'something@developer.gserviceaccount.com')  # Google API Service Account email address
 GA_PROFILE_ID = config('GA_PROFILE_ID', default='12345678')  # Google Analytics profile id for SUMO prod
```
glogiotatidis: This is used in [`googleanalytics.py`](https://github.com/mozilla/kitsune/blob/master/kitsune/sumo/googleanalytics.py) which pulls GA stats into Kitsune. 

Is GTM compatible with this? In other other words will it populate the same tables / properties (or how they are called in GA)?

glogiotatidis: Just read your comment in #3027 . Will continue the discussion there.

kitsune/sumo/static/sumo/js/analytics.js
```diff
@@ -1,119 +0,0 @@
...
-  if (_gaq) {
-    _gaq.push(['_trackEvent', category, action, value]);
-  }
-}
-
-function trackPageview(value) {
```
bensternthal: It looks like some of these functions like trackPageview are used in quite a few places. I think we should think about migrating the few custom functions here or if the new ga will collect this data, deprecate and remove.

jpetto: Yeah, this was going to be a sub-task of commit 2 (adding `data-*` attrs) referenced above - removing all old GA calls. I'll update my commit here to keep this file.

jpetto: > I think we should think about migrating the few custom functions here or if the new ga will collect this data, deprecate and remove.

The intent (as I see it) is to remove *all* of the calls to `trackPageview` and `trackEvent` and replace with `data-*` attributes as requested in the PDF. However, there may be some gaps left, which we'd only learn by going through the process. In other words, there may be questions for AP when removing current tracking calls and adding new `data-*` attributes.

bensternthal:
@jpetto ok before I go down editing a bunch of stuff can you sanity check the todo list for the next part of the pr:

- [ ] Find and update all uses of ```a[data-ga-click]```
- [ ] Find and update all uses of ```button[data-ga-click]```
- [ ] Find and update all uses of function ```trackevent```
- [ ] Find and update all uses of function```trackPageview```
- [ ] Add the new tracking AP specified in their doc.

Two clarifications

1. Can you verify if [this](https://docs.google.com/presentation/d/1UEUXYVMXD7w12YL7Io4SxkJ4CGd3-v3ewMN-A6kOH2s/edit#slide=id.g2b8b5c67bb_0_11) is the spec to use when specifying html data attributes?
2. When I replace the js for trackevent and trackPageview, can you give me the format to use including the function call, example ```ga('send', 'event', '<category>', '<action>', '<label>');```

Outside of that I think we need to keep/update the [code](https://github.com/mozilla/kitsune/blob/master/kitsune/sumo/static/sumo/js/analytics.js#L88) that tracks reads. I dont see a way of accomplishing this without the custom code that does the time outs.

jpetto:
A few notes on the list:

- I see some elements with `data-ga-click` attributes that would not be covered under the requirements set by AP (e.g. the form in `new_question.html`). You may want to double check the removal of these with AP.
- For each removal of a `trackEvent` call, make sure that the new GTM implementation will (via `data-*` attributes) continue to gather the same data, or verify the data capture being removed is no longer needed.
- The `trackPageview` calls appear to be a little different, as they seem to solely be tied to searches. It may be worth asking AP how to handle that data collection with GTM. (If anything, it would probably be a JS call on our end to `gtag` with specific arguments.)

In terms of the clarifications:

1. The data attributes in the spec are just (solid) suggestions that we could tweak if we wanted to. I think what they've provided is fine. (For example, we could use `data-action` instead of `data-event-action` to save on file size/typing. We'd just need to tell AP which attribute names we've picked.)
2. I don't think you need to do any JS replacement - only removal. On moz.org, for basic tracking, all we have are `data-*` attributes. The JS loaded from GTM sees those attributes and makes the call to `gatag` for you. The only times we make specific JS calls on moz.org are for non-basic interactions like playing/pausing a video or scrolling down a page.

In terms of the code that tracks reads, I think this again should be run by AP/the Mozilla analytics person handling this work. Maybe they've decided that data isn't important? Or, if we want/need to keep it, it's again probably just a JS call on our end to `gtag` with specific arguments.

The way I am reading the spec from AP is that it covers the most important data to be captured. We may discover gaps or introduce new things we want to track, in which case we'd either add more `data-*` attributes or write some custom JS on our end.

glogiotatidis:
Replaced by #3076

Issue #3039, opened 2018-02-02
update_visitors_metric cron job raises error


glogiotatidis:
The cron job `update_visitors_metric` throws an errors on both SCL3 and AWS deployments and thus has been disabled. 

```
Traceback (most recent call last):
  File "/app/manage.py", line 32, in <module>
    execute_from_command_line(sys.argv)
  File "/venv/lib/python2.7/site-packages/django/core/management/__init__.py", line 354, in execute_from_command_line
    utility.execute()
  File "/venv/lib/python2.7/site-packages/django/core/management/__init__.py", line 346, in execute
    self.fetch_command(subcommand).run_from_argv(self.argv)
  File "/venv/lib/python2.7/site-packages/django/core/management/base.py", line 394, in run_from_argv
    self.execute(*args, **cmd_options)
  File "/venv/lib/python2.7/site-packages/django/core/management/base.py", line 445, in execute
    output = self.handle(*args, **options)
  File "/venv/lib/python2.7/site-packages/cronjobs/management/commands/cron.py", line 64, in handle
    registered[script](*args)
  File "/app/kitsune/kpi/cron.py", line 54, in update_visitors_metric
    visitors = googleanalytics.visitors(start, end)
  File "/app/kitsune/sumo/googleanalytics.py", line 68, in visitors
    metrics='ga:visitors').execute()['rows'][0][0])
KeyError: 'rows'
```

PR #3040, opened 2018-02-02
Get cron jobs and celery workers working

glogiotatidis:
A number of fixes required to get cron jobs running (using APScheduler) and Celery working.

This has been manually pushed to sumo-dev.frankfurt and appears to be working OK


kitsune/settings.py
```diff
@@ -928,7 +924,11 @@ def JINJA_CONFIG():
 
 
 # Google Analytics settings.
-GA_KEY = config('GA_KEY', default='longkey')  # Google API client key
+# GA_KEY is expected b64 encoded.
```
glogiotatidis: GA_KEY is binary. In SCL3 we have in `settings_local.py` which worked OK but we can't put it as is in an env variable. Base64 encoding it does the trick

kitsune/settings.py
```diff
@@ -1088,11 +1088,10 @@ def JINJA_CONFIG():
 DMS_PROCESS_EXIT_SURVEYS = config('DMS_PROCESS_EXIT_SURVEYS', default=None)
 DMS_SURVEY_RECENT_ASKERS = config('DMS_SURVEY_RECENT_ASKERS', default=None)
 DMS_CLEAR_EXPIRED_AUTH_TOKENS = config('DMS_CLEAR_EXPIRED_AUTH_TOKENS', default=None)
-DMS_UPDATE_VISITORS_METRIC = config('DMS_UPDATE_VISITORS_METRIC', default=None)
+# DMS_UPDATE_VISITORS_METRIC = config('DMS_UPDATE_VISITORS_METRIC', default=None)
```
glogiotatidis: See also #3039

bin/run-celery-worker.sh
```diff
@@ -0,0 +1,3 @@
+#!/bin/bash
+
+exec newrelic-admin run-program python manage.py celeryd --maxtasksperchild=${CELERY_MAX_TASKS_PER_CHILD:-10}
```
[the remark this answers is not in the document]
pmac: 10 does seem pretty low. These tasks are things like updating the search indexes and processing uploaded images. Maybe something like 100 would be more reasonable?

kitsune/settings.py
metadave: just a heads up, K8s secrets are base64-encoded, so this value will be double base64 encoded. I don't believe there is any impact here, just wanted to note.

```
echo "foo" | base64 | base64 | base64 -D | base64 -D
```

bin/run-celery-worker.sh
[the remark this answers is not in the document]
glogiotatidis: OK, changed default to 100

PR #3047, opened 2018-02-07
Adding Guarani (gn) and Georgian (ka) to available KB locales

vesper-mozilla:
As per discussions and requests from contributors/l10n, this is to expand available KB locales by adding Guarani (gn) and re-adding Georgian (ka) to the available pool of locales.


glogiotatidis:
Thanks i'll follow up with a couple of things needed to enable the locale and merge into `legacy` as well.

vesper-mozilla:
Thank you, Giorgos!

PR #3050, opened 2018-02-08
Fix read only tests

glogiotatidis:
This PR makes CircleCI to fail if tests fail. It surfaces a problem with product_details which try to use database despite the file storage setting.

@pmac do you have any ideas why this is failing?

Here's a previous run with all the settings printed https://circleci.com/gh/glogiotatidis/kitsune/62


pmac:
I looked back at all of the Circle builds of the master branch, and all tests were passing before the removal of the `customercare` stuff. Looking at the builds it appears as though the changing of how the multidb router stuff was configured in #3019 is what made the tests fail to start with the DB table errors. My changes in the [glogiotatidis-fix-tests-with-pd-storage](https://github.com/mozilla/kitsune/tree/glogiotatidis-fix-tests-with-pd-storage) branch (based on your branch in #3050) fix some of this, but other things have cropped up in the interim. Perhaps you'll recognize some of the failures. I fixed an `UnpicklingError` in Bable by upgrading Babel. Now there are a lot of `DatabaseError: You can't do this in read-only mode.` failures, which appear to be signal handlers attached in previous tests that aren't being detached, but I don't know why yet. You can see that in the [tests on my branch](https://circleci.com/gh/mozilla/kitsune/113).

pmac:
More details:

* [Test run 63](https://circleci.com/gh/mozilla/kitsune/63) was the last fully successful unit test run
* [Test run 64](https://circleci.com/gh/mozilla/kitsune/64) was pretty good. Only tests failing are due to the removal of the `customercare` urls. Should be easy to fix those.
* [Test run 66](https://circleci.com/gh/mozilla/kitsune/66) is where the database failures start.

glogiotatidis:
Turned out to be read_only tests causing this mess. I fixed a couple of things here and there as well.

@pmac r?

docker/composefiles/test.yml
```diff
@@ -2,6 +2,6 @@ version: '3.4'
 services:
   web:
     image: kitsune:full-no-locales-latest
-    command: ./manage.py test --noinput --nologcapture -a '!search_tests' --with-nicedots
+    command: bash
```
pmac: Did you mean to leave this change in? Looks leftover from debugging.

glogiotatidis:
> I just found that one question about bash being the command in the docker-compose file. But looking at the test run on circle I do see a lot of tracebacks that look like:
> 
> NotFoundError: TransportError(404, {u'index': u'sumo_metrics', u'resource.type': u'index_or_alias', u'root_cause': [{u'index': u'sumo_metrics', u'reason': u'no such index', u'type': u'index_not_found_exception', u'resource.id': u'sumo_metrics', u'resource.type': u'index_or_alias'}], u'reason': u'no such index', u'type': u'index_not_found_exception', u'resource.id': u'sumo_metrics'})
> 
> I'm not sure what those are or why they're not causing the tests to fail, but I don't see those in the older runs that were passing. If they're not causing problems then perhaps they're fine, but just FYI. I'm approving this PR, but fix up things if you think it's required before merge.

Those don't fail b/c we're handling exceptions https://github.com/mozilla/kitsune/blob/master/kitsune/questions/views.py#L302

Since we have disabled search tests related we don't run the code that generates the indices https://github.com/mozilla/kitsune/blob/master/kitsune/search/tests/__init__.py#L41

and while I's at it i found out the a typo https://github.com/mozilla/kitsune/pull/3050/commits/546c70d8ac83eb14d77b78c99c852467eb848554

glogiotatidis:
@pmac I fixed the ES related messages and in the process improved the top_contributors_* functions as well. Please take a look at the new commits and merge at will if ok.

Thanks!

glogiotatidis:
Thanks!

PR #3052, opened 2018-02-08
fix cron deployment replicas in K8s automation

metadave:
The cron deployment shared the same replica variable as the web deployment. This PR gives cron deployments their own replica setting, and sets a default value of 1.


metadave:
self-merging so I can continue to iterate.

PR #3055, opened 2018-02-13
Setup Django's SecurityMiddleware

pmac:
* Exempt /healthz and /readiness URLs from HTTPS redirect
* Remove WSGI HTTPS=on hack
* Remove old commonware security middlewares

PR #3056, opened 2018-02-13
add security middleware env to web deployment

metadave:
https://github.com/mozilla/kitsune/pull/3055


glogiotatidis:
This looks fine, thanks @metadave. As [pmac mentioned](https://github.com/mozilla/kitsune/pull/3055#issuecomment-365351851) those are the default values so we can do without this PR. If you like explicitly defining those go ahead and merge this PR.

jgmize:
> This looks fine, thanks @metadave. As pmac mentioned those are the default values so we can do without this PR. If you like explicitly defining those go ahead and merge this PR.

I have no objections to merging this either, but I generally prefer to only add environment variables when we want to override the defaults.

metadave:
If I don't merge it, we'll need it. If I merge it, we won't need it 😃

PR #3059, opened 2018-02-15
Add AWS internal IP in ALLOWED_HOSTS

glogiotatidis:
This will allow us to set proper ALLOWED_HOSTS and ELBs will continue to
work.

from https://gist.github.com/dryan/8271687

/cc @metadave

### DIFF
--- a/kitsune/settings.py
+++ b/kitsune/settings.py
@@ -5,15 +5,13 @@ import logging
 import os
 import platform
 import re
+from datetime import date
 
 import dj_database_url
 import django_cache_url
-
-
-from datetime import date
-from decouple import Csv, config
-
 import djcelery
+import requests
+from decouple import Csv, config
 
 from bundles import PIPELINE_CSS, PIPELINE_JS
 from kitsune.lib.sumo_locales import LOCALES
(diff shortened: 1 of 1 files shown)


kitsune/settings.py
```diff
@@ -1042,6 +1040,15 @@ def JINJA_CONFIG():
 ]
 
 ALLOWED_HOSTS = config('ALLOWED_HOSTS', default='', cast=Csv())
+
+# Add AWS internal IP to allowed hosts for ELB health check to work.
+try:
+    EC2_PRIVATE_IP = requests.get('http://169.254.169.254/latest/meta-data/local-ipv4',
```
metadave: We block access to the metadata server in our K8s clusters. Let's discuss options during our SUMO meeting today.

jgmize: we can get that from the k8s api from within the pods: https://kubernetes.io/docs/tasks/administer-cluster/access-cluster-api/#accessing-the-api-from-a-pod

glogiotatidis:
I'll close this one in favor of the discussion in https://github.com/mozmeao/infra/issues/735

PR #3061, opened 2018-02-15
remove readiness probes from cron and celery deployments

metadave:


glogiotatidis:
r+ thanks @metadave 

ignore the failing tests: mysql sometime "goes away" while testing. this has nothing to do with it.

PR #3062, opened 2018-02-16
Add django-allow-cidr to fix ELB health checks

pmac:
This allows us to set the `ALLOWED_CIDR_NETS` env var to IP ranges which will allow the health checks to succeed for any node in the cluster.

We may still want to do something fancy to automatically get the CIDR for the cluster.

Refs https://github.com/mozmeao/infra/issues/735


metadave:
:+1:, I added the K8s piece [here](https://github.com/mozilla/kitsune/pull/3063)

glogiotatidis:
Nice work! :evergreen_tree:

PR #3063, opened 2018-02-20
add K8s secret specifying a CIDR for ALLOWED_HOSTS

metadave:
this is the K8s side of https://github.com/mozilla/kitsune/pull/3062


glogiotatidis:
Thanks dave!

PR #3064, opened 2018-02-21
K8s updates

glogiotatidis:
/cc @metadave


metadave:
Why was `ALLOWED_CIDR_NETS` removed? It was just added in https://github.com/mozilla/kitsune/pull/3063

glogiotatidis:
From the comment in the commit 
```
django-allow-cidr needs to be updated to correctly parse Host with port number.
Remove the setting to disable middleware until fixed.```

See also https://github.com/mozmeao/django-allow-cidr/pull/1

PR #3067, opened 2018-02-21
Upgrade django-allow-cidr to 0.3.0

pmac:
Handles host header values with ports.


jgmize:
CircleCI test failure was due to 'MySQL server has gone away', not related to this change.

glogiotatidis:
Thanks!

PR #3069, opened 2018-02-22
Fix mysql gone away

glogiotatidis:


metadave:
I see the variable added in [this](https://github.com/mozilla/kitsune/pull/3068) PR as well.

glogiotatidis:
> I see the variable added in this PR as well.

good point, it's the same commit

PR #3077, opened 2018-02-23
stage/prod settings files for Oregon cluster

metadave:
yes, this is mostly a copy/paste of the dev settings, but we need them to get started. Prod will most likely have higher replica settings etc. I'm starting with a single `oregon` directory to keep both oregon-a and oregon-b settings, but I could see the need to split them out soonish.


glogiotatidis:
Thanks Dave!

k8s/regions/oregon/prod.yaml
```diff
@@ -0,0 +1,74 @@
+# General settings
+TARGET_ENVIRONMENT: prod
+AWS_REGION: us-west-2
+AWS_RESOURCE_STACK: SUMO-prod
```
glogiotatidis: Are those two AWS_* variables used somewhere?

metadave: doesn't look like it, it was carried over from MDN automation.

metadave: removed in https://github.com/mozilla/kitsune/pull/3085

PR #3080, opened 2018-02-26
Add HostnameMiddleware.

glogiotatidis:


kitsune/settings.py
```diff
@@ -84,6 +84,7 @@ def path(*parts):
 SERVER_EMAIL = config('SERVER_EMAIL', default='server-error@support.mozilla.org')
 
 PLATFORM_NAME = platform.node()
+K8S_DOMAIN = config('K8S_DOMAIN', default='')
```
pmac: Will this include the "app name" (I guess k8s "namespace" is what it's called now)?

kitsune/sumo/middleware.py
```diff
@@ -222,3 +222,16 @@ def process_response(self, request, response):
             response.set_cookie(mobility.middleware.COOKIE, 'on')
 
         return response
+
+
+class HostnameMiddleware(object):
```
pmac: I know we don't have to yet, but it would be nice if we started writing middleware in a way that's forward compatible with Django 1.10+, which basically just means having it inherit from `MiddlewareMixin` like I did for django-allow-cidr: https://github.com/mozmeao/django-allow-cidr/blob/master/allow_cidr/middleware.py#L7-L10. Not required, but would be nice to start doing this in prep for the upgrade to 1.11.

glogiotatidis: Nice tip, thanks

kitsune/settings.py
glogiotatidis: with k8s `platform.node()` returns more meaniful answers like `sumo-dev-web-1092229904-8xjsl`. So I was planning `K8S_DOMAIN` to be the domain like, `frankfurt.moz.works`

glogiotatidis: Also updated with k8s app settings to include `k8s_domain`

pmac: Ah! cool. thanks for the info.

PR #3084, opened 2018-02-26
Unify deployment template, simplify env variable management.

glogiotatidis:
- Unifies web, cron and celery templates into one template.

   Maintaining 3 different almost identical templates is boring and error prone.

 - Allows per app (web, cron, celery) customization.

 - Reads env variables from the yaml. If value is SECRET, links to app secrets.

bonus: removed a few unused variables


glogiotatidis:
There's something I don't get in the original templates:
 - I didn't see anything about liveness probes, only readiness
 - Readiness probe con doesn't define the URL to hit. How does k8s know to ping /readiness?

@metadave do you have any hints?

glogiotatidis:
Also btw @metadave @pmac do you have strong feelings about python-invoke? It seems half baked  and I'd like to try to build something similar with `click`. sound good?

metadave:
I believe what's there for liveness/readiness was pulled from the Deis deployment. I was waiting until the ELBs were created until working on liveness/readiness.

pmac:
> Also btw @metadave @pmac do you have strong feelings about python-invoke? It seems half baked and I'd like to try to build something similar with click. sound good?

I'm not married to anything. What I liked about invoke was that it has a common interface (i.e. "invoke"), it has some helpful tools for dealing with the system (the `cd` context manager, running shell commands easily, etc.), and it was a lot less magic than the other system we looked at. We could definitely do our own fully custom package with click but it just seemed like more work. I'm curious what you feel about invoke isn't fully baked?

metadave:
Liveness/readiness updates: https://github.com/mozilla/kitsune/pull/3085

metadave:
+1

glogiotatidis:
Updated: 
 - to include the readiness and liveness probes 
 - add `--tag` flag to specify docker tag to push
 - add `create-all` and `delete-all` commands

So... readiness probe appears to be working but I couldn't get liveness probe to work. I even tried the [documentation example](https://v1-6.docs.kubernetes.io/docs/tasks/configure-pod-container/configure-liveness-readiness-probes/#define-a-liveness-http-request)

```
kubectl create -f https://k8s.io/docs/tasks/configure-pod-container/http-liveness.yaml
```

and k8s kept killing it as unhealthy. 

Any ideas @metadave ? Maybe related to v1.6?

metadave:
@glogiotatidis  I see that MDN is running in Frankfurt and has a liveness probe defined, the pods appear to be up and running:

```
    ...
    livenessProbe:
      failureThreshold: 6
      httpGet:
        httpHeaders:
        - name: Host
          value: developer.mozilla.org
        path: /healthz
        port: 8000
        scheme: HTTP
      initialDelaySeconds: 20
      periodSeconds: 12
      successThreshold: 1
      timeoutSeconds: 10
    ...
```

k8s/templates/sumo-app.yaml.j2
```diff
@@ -0,0 +1,85 @@
...
+          timeoutSeconds: {{ kubernetes.readiness.timeout_seconds }}
+        {%- endif %}
+        {%- if kubernetes.liveness %}
+        livenessProbe:
+          httpGet:
+              port: {{ app.port }}
+              path: /healthz/
+              httpHeaders:
+                  - name: Host
+                    value: {{ app.allowed_hosts.split(',')[0] }}
```
metadave: How does this work with django-allow-cidr?

glogiotatidis: Here we use the first hostname from allowed_hosts and django-allow-cidr respects allowed_hosts.

glogiotatidis: > @glogiotatidis I see that MDN is running in Frankfurt and has a liveness probe defined, the pods appear to be up and running:

Used similar settings and it works now, thanks!

glogiotatidis:
> @glogiotatidis I see that MDN is running in Frankfurt and has a liveness probe defined, the pods appear to be up and running:

Used similar settings and it works now, thanks!

glogiotatidis:
Please take another look @metadave ?

metadave:
:+1:

glogiotatidis:
Thanks!

PR #3087, opened 2018-02-27
copy frankfurt dev settings to stage and prod

metadave:
note: `deployments.create-nodeport` is broken so I manually created frankfurt stage and prod nodeports. Other than replacing `dev` with `stage` and `prod`, there are no env changes compared to dev.


glogiotatidis:
Thanks Dave. I'll go through all settings for all apps and clusters and validate them.

PR #3088, opened 2018-02-28
Fix #3030: Add time-ago options to esreindex management command

### LINKED ISSUE #3030: Setup search indexing for hot-spare region
For our read-only spare region we won't be able to use the ORM model save signals to trigger indexing of new and modified questions and articles. We can reindex all of the articles nightly, but the questions are too numerous to do in that way. For now we'll just do an initial full indexing run, then set up a nightly job to freshen the index by reindexing all of the articles and all of the questions that have a modified date within the past two days. This should cover things if the nightly job fails one time.

I'll be writing up another issue to cover how we'd like to set up search indexing in future to more reliably avoid these issues.

pmac:
This is a WIP since I've not gotten to test with a real DB much
and haven't yet added unit tests, but I believe it is working.

### DIFF
--- a/kitsune/search/models.py
+++ b/kitsune/search/models.py
@@ -126,7 +126,7 @@ class SearchMappingType(MappingType, Indexable):
         qs = cls.get_model().objects.order_by('pk').values_list('pk', flat=True)
         if seconds_ago:
             if cls.seconds_ago_filter:
-                dt = datetime.datetime.utcnow() - datetime.timedelta(seconds=seconds_ago)
+                dt = datetime.datetime.now() - datetime.timedelta(seconds=seconds_ago)
                 qs = qs.filter(**{cls.seconds_ago_filter: dt})
             else:
                 # if seconds_ago is specified but seconds_ago_filter is falsy don't index anything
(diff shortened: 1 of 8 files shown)


kitsune/search/models.py
```diff
@@ -117,13 +118,21 @@ def get_localized_fields(cls):
...
         # Some models have a gazillion instances. So we want to go
         # through them one at a time in a way that doesn't pull all
         # the data into memory all at once. So we iterate through ids
         # and pull objects one at a time.
-        return cls.get_model().objects.order_by('pk').values_list(
-            'pk', flat=True)
+        qs = cls.get_model().objects.order_by('pk').values_list('pk', flat=True)
+        if seconds_ago:
+            if cls.seconds_ago_filter:
+                dt = datetime.datetime.utcnow() - datetime.timedelta(seconds=seconds_ago)
```
glogiotatidis: (unfortunately) we use `datetime.now` in our models.

glogiotatidis:
This is still WIP?

pmac:
Fixed the `datetime.now` thing and removed WIP label. We'll discuss how and when to run the cron tasks in #3098

glogiotatidis:
Nice work! Thanks pmac!

PR #3092, opened 2018-02-28
Fix bug 1427813: Only minimal HTML in user bios

pmac:


kitsune/sumo/templatetags/jinja_helpers.py
```diff
@@ -113,6 +113,14 @@ def wiki_to_html(wiki_markup, locale=settings.WIKI_DEFAULT_LANGUAGE,
                                              nofollow=nofollow))
 
 
+@library.filter
+def wiki_to_safe_html(wiki_markup, locale=settings.WIKI_DEFAULT_LANGUAGE,
```
glogiotatidis: This filter needs to be able to strip the disallowed tags from the final output. 

Here's what I get when saving a test profile.

image

instead of 

image

kitsune/sumo/templatetags/jinja_helpers.py
```diff
@@ -113,6 +113,14 @@ def wiki_to_html(wiki_markup, locale=settings.WIKI_DEFAULT_LANGUAGE,
...
+                      nofollow=True):
+    """Wiki Markup -> HTML jinja2.Markup object with limited tags"""
+    return jinja2.Markup(parser.wiki_to_html(wiki_markup, locale=locale,
+                                             nofollow=nofollow, tags=bleach.ALLOWED_TAGS))
```
glogiotatidis: Maybe we add `<p>` to the allowed tags?. The current editor wraps every paragraph in `<p>`

kitsune/users/tests/test_templates.py
```diff
@@ -446,6 +446,15 @@ def test_bio_links_nofollow(self):
...
         eq_(2, len(doc('.bio a[rel="nofollow"]')))
 
+    def test_bio_links_no_img(self):
+        # bug 1427813
+        self.profile.bio = 'my image <img src="https://www.example.com/the-dude.jpg">'
+        self.profile.save()
+        r = self.client.get(reverse('users.profile', args=[self.u.username]))
+        eq_(200, r.status_code)
+        doc = pq(r.content)
+        eq_(0, len(doc('.bio img')))
```
glogiotatidis: In cases like this I believe checking against the exact final text is safer.

glogiotatidis:
Also add a migration to convert bleach existing bio data?

kitsune/users/tests/test_templates.py
pmac: I considered that but that does not seem to be how it works now. It looks like we store the raw input and wiki process it at render time. I usually like to do rendering and cleaning on update then just serve it straight out of the DB but that would be a good bit more work.

kitsune/sumo/templatetags/jinja_helpers.py
pmac: Cool.

kitsune/sumo/templatetags/jinja_helpers.py
pmac: Ah... hmm... I thought bleach was stripping tags. I was going to add more bleach processing to the output of `wiki_to_html` but discovered that it's already doing bleach.clean, just with a lot more tags. I can look at making it strip tags instead of converting them to HTML entities. It's also possible we don't need to worry too much since people can just edit again to remove the HTML.

glogiotatidis:
> Also add a migration to convert bleach existing bio data?

Answering to myself: As pmac says in other comment, we're processing at render time, so no need.

glogiotatidis:
@pmac the test failure is legit 
```
FAIL: kitsune/users/tests/test_templates.py:ViewProfileTests.test_bio_links_nofollow
  File "/app/kitsune/users/tests/test_templates.py", line 447, in test_bio_links_nofollow
    eq_(2, len(doc('.bio a[rel="nofollow"]')))
  File "/venv/lib/python2.7/site-packages/nose/tools.py", line 31, in eq_
AssertionError: 2 != 0
```

pmac:
Thanks Giorgos. Looking at it now.

glogiotatidis:
Thanks pmac!

glogiotatidis:
Also merged in legacy c592b6778ca9532df13919588062f89560609a2f

PR #3097, opened 2018-03-02
Skip cron jobs when READ_ONLY=True

glogiotatidis:
Decorator to skip jobs when `READ_ONLY=True`. Currently only jobs related to ES re-indexes are allowed.

### DIFF
--- a/scripts/cron.py
+++ b/scripts/cron.py
@@ -49,6 +49,21 @@ class scheduled_job(object):
         print(msg, file=sys.stderr)
 
 
+class skip_read_only(object):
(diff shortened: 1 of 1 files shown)


glogiotatidis:
Renamed `skipReadOnly` to `skip_read_only`

glogiotatidis:
OK i think i'll go ahead and extend `scheduled_job` as you suggest to also add `STAGE=True` used in some cron jobs.

Let's first merge #3030 and #3098 and I'll PR again with fresh code.

Thanks for your time!

PR #3098, opened 2018-03-02
Add reindex_questions and reindex_threads cron jobs.

glogiotatidis:
Add reindex commands that run every two hours for questions and forums threads. KB is reindexing daily.

Merge after #3030

### DIFF
--- a/kitsune/forums/cron.py
+++ b/kitsune/forums/cron.py
@@ -1,0 +1,15 @@
+import logging
+from kitsune.forums.models import ThreadMappingType
+from kitsune.search.tasks import index_task
+
+import cronjobs
+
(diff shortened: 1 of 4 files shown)


pmac:
I was thinking more along the lines of using just the `manage.py esreindex --hours-ago 2` somewhere. If there's an advantage to doing them all individually though I'm happy to have it done this way as well.

glogiotatidis:
> I was thinking more along the lines of using just the manage.py esreindex --hours-ago 2 somewhere. If there's an advantage to doing them all individually though I'm happy to have it done this way as well.

good idea. closing this one in favor of #3103

PR #3102, opened 2018-03-05
Skip cron jobs based on READ_ONLY and STAGE settings.

glogiotatidis:
Replaces #3097 

Also there're some cron jobs that still check internally whether `STAGE=True`. Those cronjobs are commented out in `scripts/cron.py` b/c they are broken and need a fresh approach altogether.

### DIFF
--- a/scripts/cron.py
+++ b/scripts/cron.py
@@ -23,12 +23,15 @@ class scheduled_job(object):
     def __init__(self, *args, **kwargs):
         self.args = args
         self.kwargs = kwargs
+        self.skip = self.kwargs.pop('skip', False)
 
     def __call__(self, fn):
-        job_name = fn.__name__
-        self.name = job_name
+        self.name = fn.__name__
+        if self.skip:
+            self.log('Skipped, not registered.')
+            return None
         self.callback = fn
-        schedule.add_job(self.run, id=job_name, *self.args, **self.kwargs)
+        schedule.add_job(self.run, id=self.name, *self.args, **self.kwargs)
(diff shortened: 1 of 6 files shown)


scripts/cron.py
```diff
@@ -50,76 +53,89 @@ def log(self, message):
...
+@scheduled_job('cron', month='*', day='*', hour='*', minute='30',
+               max_instances=1, coalesce=True, skip=settings.READ_ONLY)
 @babis.decorator(ping_after=settings.DMS_SEND_WELCOME_EMAILS)
 def job_send_welcome_emails():
     call_command('cron send_welcome_emails')
 
 
-@scheduled_job('cron', month='*', day='*', hour='*', minute='59', max_instances=1, coalesce=True)
+@scheduled_job('cron', month='*', day='*', hour='*', minute='59',
+               max_instances=1, coalesce=True, skip=(settings.READ_ONLY | settings.STAGE))
```
pmac: shouldn't this be `(settings.READ_ONLY or settings.STAGE)`? I'm sure this works, but it seems a bit less readable than just the or operator. Just a NIT though, not a blocker if you really prefer it this way.

pmac:
Oh, and flak8 is complaining about an unused import of `settings` in `users/cron.py`.

scripts/cron.py
glogiotatidis: Sure, why not. updated

glogiotatidis:
Fixed the flake error. Tests fail due to mysql error, unrelated.

PR #3103, opened 2018-03-05
Index all new data every hour.

glogiotatidis:


scripts/cron.py
```diff
@@ -68,6 +68,12 @@ def job_send_welcome_emails():
     call_command('cron send_welcome_emails')
 
 
+@scheduled_job('cron', month='*', day='*', hour='*', minute='45', max_instances=1, coalesce=True)
+@babis.decorator(ping_after=settings.DMS_REINDEX)
+def job_reindex():
+    call_command('esreindex --hours-ago 1')
```
pmac: I think we may miss some if it both runs every hour and only checks for an hour back. I'd suggest having a bit of overlap. For example, maybe we have it run every hour, but have it pass in `--minutes-ago 90`.

glogiotatidis:
very good point, updated

pmac:
Circle failure due to unrelated DB container issue after testing.

PR #3104, opened 2018-03-05
Fix web command for stage-oregon-{a,b}

jgmize:
./bin/run-stage.sh does not exist.


jgmize:
Note that Frankfurt stage was already configured to use ./bin/run-prod.sh

jgmize:
This change has been deployed to https://stage-oregon-a.sumo.moz.works and https://stage-oregon-b.sumo.moz.works

metadave:
:+1:

PR #3113, opened 2018-03-06
Fix #3112: Add ability to set cache headers

### LINKED ISSUE #3112: Add middleware to add cache headers
This will mostly be enabled when `READ_ONLY` is `True` since we'll be using a CDN in front of the site in `READ_ONLY` mode to reduce the load on our hot-spare cluster in the event of a failover.

pmac:
By default it will set a max-age of 2 hours if READ_ONLY is True.

### DIFF
--- a/kitsune/settings.py
+++ b/kitsune/settings.py
@@ -90,6 +90,10 @@ CACHES = {
     },
 }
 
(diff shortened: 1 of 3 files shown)


pmac:
I also added the ability for us to turn on Etags since that middleware is already enabled. It's off by default for now.

Issue #3091, opened 2018-02-28
Set READ_ONLY= False


glogiotatidis:
After switching to AWS set `READ_ONLY=False` on prod in oregon-a and oregon-b only. Frankfurt must remain `READ_ONLY=True`

PR #3124, opened 2018-03-13
Fixed(2)

akashgosai:


akashgosai:
In response to #3091

glogiotatidis:
Thanks for your contribution @akashgosai 

While this is indeed the change requested in the #3091 as the issue explains this is to be made after we complete the migration to AWS. Thus I'm going to close this PR now.

Currently Kitsune is in fix-only mode but starting next month we will welcome contributors and contributors to help us build the platform together.

PR #3127, opened 2018-03-13
Add setting for using the x-forwarded-host header

pmac:
Turn on said setting for frankfurt prod. Also change the static_url setting to just SUMO in frankfurt prod since it is a CDN.

Ref https://github.com/mozmeao/infra/pull/747


metadave:
:+1:

pmac:
Tests passed, but we hit that "DB missing" thing again :/ Still haven't figured that one out...

PR #3130, opened 2018-03-14
Fix Bug 1445673: Remove optimizely code from old account.

bensternthal:


bensternthal:
Odd test failure, I do not think it is caused by this code change:

``django.db.utils.OperationalError: (2006, 'MySQL server has gone away')
Exited with code 1``

glogiotatidis:
Thanks Ben! :taco:

glogiotatidis:
This is on stage, waits for GTM o be reviewed before getting merged to production

PR #3133, opened 2018-03-15
Remove DB migrations run from app start

pmac:
Rename run-common.sh to run-post-deploy.sh to indicate that these tasks (currently just migrations) should be run only once post deployment. We still need to set up the post-deployment tasks when we get automated deployment.


glogiotatidis:
Thanks pmac

PR #3140, opened 2018-03-20
Add middleware to disable cache for vary: accept-language

pmac:
Also disable mobile detection middleware in READ_ONLY mode.

### DIFF
--- a/kitsune/sumo/middleware.py
+++ b/kitsune/sumo/middleware.py
@@ -104,6 +104,26 @@ class Forbidden403Middleware(object):
         return response
 
 
+class VaryNoCacheMiddleware(MiddlewareMixin):
+    """
+    If enabled this will set headers to prevent the CDN (or other caches) from
+    caching the response if the response was set to vary on accept-langauge.
+    This should be near the top of the list of middlewares so it will be able
+    to inspect the near-final response since response middleware is processed
+    in reverse.
+    """
+    def __init__(self):
+        if not settings.ENABLE_VARY_NOCACHE_MIDDLEWARE:
+            raise MiddlewareNotUsed
+
+    @staticmethod
+    def process_response(request, response):
+        if 'vary' in response and 'accept-language' in response['vary'].lower():
+            add_never_cache_headers(response)
+
+        return response
+
+
 class CacheHeadersMiddleware(MiddlewareMixin):
     """
     Sets no-cache headers normally, and cache for some time in READ_ONLY mode.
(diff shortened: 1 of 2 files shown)


kitsune/sumo/middleware.py
```diff
@@ -199,8 +219,12 @@ def safe_query_string(request):
 # This is a modified version of 'mobility.middleware.DetectMobileMiddleware'.
 # We want to exclude tablets from being detected as MOBILE and there is
 # no way to do that by just overriding the detection regex.
-class DetectMobileMiddleware(object):
+class DetectMobileMiddleware(MiddlewareMixin):
     """Looks at user agent and decides whether the device is mobile."""
+    def __init__(self, *args, **kwargs):
+        if settings.SKIP_MOBILE_DETECTION:
```
glogiotatidis: I think we should do the same for [`MobileSwitchMiddleware`](https://github.com/mozilla/kitsune/pull/3140/files#diff-6f05e342c77cc4322987a6ff3c3edd64R241) which will allow users to switch to mobile by passing `?mobile=1` in the URL

kitsune/sumo/middleware.py
```diff
@@ -104,6 +104,26 @@ def process_response(self, request, response):
...
+    to inspect the near-final response since response middleware is processed
+    in reverse.
+    """
+    def __init__(self):
+        if not settings.ENABLE_VARY_NOCACHE_MIDDLEWARE:
+            raise MiddlewareNotUsed
+
+    @staticmethod
+    def process_response(request, response):
+        if 'vary' in response and 'accept-language' in response['vary'].lower():
```
jgmize: Does the response object *always* have a 'vary' key, or should this be `response.get('vary', '')`

kitsune/sumo/middleware.py
pmac: Is there a reason to disallow that for the CDN though? It doesn't rely on headers we'd have to worry about and is technically a different URL if it has different parameters.

pmac: Also in my searching it looks like a lot of code expects `request.MOBILE` to be set, and the `MobileSwitchMiddleware` is the only thing I see that sets it. I believe a lot of things would break if it were disabled. We could have it just set `request.MOBILE = False` and skip the other functions in READ_ONLY if that'd be better.

glogiotatidis: > Is there a reason to disallow that for the CDN though? It doesn't rely on headers we'd have to worry about and is technically a different URL if it has different parameters.

My concern is that if someone visits with `mobile=1` the cache would store the mobile version of the page?

glogiotatidis: We cache (and we pass) no querystrings atm

image

Since we're not passing them to the app I guess we can leave this middleware as is or hardcode to `False` as you suggest. What do you think?

pmac: huh... does no other part of the site rely on query strings? I believe searching does right? e.g.: 

https://support.mozilla.org/en-US/search/advanced?q=insecure&language=en-US&sortby_documents=relevance&a=1&w=1

I think we need to turn that back on actually.

glogiotatidis:
After discussing we agree to 
 - Change CDN to pass all query parameters and thus mobile=1 will work
 - No code changes needed

Nice work @pmac :+1: :)

pmac:
Thanks! Merging.

PR #3143, opened 2018-03-21
Upgrade django-watchman to 0.15.0

pmac:
Adds respect for the WATCHMAN_DISABLE_APM setting


glogiotatidis:
Looks good, tests fail due with "mysql gone away"

PR #3148, opened 2018-03-26
Add OIDC auth backend instead of replace defaults

pmac:
Should allow both traditional and OIDC auth to work.


glogiotatidis:
Thanks!

PR #3152, opened 2018-03-27
Fix #3151: Require OIDC auth for admin

### LINKED ISSUE #3151: Support both OIDC and DB logins
And restrict use of `/admin/` to users authenticated with OIDC only. I'll be investigating whether this is possible in this issue. If we can't support this quickly in a single instance we will deploy an admin-only version of kitsune that will only work with OIDC auth and allow use of the Admin, and turn off `/admin/` on SUMO proper.

/cc @glogiotatidis

pmac:
Also only refresh OIDC token on admin URLs.

### DIFF
--- a/kitsune/settings.py
+++ b/kitsune/settings.py
@@ -502,6 +502,7 @@ MIDDLEWARE_CLASSES = (
     # loaded before the LocaleURLMiddleware
     'commonware.middleware.NoVarySessionMiddleware',
     'django.contrib.auth.middleware.AuthenticationMiddleware',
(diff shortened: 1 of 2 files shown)


pmac:
@glogiotatidis this is as far as I got, but I think it will work. I believe it should:

* limit the OIDC refresh middleware to only acting on the admin URLs (other than `/admin/login/`)
* require OIDC auth for other `/admin/*` URLs.
* allow normal auth to function for the rest of the site

I've not tested this as it's difficult locally, but it's not overly complex. I'm hoping we can build an image and test it out in one of our non-prod environments. Feel free to take this and run with it if you see needed improvements.

PR #3158, opened 2018-03-28
Change session cookie name.

glogiotatidis:


safwanrahman:
It will mitigate the problem but I dont know the reason to have the problem.
Do you have any idea?

glogiotatidis:
> It will mitigate the problem but I dont know the reason to have the problem.
> Do you have any idea?

This is correct. we're still investigating

PR #3175, opened 2018-04-06
[Fix #3174] Improve performance of _get_creator_counts util function

### LINKED ISSUE #3174: Improve performance of _get_creator_counts util function
`kitsune.community.utils._get_creator_counts` until function is DB heavy and takes a lot of time to execute. Evaluate its usefulness and provide a way to optimize the query and/or cache the results.  

This issue is related to the degraded performance SUMO experienced on Fri March 30th ([NR Error](https://rpm.newrelic.com/accounts/1299394/applications/45097089/downtime/34422892))

safwanrahman:
Fix #3174
I have tested it in local machine with test database.
By Using elasticsearch, DB queries can be reduced heavily.

This functionality was actually written using elasticsearch, but later because of `facet` removed from elasticsearch 2.x, it was rewritten using queryset.

Now, we should use DB queries for counting the numbers and use elasticsearch for getting users information. Later, after changing/upgrading the client library, this thing need to be fully rewritten using elasticsearch.

@jgmize @glogiotatidis @pmac r?

### DIFF
--- a/kitsune/community/utils.py
+++ b/kitsune/community/utils.py
@@ -7,8 +7,7 @@ from django.db.models import Count, F
 
 from kitsune.products.models import Product
 from kitsune.questions.models import Answer
-from kitsune.users.models import User
-from kitsune.users.templatetags.jinja_helpers import profile_avatar
+from kitsune.users.models import User, UserMappingType
 from kitsune.wiki.models import Revision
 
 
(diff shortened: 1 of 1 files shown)


pmac:
This looks okay to me, though I haven't had a chance to test it locally. I'd like Giorgos to have a look before we merge as well. But I wanted to go ahead and say thanks! Hopefully we'll get this in later this week.

jgmize:
This looks like it should reduce the load on the db, but given the expense of this function and the frequency with which it is called, I wonder if it would be better to persist the results of the function in the DB and only call this function from a cron job to eliminate the “[thundering herd](http://www.ehcache.org/documentation/2.8/recipes/thunderingherd.html)” or “[cache miss storm](https://www.percona.com/blog/2010/09/10/cache-miss-storm/)” problems we've been observing.

safwanrahman:
@jgmize Yeah. At first I have also thought about calling the function from cron job and storing the result. But after seeing the usage, I have found that there are a lot of combinations that can be returned from this function and making it run via cron job will lead to refactoring medium portion of code.
I will investigate and look forward the code and find out if its possible to do it in that way. But untill then, for rudicing the DB load, we can get this PR landed.

Regarding *expensive*, do you mean DB expensive or process expensive?

jgmize:
I meant expensive in terms of time it takes to execute, regardless of whether that time is due to waiting on mysql, elasticache, or some combination.

glogiotatidis:
This is good stuff, thanks @safwanrahman! :cake:

Issue #3173, opened 2018-04-04
Improve performance of question_details view (mysql questions_question select)


glogiotatidis:
question_details view executes SQL queries that take a lot of time. Evaluate whether those queries can be removed, optimized and/or cached to improve performance.

This issue is related to the degraded performance SUMO experienced on Fri March 30th ([NR Error](https://rpm.newrelic.com/accounts/1299394/applications/45097089/downtime/34422892))

 - [question_details](https://rpm.newrelic.com/accounts/1299394/applications/45097089/transactions#id=5b225765625472616e73616374696f6e2f46756e6374696f6e2f6b697473756e652e7175657374696f6e732e76696577733a7175657374696f6e5f64657461696c73222c22225d) transaction
 - [questions_question](https://rpm.newrelic.com/accounts/1299394/applications/45097089/datastores#/overview/MySQL/drilldown?metric=Datastore%252Fstatement%252FMySQL%252Fquestions_question%252Fselect&value=total_call_time_per_minute) select transaction
 - [upload_imageattachment](https://rpm.newrelic.com/accounts/1299394/applications/45097089/datastores#/overview/MySQL/drilldown?metric=Datastore%2Fstatement%2FMySQL%2Fupload_imageattachment%2Fselect) select transaction

jgmize:
There was another degraded performance issue just now related to DB slowness that is related to this and also #3174.

glogiotatidis:
We-re not done here yet, re-opening 

This is live on https://dev.sumo.moz.works 

@safwanrahman can you please verify that this is working ok?

safwanrahman:
@glogiotatidis It seems to be working perfectly!
https://dev.sumo.moz.works/en-US/questions/1204106

glogiotatidis:
OK thanks @safwanrahman 

I pushed this to production, also marked the deployment tag in NewRelic ([oregon-a](https://rpm.newrelic.com/accounts/1299394/applications/45098028/deployments/17459534) and [oregon-b](https://rpm.newrelic.com/accounts/1299394/applications/45097089/deployments/17459556))

I'll make a note to revisit in a few days to if I can find performance differences.

glogiotatidis:
> I'll make a note to revisit in a few days to if I can find performance differences.

While this was a valid micro-optimization unfortunately it didn't create a visible difference in the graphs (vertical line is the deployment), thus I'm re-opening this. I suggest that next step is to look into #3228.

[...]

PR #3185, opened 2018-04-12
Revert "Decrease number of pods to 40 for oregon-a and oregon-b."

glogiotatidis:
This reverts commit 6345c4dc1c972ddad4e3e985c531bd6804fc793f.

We got NR alerts, increased latency and high RDS  cpu. Maybe related to #3173 and #3174 and not to the decrease of the number of pods but the timing was suspicious so we reverted.


metadave:
see also: 
- https://github.com/mozmeao/infra/issues/777
- https://github.com/mozmeao/infra/issues/779

Issue #3179, opened 2018-04-11
IOError: [Errno 2] No such file or directory: '/app/media/uploads/products/logo-sprite-2x.png'


glogiotatidis:
https://sentry.prod.mozaws.net/operations/sumo-aws-prod/issues/3484380/

```python
IOError: [Errno 2] No such file or directory: '/app/media/uploads/products/logo-sprite-2x.png'
  File "django/core/handlers/base.py", line 132, in get_response
    response = wrapped_callback(request, *callback_args, **callback_kwargs)
  File "newrelic/hooks/framework_django.py", line 527, in wrapper
    return wrapped(*args, **kwargs)
  File "django/contrib/admin/options.py", line 618, in wrapper
    return self.admin_site.admin_view(view)(*args, **kwargs)
  File "django/utils/decorators.py", line 110, in _wrapped_view
    response = view_func(request, *args, **kwargs)
  File "django/views/decorators/cache.py", line 57, in _wrapped_view_func
    response = view_func(request, *args, **kwargs)
  File "django/contrib/admin/sites.py", line 233, in inner
    return view(request, *args, **kwargs)
  File "django/utils/decorators.py", line 34, in _wrapper
    return bound_func(*args, **kwargs)
  File "django/utils/decorators.py", line 110, in _wrapped_view
    response = view_func(request, *args, **kwargs)
  File "django/utils/decorators.py", line 30, in bound_func
    return func.__get__(self, type(self))(*args2, **kwargs2)
  File "django/contrib/admin/options.py", line 1613, in changelist_view
    self.save_model(request, obj, form, change=True)
  File "django/contrib/admin/options.py", line 1080, in save_model
    obj.save()
  File "kitsune/products/models.py", line 106, in save
    'logo-sprite-2x.png'))
  File "PIL/Image.py", line 1925, in save
    fp = builtins.open(filename, "w+b")
```

glogiotatidis:
Related bug https://bugzilla.mozilla.org/show_bug.cgi?id=1453388

glogiotatidis:
https://github.com/mozilla/kitsune/pull/3188

jcristau:
Presumably this wasn't supposed to comment on https://bugzilla.mozilla.org/show_bug.cgi?id=3179#c3?

glogiotatidis:
> Presumably this wasn't supposed to comment on https://bugzilla.mozilla.org/show_bug.cgi?id=3179#c3?

[...]

PR #3188, opened 2018-04-16
[fix issue #3179] Switch product icons from sprite to real product icon.

glogiotatidis:
Kitsune Products featured code to auto-generate a sprite with all product icons
on every product save. This code broken when we switched to the AWS S3 backed
user media backend.

Instead of fixing this code, we opted for removing the sprite functionality
altogether because:

  - it was not used in all places (i.e. questions used a different manually
    created sprite)

  - it raised serious concerns on how we can atomically update the auto
    generated sprite on S3 while avoiding errors.

  - The whole implementation was a hack which saved files without using
    default_storage engine.

Now product images are directly used on desktop and mobile and as a bonus the
same images -which can be managed through the admin interface- are used in both
product listing and AAQ.

Partially fixes bug 1196475 too.

### DIFF
--- a/docs/questions.rst
+++ b/docs/questions.rst
@@ -33,13 +33,6 @@ Next, Add a new item to the ``products`` dictionary using something like::
 
 ``'product-slug'`` should be the slug of the ``Product`` object for this product.
 
-You will also need to add a new 96x96 icon to the ``img/logos.large.sprite.png`` and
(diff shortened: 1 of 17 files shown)


glogiotatidis:
@pmac thoughts on the approach and r?

glogiotatidis:
Thanks!

PR #3187, opened 2018-04-13
Double WSGI_NUM_WORKERS & halve # of pods for mozmeao/infra#782

jgmize:

PR #3189, opened 2018-04-16
Double WSGI_NUM_WORKERS & halve pods in oregon-b

jgmize:
Apply same settings as in oregon-a by #3187 for mozmeao/infra#782


metadave:
:+1:

PR #3194, opened 2018-04-20
update k8s deployment docs to reflect new Frankfurt cluster version

metadave:
See also: https://github.com/mozmeao/infra/issues/796


glogiotatidis:
Thanks Dave!

PR #3205, opened 2018-05-16
Promotions experiment splits clicks to 3 destinations

hoosteeno:
This addresses https://bugzilla.mozilla.org/show_bug.cgi?id=1441583#c2.

Functional testing I did:

- [x] Works on Firefox, not on other browsers
- [x] Works on Firefox 40+, not Firefox 39 (UA spoofing)
- [x] Works when signed out of sync, not when signed in
- [x] Works on particular URLs, not on all URLs
- [x] Works in en-US, not in different locale
- [x] Text as expected
- [x] Link variations as expected
- [x] GA events fire as expected

### DIFF
--- a/kitsune/sumo/jinja2/base.html
+++ b/kitsune/sumo/jinja2/base.html
@@ -12,22 +12,6 @@
 
   {{ favicon() }}
 
-  {# TrafficCop experiment_fxa_cta_topbar #}
-  {% set run_experiments = False %}
-  {% if waffle.flag('experiment_fxa_cta_topbar') %}
-    {% if request.path in ['/en-US/kb/install-firefox-android-device-using-google-play',
-                           '/en-US/kb/will-firefox-work-my-mobile-device',
-                           '/en-US/kb/get-started-firefox-overview-main-features',
-                           '/en-US/kb/update-firefox-latest-version',
-                           '/en-US/kb/install-firefox-your-ipad-iphone-or-ipod'] %}
-      {% set run_experiments = True %}
-    {% endif %}
-  {% endif %}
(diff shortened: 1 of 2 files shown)


hoosteeno:
@jpetto you reviewed the previous incarnation of this experiment. Would you be up for reviewing this one?

kitsune/sumo/jinja2/base.html
```diff
@@ -75,6 +59,23 @@
   <![endif]-->
 
   {% include "includes/google-analytics.html" %}
+
+  {# TrafficCop experiment_fxa_cta_topbar #}
+  {% set run_experiments = False %}
+  {% if waffle.flag('experiment_fxa_cta_topbar') %}
+    {% if request.path in ['/en-US/kb/install-firefox-android-device-using-google-play',
```
jpetto: You should be able to do a single `if` here (with an `and`),  but if you think it's easier to read this way, I don't have a problem leaving as-is.

kitsune/sumo/static/sumo/js/experiment-fxa-cta-topbar.js
```diff
@@ -12,37 +12,48 @@
...
         id: 'experiment-fxa-cta-topbar',
         customCallback: showPromotion,
         variations: {
-          'a': 25,
-          'b': 25,
-          'c': 25,
-          'd': 25
+          'a': 34,
+          'b': 33,
+          'c': 33,
```
jpetto: Nit - drop the trailing comma.

glogiotatidis:
That's now in prod

PR #3208, opened 2018-05-18
Re issue #3173: Add a max limit to ImageAttachment queries

pmac:


jgmize:
merging with existing tests passing; @pmac said he will follow up with additional tests in a follow-up PR via IRC

PR #3211, opened 2018-05-19
Add limit to answers returned in question feeds

pmac:


kitsune/questions/feeds.py
```diff
@@ -112,7 +112,8 @@ def description(self, question):
         return self.title(question)
 
     def items(self, question):
-        return question.answers.filter(is_spam=False).order_by('-created')
+        return question.answers.filter(
+            is_spam=False).order_by('-created')[:config.ANSWERS_PER_PAGE]
```
safwanrahman: I think another optimization can be done by using `select_related('creator')` and

pmac: could be. I'll take a look. thanks!

PR #3229, opened 2018-05-22
Re-enable questions list for all products

jgmize:


metadave:
:+1:

glogiotatidis:
Thanks!

glogiotatidis:
We keep feeds disabled? 

/cc @pmac

pmac:
Well they don't currently work even when enabled. I'm hoping to fix that in the next couple days and then we can enable feeds that also work :)

glogiotatidis:
pushed to prod

AliceWyman:
> We keep feeds disabled?
> /cc @pmac

> Well they don't currently work even when enabled. I'm hoping to fix that in the next couple days and then we can enable feeds that also work :)

Related bug: https://bugzilla.mozilla.org/show_bug.cgi?id=1396792

PR #3240, opened 2018-05-30
Fixing two typos

vesper-mozilla:
Policies for Firefox Enterprises > Policies for Firefox Enterprises
Firefox Quantum for businesses. > Firefox Quantum for businesses

Thanks to Noah and Tonnes


glogiotatidis:
Thank you @vesper-mozilla, Tonnes and Noah for this.

- `db-strings.py` contains strings that are extracted from the database. 
- We extract the strings into a python file, so we can then generate `.po` files and give them out for translation.
- This is a one-way process, i.e. strings from `db-strings.py` are not imported in the db

I'm going to close this PR, not because the suggestions are not alright but because these need to happen in SUMO itself. In other words through the Django admin or other editing interface. Then I'm going to extract the strings and `db-strings.py` will get updated.

If something is unclear let me know

Thanks!

Issue #2967, opened 2017-11-21
Setup Jenkins pipeline


glogiotatidis:
Setup Jenkins pipeline to:

 - Build
 - Test
 - Upload to docker hub
 - Deploy to k8s

akatsoulas:
This issue is too old and probably resolved by now. Please open a new issue if this is not the case.

PR #3245, opened 2018-06-06
[Issue #2967] Update Jenkinsfile.

glogiotatidis:


Kase33:
thank you! There are so many of you out there who makw it all possible!

Jenkinsfile
```diff
@@ -1,62 +1,89 @@
...
+            dockerImagePush("${config.project.docker_name}:full-${GIT_COMMIT_SHORT}", "mozjenkins-docker-hub")
+            dockerImagePush("${config.project.docker_name}:full-no-locales-${GIT_COMMIT_SHORT}", "mozjenkins-docker-hub")
+            dockerImagePush("${config.project.docker_name}:locales-${GIT_COMMIT_SHORT}", "mozjenkins-docker-hub")
+            dockerImagePush("${config.project.docker_name}:staticfiles-${GIT_COMMIT_SHORT}", "mozjenkins-docker-hub")
+            dockerImagePush("${config.project.docker_name}:base-dev-${GIT_COMMIT_SHORT}", "mozjenkins-docker-hub")
+            dockerImagePush("${config.project.docker_name}:base-${GIT_COMMIT_SHORT}", "mozjenkins-docker-hub")
+        }
 
-  milestone()
+        stage("Run Flake8") {
```
pmac: To me it makes more sense to run these tests before pushing the images to docker hub? Is there a reason to push images that fail tests?

glogiotatidis: Yes and no.

Ideally I agree we shouldn't be pushing images before tests pass. And that's why we don't mark images as `latest` unless tests pass.

In CircleCI we got this annoying `mysql has gone away` error msg and I changed the code to push images before running the tests anyway, so I could get an image and test locally or push directly if it's about a change that didn't require running the tests.

The problem with Kitsune tests is that they take toooo long so we need to make steps towards improving the speed (`test --parallel` with Django 1.11 or fancier things with `py.test`) so pushing early makes things available  to start the manual work  earlier.

glogiotatidis: On second though:

- This pipeline runs on commits pushed to `mozilla/kitsune` so why not have images pushed to docker hub independently of the status of the tests? It shouldn't hurt and it's possible helpful with debugging?

pmac: That seems fair. Proceed!

PR #3249, opened 2018-06-13
[Bug 1460649]: Display changes in translation review side by side

MikkCZ:
I have tried to set up Kitsune Docker as per the instruction in the README, but even after I somehow bent pip to install all dependencies despite complaining about some incompatible versions, I was unable to make application start. So I this patch is more reverse engineered and tested manually by editing the HTML and CSS generated by Less via DevTools. For that reason I also limited the changes only to the review_translation view and not any another.

@glogiotatidis r?

Note: all the HTML templates look messy to me, and many times duplicated. Is there some plan to change that (make better use of semantic HTML5 etc.)?

### DIFF
--- a/kitsune/sumo/static/sumo/less/wiki.less
+++ b/kitsune/sumo/static/sumo/less/wiki.less
@@ -695,6 +695,11 @@ textarea[name="content"] {
       }
     }
   }
+
+  .half-width {
+    display: inline-block;
+    width: 49%;
+  }
 }
 
 .html-rtl {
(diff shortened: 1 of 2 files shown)


glogiotatidis:
thanks for the pr @MikkCZ.  Please note that I'll be focusing on other projects for the next month(s). other folks will spend time reviewing and merging prs meanwhile.

MikkCZ:
This PR is stale for too long, so I am closing this. The history should be preserved, so if anyone want's to update it or use as inspiration, feel free to.

PR #3252, opened 2018-06-15
Change product cards style on home page based on new mockup.

mlagace:
* Merge may-2018-redesign (#3230) to resolve conflicts
* Implement new design for product cards on home page

sumo-product-cards-new


mlagace:
The new icons were added in the commit but they will need to be added through the /admin console.

PR #3253, opened 2018-06-18
Update search results page style for redesign

mlagace:
* Updated the search results page style based on the mockup redesign we received.

image


jpetto:
I'm getting some odd display results:

screen shot 2018-06-29 at 12 32 12 pm

Any ideas?

mlagace:
This seems to be some odd behaviour with cached content. If I switch between branches that don't have the new search styles, and do some searches and then switch back to the **feature/search-results-redesign** branch, I get a messed up result like you pasted above when I do a search. After clearing all my history/cache/cookies, I do a new search and I now see the proper styles like the image I added in my first comment.

jpetto:
Hmm, I've stopped Docker, killed all remaining containers, restarted, and tried the search from private windows/tabs in Firefox, Chrome, and Safari. All return the same result from my last screenshot.

Do you think this has anything to do with my local install only having 3 products?

mlagace:
Rebased and I fixed the issue @jpetto. I forgot to remove the `doctype-filter` list in the sidebar since it's now in search-results-list.html.

kitsune/search/jinja2/search/results.html
```diff
@@ -33,23 +33,6 @@
         {% if product %}
```
jpetto: Do we need any of this `base_doctype_url` logic here anymore? Doesn't look like it's used now...

kitsune/sumo/static/sumo/tpl/search-results.html
```diff
@@ -4,24 +4,6 @@
     {% if product %}
       {% set base_doctype_url = base_doctype_url|urlparams(product=product.slug) %}
```
jpetto: Same question here - can we remove this `base_doctype_url` logic?

kitsune/sumo/static/sumo/tpl/search-results.html
```diff
@@ -4,24 +4,6 @@
...
-      <li {{ w|class_selected(2) }}>
-        <a href="#" data-instant-search-set-params="w=2" data-instant-search="link" data-href="{{ base_doctype_url|urlparams(w=2)|encodeURI }}">
-          {{ _('Community Discussion Only') }}
-        </a>
-      </li>
-    </ul>
-
     {% set base_product_url = base_url %}
     {% if w != 3 %}
       {% set base_product_url = base_product_url|urlparams(w=w) %}
```
jpetto: I'm not seeing any UI response when I click on one of the products in the sidebar list when searching via the sticky nav. I see a network request is made, but there's no visual indication that one of the selections (e.g. "All Products", "Firefox") has been chosen/applied as a filter.

mlagace: The product gets highlighted when you click on it and it shows a black left border meaning it's selected though. I know before the search title wasn't all bold and it showed the product selected in bold in the title. Now it's in the title but we can't really tell what is the product unless you look at the selected product in the sidear. Maybe show the product as a link in the title?

image

jpetto: Are you getting the UI from the above screenshot using the instant/live search? I don't get any bolding - from this PR locally or even in prod. It seems the `selected` class attribute is being rendered incorrectly:

screen shot 2018-07-06 at 3 26 03 pm

As this appears to be a pre-existing bug, I suppose we don't need to block this PR. I'll file a separate issue.

jpetto: A quick test shows that changing `return ' class="selected" ';` to `return ' class=selected ';` in [`/kitsune/sumo/static/js/nunjucks.js#53`](https://github.com/mozilla/kitsune/blob/master/kitsune/sumo/static/sumo/js/nunjucks.js#L53) solves the issue.

If you want to double-check that and include it here, please feel free. Otherwise, will file an issue and separate PR.

Thanks!

mlagace: Yeah I got that screenshot from the instant/live search. I don't see the second set of double quotes like you do. I see:

image

PR #3255, opened 2018-06-18
Update Product Help Topics page style for redesign

mlagace:
* Updated the Product help topics page style based on mockup for redesign.

The implemented style looks like this:

image

The mockup looks like this:

image

As you can see, the icons in the mockup are square in the card title but the current icons aren't. The current icons won't look good with the new design so they'll have to be updated with new icons.

Also, as you can see in the mockup, in each card there's a list of subtopics in lightgrey. The current code doesn't look like it supports retrieving subtopics on the Product Help Topics page so the code would have to be changed to retrieve the top 2-3 popular subtopics for each topic.


jpetto:
Without the subtopics being available, I don't think the new UI/UX holds up. This is a rhetorical question, but why not just have the topic name/title be the link? As is, "See all" doesn't offer any value, and IMO makes the UX _worse_.

If updating the code to retrieve the top 3 subtopics isn't in scope, I think we need to re-assess the proposed design.

Also, assuming we continue with at least some version of the mockup, what is the proposed plan for updating the icons?

mlagace:
Updating the code to retrieve the top 3 subtopics wasn't in scope but we thought maybe it would be done in a next phase so that is why I just implemented the design as in the mockup for now but I agree that just having "See all" doesn't look great. For now I can remove the "See all" part and just show the icon title with the whole card being light grey.

As for the icons, they'll have to be updated in the /admin. I added all the new icons they gave us with the mockup in this commit but there wasn't any details in which icon should be used for which topic though.

jpetto:
Just wanted to give an update that I'm checking with creative on how we should move forward on some of these design compromises. Will get back to you as soon as I have more info.

mlagace:
Thanks @jpetto !

Dinodog:
Can we leave out the list of article, remove the "see all" link and just link to the sub  topic page?

sgarrity:
> Can we leave out the list of article, remove the "see all" link and just link to the sub topic page?

@Dinodog To confirm, is something like this what you have in mind (removing the sub-list and 'see all')? Looks reasonable, I think:

image

madasan:
re:icons @mlagace here's the link to the new icons: https://github.com/feathericons/sketch-feather

jpetto:
@mlagace - Is there anything else you need here to move forward?

mlagace:
No @jpetto I will make the changes. Thanks  @Dinodog , @jpetto and @madasan.

jpetto:
Looking good for a merge after a rebase. 👍

mlagace:
Rebased

PR #3256, opened 2018-06-18
Update Product documents page style for redesign.

mlagace:
Updated the Product documents.html page style based on the mockup for the redesign.

Implemented style:
image

In the mockup there wasn't any design for subtopics in the sidebar when a subtopic is selected so for now I implemented it as you see above. If there are no subtopics selected you will see this instead:
image

Also in the mockups all icons are included in the topic sidebar but not in the search results product sidebar so for now since the current topic icons didn't look nice with the topic sidebar, I didn't add the topic icons to the topic sidebar to make it a bit more consistent with the search results sidebar as well.

Here's the mockup for it:
image

I also didn't get time to add the breadcrumbs over the document list like in the mockup so I left the breadcrumbs as is for now. The fonts should also look nicer with the right weights once they're merged with the Zilla Slab font changes.


jpetto:
@Dinodog and/or @madasan - Can you weigh in on the changes from the mockups here?

Dinodog:
These changes look fine @jpetto.

jpetto:
This looks good to merge after a rebase. 👍

mlagace:
Rebased

Issue #3017, opened 2018-01-16
[Bug 396792] Intermittant RSS Issue


bensternthal:
Intermittent issue with RSS:

https://bugzilla.mozilla.org/show_bug.cgi?id=1396792

bensternthal:
@pmac any chance you can take a look at this sometime this week?

pmac:
Sure.

pmac:
Progress: I've narrowed it down to something to do with the Meinheld Gunicorn worker we're using. The feeds work fine using both Django's development server and the default `sync` Gunicorn worker. Investigation continues.

pmac:
Feeds also work fine with the officially supported `gevent` worker (which uses `greenlet` like Meinheld). I'm not sure how the rest of the site would work with the `gevent` worker, but it might be worth a shot. Meinheld's maintenance is slow-to-none as well (last release in Nov 2016), so that's another reason to try something else.

Thoughts @jgmize @metadave @glogiotatidis @jpetto ?

pmac:
I could put my changes in a branch, deploy that to one of the stage instances (maybe frankfurt) and we could test if you think it's worth it. I'm also still looking for issues in Meinheld to see if I can explain the behavior. I'm still not sure what makes the feed views special in this way.

pmac:
PR above added for experiment with `gevent` Gunicorn worker class.

jgmize:
While we may not have tested gevent with kitsune, we did with snippets-service and ran into errors. If a different gunicorn worker resolves the rss issue but introduces other errors on other urls, then we could put a path based ingress that directs just the rss urls to a new service backed by a separate deployment if needed.

pmac:
> It might be worth building a release of our own from the master branch, which has commits from this year.

I considered that, but I'm really quite uncomfortable with it. I don't think the changes are related to our issues, and that would mean running an unreleased version of a beta server... I just don't think it's a good idea.

jgmize:
meinheld  is far from perfect, but it's been more performant and reliable than any of the other gunicorn workers I've tried, including gevent. I've tested gevent with several projects over the years, and each time it created more problems than it solved, often due to its monkeypatching of the standard library. Feel free to test it again though, it's been a while since I tried it last and things might be better now.

pmac:
[...]

PR #3265, opened 2018-07-07
Upgrade meinheld to latest commit on master branch for #3017

jgmize:


jgmize:
Replying to https://github.com/mozilla/kitsune/issues/3017#issuecomment-403073130
> > It might be worth building a release of our own from the master branch, which has commits from this year.

> I considered that, but I'm really quite uncomfortable with it. I don't think the changes are related to our issues, and that would mean running an unreleased version of a beta server... I just don't think it's a good idea.

I disagree. I've reviewed https://github.com/mopemope/meinheld/compare/0427b44e1b25c6ae67975a403997ea1148a7310b...master and I believe it's ready for production. Whether it will resolve #3017 is another question, but I believe it's worth testing the docker image built with this PR to find out.

pmac:
I'm not comfortable doing this. I don't think the Meinheld project is worth us spending time on and I don't trust it. If you are comfortable with it and do want to take on figuring out the WSGI issues (mostly #3017) then go for it. But for the record my vote is no.

PR #3267, opened 2018-07-09
Squash feeds response object prior to response

### LINKED ISSUE #3017: [Bug 396792] Intermittant RSS Issue

pmac:
Something about how Meinheld works doesn't like a Response object
that has been written to more than once. This should fix Feeds.

Fix #3017

### DIFF
--- a/kitsune/questions/feeds.py
+++ b/kitsune/questions/feeds.py
@@ -1,4 +1,3 @@
-from django.contrib.syndication.views import Feed
 from django.shortcuts import get_object_or_404
 from django.utils.feedgenerator import Atom1Feed
 from django.utils.html import strip_tags, escape
@@ -9,6 +8,7 @@ from taggit.models import Tag
 from kitsune.products.models import Product, Topic
 from kitsune.questions import config
 from kitsune.questions.models import Question
+from kitsune.sumo.feeds import Feed
 from kitsune.sumo.urlresolvers import reverse
 from kitsune.sumo.templatetags.jinja_helpers import urlparams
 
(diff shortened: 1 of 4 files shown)


kitsune/questions/feeds.py
```diff
@@ -13,7 +13,23 @@
...
+    nature of the response cycle doesn't properly handle responses that have more than
+    one item in the `Response._container` list. This class squashes that list before
+    returning the HttpResponse object.
+
+    See https://github.com/mozilla/kitsune/issues/3017
+    """
+    def __call__(self, request, *args, **kwargs):
+        response = super(SyncFeed, self).__call__(request, *args, **kwargs)
+        # this squashes the response._container list to a single item
+        response.content = response.content
```
jgmize: I wouldn't have realized that this did anything at all without the comment above. Interesting.

jpetto: But _how_ is the above working?

pmac: Yeah. It's weird. In the `HttpResponse` object `response.content` [calls a property method](https://github.com/django/django/blob/stable/1.8.x/django/http/response.py#L354) that returns a join of all the items in the `_container` list. [Assigning to that property](https://github.com/django/django/blob/stable/1.8.x/django/http/response.py#L367) puts the given string into a list of length 1.

pmac: The code isn't clear, but it is the most succinct way of reducing that list to length 1 using the built-in methods.

PR #3272, opened 2018-07-19
Fix #3271: Upload static assets to s3

### LINKED ISSUE #3271: Move static asset hosting to s3
The static files (JS, CSS, Images, etc.) are currently being served from the app servers. This can be an issue during deployment when some pods are serving new HTML while others don't yet have the new media files. This mismatch period can be several minutes since we deploy the 2 clusters separately.

See https://github.com/mozilla/bedrock/pull/5932

pmac:
**NOTE**: Depends on https://github.com/mozmeao/infra/issues/890 to actually work.


metadave:
buckets and IAM stuff all set.

pmac:
> buckets and IAM stuff all set.

Thanks Dave!

pmac:
I've changed tack for the separation of hashed vs. normal filenames. I wrote a python script that loops over the hashed files in the `staticfiles.json` file and moves them to another directory for use with the `aws s3 sync` command. This should be much more reliable. I'll port it to bedrock once this is proved to work and lands successfully here.

Also it turns out that kitsune wasn't using manifest staticfiles storage, but cached. This means that I believe every call to a static file in a template was causing a cache lookup at a minimum, and if the cache missed it would recalculate the hash of the file and then set it in cache.... so this includes the switch to manifest staticfiles storage.

/cc @glogiotatidis @metadave @jgmize @jpetto

pmac:
Thanks Dave! I'm going to go ahead and merge this so that I can begin testing on stage.

PR #3285, opened 2018-08-03
[fix bug 1416244] Enable new email collection survey.

jpetto:


jpetto:
@pmac I believe this is now ready for a final review.

Only open question on my end is if we should keep the changes to the deployments (looking for SurveyGizmo API secrets).

pmac:
> Only open question on my end is if we should keep the changes to the deployments (looking for SurveyGizmo API secrets).

I don't think we should keep those. Stage doesn't need them. We don't necessarily have to remove them from the secrets repo, but removing them from the stage.yml file in the repo seems appropriate.

PR #3290, opened 2018-08-09
[fix 3289] Update requirements.

jpetto:


jpetto:
Getting Mocha test failures when updating jQuery, so I'll do that in a separate PR.

PR #3297, opened 2018-08-14
Port relevant bits of django-badger into kitsune (Django upgrade pre-req)

jpetto:


jpetto:
@pmac While we wait for the redesign to get launched (and free up dev), I think this is in a spot for at least a good once over in terms of code review. Catching major/obvious issues prior to opening up QA on dev for stakeholders is probably a good idea, with final code review coming after manual QA by Madalina/Patrick/et al.

(Please ignore the changes to `Jenkinsfile` for now - those will be removed prior to merge.)

I tried to not rewrite Badger and only make obvious/easy modifications that we can launch _before_ doing the Django upgrade. Once we're confident all is well on Django 1.8, it'll (again, hopefully) be easier and faster to update for the latest Django.

Thanks!

pmac:
Excellent. I'll start looking. 🎉

scripts/update/deploy.py
```diff
@@ -60,6 +60,9 @@ def db_migrations(ctx):
     with ctx.lcd(settings.SRC_DIR):
         ctx.local('python2.7 manage.py migrate --noinput actstream --fake')
         ctx.local('python2.7 manage.py migrate --noinput authority --fake')
+        # relevant bits of django-badger were migrated to kitsune.kbadge, meaning the
+        # database tables already exist, so we can skip the initial migration
+        ctx.local('python2.7 manage.py migrate --noinput kbadge --fake-initial')
```
pmac: I don't believe this script is ever run anymore, but points for thoroughness 💯

jpetto:
This has been rebased, squashed, and cleaned (e.g. changes to `Jenkinsfile` reverted). I think we need to reserve demo for the re-design fixes, so maybe this has to sit in the queue a while longer?

pmac:
hmm... I dunno. If you think we're just about ready we can take over dev for a bit to get this out, then rebase those redesign fix branches and just move forward. it would be nice to get this landed.

jpetto:
Sent an email to the stakeholders to test on dev. Will follow-up in two days if I don't hear back before then.

jpetto:
Roland has confirmed things are looking good on stage. We are set to merge/push to prod when ready - which is probably Monday 10/14 to be safe.

metadave:
nice work @jpetto!

PR #3300, opened 2018-08-24
[WIP] Indexing data using elasticsearch-dsl

safwanrahman:
This PR is still work in progress. It creates seperate index for mapping speicif locales as well as index objects to proper index.
Things need to be done
- [x] Add analyzer for specified locales
- [x] Index Questions objects
- [ ] Index users objects

@pmac Can you please take a look and let me know if the approach is correct? I am using a wrapper around `elasticsearch-dsl` named [`django-elasticsearch-dsl`](https://github.com/sabricot/django-elasticsearch-dsl). What do you think about using the wrapper?


pmac:
Sorry for the delayed response. I'm taking it in. I think the approach with django-elasticsearch-dsl can work fine for us. The tests in that project look good as well as the python, django, and elasticsearch-dsl version support. I'm a little concerned with the activity and size of the project however. Do you think it gets us enough of an advantage that it'd be worth it to use it instead of manually doing the indexing ourselves? It looks like a lot of code is required regardless, and introducing a django integration dependency is a risk (as we've seen with elasticutils).

safwanrahman:
@pmac Thats true actually. For our specific kind of feature (different analyzer for different locale), we need to overwrite many code of `django-elasticsearch-dsl`. I think we can make our implementation by taking idea from django-elasticsearch-dsl and implement it using elasticsearch-dsl. What do you think?

safwanrahman:
@pmac and I have discussed over IRC and come to decision to use `django-elasticsearch-dsl` package because its just a thin wrapper around `elasticsearch-dsl-py` and we need a generalize approach to index all the models. 👍

kitsune/search/config.py
```diff
@@ -1,3 +1,57 @@
...
+    'ja': 'cjk',
+    'pl': 'polish',
+    'th': 'thai',
+    'zh-CN': 'chinese',
+    'zh-TW': 'chinese',
+}
+
+# Locales that uses snowball tokenizer
+ES_SNOWBALL_LOCALES = {
+        'eu': 'Basque',
```
pmac: NIT: these items are over indented.

pmac:
@safwanrahman this is looking good. I need to get it setup and running to see where we are.

safwanrahman:
I have pushed the document for indexing `Question` objects also, and it seems to be working fine!

-------------

image

kitsune/search/views.py
```diff
@@ -15,8 +15,8 @@
 
 import bleach
 import jinja2
-from elasticutils.utils import format_explanation
-from elasticutils.contrib.django import ES_EXCEPTIONS
+# from elasticutils.utils import format_explanation
+# from elasticutils.contrib.django import ES_EXCEPTIONS
```
pmac: Just delete these imports

kitsune/search/views.py
```diff
@@ -176,6 +176,8 @@ def simple_search(request, template=None):
 
     searcher = generate_simple_search(search_form, language, with_highlights=True)
     searcher = searcher[:settings.SEARCH_MAX_RESULTS]
+    import json
+    print(json.dumps(searcher.build_search()))
```
pmac: debugging statement? 🖨

pmac:
Thanks again @safwanrahman. I'm going to go ahead and merge into `es_upgrade` and we can continue to work from there. I'll also rebase that branch on master after this merges.

PR #3302, opened 2018-08-24
Raise instead of return HttpResponseForbidden in middleware

jgmize:


pmac:
`HttpResponseForbidden` is a subclass of `HttpResponse` and should be returned by a view or middleware. I'm not sure what this would do. Have you tried it?

jgmize:
No, I saw it raised in [an example](https://www.programcreek.com/python/example/52376/django.http.HttpResponseForbidden) and thought we might test and see if this returns a 401 instead of a 504.

PR #3309, opened 2018-09-03
Adding two non-supported locales for redirects to work + fixing two spelling issues

vesper-mozilla:
https://bugzilla.mozilla.org/show_bug.cgi?id=1483506


pmac:
The locale additions look fine. The modifications to the `db_strings.py` file I don't think should be included. That file is generated by a script that we run to get strings out of the database and then extract them for translation. In order to fix typos there they need to be updated in the database first, then we can extract them. If they're updated here but not the DB then the fixes will be overwritten during the next extract.

vesper-mozilla:
Gotcha, @pmac, reverted the changes.

pmac:
Merged to master as e399b0e147752bde0737062f9ac2587282e0b10d. Thanks again!

pmac:
@vesper-mozilla also I've fixed those typos in the database and will be updating that file soon. Thanks for pointing those out.

vesper-mozilla:
@pmac Awesome, thank you! I assume that this means the code is already live and if it works as it should, these redirects should go to English instead of 404ing?

https://support.mozilla.org/1/mobile/3.1/Android/jv/screenshot-telemetry
https://support.mozilla.org/1/mobile/3.1/Android/su/screenshot-telemetry

pmac:
That should be the case once it's deployed. I'll try to get a deployment done today. These changes are not yet in production. See the "[what is deployed](https://whatsdeployed.io/s-iiX)" tool for current status.

pmac:
This is now in production and seems to work well! Thanks again.

PR #3312, opened 2018-09-11
May 2018 redesign

jpetto:
All work included here was reviewed in smaller PRs before being merged to this branch. Would be good to take one more sanity check look over things though.


jpetto:
@glogiotatidis If you could just give this code a good once over, I think that would be sufficient. No need to dive into the details - just make sure nothing jumps out at you.

Thanks!

safwanrahman:
I think #3288 not merged yet!

kitsune/products/views.py
```diff
@@ -43,16 +44,19 @@ def product_landing(request, template, slug):
         else:
             latest_version = 0
 
-    # Don't show firefox download button at header at firefox product page
-    hide_fx_download = product.slug == 'firefox'
+    user_agent = request.META.get('HTTP_USER_AGENT', '')
+    browser = get_browser(user_agent)
+
+    # Only show firefox download button when product is firefox and browser is not Firefox
+    show_fx_download = (product.slug == 'firefox' and browser != 'Firefox')
```
glogiotatidis: nit (and maybe not important): `get_browser` does case insensitive browser matching but here we only check for `Firefox`. Maybe it's worth using `lower()`

jpetto: [The tests](https://github.com/mozilla/kitsune/blob/master/kitsune/sumo/tests/test_utils.py#L188) check for `Firefox`, so I think we're good. But you raise a valid point - we're relying on the UA string capitalizing browser names, which is true now and probably will be in the future, but is still a potentially brittle check. I have to do some follow-up on this branch - I'll see about standardizing on lowercase for all instances.

madasan:
A few things:

1. There’s no Firefox download button in the header on any page except Firefox product pages. 
2. Firefox button appears twice in the product page (both in the header and body). 

There has been a bit of confusion around this and we apologize for this. The logic should be as follows:

Show Firefox download button in header on all pages EXCEPT:

- If the visitor is already using Firefox (on any page)

- If the current page is the Firefox product page, do not show (as there is already a larger download button on that page). This includes all category and article pages under the Firefox product, that should all have the larger download button in the body of the page.

home page

product page

category page 2

article page

madasan:
3. All the tiles on the category page have the same cogwheel icons

screenshot_2018-09-14 firefox help

Expected behaviour:

- There should be different icons for each category

madasan:
4. The product logo is showing as a "?" on all topic pages

screenshot_2018-09-14 firefox for android help

screenshot_2018-09-14 firefox for fire tv help

5. On the "Firefox for Fire TV" topic page, the title is bumping into the product name. See above ^

rtanglao:
icons are squashed on fennec:
https://support.mozilla.org/en-US/forums/contributors/713162?last=74688
2018-09-14-08-14-28-3132ac-fennec-sumo-squashed-icons

can we fix this before launch?

madasan:
6. More squashed icons. This time on Desktop when visiting the AAQ (Get Community Support)
screenshot_2018-09-17 ask a question mozilla support

madasan:
7. The white background breaks the forum experience a bit. See: https://dev.sumo.moz.works/en-US/questions/all?show=all

vesper-mozilla:
Localization issues: Some strings seem to be hard-coded - they are not localized or localizable (I can't see them in Pontoon, our UI localization tool (https://pontoon.mozilla.org/projects/sumo/info/)

(screenshots for the Polish version with any visible English text being the issue)

 "Find help" in the search box and "Help articles" on top of the product list:
screenshot_2018-09-18 pomoc mozilli

"Get community support" and "Volunteer" in the top bar:
screenshot_2018-09-18 pomoc mozilli 1

Parts of the footer:
screenshot_2018-09-18 pomoc mozilli 2

"Search and navigate easily with these essential features" on the topmost level of product KB navigation:
screenshot_2018-09-18 pomoc dla programu firefox

"These fine people helped write this article:" and "You can help too - find out how" at the bottom of each KB article page.
screenshot_2018-09-18 jak dzialaja miniaturki pomoc dla programu firefox

vesper-mozilla:
Localization issue: the font used on the subpages of https://dev.sumo.moz.works/get-involved for locales that include diacritics does not render well and looks rather bad. Could we have another (full internationalized) font instead?

screenshot_2018-09-18 pomoc mozilli 3
screenshot_2018-09-18 pomoc mozilli 4

jpetto:
All in this PR has been verified on stage. We are good to merge!

PR #3323, opened 2018-09-18
Upgrade Node to 6.x and use Yarn for install

### LINKED ISSUE #3322: Switch from NPM to Yarn for Node dependency installation
Yarn will lock and cryptographically verify all dependencies so that we can get more repeatable deployments.

pmac:
Fix #3322


jgmize:
Please read the articles linked in https://github.com/mozilla/kitsune/issues/3322#issuecomment-422457326 before moving forward with this.

docker/dockerfiles/staticfiles
```diff
@@ -3,18 +3,19 @@ FROM python:2-stretch
...
     rm -rf /var/lib/apt/lists/*
 
 ENV PATH="/venv/bin:$PATH"
 
-COPY ./package.json /app/package.json
-COPY ./bower.json /app/bower.json
+COPY package.json bower.json yarn.lock /app/
 
-RUN npm install
+RUN yarn install --pure-lockfile
```
jpetto: Should we be using [--frozen-lockfile](https://yarnpkg.com/lang/en/docs/cli/install/#toc-yarn-install-frozen-lockfile) instead?

pmac: Hmm.. not sure. Sounds like it might fail more. You think we should?

PR #3333, opened 2018-09-24
[fix #3281] Update Fx download button display logic.

### LINKED ISSUE #3281: Add download button for non-Firefox browsers on home page
It appears the header Firefox download button (that only shows in non-Firefox browsers) is not appearing on the home page. It should!

From this comment: https://github.com/mozilla/kitsune/issues/3269#issuecomment-407227694

jpetto:


jpetto:
I'm pretty sure the logic here is now finalized. Should definitely push to dev when code is approved to allow stakeholders to verify implementation.

kitsune/products/jinja2/products/documents.html
```diff
@@ -27,7 +27,9 @@
     {{ topic_sidebar(topics[:10], subtopics, product, topic, subtopic) }}
   </aside>
   <div class="grid_9">
-    {% if show_fx_download %}
+  {% if product.slug == 'firefox' %}
```
safwanrahman: why only show in firefox product page? can you explain where its decided to show in the firefox products document only?

jpetto: This is the requirement set by @madasan. Perhaps she can shed some light on the decision?

kitsune/products/jinja2/products/documents.html
```diff
@@ -27,7 +27,9 @@
...
+    {# show-fx-download.js will remove the 'hidden' class below if the visitor is *NOT* using Firefox #}
```
pmac: 🎉 for client-side detection 🎉

kitsune/sumo/templatetags/jinja_helpers.py
```diff
@@ -507,3 +507,22 @@ def image_for_product(product_slug):
     except Product.DoesNotExist:
         return os.path.join(settings.STATIC_URL, 'products', 'img', 'product_placeholder.png')
     return obj.image_url
+
+
+@jinja2.contextfunction
+@library.global_function
+def show_fx_download(context):
```
pmac: Looks like you decided against using this?

kitsune/sumo/templatetags/jinja_helpers.py
```diff
@@ -507,3 +507,22 @@ def image_for_product(product_slug):
...
+    product being displayed.
+
+    This function is used in conjunction with show-fx-download.js. This function simply adds
+    markup to the template wrapped in a 'hidden' class. The JS removes the 'hidden' class if
+    the visitor is *NOT* already using Firefox.
+    """
+    product = context.get('product', None)
+
+    if product:
+        return product.slug != 'firefox'
```
pmac: Isn't this backward?

kitsune/sumo/templatetags/jinja_helpers.py
jpetto: It's used in `sumo/jinja2/base.html` - only for the download button in the header.

kitsune/sumo/templatetags/jinja_helpers.py
jpetto: Sigh, no, but it is very confusing. tl;dr - I think I'll rename this method to `show_header_fx_download`.

The download button in the header should always be shown _unless_ the visitor is on a Firefox product page (or any derivative thereof - e.g. a KB article). If the visitor lands on a page where `context['product'].slug == 'Firefox'`, the big, in-page download button will be shown - hence hiding the header download button with this function.

pmac: Ah! Makes sense now. It's supposed to be "backward" from what I was thinking :)

kitsune/products/jinja2/products/documents.html
madasan: @safwanrahman There's two Firefox download buttons: one in the body of the page, one in the header.

The one in the body should be displayed on all Firefox product pages
The one in the header should be displayed on all the other pages (not on the Firefox product ones we already have a button there).

That sounds easy but getting that logic right in the code might be a bit complicated :)

PR #3334, opened 2018-09-25
Update mixins.less

LunarTwilight:
Fixes 
image


pmac:
This appears to be a legit CSS technique for targeting IE7 and below. See:

https://stackoverflow.com/a/14927670

But I'll let @jpetto check me in case I'm very wrong.

PR #3340, opened 2018-09-28
Allow IPv4 addresses when ENFORCE_HOST is set

### LINKED ISSUE #2998: Redirect support.mozilla.com to support.mozilla.org

pmac:
Fix #2998


pmac:
The app we use for the redirects doesn't strip the port from the host for checking, so `support.mozilla.org:8080` would redirect. This isn't a big deal, but I did file a PR to fix this:

https://github.com/dabapps/django-enforce-host/pull/1

PR #3349, opened 2018-10-03
Add sumo architecture diagrams

metadave:


metadave:
https://github.com/mozmeao/infra/issues/944

jgmize:
- I don't think we need the plist or tiff files, just the png
- the sumo-prod ELB should point directly at the kubernetes clusters
  - traffic from the sumo-prod ELB doesn't go through the sumo-prod-a or sumo-prod-b ELBs

metadave:
diagram updated. 

- the plist/tiff/directory structure is for opening in Omnigraffle (right click on .graffle file, click Show Package Contents). I'd prefer to keep the original file next to the image so we don't have to redraw the diagram in the future,

PR #3350, opened 2018-10-03
Add running of post-deployment tasks to deployment

pmac:
Re #3133

### DIFF
--- a/k8s/commander.sh
+++ b/k8s/commander.sh
@@ -14,13 +14,14 @@ function deploy {
     REGION_ENV=${3}
     COMMIT_HASH=${4}
     DEPLOY_SECRETS=${5:-NO}
+    K8S_NAMESPACE="sumo-${REGION_ENV}"
 
     export KUBECTL_BIN="./regions/${REGION}/kubectl"
     export KUBECONFIG="./regions/${REGION}/kubeconfig"
 
     if [[ "${DEPLOY_SECRETS}" == "secrets" ]]; then
         echo "Applying secrets";
-        ${KUBECTL_BIN} -n sumo-${REGION_ENV} apply -f "regions/${REGION}/${REGION_ENV}-secrets.yaml"
+        ${KUBECTL_BIN} -n "${K8S_NAMESPACE}" apply -f "regions/${REGION}/${REGION_ENV}-secrets.yaml"
     else
         echo "Secrets will *NOT* be applied";
     fi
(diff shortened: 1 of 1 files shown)


k8s/commander.sh
```diff
@@ -31,6 +32,12 @@ function deploy {
     invoke  -f "regions/${REGION}/${REGION_ENV}.yaml" rollouts.status-cron
     invoke  -f "regions/${REGION}/${REGION_ENV}.yaml" deployments.create-web --apply --tag full-${COMMIT_HASH}
     invoke  -f "regions/${REGION}/${REGION_ENV}.yaml" rollouts.status-web
+
+    # run post-deployment tasks
+    echo "Running post-deployment tasks"
+    SUMO_POD=$(${KUBECTL_BIN} -n "${K8S_NAMESPACE}" get pods | grep "${K8S_NAMESPACE}" | cut -d ' ' -f 1 | head -n 1)
```
metadave: it might be easier to use the jsonpath output filter of kubectl to get the first pod:
```
kubectl -n sumo-stage get pods -o=jsonpath={.items[0].metadata.name}
```

See also: https://kubernetes.io/docs/reference/kubectl/jsonpath/

pmac: Oooooo... nice! Thanks. I'll update. That is better. 🌮

pmac: Updated 🎉

metadave: :taco: :taco: :taco:

metadave:
:+1:

jgmize:
Running migrations post-deployment instead of pre-deployment could result in errors during the deployment if you have a new model or field.

k8s/commander.sh
```diff
@@ -31,6 +32,12 @@ function deploy {
     invoke  -f "regions/${REGION}/${REGION_ENV}.yaml" rollouts.status-cron
     invoke  -f "regions/${REGION}/${REGION_ENV}.yaml" deployments.create-web --apply --tag full-${COMMIT_HASH}
     invoke  -f "regions/${REGION}/${REGION_ENV}.yaml" rollouts.status-web
+
+    # run post-deployment tasks
+    echo "Running post-deployment tasks"
+    SUMO_POD=$(${KUBECTL_BIN} -n "${K8S_NAMESPACE}" get pods -o=jsonpath={.items[0].metadata.name})
```
jpetto: [The docs](https://kubernetes.io/docs/reference/kubectl/jsonpath/) have single quotes around the value passed to `jsonpath`, e.g.

`kubectl get pods -o=jsonpath='{.items[0].metadata.name}'`

Are those needed here?

pmac: I tried running exactly this and it seemed to work. might be better to quote. I'll try it.

pmac:
> Running migrations post-deployment instead of pre-deployment could result in errors during the deployment if you have a new model or field.

Running migrations pre-deployment wouldn't do anything since the new migrations will be part of the thing that is being deployed. Or am I misunderstanding what you're saying?

k8s/commander.sh
pmac: with quotes works too and seems safer. updated.

PR #3358, opened 2018-10-04
[fix #3356] Update .editorconfig to default to 2 spaces for html and less files

### LINKED ISSUE #3356: Update .editorconfig
The existing `.editorconfig` file specifies a default of 4 spaces, but nearly all files use 2 spaces.

jpetto:


.editorconfig
```diff
@@ -3,7 +3,7 @@
 
 [*]
 indent_style = space
-indent_size = 4
+indent_size = 2
```
pmac: All of our Python across all of our apps uses 4 spaces. If you wanna do this you should add a section for Python files.

jpetto:
Yah, sorry. I was typing furiously and have been meaning to change this for some time. Updated to specify indent of 2 spaces for `.html` and `.less` files only.

PR #3362, opened 2018-10-08
[fix #3347] Add UTM params to Fx download buttons.

### LINKED ISSUE #3347: Add UTM params to Firefox download buttons
As reported in this https://bugzilla.mozilla.org/show_bug.cgi?id=1495841 UTM tracking is now missing from Kitsune.

Poking a little deeper I see the following github prs that looks to revert a lot of the custom JS tracking we had implemented as part of the SCL3 migration.  It looks like this was meant to be temporary but I do not see it being re-enabled (unless i am on the wrong branch).

https://github.com/mozilla/kitsune/pull/3138
https://github.com/mozilla/kitsune/pull/3137

jpetto:
For reference, here's the [original bugzilla comment](https://bugzilla.mozilla.org/show_bug.cgi?id=1495841#c1). (Note the updated URLs in a [later comment](https://bugzilla.mozilla.org/show_bug.cgi?id=1495841#c3).)

### DIFF
--- a/kitsune/products/jinja2/products/documents.html
+++ b/kitsune/products/jinja2/products/documents.html
@@ -32,7 +32,7 @@
     <div class="firefox-download-button hidden">
       <div class="documents-download-buttons download-buttons">
         <div class="download-firefox">
-          <a href="https://www.mozilla.org/firefox/download/thanks/" class="download-button"
+          <a href="https://www.mozilla.org/firefox/download/thanks/?utm_source=support.mozilla.org&amp;utm_medium=referral&amp;utm_campaign=non-fx-button&amp;utm_content=body-download-button" class="download-button"
             data-event-category="Download Button"
             data-event-action="Panel Opening"
             data-event-label="{{ latest_version }}">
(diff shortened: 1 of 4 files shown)


kitsune/products/jinja2/products/documents.html
```diff
@@ -32,7 +32,7 @@
     <div class="firefox-download-button hidden">
       <div class="documents-download-buttons download-buttons">
         <div class="download-firefox">
-          <a href="https://www.mozilla.org/firefox/download/thanks/" class="download-button"
+          <a href="https://www.mozilla.org/firefox/download/thanks/?utm_source=support.mozilla.org&amp;utm_medium=referral&amp;utm_campaign=non-fx-button&amp;utm_content=body-download-button" class="download-button"
```
pmac: I didn't think ampersands in URLs in a tags needed to be escaped to `&amp;`. I could be wrong on that thought.

jpetto: It's probably no longer required - I think modern browsers will compensate. We do it on bedrock, so my muscles are trained. 🤷‍♂️

pmac: Great! Works for me then. 👍

PR #3363, opened 2018-10-09
[fix #3341] Improve re-design RTL UX.

### LINKED ISSUE #3341: RTL problems of 2018 redesign
There is a problem in language(locale) selector of newer version of #sumo when a RTL language selected.
Name of locale is on the earth and only left side of language selector is active and responsible!
Look at top of page at
https://support.mozilla.org/fa/
https://support.mozilla.org/ar/
https://support.mozilla.org/he/

jpetto:
There are likely some issues I missed, but I tried to hit the major issues on major pages. If additional issues are found, we can do follow-up PRs.


pmac:
The tests passed, it's just the old "MySQL server has gone away" failure.

PR #3370, opened 2018-10-15
make # of replies the most important meta property

philipp-sumo:
when i browse the list of questions in the support forum as a contributor, the # of replies field is one of the most important factors to determine if i should go into a thread - if there are no replies the user should get an answer as soon as possible; if there are already many replies it's likely a thread with a long history so i'll leave it to those already involved if i don't want to read up on it.

i think it would be more logical in the new redesign to show the number of replies on the left and then have it followed by votes and views (both some measurements of popularity of a thread) - sample:

grafik


jpetto:
Makes sense to me. :shipit:

PR #3375, opened 2018-10-17
[fix #3373] Replace top contributors list with link on /questions pages.

### LINKED ISSUE #3373: top_contributors_questions DB queries are too expensive to include in questions_list view
We have had repeated performance problems where the DB CPU gets stuck near 100% due to queries similar to the following:
```sql
SELECT COUNT(*) FROM 
(SELECT `auth_user`.`id` AS Col1, COUNT(`questions_answer`.`id`) AS `query_count` 
 FROM `auth_user` 
 INNER JOIN `questions_answer` ON 
 (`auth_user`.`id` = `questions_answer`.`creator_id` ) 
 WHERE (`questions_answer`.`id`) IN 
 (SELECT U0.`id` FROM `questions_answer` U0 INNER JOIN `questions_question` U1 
  ON ( U0.`question_id` = U1.`id` ) 
  INNER JOIN `products_product` U4 ON ( U1.`product_id` = U4.`id` ) 
  WHERE (NOT (U0.`is_spam` = %s) AND NOT (U1.`is_spam` = %s) AND NOT (U0.`creator_id` = (U1.`creator_id`)) AND U0.`created` >= %s AND U1.`locale` = %s AND U4.`slug` = %s)) 
 GROUP BY `auth_user`.`id` ORDER BY NULL)
```

While the CPU is spiking, I could see many, many examples of this query in a MySQL `show processlist`, and you can also see an example of this query (and the ~20 sec execution time) [in NR](https://rpm.newrelic.com/accounts/1299394/applications/45097089/datastores#/overview/MySQL/trace/?id=9c10a0d9-d1b3-11e8-9f71-0242ac110007_0_2025&metric=Datastore%252Fstatement%252FMySQL%252Fauth_user%252Fselect)

I believe this query is from [top_contributors_questions](https://github.com/mozilla/kitsune/blob/master/kitsune/community/utils.py#L16) utility function [called by the question_list view](https://github.com/mozilla/kitsune/blob/master/kitsune/community/utils.py#L16). The results of this query are cached for 15 minutes, but when there is a cache miss the repeated attempts to update the cache (especially when particularly aggressive bots are scanning the site) cause the DB to max out the CPU and have required manual intervention to bring the site back to acceptable performance levels.

jpetto:
- Also remove top contributors list from mobile view.

### DIFF
--- a/kitsune/community/utils.py
+++ b/kitsune/community/utils.py
@@ -51,7 +51,7 @@ def top_contributors_questions(start=None, end=None, locale=None, product=None,
     counts = _get_creator_counts(users, count, page)
 
     if use_cache:
-        cache.set(cache_key, counts, 60*15)  # 15 minutes
+        cache.set(cache_key, counts, 60*180)  # 3 hours
     return counts
 
 
@@ -99,7 +99,7 @@ def top_contributors_l10n(start=None, end=None, locale=None, product=None,
     counts = _get_creator_counts(users, count, page)
 
     if use_cache:
-        cache.set(cache_key, counts, 60*15)  # 15 minutes
+        cache.set(cache_key, counts, 60*180)  # 3 hours
     return counts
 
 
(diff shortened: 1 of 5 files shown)


pmac:
This is looking great so far. I'd also like to increase the cache timeout for the `top_contributors_questions` function to at least 3 hours.

pmac:
Also for the `top_contributors_l10n` function in that same file.

PR #3376, opened 2018-10-17
[fix #3371] Fix kbadge file/image uploads.

### LINKED ISSUE #3371: Fix badges media storage/file upload
As a carry-over from django-badger, file uploads in the `kbadge` app currently do not work/support S3 storage.

Should update `kbadge` to use the standard S3 file storage.

See #3045.

jpetto:


jpetto:
Roland tested this on dev (as have I) and all looks good.

kitsune/kbadge/admin.py
```diff
@@ -25,7 +19,7 @@ def show_unicode(obj):
 def show_image(obj):
     if not obj.image:
         return 'None'
-    img_url = "%s%s" % (UPLOADS_URL, obj.image)
+    img_url = "%s%s" % (settings.MEDIA_URL, obj.image)
```
pmac: you should just be able to use the `url` property of the image field.

```suggestion
    img_url = obj.image.url
```

kitsune/kbadge/models.py
```diff
@@ -157,36 +85,16 @@ def __call__(self, instance, filename):
                                 ext=self.ext)
 
 
-class JSONField(models.TextField):
```
pmac: I guess this was just already superfluous?

jpetto: Yep, this was something that could have been removed in the original `django-badger` port - I just missed it.

kitsune/kbadge/models.py
```diff
@@ -124,20 +65,7 @@ def slugify(txt):
...
-        for m in methods
-    )
-    return perms
-
-
-# We need to do this in django 1.7 because migrations serialize the fields.
-# See https://code.djangoproject.com/ticket/22999
+# This is only retained to ensure migrations don't fail
 @deconstructible
 class UploadTo(object):
```
jpetto: @pmac Curious to see what your thoughts are here. This class is no longer used, but migrations fail if it's not present. Do we only need to keep this around for _this_ migration, or, because it's referenced in the initial migration, do we need to do something more here (e.g. manually edit the initial migration)?

pmac: I think it's fine to leave it. We have a task to clean up migrations and we can delete it along with that work. Maybe make a note in the migration itself that we should delete that class?

kitsune/kbadge/migrations/0002_auto_20181023_1319.py
```diff
@@ -0,0 +1,17 @@
...
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('kbadge', '0001_initial'),
+    ]
+
+    operations = [
+        migrations.RunSQL(
+            "UPDATE badger_badge SET image = CONCAT('uploads/', image)"
```
jpetto: @pmac Do you think we need a `WHERE` clause here? Something like `WHERE image NOT LIKE 'uploads/%'`?

pmac: That does seem like a good idea.

PR #3378, opened 2018-10-19
Add proper variable for static url in email base template

pmac:
Fix broken image in SUMO emails.

Fix bug 1262192


pmac:
I saw that the header image in emails I was getting was broken and found the old bug filed about it. I think this should fix it. The image in the email needs updating for branding purposes however.

jpetto:
Yeah, if "Support" is part of the image, we should re-render it using Zilla Slab.

pmac:
may need your image manipulation expertise on that. I can give it a shot if you're too busy.

jpetto:
On it. 😀

jpetto:
Here ya go!

mozilla-support

jpetto:
And an alternate, just in case:

mozilla-support-alt

pmac:
@jpetto based on https://mozilla.design/mlogo I thought it was supposed to be the latter, but I don't really care. Perhaps we should ask Patrick or Madalina?

pmac:
And I still have to fix my failing test.

pmac:
Also thanks for those images!!! 🥇

jpetto:
Ah, yes, I believe you are right - let's use the second logo posted, the one with the black box around the wordmark.

pmac:
Tests passed! (db gone away problem again)

PR #3379, opened 2018-10-19
Use urlwait for circle tests

pmac:
Hopefully fix the "database has gone away" problems.


pmac:
Tests passed this time at least. Don't know if this will be more reliable on Circle or not, but I think it can't hurt? Thoughts @glogiotatidis?

Issue #3380, opened 2018-10-21
Setup Issue: Bower version error


omkar-dsd:
On executing the setup command:
```
sudo docker-compose -f docker-compose.yml -f docker/composefiles/dev.yml run web ./node_modules/.bin/bower install --allow-root
```
Gives the following error.

```
/app/node_modules/bower/node_modules/tar-fs/node_modules/chownr/chownr.js:2
const fs = require('fs')
^^^^^
bower error         Use of const in strict mode.
```

I tried updating Node version to the latest (10.12.0) and latest LTS (8.12.0). Error still persists.

### Complete Error:
```
Starting kitsune_mariadb_1 ... done
Starting kitsune_elasticsearch_1 ... done
Starting kitsune_redis_1         ... done
? May bower anonymously report usage statistics to improve the tool over time? Yes

/app/node_modules/bower/node_modules/tar-fs/node_modules/chownr/chownr.js:2
const fs = require('fs')
^^^^^
bower error         Use of const in strict mode.

Stack trace:
SyntaxError: Use of const in strict mode.
    at Module._compile (module.js:439:25)
    at Object.Module._extensions..js (module.js:474:10)
    at Module.load (module.js:356:32)
    at Function.Module._load (module.js:312:12)
    at Module.require (module.js:364:17)
    at require (module.js:380:17)
    at Object.<anonymous> (/app/node_modules/bower/node_modules/tar-fs/index.js:1:76)
    at Module._compile (module.js:456:26)
    at Object.Module._extensions..js (module.js:474:10)
    at Module.load (module.js:356:32)

Console trace:
Error
    at StandardRenderer.error (/app/node_modules/bower/lib/renderers/StandardRenderer.js:82:37)
    at Logger.<anonymous> (/app/node_modules/bower/bin/bower:110:22)
    at Logger.emit (events.js:95:17)
    at Logger.emit (/app/node_modules/bower/node_modules/bower-logger/lib/Logger.js:29:39)
    at /app/node_modules/bower/lib/commands/index.js:45:20
    at _rejected (/app/node_modules/bower/node_modules/q/q.js:864:24)
    at /app/node_modules/bower/node_modules/q/q.js:890:30
    at Promise.when (/app/node_modules/bower/node_modules/q/q.js:1142:31)
    at Promise.promise.promiseDispatch (/app/node_modules/bower/node_modules/q/q.js:808:41)
    at /app/node_modules/bower/node_modules/q/q.js:570:49
System info:
[...]

PR #3386, opened 2018-10-23
Upgrade node version on dev image to match that on prod

pmac:
The staticfiles image node version was updated to 6.x some time ago but apparently the dev version was overlooked. This is the real source of the errors in #3380


pmac:
Tests passed. Merging to fix master.

PR #3391, opened 2018-10-28
Fixes #3335: Removed line which was creating bug

### LINKED ISSUE #3335: JS error on login page
Go to https://support.mozilla.com/en-US/users/auth and open the console
image
(screenshot b/c it wouldn't let me copy the full message)

navgurukul-shivani18:
Fixes #3335: Removed line which was creating bug

### DIFF
--- a/kitsune/sumo/jinja2/includes/common_macros.html
+++ b/kitsune/sumo/jinja2/includes/common_macros.html
@@ -6,8 +6,13 @@
         <input type="hidden" name="{{ k }}" value="{{ v }}" />
       {% endfor %}
     {% endif %}
-    <input type="search" name="q" required="required" aria-required="true" placeholder="{{ _('Find help...')|truncate(12) }}" class="searchbox" id="{{id+'shivani'}}" />
-    <button type="submit" title="{{ _('Search') }}" class="submit-button">{{ _('Search') }}</button>
+    {% if id!=False}
+      <input type="search" name="q" required="required" aria-required="true" placeholder="{{ _('Find help...')|truncate(12) }}" class="searchbox" id="{{'search-'+id}}" />
+      <button type="submit" title="{{ _('Search') }}" class="submit-button">{{ _('Search') }}</button>
+    {% else %}
+      <input type="search" name="q" required="required" aria-required="true" placeholder="{{ _('Find help...')|truncate(12) }}" class="searchbox" id="{{'search-'+(Math.random()*Math.random())}}" />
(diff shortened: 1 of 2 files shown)


kitsune/users/jinja2/users/includes/auth_contributor.html
```diff
@@ -3,7 +3,6 @@
 <div class="grid_12">
   <article id="register-form" class="main">
     <h1>{{ _('Register to be a Contributor') }}</h1>
-    <span>{{ _('You don\'t need to register to <a href="{url}">ask a question</a>.')|fe(url=url('questions.aaq_step1')) }}</span>
```
vaishnavipamulapati: How does this affect the presence of two elements with ID search-q, as reported in the related issue?

pmac:
Hi. Thanks for the pull-request. As @vaishnavipamulapati pointed out however, I think the patch for this PR isn't the right one as it doesn't seem to address the related issue. Let me know when you've updated your patch and I'll be happy to review again.

kitsune/sumo/jinja2/includes/common_macros.html
```diff
@@ -6,7 +6,7 @@
         <input type="hidden" name="{{ k }}" value="{{ v }}" />
       {% endfor %}
     {% endif %}
-    <input type="search" name="q" required="required" aria-required="true" placeholder="{{ _('Find help...')|truncate(12) }}" class="searchbox" id="search-q" />
+    <input type="search" name="q" required="required" aria-required="true" placeholder="{{ _('Find help...')|truncate(12) }}" class="searchbox" id="{{id+'shivani'}}" />
```
pmac: A couple of problems with this. 

1. `id` is not necessarily a string. you can see in the macro signature that it's `False` by default.
2. adding your name to the ID of an element on the site is not appropriate.

navgurukul-shivani18: @pmac Hi
Thanks for the review.
>1. id is not necessarily a string. you can see in the macro signature that it's False by default.

I thought id would be different for each, hence, wanted to incorporate into the code. I'll try to add it in a different way :)
>2. adding your name to the ID of an element on the site is not appropriate.

I wanted to include a random string instead of "search" as this error may arise again due to same id. I'll remove my name, please suggest me a set of characters that would be good?

akatsoulas:
Closing due to inactivity. Please feel free to open a new PR if the issue is still relevant.

PR #3392, opened 2018-10-28
Fixed Grammar

g1tman:


pmac:
Thanks for your contribution, but it's more correct to say "there is documentation online" rather than "there is a documentation online". I will agree that the wording isn't the best though. What if we changed it to: "You can find the full documentation online."

akatsoulas:
Thanks @g1tman for this PR. I am closing it since it's inactive for some time. Feel free to open a new one if you interested in fixing this.

PR #3394, opened 2018-10-29
Migrate from raven to sentry-sdk

### LINKED ISSUE #3393: Upgrade to new Sentry SDK
Sentry dropped raven package in favor of a new unified SDK.

 - New Package: https://github.com/getsentry/sentry-python
 - Docs: https://docs.sentry.io/quickstart/?platform=python

While I couldn't find a deprecation schedule, please plan to upgrade to the new SDK in the near future to avoid incompatibility issues with the server.

vaishnavipamulapati:
Fix #3393

- Removed raven / sentry config from `kitsune.log_settings`, since the logging info is now set up by default in the new package.
- `sentry-sdk` has `certifi` as a peer dependency, added that to specify hash. Also it required an upgrade of `urllib3`
- No separate WSGI configuration needed.

Confirmed working using `runserver` and `gunicorn`

### DIFF
--- a/kitsune/log_settings.py
+++ b/kitsune/log_settings.py
@@ -7,10 +7,6 @@ from django.utils.log import dictConfig
 config = {
     'version': 1,
     'disable_existing_loggers': True,
-    'root': {
-        'level': logging.ERROR,
-        'handlers': ['sentry'],
-    },
     'formatters': {
         'default': {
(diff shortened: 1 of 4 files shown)


vaishnavipamulapati:
Although I now see some warnings in the console which were previously suppressed. Not sure about that part.

Issue #3330, opened 2018-09-20
Standardize case in/when using get_browser function


jpetto:
The [`get_browser` function](https://github.com/mozilla/kitsune/blob/69253e763ada10422d3c88e681bcb4af00e1eb4f/kitsune/sumo/utils.py#L334) matches ignoring case, but returns whatever case is present in the user agent string. To make this function more stable, we should return only lower case strings.

LucianaMarques:
Hi! I just submitted a [pull request](https://github.com/mozilla/kitsune/pull/3395l)!

akatsoulas:
Closing as stale.

PR #3395, opened 2018-10-30
return only lower cases strings in get_browser

LucianaMarques:
I'm submitting this pull request to solve issue #3330 .

I understand that this function returns a string accordingly to match. If the string returned is not None, adding .lower() should suffice ([reference](https://docs.python.org/2/library/string.html)).

I'm not sure, though, if this works for all python versions.


omkar-dsd:
You need to change the Tests accordingly.

LucianaMarques:
> You need to change the Tests accordingly.

Thank you, I'm a little bit busy but I'll certainly keep working on this.

PR #3396, opened 2018-10-30
Allow filtering the questions API by date range

### LINKED ISSUE #3332: Allow filtering the questions API by date range
The `created` and `updated` datetime fields should be filterable by ranges or other date-appropriate Django filter types (e.g. `month`, `hour`, `year`, etc.).

vaishnavipamulapati:
Fixes #3332

cc @pmac

### DIFF
--- a/kitsune/questions/api.py
+++ b/kitsune/questions/api.py
@@ -144,6 +144,8 @@ class QuestionFilter(django_filters.FilterSet):
     metadata = django_filters.MethodFilter(action='filter_metadata')
     solved_by = django_filters.MethodFilter(action='filter_solved_by')
     taken_by = django_filters.CharFilter(name='taken_by__username')
+    created_between = django_filters.MethodFilter(action='filter_created_between')
+    updated_between = django_filters.MethodFilter(action='filter_updated_between')
 
     class Meta(object):
         model = Question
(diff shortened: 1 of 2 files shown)


vaishnavipamulapati:
- Is it okay to assume that the range input will be a python list of `datetime` objects, or should I parse the dates from strings?
- Should this put to use somewhere in specific on the frontend?

kitsune/questions/api.py
```diff
@@ -224,6 +228,14 @@ def filter_metadata(self, queryset, value):
 
         return queryset
 
+    def filter_created_between(self, queryset, range):
+        questions_created_between = Question.objects.filter(created__range=range)
```
safwanrahman: do not use built in variable `range`

vaishnavipamulapati: thanks!

akatsoulas:
Hi all. The API now supports `lt` and `gt` filters in the queryset for both `created` and `updated` fields.
The query can have the form of `YYYY-MM-DD HH:MM:SS` and the time part can be optional. A fully formulated query can look like `created__lt=YYYY-MM-DD HH:MM:SS`. By combining both `lt` and `gt` filters you can get a date range filtering. Because of this I am closing this PR. Thank you @vaishnavipamulapati for submitting this patch!

akatsoulas:
The filtering works for both Question and Answers.

PR #3400, opened 2018-11-02
Add text search to inproduct redirects in the admin

pmac:
Fix bug 1504289


pmac:
This is a very small change. Going to go ahead and get this in.

PR #3401, opened 2018-11-02
update newrelic plugin to latest

metadave:
untested


pmac:
Tests passed. Circle failed because we still haven't figured out how to fix the "DB server went away" thing :( Looks okay to me. Thanks Dave!

PR #3403, opened 2018-11-05
Announcements, CustomerCare, Dashboard: Functional timezone in sync w…

omkar-dsd:
In Accordance to [Django Documentation](https://docs.djangoproject.com/en/1.8/ref/utils/#module-django.utils.timezone), syncing the timezone with the one in Django's settings.

### DIFF
--- a/authority/models.py
+++ b/authority/models.py
@@ -1,8 +1,8 @@
-from datetime import datetime
 from django.db import models
 from django.contrib.contenttypes.models import ContentType
 from django.contrib.contenttypes import generic
 from django.contrib.auth.models import Group
+from django.utils import timezone
 from django.utils.translation import ugettext_lazy as _
 
 from authority.compat import user_model_label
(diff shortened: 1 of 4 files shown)


omkar-dsd:
#3412

pmac:
Hi! My apologies for the delay in response. This does look good, but I'm going to have to research the implications a bit more. I'm afraid that changing what that field means at this point would make using it with the existing data confusing without also converting all of that existing data.

omkar-dsd:
@pmac DateTimeField is timezone aware, might not cause any issue. Still, to be on a safer side, I will also go through the related consequences, if any.

omkar-dsd:
@pmac any update on this ?

akatsoulas:
Thank  you @omkar-dsd for this PR. Indeed this is a desired change but as Pmac mentioned in his last comment this will require to update our current database entries which is not currently possible b/c the project is in the middle of a big upgrade. Thank you again for your contribution.

Issue #3294, opened 2018-08-10
Move from elasticutils to elasticsearch-dsl


pmac:
This will enable us to fix #3260 by supporting ESv6 and #3177 since elasticutils doesn't support recent Djangos.

jgmize:
required for https://github.com/mozilla/kitsune/issues/2990 also

akatsoulas:
This work is being tracked [here](https://github.com/mozilla/sumo-project/issues?q=is%3Aopen+is%3Aissue+milestone%3AElasticSearch7)

PR #3415, opened 2018-11-07
Index the answer objects

safwanrahman:
@pmac r?

### DIFF
--- a/kitsune/search/documents.py
+++ b/kitsune/search/documents.py
@@ -1,7 +1,8 @@
+from django.db.models import F
 from django.utils import timezone
 from django_elasticsearch_dsl import DocType, Index, fields
 
-from kitsune.questions.models import Question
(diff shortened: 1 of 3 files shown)


kitsune/search/documents.py
```diff
@@ -1,7 +1,8 @@
+from django.db.models import F
```
pmac: I don't see where this is used.

safwanrahman: @pmac fixed!

safwanrahman:
@pmac I have fixed the issue. can you review again?

pmac:
Yup. I'll get it in today hopefully.

pmac:
Thanks safwan! I'm going to hopefully land #3416 shortly and then rebase `es_upgrade` on current master so we can keep up and use the new docker stuff.

pmac:
re #3294

PR #3416, opened 2018-11-08
Move to a single Dockerfile and docker-compose file

pmac:
Use make to create easy to use and remember commands for building and running things.


pmac:
I'm also interested in your thoughts if you have some time @glogiotatidis.

pmac:
Example [successful Jenkins build](https://ci.us-west.moz.works/blue/organizations/jenkins/Kitsune/detail/improve-docker-dev/5/pipeline).

Dockerfile
```diff
@@ -0,0 +1,142 @@
+################################
+# Python dependencies builder
+#
+FROM python:2-stretch AS base
```
jgmize: as a future experiment, we may want to try one of the images linked from https://github.com/mozmeao/infra/issues/960#issuecomment-433537794, specifically `revolutionsystems/python:2.7-wee`

pmac: We can experiment with that. I think perhaps a more simple site would be a better place to start though.

Makefile
```diff
@@ -0,0 +1,123 @@
...
+	@echo "run           - docker-compose up the entire system for dev"
+	@echo ""
+	@echo "pull          - pull the latest production images from Docker Hub"
+	@echo "init          - initialize the database and install Node and Bower packages"
+	@echo "djshell       - start a Django Python shell (ipython)"
+	@echo "shell         - start a bash shell"
+	@echo "clean         - remove all build, test, coverage and Python artifacts"
+	@echo "rebuild       - force a rebuild of all of the docker images"
+	@echo "lint          - check style with flake8, jshint, and stylelint"
+	@echo "test          - run tests against local files"
```
jpetto: I get a bunch of errors about missing static files running this, but that may not be related to this PR. (Local testing has never been solid for me.)

jpetto: Also failed with `MySQL server has gone away`, but I'm pretty sure that's not related to this PR.

README.md
```diff
@@ -11,69 +11,54 @@ See [what's deployed](https://whatsdeployed.io/s-J18)
...
-   docker-compose -f docker-compose.yml -f docker/composefiles/build.yml build base
-   docker-compose -f docker-compose.yml -f docker/composefiles/build.yml build dev
+   make init
    ```
 
-1. Copy .env-dist to .env
+3. Run Kitsune.
    ```
-   cp .env-dist .env
+   make run
```
metadave: There's a lot of output from the commands listed on this page, some success output might be nice to guide the user:

For example:

The server is ready if you see this:

```
web_1              | System check identified no issues (3 silenced).
web_1              | November 09, 2018 - 06:44:42
web_1              | Django version 1.8.15, using settings 'kitsune.settings'
web_1              | Starting development server at http://0.0.0.0:8000/
web_1              | Quit the server with CONTROL-C.
```

README.md
```diff
@@ -11,69 +11,54 @@ See [what's deployed](https://whatsdeployed.io/s-J18)
...
-   ```
-   docker-compose -f docker-compose.yml -f docker/composefiles/dev.yml exec web ./manage.py update_product_details
-   ```
+* Update product details
+  ```
+  docker-compose exec web ./manage.py update_product_details
+  ```
 
-1. (Optional) Get search working
+* Get search working
```
metadave: I couldn't get search to work following these docs. I enter a search term into the dialog, hit enter/magnifying glass and then... nothing.

README.md
```diff
@@ -11,69 +11,54 @@ See [what's deployed](https://whatsdeployed.io/s-J18)
...
 
    First, make sure you have run the "Create some data" step above.
 
-   1. Enter the web container: `docker exec -it kitsune_web_1 /bin/bash`
+   1. Enter the web container: `docker-compose exec web bash`
```
metadave: is there any reason that this isn't done automatically during `make run`?

safwanrahman:
This was the most needed thing! First when we dockerize the application, the initial plan of mine was something like this!

Dockerfile
```diff
@@ -0,0 +1,142 @@
...
+            libmariadbclient-dev mariadb-client && \
+    rm -rf /var/lib/apt/lists/*
+
+COPY ./requirements/*.txt /app/requirements/
+
+RUN pip install --no-cache-dir --require-hashes -r requirements/default.txt && \
+    pip install --no-cache-dir --require-hashes -r requirements/dev.txt && \
+    pip install --no-cache-dir --require-hashes -r requirements/test.txt
+
+ARG GIT_SHA=head
```
safwanrahman: I think the `ARG` should be at begining of the file

pmac: my understanding is that it just needs to come before you use it. This is mostly copied straight from the split up dockerfiles.

README.md
pmac: I think the idea is that many (most) people won't need this and so it's better to avoid the time it takes to do these steps at first. I could be convinced otherwise though. Also these require some user input, so I don't mind them being separate, but could be better incorporated I'm sure.

README.md
pmac: That's entirely possible. These have been like this for a while so it's possible they're no longer accurate. I can check it out.

README.md
pmac: That's a good point.

Makefile
pmac: hmmm... maybe we do need a collectstatic step? And I have been trying to figure out the "MySQL server has gone away" thing, but it's still not fixed :( My suspicion right now is some strange interaction with django-nose. It is not a quick and easy thing to move to pytest for a test though.

Makefile
```diff
@@ -0,0 +1,123 @@
...
+	@echo "You need to specify a subcommand."
+	@exit 1
+
+help:
+	@echo "build         - build docker images for dev"
+	@echo "run           - docker-compose up the entire system for dev"
+	@echo ""
+	@echo "pull          - pull the latest production images from Docker Hub"
+	@echo "init          - initialize the database and install Node and Bower packages"
+	@echo "djshell       - start a Django Python shell (ipython)"
```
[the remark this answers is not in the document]
pmac: Sure. I personally use the django shell and the bash shell mostly, but that's another good one.

Makefile
```diff
@@ -0,0 +1,127 @@
...
+	touch .docker-build-full
+
+pull: .env
+	-GIT_COMMIT= ${DC} pull base base-dev staticfiles locales full-no-locales full mariadb elasticsearch redis
+	touch .docker-build-pull
+
+rebuild: clean build
+
+run: .docker-build-pull
+	${DC} up web
```
glogiotatidis: Personally I prefer to use

`docker-compose run web --service-ports bash`

and then in the shell run the server.

I believe this way you can debug better and also restart easier when things fail.

I'm not a huge user of DC though and i'd love comments on this.

pmac: huh... that's a good idea. I've never done it that way. I don't see why that wouldn't work, but it is a bit more to do. Perhaps it would be best to document this and add `--service-ports` to the `make shell` command if people want to do this?

glogiotatidis: You can only add `--service-ports` to one active shell command, otherwise DC will complain that it can't bind to the ports. So not sure if changing `make shell` is the right thing to do. 

That said feel free to go whichever way you prefer, I just dropped an idea :)

pmac: @glogiotatidis thanks for the suggestion! I added a note in the README about this technique and added `--service-ports` to the `make shell` command. I think this way both methods are well supported.

pmac: hmm... I assume it's complaining in the same way that anything complains when ports are in use? I guess it's possible you'd want a shell while you already have other things running. I could instead make another command for this specificially; `make runshell` maybe. That way you could decide whether or not you wanted that. I could also be making the makefile do too much and we could just document how to use docker-compose directly in this case. Not sure.

pmac: I updated things to add a `make runshell` command that binds ports, and removed port binding from `make shell`. I think that's enough make automation. Anything else can be done with one of those two commands. Thoughts?

glogiotatidis: Yap, `runshell` is probably the best thing. thanks for taking the time

pmac:
Things are building correctly and people seem pleased. I'm going to merge this. Thanks all!

PR #3417, opened 2018-11-12
Adding search class for simplesearch

safwanrahman:
This PR adds a `SimpleSearch` class for building the `Search` object of simple search.
So it should work something like this
```python
from kitsune import search as constants
simple_search = SimpleSearch(query="firefox",
                             doc_type=constants.WHERE_WIKI, locale="en-US")
search = simple_search.get_search()
search.execute()
```
I have tried to follow factory pattern in `SimpleSearch`, but dont know if totally follow the factory pattern.

This class can be used in `simple_search` view to build the Search object then paginate and execute. Need to be verified that all the **boosting are correct and fields are covered** according to previous [`generate_simple_search`](https://github.com/mozilla/kitsune/blob/master/kitsune/search/search_utils.py#L29-L121) function.

Highlight and others are not added yet.

This will made following elasticsearch query. Looking at the json, I think it make sense. Maybe we can compare the old elasticsearch query with this query.
```json
{
  "query": {
    "bool": {
      "should": [
        {
          "bool": {
            "minimum_should_match": 1,
            "filter": [
              {
                "terms": {
                  "document_category": [
                    10,
                    20
                  ]
                }
              },
              {
                "term": {
                  "document_locale": "en-US"
                }
              },
              {
                "term": {
                  "document_is_archived": false
                }
              },
              {
                "term": {
                  "index_name": "sumo_wiki_document"
                }
              }
            ],
            "should": [
              {
                "multi_match": {
                  "fields": [
                    "document_summary^2.0",
                    "document_keywords^8.0",
                    "document_title^6.0",
                    "document_content^1.0"
                  ],
                  "query": "Sample"
                }
              },
              {
                "multi_match": {
                  "fields": [
                    "document_summary^2.0",
                    "document_keywords^8.0",
                    "document_title^10.0",
                    "document_content^8.0"
                  ],
                  "type": "phrase",
                  "query": "Sample"
                }
              }
            ]
          }
        },
        {
          "bool": {
            "minimum_should_match": 1,
            "filter": [
              {
                "term": {
                  "question_is_archived": false
                }
              },
              {
                "term": {
                  "question_has_helpful": true
                }
              },
              {
                "term": {
                  "index_name": "sumo_question"
                }
              },
              {
                "term": {
                  "question_locale": "en-US"
                }
              }
            ],
            "should": [
              {
                "multi_match": {
                  "fields": [
                    "question_title^4.0",
                    "question_content^3.0",
                    "question_answer_content^3.0"
                  ],
                  "query": "Sample"
                }
              },
              {
                "multi_match": {
                  "fields": [
                    "question_title^4.0",
                    "question_content^3.0",
                    "question_answer_content^3.0"
                  ],
                  "type": "phrase",
                  "query": "Sample"
                }
              }
            ]
          }
        }
      ]
    }
  }
}
```

@pmac r?

### DIFF
--- a/kitsune/search/documents.py
+++ b/kitsune/search/documents.py
@@ -34,6 +34,10 @@ class WikiDocumentType(KitsuneDocTypeMixin, DocType):
     document_recent_helpful_votes = fields.IntegerField()
     document_display_order = fields.IntegerField(attr='original.display_order')
 
+    # Store the index name to the index until elastic/elasticsearch#23306 get fixed
+    # https://github.com/elastic/elasticsearch/issues/23306
+    index_name = fields.KeywordField()
+
     # Custom configuration for kitsune to have separate analyzer for supported locales
     supported_locales = settings.SUMO_LANGUAGES
 
@@ -69,6 +73,9 @@ class WikiDocumentType(KitsuneDocTypeMixin, DocType):
         else:
             return 0
 
+    def prepare_index_name(self, instance):
+        return config.WIKI_DOCUMENT_INDEX_NAME
+
 
 class QuestionDocumentType(KitsuneDocTypeMixin, DocType):
     url = fields.KeywordField(attr='get_absolute_url')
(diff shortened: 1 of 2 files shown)


safwanrahman:
Thanks @pmac

pmac:
re #3294

PR #3419, opened 2018-11-16
Integrating with simplesearch view and fix the tests

safwanrahman:
This will integrate `SimpleSearch` (implemented in #3417) with the `simple_search` view. All the tests of `simple_search` view seems to be passing! 💯 
Maybe we have progress one step towards upgrading to Elasticsearch 6.x! 😄 
@pmac what do you think?

### DIFF
--- a/kitsune/search/tasks.py
+++ b/kitsune/search/tasks.py
@@ -9,8 +9,8 @@ from django_elasticsearch_dsl.registries import registry
 from django_statsd.clients import statsd
 from multidb.pinning import pin_this_thread, unpin_this_thread
 
-from kitsune.search.es_utils import index_chunk, UnindexMeBro, write_index, get_analysis, \
-    es_analyzer_for_locale
+from kitsune.search.es_utils import (index_chunk, UnindexMeBro, write_index,
+                                     get_analysis, es_analyzer_for_locale)
 from kitsune.search.utils import from_class_path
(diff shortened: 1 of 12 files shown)


pmac:
re #3294

kitsune/search/es_utils.py
```diff
@@ -32,10 +32,15 @@ class UntypedS(object):
 def F(*args, **kwargs):
     pass
 
+
+def get_es():
```
pmac: We should at least have this print a warning or something. It still needs to go away entirely.

kitsune/search/tasks.py
```diff
@@ -178,7 +178,7 @@ def unindex_task(cls, id_list, **kw):
 @task()
 @timeit
 def update_synonyms_task():
-    es = get_es()
+    es = get_es()  # noqa
```
pmac: Is this being ignored by flake8 because it's not imported? I don't think we should paper over it like this as we might forget to remove it later.

safwanrahman: Yeah. I think I can add a comment here with `TODO:` to remove it.
I dont know any other easy way to make the flake8 happy for this.

kitsune/search/es_utils.py
safwanrahman: Maybe can add a comment with `TODO:`

PR #3421, opened 2018-11-17
Remove readiness probes due to cascading failure

jgmize:


safwanrahman:
@jgmize can we investigate why the probes are failing?

jgmize:
The readiness probes, as [configured in kitsune](https://github.com/mozilla/kitsune/blob/master/kitsune/settings.py#L565), depend on the DB and the cache. When we experience problems with the DB or the cache, the failure of the readiness probes exacerbates the problems with the site, as we observed earlier today. In order to restore service to the site I manually deleted the probes with `kubectl edit`; this PR is making the change permanent.

PR #3426, opened 2018-11-30
Update survey with new IDs for Q4

pmac:
Fix bug 1510201

### DIFF
--- a/kitsune/kpi/surveygizmo_utils.py
+++ b/kitsune/kpi/surveygizmo_utils.py
@@ -32,9 +32,17 @@ SURVEYS = {
         'exit_survey_campaign_id': 2208951,
         'active': False,  # allows cron job to skip this survey
     },
+    # bug 1416244
     'general-q3-2018': {  # General survey for en-US users enabled in Q3 of 2018
         'email_collection_survey_id': 4494159,
         'exit_survey_id': 4456859,
+        'exit_survey_campaign_id': 7259518,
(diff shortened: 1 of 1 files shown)


kitsune/kpi/surveygizmo_utils.py
```diff
@@ -32,10 +32,18 @@
...
     },
+    # bug 1416244
     'general-q3-2018': {  # General survey for en-US users enabled in Q3 of 2018
         'email_collection_survey_id': 4494159,
         'exit_survey_id': 4456859,
         'exit_survey_campaign_id': 7259518,
+        'active': False,
+    },
+    # bug 1510201
+    'general-q4-2018': {  # General survey for en-US users enabled in Q4 of 2018
```
pmac: I'm guessing that adding a new ID here is the way to go. Could just update the Q3 one, but it needs to change being date-based I'm thinking. I can't find where this is referenced on the front-end however. Maybe it just works since this is the only active one?

jpetto: Agree with adding a new survey entry as you did.

For the front-end, iirc, all we need to do is make sure the [`surveygizmo` waffle flag](https://github.com/mozilla/kitsune/blob/master/kitsune/sumo/jinja2/base.html#L3) is on, which will load the Survey Gizmo JS, which controls the rest of the script/survey loading.

PR #3427, opened 2018-12-01
Update survey campaign ID

pmac:
Late update to the previous commit.

Fix bug 1510201


pmac:
Followup to #3426 with new info from @geroter.

PR #3428, opened 2018-12-04
[minor fix] Sudanese -> Sundanese

vesper-mozilla:
Apologies for this minute fix, but I feel it's important to get the locale name right, even if it requires two attempts...


pmac:
@vesper-mozilla same issue here as #3429 with git.

PR #3429, opened 2018-12-05
Adding fallback for jv and su (to Indonesian / id) + fixing locale addition

vesper-mozilla:
As discussed with Fauzan (l10n lead for Indonesia) during Mozlando - adding fallback (using Safwan's feature) for jv and su into id.

Also, adding both jv and su to the correct array this time 🤦‍♂️ (pebkac)


kitsune/wiki/config.py
```diff
@@ -36,6 +36,8 @@
     'gl': ['es'],  # Bug 800880
     'wo': ['fr'],  # Bug 800880
     'fy-NL': ['nl'],  # Bug 800880
+    'su': ['id'], # Bug 800880
```
safwanrahman: You should add a bug about the changes, then the bug number in comments.

pmac:
@vesper-mozilla there are a lot of commits included here which were already merged in other PRs, and there are merge conflicts. Rebase your work on current master and push again please. And let me know if you can't get to that soon and I'll be happy to give it a shot myself. I just want to be sure the end result is what you wanted.

PR #3437, opened 2019-01-03
Fix build issues

pmac:
* Ensure docker environment is setup for l10n linting (use make)
* Ensure docker build args are passed to docker (GIT_SHA, and LOCALE_ENV)

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -51,6 +51,7 @@ services:
   lint-l10n:
     image: mozmeao/kitsune:base-${GIT_COMMIT_SHORT:-latest}
     command: dennis-cmd lint --errorsonly locale
+    env_file: .env-test
     volumes:
       - ./locale:/app/locale:delegated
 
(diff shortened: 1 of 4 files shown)


pmac:
@jpetto I discovered that the new l10n automation wasn't working when a `.env` file wasn't present for docker-compose ¯\_(ツ)_/¯. So I've fixed that by using make which ensures that file is created. Also it turns out that the `GIT_SHA` and `LOCALE_ENV` variables weren't making it to the docker images and were therefore not working. This fixes that as well.

docker-compose.yml
```diff
@@ -59,36 +60,52 @@ services:
...
     image: mozmeao/kitsune:locales-${GIT_COMMIT_SHORT:-latest}
 
   full-no-locales:
     build:
       context: .
       target: full-no-locales
+      args:
+        - GIT_SHA
+        - LOCALE_ENV
+    image: mozmeao/kitsune:locales-${GIT_COMMIT_SHORT:-latest}
```
jpetto: Two images? Is this copy pasta?

pmac: Indeed it is! Good catch!

pmac:
Fixed. Thanks for the quick r? 🎆

PR #3445, opened 2019-01-16
[fix #3436] Auto detect Firefox Lite on mobile AAQ start page.

### LINKED ISSUE #3436: Auto-detect Firefox Lite on mobile AAQ start page
See https://github.com/mozilla-tw/Rocket/issues/3004 for all of the details.

Basically the mobile version of https://support.mozilla.org/questions/new does some auto-detection to highlight a best-guess of the product in use at the time. We should auto-select Firefox Lite when viewed from Lite.

jpetto:
I _think_ this will do the trick, but should test on dev first.

UA string for Lite Nightly (Night Litely?) provided in #rocket: 

`Mozilla/5.0 (Linux; Android 8.0.0; moto e5 Build/OPP27.91-176; rv) AppleWebKit/537.36 (KHTML, like Gecko) Version/4.0 Rocket/1.0.9(9399).nightly Chrome/71.0.3578.99 Mobile Safari/537.36`

### DIFF
--- a/kitsune/questions/views.py
+++ b/kitsune/questions/views.py
@@ -519,14 +519,19 @@ def aaq(request, product_key=None, category_key=None, showform=False,
     if product_key is None:
         product_key = request.GET.get('product')
         if request.MOBILE and product_key is None:
+            ua = request.META.get('HTTP_USER_AGENT', '').lower()
+
             # Firefox OS is weird. The best way we can detect it is to
             # look for a mobile Firefox that is not Android.
-            ua = request.META.get('HTTP_USER_AGENT', '').lower()
             if 'firefox' in ua and 'android' not in ua:
                 product_key = 'firefox-os'
+            # 'Rocket' is currently in the UA, but may be replaced with 'Lite' later on
+            elif any(s in ua for s in ['rocket', 'lite']):
+                product_key = 'firefox-lite'
             elif 'fxios' in ua:
                 product_key = 'ios'
             else:
+                # android
                 product_key = 'mobile'
 
     product_config = config.products.get(product_key)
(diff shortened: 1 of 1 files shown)


jpetto:
It _might_ be a good idea to also check for `lite` in the UA string - waiting on word from the team in #rocket.

kitsune/questions/views.py
```diff
@@ -519,14 +519,18 @@ def aaq(request, product_key=None, category_key=None, showform=False,
...
         product_key = request.GET.get('product')
         if request.MOBILE and product_key is None:
+            ua = request.META.get('HTTP_USER_AGENT', '').lower()
+
             # Firefox OS is weird. The best way we can detect it is to
             # look for a mobile Firefox that is not Android.
-            ua = request.META.get('HTTP_USER_AGENT', '').lower()
             if 'firefox' in ua and 'android' not in ua:
                 product_key = 'firefox-os'
+            elif 'rocket' in ua:
```
pmac: will the above `if` statement not catch rocket first, or does it report "android"?

pmac: also are they going to keep "rocket" in the UA for the foreseeable future, or will they move to "lite"?

jpetto: Just posted the UA string example I received in the opening comment. It shows `firefox` is *not* present, but `android` is, so I believe this logic will work as expected.

jpetto:
Just pushed a forced commit checking for `rocket` or `lite`. I think I'll take this to dev for folks to test while we wait for feedback on the future of the UA string.

jpetto:
Got confirmation that `Rocket` is planned to stay in the UA string long term (and made a commit to that effect).

I think we're good for final review/merge here.

PR #3446, opened 2019-01-21
[Bug 1521551] Updating support.mozilla.org locales

vesper-mozilla:
kitsune/lib/languages.json:
- mostly updates to include previously unlisted locales

kitsune/settings.py:
- adding new locales
- removing locales to be temporarily disabled
- replacing arrays of locales with test locale (xx) to disable the features that are no longer necessary as active (I wasn't sure if I can just delete or comment out the whole array

kitsune/wiki/config.py:
- adding new fallback options to include previously unlisted locales


safwanrahman:
@vesper-mozilla Thanks for your contribution. But I think there need to be some data migration in order to remove the locales. I think for this task, you need to setup the kitsune locally.

vesper-mozilla:
Hi Safwan,

Thank you for looking at this PR.

In the interest of time and to avoid the lack of technical expertise on my side to delay this change even further, is this something you could perform if you had the time?

I am OK with not being credited with the pull request - it's more important to me to see the change go live rather than to see it go live with my name attached to it.

If you can't make it, I'll ask @pmac or another web dev within Mozilla for a bit of their time.

safwanrahman:
@vesper-mozilla Can you explain in the bug what changes you are looking for? Like which locale you would like to add, which locale you would like to remove and which locale you would like to change the locale code?

PR #3450, opened 2019-01-24
Integrate ux experiments into SuMo.

akatsoulas:


akatsoulas:
@pmac r?

I tried to break this PR into logical commits. The first commit is the `dist` folder taken from directly from the vendor (with a few minor modifications) [0]. cc @johngian 

[0]  https://github.com/mozilla/sumo-experiments/

jpetto:
Hi @akatsoulas - Is there a bug or issue associated with this PR I could read for context?

Can you give me a quick summary on the purpose of the the [sumo-experiments](https://github.com/mozilla/sumo-experiments) repo?

Thanks!

akatsoulas:
Hi @jpetto . I think @hmitsch covered almost everything here. Basically it's an experiment that will run for a short period to gather some data regarding the UI design in the page and then it will be thrown away. The only thing that I would add to what @hmitsch said, is that instead of TrafficCop this experiment is using django-waffle.

The [sumo-experiments](https://github.com/mozilla/sumo-experiments) repo is just a place for the vendor to provide the UI for the two KB articles.

The key elements of this PR are:
* Adds the static files inside Kitsune and uses django-pipeline for the static handling.
* A waffle flag to load balance between the original and the static version.
* Uses Kitsune's GA implementation for tracking.
* Uses Google forms for feedback.

Please let me know if you have any more questions regarding the nature of this experiment.

akatsoulas:
@pmac I would say a month but I will let @madasan verify this. Regarding the percentage, it's 50/50 distribution for the current and the experimental version for both articles.

@jpetto thanks for taking the time going through this PR. I will ping you after addressing these comments.

madasan:
The experiment should run for two weeks with the possibility of extending it for another 2, depending on the data collected

akatsoulas:
@jpetto pr updated

A couple of comments regarding your suggestions:

* Regarding the search modal, although it's a valid concern that can be mistaken as a live search window, I believe it's there as part of the experiment to provide some metrics on how search is used and how often people interact with it. Maybe @madasan can provide some more context about this, if you are interested.

* I fully agree with the extension of the `base.html` template. We thought a lot about this and the reason we decided not to go with it, is because we wanted a fast iteration and the minimum footprint when it comes to integration with the Kitsune. The idea is to gather some data and then just delete all the assets. For this reason we decided not to properly connect it which unfortunately has the risk of forgetting things and duplicating code

I believe most of the comments should be addressed. Please let me know if I missed anything. Thanks :)

jpetto:
I've pushed the latest changes to the dev site. Things are looking good as far as I'm concerned.

@akatsoulas Unless you'd like to wait, I'll plan to merge and push this out on Monday.

Thanks!

akatsoulas:
@jpetto this is great, thanks!

metadave:
very nice @akatsoulas!

thanks for the review @jpetto @pmac!

PR #3453, opened 2019-01-31
Fix db error after tests

pmac:
Setting REUSE_DB=1 will cause the custom test runner to avoid attemptint to tear down the database when it's done which appears to be when the "database has gone away" errors are happening.


pmac:
It's a feature of [Django Nose](https://django-nose.readthedocs.io/en/latest/usage.html#enabling-database-reuse). It should have no real effect since we're using a dockerized DB for tests.

PR #3455, opened 2019-02-01
Switch from Jenkins to GitLab CI

pmac:
GitLab should run PR tests when opened from a branch on the mozilla repo.


docker-compose.yml
```diff
@@ -105,7 +110,7 @@ services:
       args:
         - GIT_SHA
         - LOCALE_ENV
-    image: mozmeao/kitsune:full-${GIT_COMMIT_SHORT:-latest}
+    image: mozmeao/kitsune:full-${CI_COMMIT_SHORT_SHA:-latest}
```
safwanrahman: I think we should not create the variable name specific to CI. better to have the previous general name like `GIT_COMMIT_SHORT`

safwanrahman:
We can pass the `GIT_COMMIT_SHORT=CI_COMMIT_SHORT_SHA` from the ` .gitlab-ci.yml` file. What do you think @pmac

docker-compose.yml
pmac: except these are specifically for CI. Shouldn't need anything other than "latest" locally.

pmac:
Could do. It would be a single place to change it if we wanted another variable later.

docker-compose.yml
safwanrahman: This can be used locally also! sometime you may need to run specific version of the image, in that case you can set this environment variable and run that version locally!

safwanrahman:
I wonder from where we are setting the `GIT_SHA` environment variable?

safwanrahman:
>Could do. It would be a single place to change it if we wanted another variable later.

I think that would be better. So we can use `GIT_COMMIT_SHORT` in the place of `GIT_SHA`

.gitlab-ci.yml
```diff
@@ -0,0 +1,82 @@
+stages:
+  - build
+  - test
+  - push
+
+variables:
+  COMPOSE_PROJECT_NAME: "kitsune-${CI_JOB_ID}-${CI_COMMIT_SHORT_SHA}"
```
safwanrahman: This is interesting! Can you explain a bit more the reason to change the project name?

pmac: This allows multiple runs of the same docker-compose on the same machine since the test jobs will run in parallel. The job ID being in there keeps them separate and allows them to be easily identified and killed after.

pmac: See https://docs.docker.com/compose/reference/envvars/#compose_project_name

pmac:
This is pretty much ready to go. We've got a runner box going for our jobs, and this is all hooked up and doing everything that Jenkins is doing. It should be easy to get deployment automation added to this after it merges as well. Curious for your opinions @glogiotatidis and @jpetto.

pmac:
@metadave yes. I'm planning on doing that when we add deployment automation after GitLab is building and pushing the images.

glogiotatidis:
Took a quick look at this and i like what i see. This already builds faster compared to Jenkins (10 mins vs 16 mins) :metal: 

nice work pmac!

.gitlab-ci.yml
```diff
@@ -0,0 +1,81 @@
...
+  tags:
+    - meao
+  stage: build
+  script:
+    - make build-ci
+
+docker push:
+  tags:
+    - meao
+  stage: test
```
safwanrahman: Is it should be test? should we push always without test get passed?

pmac: that's what we're doing now. the idea is that if the tests don't pass we can pull down the image and more easily figure out what's wrong. We only push the `:latest` tags after tests pass.

akatsoulas:
I am closing this PR as there is currently an ongoing effort to switch to a different setup.

PR #3456, opened 2019-02-03
Add mention of docker-compose

sau226:
Docker Compose is relied upon by all the steps listed in the getting started guide as errors will be thrown otherwise.


pmac:
Great point. Thanks for the fix! 🥇

PR #3460, opened 2019-02-14
Integrate updated UI

akatsoulas:


akatsoulas:
@jpetto / @pmac it would be great if you have some time to review this.
This PR re-integrates an updated version of the UI from the vendor and adds/updates some metrics.
It also moves the experiments in a different from the original url.

kitsune/sumo/static/sumo/kb-ux-experiment/src.682efaae.js
```diff
@@ -0,0 +1,375 @@
...
+    } else {
+        document.addEventListener('DOMContentLoaded', fn);
+    }
+}
+
+ready(fn);
+},{}],16:[function(require,module,exports) {
+'use strict';
+
+require('./styles/reset.scss');
```
pmac: What is happening with all of this `require()` and direct use of `scss` files? Is this rendering SCSS in the browser? We should not be doing that if so. We have a static asset pipeline that will pre-render this stuff.

akatsoulas: I believe currently it's ignored completely. This file is the output of the node build. Even the path here is wrong. I will include the native js version in the pipeline. Thanks for catching this!

akatsoulas:
@pmac PR updated.

kitsune/wiki/jinja2/kb-ux-experiment/enable-and-disable-cookies-website-preferences.html
```diff
@@ -35,9 +35,11 @@ <h2 class="logo">
                   <a href="/en-US/questions/new" class="ask" data-event-category="navigation" data-event-action="main navigation" data-event-label="Get Community Support">Get Community Support</a>
                 </li>
 
-                <li>
-                  <a href="/en-US/users/auth" class="sign-in register" data-event-category="navigation" data-event-action="main navigation" data-event-label="Sign In">Sign In</a>
-                </li>
+                {% if not user.is_authenticated() %}
```
pmac: It's probably better if we don't include authenticated users in this experiment at all, since a lot about the page can change for an authenticated user, and especially one that has moderation and editing privileges.

kitsune/wiki/views.py
```diff
@@ -259,6 +253,25 @@ def document(request, document_slug, template=None, document=None):
     return response
 
 
+# Temporary view to hosts the OI UX experiments for 2 articles in the KB
+def ux_experiment_view(request, document_slug):
+    # If there is a direct request to this view
+    # and the flag is not active, redirect to the document view.
+    if (not flag_is_active(request, 'ux_experiment_1') or
```
pmac: If this flag is set to less than 100% then this will run that percentage a 2nd time if they came from the main url. So you could in theory have someone enter the experiment from the main view, then get sent back here, and possibly have that happen multiple times. You need some way of making the user stick once a decision has been made. My recommendation would be to use a [waffle switch](https://waffle.readthedocs.io/en/stable/types/switch.html) here instead of a flag to indicate whether the entire thing is enabled or not. So here you'd check the switch, and in the main view you'd check both the switch and the flag.

kitsune/wiki/jinja2/kb-ux-experiment/enable-and-disable-cookies-website-preferences.html
akatsoulas: Sounds good. I will remove this check and I will allow only non-authenticated users from waffle.

akatsoulas:
@pmac PR updated.

kitsune/wiki/views.py
```diff
@@ -135,16 +135,11 @@ def document(request, document_slug, template=None, document=None):
 
     # Render the static pages if the slug matches the experiment
     if (flag_is_active(request, 'ux_experiment_1') and
+        switch_is_active('ux_experiment_1') and
```
pmac: NIT: I would reverse these. The switch is for the whole thing, and the flag will do the segmentation. No sense checking the flag if the switch is off.

PR #3463, opened 2019-02-18
Allow access to the experiments through redirects.

akatsoulas:
@pmac r?

What I am trying to do here is to allow access to the `experiment_view` only through the original (`document`) view. The reason for this is that I would prefer to block the new url unless a user already has the right flag to access it. Without this you could directly visit the url and bypass waffle. Thoughts?


pmac:
I'm trying to remember what we've done, but I think we did some things to limit sessions for anonymous users. A cookie would work just fine for this as well though. I'd suggest setting a cookie as part of the redirect response, and reading it in the experiment view. In theory this is all overkill because there will be no links to the experiment URL anywhere so there should not be any significant direct traffic to that page, but I guess it is possible.

pmac:
Actually, after searching around, I think all we did was move sessions to the cache and out of the DB. So you're probably fine doing it like this.

PR #3470, opened 2019-02-21
Allow attaching to the web container.

akatsoulas:


pmac:
Based on [the docs](https://docs.docker.com/engine/reference/run/#foreground) it looks like you don't need the stdin line, but the tty one is probably good. I think it's only really needed when you're running as `up`, but when debugging I tend to `docker-compose run` or `docker-compose exec app bash` when I need a shell on the currently running container. How are you using docker when you'd like to have a TTY?

pmac:
There is also a `make runshell` command in the Makefile that gives you a bash shell in the container and has the ports mapped, so you can just `python manage.py runserver` from there and debug as you wish.

akatsoulas:
I am mostly using this to attach to the container when there is debugger breakpoint.

pmac:
Right. So using the `make runshell` to get a real shell in the container is probably best there. But if these changes help you then I'm fine with merging them.

akatsoulas:
Thanks for reviewing!

PR #3477, opened 2019-02-28
[bug 1527337] Treat FreeBSD as "linux" in kb

philipp-sumo:


kitsune/sumo/static/sumo/js/browserdetect.js
```diff
@@ -157,7 +157,7 @@ var BrowserDetect = window.BrowserDetect = {
     },
     {
       string: navigator.platform,
-      subStrings: ['Linux'],
+      subStrings: [/Linux|X11/],
```
akatsoulas: Thanks @philipp-sumo for this PR!
If I am not mistaken, the `X11` string is present only in the `userAgent` whereas `platform` has `FreeBSD` and possible the architecture (eg amd64). Can you confirm this?

philipp-sumo: oh sorry you're right of course - i didn't pay enough attention to the useragent/platform string.
i don't have a freebsd system available for testing myself but the documentation i could find indeed suggests that `FreeBSD` would be needed instead.

akatsoulas: Thanks for the update. I will test this in a VM. Looking good though!

rleitan:
I confirm the fix on staging!

PR #3480, opened 2019-03-05
Django 1.11 updates

SmileyChris:
Work towards 1.11 upgrade


safwanrahman:
@SmileyChris Thanks for the PR. Can you please rebase your branch against master rathar than merging master on your branch? So it will be cleaner in diff as well as git log.

SmileyChris:
Ok, I've rebased the entire branch against master, this means that you'll want to force it into your `django-1.11` branch rather than merge it, since there were two commits in there before my fork which have now also been rebased.

safwanrahman:
@SmileyChris Oh! Seems like `django-1.11` branch should be rebased upon master. @pmac can you rebase the `django-1.11` branch against master? so its easy to review the PR.

pmac:
`django-1.11` branch rebased on current master. Still going to conflict. Apologies for the churn @SmileyChris.

kitsune/sumo/tests/test_locale_middleware.py
```diff
@@ -128,22 +128,20 @@ def test_anonymous_change_to_login(self):
...
 
-        response = self.client.get('/nn-no/', follow=True)
-        self.assertRedirects(response, '/no/', status_code=302)
+            response = self.client.get('/nn-no/', follow=True)
+            self.assertRedirects(response, '/no/', status_code=302)
 
-        response = self.client.get('/xy/', follow=True)
-        self.assertRedirects(response, '/en-US/', status_code=302)
+            response = self.client.get('/xy/', follow=True)
+            self.assertRedirects(response, '/en-US/', status_code=302)
```
jpetto: Nit - indentation.

SmileyChris: Derp, fixed.

jpetto:
This is all looking great to me! @SmileyChris Should we do a push to the dev server, or are there more commits imminent?

kitsune/sumo/tests/test_locale_middleware.py
```diff
@@ -127,23 +127,20 @@ def test_anonymous_change_to_login(self):
...
+@override_settings(NON_SUPPORTED_LOCALES={'nn-NO': 'no', 'xx': None})
 class NonSupportedTests(TestCase):
-    @mock.patch.object(settings._wrapped, 'NON_SUPPORTED_LOCALES',
-                       {'nn-NO': 'no', 'xx': None})
     def test_get_non_supported(self):
-        eq_('no', get_non_supported('nn-NO'))
-        eq_('no', get_non_supported('nn-no'))
-        eq_(settings.LANGUAGE_CODE, get_non_supported('xx'))
-        eq_(None, get_non_supported('yy'))
+            eq_('no', get_non_supported('nn-NO'))
```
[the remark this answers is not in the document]
pmac: Oh nm. that was already fixed :)

kitsune/gallery/forms.py
```diff
@@ -80,11 +78,6 @@ def __init__(self, *args, **kwargs):
         url = 'http://infohost.nmt.edu/tcc/help/pubs/pil/formats.html'
         self.fields['file'].help_text = msg.format(learn_more=url)
 
-    def clean(self):
```
safwanrahman: I think this should not be removed. We want only certain format file can be added. https://github.com/mozilla/kitsune/blob/master/kitsune/upload/forms.py#L34

kitsune/wiki/models.py
```diff
@@ -332,14 +332,20 @@ def __setattr__(self, name, value):
         """Trap setting slug and title, recording initial value."""
         # Public API: delete the old_title or old_slug attrs after changing
         # title or slug (respectively) to suppress redirect generation.
-        if getattr(self, 'id', None):
```
safwanrahman: Any specific reason to remove it?

SmileyChris: I made this change to avoid a recursive `__setattr__` situation that showed up during testing. The following line testing `self._state.adding` has the same logical effect.

kitsune/gallery/forms.py
SmileyChris: You're right, this is a change in behaviour. My reasoning was to rely on `django.core.validators:validate_image_file_extension`, but you may not want every single image format that pillow can open...

Note to self: This should be done with a custom image field which has `validators=[FileExtensionValidator(allowed_extensions=ALLOWED_IMAGE_EXTENSIONS)` rather than a clean method though.

metadave:
what pieces are remaining for the upgrade?

kitsune/gallery/forms.py
safwanrahman: @SmileyChris Yeah. I understand. I think in that case you need to add this to `default_validators`

SmileyChris:
@metadave I want to fix the file extension code, but that's about it

kitsune/gallery/forms.py
SmileyChris: Fixed in a0b7f0e7ca5c730d116a75d9403bb50f1082ecfd

kitsune/forums/forms.py
```diff
@@ -23,7 +22,7 @@
 
 class ReplyForm(forms.ModelForm):
     """Reply form for forum threads."""
-    content = StrippedCharField(
+    content = forms.CharField(
```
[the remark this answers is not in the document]
SmileyChris: As I said in the comment of the commit (f85daa7c6f6d1aab6c017e568fd49d26682f0de8), it's the default behaviour of the built in `CharField` now: https://docs.djangoproject.com/en/1.11/ref/forms/fields/#django.forms.CharField.strip

kitsune/flagit/tests/__init__.py
```diff
@@ -2,21 +2,15 @@
 
 from django.conf import settings
 from django.template.defaultfilters import slugify
+from django.test import override_settings
 
 from kitsune.sumo.tests import LocalizingClient, TestCase
 
 
+# Override the cache prefix with a test-specific one.
+@override_settings(
```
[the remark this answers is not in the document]
SmileyChris: You're right - this is a change in behaviour. I tried it and it worked so I didn't think anything of it. The correct way to do it if we do require a unique cache for each test would be to use the enter/exit methods of a `self.settings(...)` context manager. Your call - worth the change?

kitsune/forums/forms.py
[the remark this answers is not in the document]
pmac: DOH! Sorry. I missed that comment from earlier. Thanks!

kitsune/flagit/tests/__init__.py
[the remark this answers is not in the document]
pmac: Nah. If this works then great. I just found it interesting to point out in case issues arise in future. Maybe adding a comment to this affect would be nice if you get time and remember, but I'm not going to worry about it since it's working. Thanks!

PR #3489, opened 2019-03-08
Do not cache the UX experiments.

akatsoulas:
The document page is heavily cached and is invalidated only when the article is updated. This has as a side-effect  that the waflle flag check is not reachable thus there is no load balancing for the UX experiments. This check excludes specific slugs from caching. @pmac thoughts on this?

### DIFF
--- a/kitsune/wiki/views.py
+++ b/kitsune/wiki/views.py
@@ -71,6 +71,15 @@ def doc_page_cache(view):
             statsd.incr('wiki.document_view.cache.skip')
             return view(request, document_slug, *args, **kwargs)
 
+        # If the slug is part of the UX experiments, redirect to the experiment view.
+        # The check is here because the KB articles are heavily cached and we need to load
+        # balance based on the waffle flag for every request.
+        if (all([x not in request.META['HTTP_USER_AGENT'] for x in EXCLUDED_BROWSERS]) and
+            document_slug in SUMO_UX_EXPERIMENTS_SLUGS and request.LANGUAGE_CODE == 'en-US' and
+            switch_is_active('ux_experiment_1') and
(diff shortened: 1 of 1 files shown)


pmac:
I would just move the switch logic from the document view to the cache decorator. That way the normal pages are still served from cache if available and the redirects will happen regardless. I'd move it to below the authentication check in the decorator.

pmac:
Also the `doc_page_cache` decorator is only used on the document view, so no need to check that.

akatsoulas:
@pmac PR updated

akatsoulas:
I tested this locally and it works as it should. When the flag is active, the static template is rendered and when it's not the original *cached* version is used.

kitsune/wiki/views.py
```diff
@@ -71,6 +71,15 @@ def _doc_page_cache_view(request, document_slug, *args, **kwargs):
             statsd.incr('wiki.document_view.cache.skip')
             return view(request, document_slug, *args, **kwargs)
 
+        # If the slug is part of the UX experiments, redirect to the experiment view.
+        # The check is here because the KB articles are heavily cached and we need to load
+        # balance based on the waffle flag for every request.
+        if (switch_is_active('ux_experiment_1') and
+            flag_is_active(request, 'ux_experiment_1') and
+            all([x not in request.META['HTTP_USER_AGENT'] for x in EXCLUDED_BROWSERS]) and
+                document_slug in SUMO_UX_EXPERIMENTS_SLUGS and request.LANGUAGE_CODE == 'en-US'):
```
pmac: It would be better to move the slug and language check first so that the DB is not checked for the switch status for every document page load.

akatsoulas:
@pmac pr updated.

PR #3493, opened 2019-03-13
Allow creation of user profiles through the admin interface.

akatsoulas:


kitsune/users/admin.py
```diff
@@ -33,12 +33,8 @@ class ProfileAdmin(admin.ModelAdmin):
     form = ProfileAdminForm
     list_display = ['full_user']
     list_select_related = True
-    readonly_fields = ['user']
```
pmac: Having this be read-only seems good. No?

akatsoulas: It does but by enabling the add permissions, you cannot select a user for the new profile as the field is readonly.

pmac: Oh! Right. Good point. 👍

PR #3494, opened 2019-03-13
Check that `HTTP_USER_AGENT` exists in request.

akatsoulas:


kitsune/wiki/views.py
```diff
@@ -74,7 +74,8 @@ def _doc_page_cache_view(request, document_slug, *args, **kwargs):
         # If the slug is part of the UX experiments, redirect to the experiment view.
         # The check is here because the KB articles are heavily cached and we need to load
         # balance based on the waffle flag for every request.
-        if (all([x not in request.META['HTTP_USER_AGENT'] for x in EXCLUDED_BROWSERS]) and
+        user_agent = request.META.get('HTTP_USER_AGENT')
+        if (user_agent and user_agent not in EXCLUDED_BROWSERS and
```
pmac: this is not the same check as before. the full User-Agent string will never be in `EXCLUDED_BROWSERS`.

pmac:
```suggestion
        if (user_agent and all([x not in user_agent for x in EXCLUDED_BROWSERS]) and
```

pmac: Or you could just replace this in the original line: `request.META.get('HTTP_USER_AGENT', '')`

akatsoulas:
@pmac pr updated

PR #3501, opened 2019-03-19
[fix bug 1535689] Add FxA banner test to all pages for specified locales & fx users

jpetto:
Banner should only load for Firefox users on desktop.

### DIFF
--- a/kitsune/sumo/jinja2/base.html
+++ b/kitsune/sumo/jinja2/base.html
@@ -32,6 +32,10 @@
     {% stylesheet style %}
   {% endfor %}
 
+  {% if waffle.switch('fxa-banner-test') and request.LANGUAGE_CODE in settings.FXA_BANNER_LANGUAGES %}
+    {% stylesheet 'fxa-banner-test' %}
+  {% endif %}
+
   <!--[if lte IE 7]>
     {% stylesheet 'ie' %}
   <![endif]-->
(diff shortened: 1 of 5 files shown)


kitsune/sumo/jinja2/base.html
```diff
@@ -32,6 +32,10 @@
     {% stylesheet style %}
   {% endfor %}
 
+  {% if waffle.switch('fxa-banner-test') and request.LANGUAGE_CODE in settings.FXA_BANNER_LANGUAGES %}
```
pmac: NIT: since the waffle switch check hits the DB it'd be more efficient to check it after the language check so that it's checked less often.

kitsune/sumo/jinja2/base.html
```diff
@@ -211,6 +215,26 @@ <h2 class="logo">
   {{ breadcrumbs(crumbs, id='main-breadcrumbs') }}
 {% endblock %}
 
+{% if waffle.switch('fxa-banner-test') and request.LANGUAGE_CODE in settings.FXA_BANNER_LANGUAGES %}
```
pmac: Same NIT here as above

kitsune/sumo/jinja2/base.html
```diff
@@ -211,6 +215,26 @@ <h2 class="logo">
...
+<div id="fxa-banner">
+  <p class="copy" id="fxa-banner-send">
+  {% trans url='https://send.firefox.com/?utm_source=support.mozilla.org&utm_medium=referral&utm_campaign=sumo_try_send' %}
+    Safely share files online with encrypted links that “self-destruct.” <a href="{{ url }}">Try Firefox Send.</a>
```
pmac: Should we instead have two strings here? The pitch sentence and the link text?

kitsune/sumo/static/sumo/js/fxa-banner.js
```diff
@@ -0,0 +1,17 @@
+(function() {
+  'use strict';
+
+  var isFirefox = (/\s(Firefox)/.test(navigator.userAgent));
+
+  if (isFirefox) {
+    var banners = ['send', 'monitor', 'sync'];
+    var choice = Math.floor(Math.random() * 3);
```
pmac: NIT: better to use `banners.length` here instead of `3` for future reasons, but not a big deal for this iteration obviously.

jpetto:
Thanks for the review @pmac! Updated.

PR #3502, opened 2019-03-20
Drop OI UX experiments.

akatsoulas:


akatsoulas:
@jpetto / @pmac r when you have some time?

In the following days the experiment will end. This PR removes all the code that was added to facilitate the experiments with the 2 KB articles. This should not be merged yet so that we can keep the master branch clean until we are ready to go live with this patch to prod.

PR #3507, opened 2019-03-21
Run migrations in one deployment per branch to fix #3506

### LINKED ISSUE #3506: Modify bin/run-post-deploy.sh to only run in specific cluster
[run-post-deploy.sh](https://github.com/mozilla/kitsune/blob/master/bin/run-post-deploy.sh) currently checks a `READ_ONLY` environment variable before calling `manage.py migrate`. We could check for a new environment variable that is explicitly set in one cluster only (for example: oregon-a) so that when a new deployment  happens in the other cluster (oregon-b in this example) the migrations are not redundantly triggered, and we eliminate the possibility of errors due to parallel migrations triggered by parallel deployments to separate clusters. Since we would also not set this new environment variable in the deployment(s) that connect to read-only replica DBs, the migrations would continue not to run there.

jgmize:
- see #3506

### DIFF
--- a/k8s/regions/frankfurt/dev.yaml
+++ b/k8s/regions/frankfurt/dev.yaml
@@ -116,6 +116,7 @@ app:
     read_only: False
     redis_default_url: SECRET
     redis_helpfulvotes_url: SECRET
+    run_migrate: "true"
     secret_key: SECRET
     sentry_dsn: SECRET
     session_cookie_secure: True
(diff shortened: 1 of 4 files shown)


k8s/regions/frankfurt/dev.yaml
```diff
@@ -116,6 +116,7 @@ app:
     read_only: False
     redis_default_url: SECRET
     redis_helpfulvotes_url: SECRET
+    run_migrate: "true"
```
akatsoulas: Shouldn't this be false from Frankfurt?

pmac: Not in dev. Dev is only in Frankfurt and is RW.

PR #3520, opened 2019-04-04
Revert "Disable sentry in prod due to issues caused by increased IO"

akatsoulas:
Reverts mozilla/kitsune#3497

### DIFF
--- a/k8s/regions/frankfurt/prod.yaml
+++ b/k8s/regions/frankfurt/prod.yaml
@@ -118,6 +118,7 @@ app:
     redis_default_url: SECRET
     redis_helpfulvotes_url: SECRET
     secret_key: SECRET
+    sentry_dsn: SECRET
     session_cookie_secure: True
     stage: False
     static_url: "/static/"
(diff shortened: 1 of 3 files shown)


akatsoulas:
@jgmize @metadave thoughts on this?

metadave:
have performance issues with Sentry logging been addressed?

jgmize:
How about we play it safe and only enable it in oregon-a initially, leaving it off in oregon-b for (say) a couple of weeks?

metadave:
+1

by any chance are there any flags we can use to throttle communications w/ sentry?

akatsoulas:
@metadave let me check this

akatsoulas:
Following up on the last comment from @metadave, we can rate limit sentry under settings [0]. I tried looking into that but it looks like that I cannot access these settings. Is there an admin level access that can tweak these settings?
https://docs.sentry.io/accounts/quotas/#id1

jgmize:
We could only enable sentry in one of the active production clusters, cutting the potential volume in half, and in the event of running into IO bottlenecks due to sentry volume, it should only affect one cluster and the traffic will fail over to the other automatically at the ELB.

akatsoulas:
Closing this. Sentry is enabled only in one of the clusters in the new infra.

PR #3523, opened 2019-04-11
Update libraries to newer versions.

akatsoulas:
* Fixes warnings about deprecated libs.
* Fixes an error with MySQL-Python
* Satisfies requirement dependencies.


pmac:
Lots of test failures. Looks like something in the newer Jinja isn't happy with how we use it :/

akatsoulas:
This change was introduced in version 2.8
http://jinja.pocoo.org/docs/2.10/changelog/#version-2-8

akatsoulas:
@pmac this PR looks a bit better now. Quick r? when you have some time? Interested in your thoughts about the changes.

kitsune/notifications/api.py
```diff
@@ -31,7 +31,7 @@ def has_object_permission(self, request, view, obj):
 
 class NotificationSerializer(serializers.ModelSerializer):
     is_read = serializers.ReadOnlyField()
-    timestamp = DateTimeUTCField(source='action.timestamp')
+    timestamp = DateTimeUTCField('%Y-%m-%dT%H:%M:%SZ', source='action.timestamp')
```
pmac: Why was this change necessary? Is this a fix from the previous upgrade of DRF?

akatsoulas: I believe this is coming from the replacement of `MySQL-Python` with `mysqlclient`. Probably the former did not return any microseconds.

pmac: Oh right. Interesting. Thanks.

PR #3524, opened 2019-04-11
Fix warnings in project's urls.

akatsoulas:


akatsoulas:
It actually does. Lots of failing tests because of this. I will spend some time looking closer to each url in order to fix this properly. Adding a DNM label for now.

akatsoulas:
On second thought, the only feasible way to fix this, due to time availability, is by doing so gradually and per app. I am closing this and I will open smaller PRs

PR #3530, opened 2019-04-17
Use the mozilla_django_oidc lib to authenticate through FxA.

akatsoulas:


akatsoulas:
@johngian quick r?

kitsune/settings.py
```diff
@@ -571,6 +571,8 @@ def parse_conn_max_age(value):
 # Auth
 AUTHENTICATION_BACKENDS = (
     # ``ModelBackendAllowInactive`` replaces ``django.contrib.auth.backends.ModelBackend``.
+    'kitsune.users.auth.FXAAuthBackend',
+    'mozilla_django_oidc.auth.OIDCAuthenticationBackend',
```
johngian: Please add a comment that this is for /admin. Or 
```
from mozilla_django_oidc.auth import OIDCAuthenticationBackend as AdminBackend.
```

kitsune/settings.py
```diff
@@ -579,16 +581,27 @@ def parse_conn_max_age(value):
     OIDC_ENABLE = False
     ENABLE_ADMIN = False
 else:
-    OIDC_ENABLE = config('OIDC_ENABLE', default=False, cast=bool)
+    OIDC_ENABLE = config('OIDC_ENABLE', default=True, cast=bool)
```
johngian: Why are you changing the defaults here ?

kitsune/sumo/jinja2/base.html
```diff
@@ -182,12 +182,6 @@ <h2 class="logo">
       <ul>
         {{ aux_nav(user, hide_aaq_link) }}
 
-        {% if not hide_locale_switcher %}
```
johngian: How is this relevant to the OIDC backend ?

kitsune/users/views.py
```diff
@@ -772,3 +776,41 @@ def validate_field(request):
...
+    @staticmethod
+    def get_settings(attr, *args):
+        """Override settings for Firefox Accounts.
+
+        The default values for the OIDC lib are used for the SSO login in the admin
+        interface. For Firefox Accounts we need to pass different values, pointing to the
+        correct endpoints and RP specific attributes.
+        """
+
+        FXA_CONF = {
```
johngian: I suggest you deduplicate this to a helper method since you are pretty much using the same settings.

kitsune/users/views.py
```diff
@@ -772,3 +776,41 @@ def validate_field(request):
...
+        return super(FXAAuthenticateView, FXAAuthenticateView).get_settings(attr, *args)
+
+
+class FXALogoutView(OIDCLogoutView):
+
+    @staticmethod
+    def get_settings(attr, *args):
+        """Override the logout url for Firefox Accounts."""
+
+        if attr == 'LOGOUT_REDIRECT_URL':
```
johngian: Same here. I would prefer to have a function that imports all the settings instead just for consistency. To avoid any future errors where we forget to change a config in one of the 3 `get_settings`

kitsune/sumo/jinja2/base.html
akatsoulas: It's not. I will remove this commit from this PR

kitsune/settings.py
akatsoulas: If it's set to readonly then this value if False. Moving forward with FxA it makes sense to have this enabled by default.

PR #3532, opened 2019-04-18
Django 1.11 cronjob changes

SmileyChris:
Django-cronjobs isn't 1.11 compatible, so here's a conversion of all the cron methods into proper management commands.


pmac:
Thanks @SmileyChris! There are [2 failing tests in Circle](https://circleci.com/gh/mozilla/kitsune/1020), but apart from that this is looking good. I'm still going through it, but hopefully we can land it shortly.

SmileyChris:
Forgot to save that file before committing. Fixed now :)

PR #3533, opened 2019-04-18
Fixed bad type in ES indexing

nicolaslara:
reindexing wasn't working as per the documentation. Fixed by passing a callable to the type parameter (https://docs.python.org/2.7/library/argparse.html#type)


akatsoulas:
cc @pmac @jpetto

PR #3535, opened 2019-04-19
Create sumo accounts fxa

akatsoulas:


kitsune/sumo/middleware.py
```diff
@@ -53,7 +53,7 @@ class HttpResponseRateLimited(HttpResponse):
 
 class SUMORefreshIDTokenAdminMiddleware(SessionRefresh):
     def __init__(self, *args, **kwargs):
-        if not settings.OIDC_ENABLE:
+        if not settings.OIDC_ENABLE or settings.DEV:
```
johngian: Maybe it would be less error prone if you add a config entry for the refresh middleware specifically.

kitsune/users/auth.py
```diff
@@ -88,3 +90,38 @@ def get_settings(attr, *args):
...
+
+    def filter_users_by_claims(self, claims):
+        """Match users by FxA uid or email."""
+        fxa_uid = claims.get('uid')
+        user_model = get_user_model()
+        user = user_model.objects.none()
+
+        # something went terribly wrong. Return None
+        if not fxa_uid:
+            return user
```
johngian: Can you log this somehow just to keep track off this edge case ?

kitsune/users/auth.py
```diff
@@ -88,3 +90,38 @@ def get_settings(attr, *args):
...
+
+        user = user_model.objects.filter(profile__fxa_uid__iexact=fxa_uid)
```
johngian: why ``iexact`` and not ``profile__fxa_uid = fxa_uid`` ?

kitsune/users/auth.py
```diff
@@ -88,3 +90,38 @@ def get_settings(attr, *args):
...
+
+        if not user:
+            # We did not match any users so far. Let's call the super method
+            # which will try to match users based on email
+            user = super(FXAAuthBackend, self).filter_users_by_claims(claims)
+            # Probably this is the first time FxA is used. Let's migrate the username
+            if user:
+                user.profile.is_fxa_migrated = True
```
johngian: How do you store here the id and extra information from FxA ?

kitsune/users/tests/test_auth.py
```diff
@@ -0,0 +1,96 @@
...
+        ok_(not self.backend.filter_users_by_claims(claims))
+
+    def test_login_existing_user_by_fxa_uid(self):
+        """Test user filtering by FxA uid."""
+        user = UserFactory.create(profile__fxa_uid='my_unique_fxa_id')
+        claims = {
+            'uid': 'my_unique_fxa_id',
+        }
+
+        request_mock = Mock(spec=HttpRequest)
```
johngian: I would also add a line to assert the count of objects before and after.

kitsune/sumo/middleware.py
akatsoulas: The OIDC_ENABLE is already used and I would prefer not to alter this. The only other use case that we have is to disable the oidc only for the admin interface for dev purposes.

kitsune/users/tests/test_auth.py
akatsoulas: This test checks for an existing user so it doesn't modify the db.

akatsoulas:
@johngian PR updated. I added the `update_user` method in the FXAAuthBackend to update the user profile with data from the profile scope.

kitsune/users/tests/test_auth.py
johngian: Ok

kitsune/users/auth.py
```diff
@@ -88,3 +94,51 @@ def get_settings(attr, *args):
...
+        if not profile.is_fxa_migrated:
+            # If it's not migrated, we can assume that there isn't an FxA id too
+            profile.is_fxa_migrated = True
+            profile.fxa_uid = fxa_uid
+
+        if not profile.avatar:
+            # Best effort to get an avatar from FxA
+            profile.avatar = claims.get('avatar', '')
+
+        profile.save()
```
johngian: Maybe you want to update instead of save, to avoid side effects

akatsoulas: I was going to but since we already have the object, I decided not to query the db to update the entries.

akatsoulas: Also there are no signals here that will cause side-effects.

akatsoulas:
Thanks @johngian :beers:

PR #3546, opened 2019-04-23
[fix #1085] Update New Relic package to 4.18.0.118

jpetto:


akatsoulas:
@jpetto should this be merged?

jpetto:
> @jpetto should this be merged?

I think this is good to be merged and test on dev. If all seems well, we can merge #3558 as well and push to production.

akatsoulas:
@jpetto How about testing master in stage and if everything looks good, we can tag production branch to current master? This way we won't have to update all the branches separately.

PR #3553, opened 2019-04-24
Allow logged in users to link their Firefox Account.

akatsoulas:


kitsune/users/auth.py
```diff
@@ -149,3 +151,64 @@ def update_user(self, user, claims):
...
+
+        Users in Kitsune can start the oauth/oidc flow when they are already authenticated.
+        In this case we have a user who is trying to connect the current SuMo account with FxA.
+        If the user is not authenticated we have a login/signup attempt.
+        """
+
+        # If the request has the /oidc/callback/ path then probably there is a login
+        # attempt in the admin interface. In this case just return None and let
+        # the OIDC backend handle this request.
+        if request.path == u'/oidc/callback/':
```
akatsoulas: @pmac @johngian Thoughts on this? Any ideas on how to do this differently? Basically what we want here is to skip the FxA auth backend if the login is targeting the admin interface.

kitsune/users/auth.py
```diff
@@ -149,3 +151,64 @@ def update_user(self, user, claims):
...
+            return None
+
+        # A user is attempting to connect a Firefox Account
+        # NOTE: this section will be dropped when the migration is complte
+        if request and request.user and request.user.is_authenticated():
```
johngian: Instead of overriding the whole flow here which will diverge everything from the upstream OIDC lib code let's move everything to ``get_or_create_user``. This way we don't touch the OIDC flow and we have a minor ``if`` case that checks if the user is authenticated then migrate to FxA.

kitsune/users/auth.py
```diff
@@ -5,6 +5,8 @@
 from django.contrib.auth.backends import ModelBackend
 from django.contrib.auth.tokens import default_token_generator
 from django.contrib.auth.models import User
+from django.core.exceptions import SuspiciousOperation
```
johngian: You don't need this import.

Issue #3541, opened 2019-04-23
[FxA] Disable registration of new profiles with username/password.


akatsoulas:
New users should be created only through Firefox Accounts.

akatsoulas:
The profile creation in the AAQ needs to be removed completely. In the first iteration only logged in users will be able to ask a question.

akatsoulas:
This ticket should:
* Hide all the instances of the AAQ flow behind authentication. Users need to login first before asking a question
* If there is direct visit to the `/questions/new` url, user should be redirected to login first  
* The step that asks the user to create a new account in the AAQ flow needs to be removed
* Users should no longer have the option to create a new account with a username/password

@mcranston18 if there is anything missing feel free to add to the list

mcranston18:
@akatsoulas This ticket is almost complete. In regards to AAQ links (e.g. screenshot below); should these links be hidden to a non-authenticated user? Or should they be visible and the non-authenticated user will be taken to the log in screen if they click it?

![Screen Shot 2019-04-26 at 8 49 33 AM](https://user-images.githubusercontent.com/1489696/56919969-faacb180-6a8f-11e9-82fc-fc902930cf42.png)

I spoke with Rina but she said to leave this decision for you. I think we should leave the links in the UI, and if non-authenticated user clicks on them, they will be sent to the log in.

PR #3561, opened 2019-04-30
Remove user registration from AAQ flow (#3541)

mcranston18:
- Remove user login and registration in the AAQ flow
- Add `login_required` to the AAQ view

@akatsoulas  I have made comments on particular lines just to draw your attention to a few things.

### DIFF
--- a/kitsune/questions/tests/test_templates.py
+++ b/kitsune/questions/tests/test_templates.py
@@ -30,7 +30,6 @@ from kitsune.sumo.urlresolvers import reverse
 from kitsune.tags.tests import TagFactory
 from kitsune.products.tests import TopicFactory
 from kitsune.upload.models import ImageAttachment
-from kitsune.users.models import RegistrationProfile
 from kitsune.users.tests import UserFactory, add_permission
 from kitsune.wiki.tests import DocumentFactory, ApprovedRevisionFactory
 
@@ -1402,6 +1401,16 @@ class AAQTemplateTestCase(TestCaseBase):
         version = question.metadata['ff_version']
         eq_('18.0.2', version)
 
+    def test_full_workflow_inactive(self):
+        """
+        Test that an inactive user cannot create a new question
+        """
+        u = self.user
+        u.is_active = False
+        u.save()
+        self._post_new_question()
+        eq_(0, Question.objects.count())
+
     def test_localized_creation(self):
         response = self._post_new_question(locale='pt-BR')
         eq_(200, response.status_code)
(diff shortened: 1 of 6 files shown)


kitsune/questions/views.py
```diff
@@ -754,7 +708,8 @@ def aaq(request, product_key=None, category_key=None, showform=False,
             url = reverse('questions.details', kwargs={'question_id': question.id})
             return HttpResponseRedirect(url)
 
-        return HttpResponseRedirect(reverse('questions.aaq_confirm'))
+        # What to do here since aaq_confirm is not needed anymore?
+        # return HttpResponseRedirect(reverse('questions.aaq_confirm'))
```
mcranston18: How should I handle the following:
```
if request.user.is_active:
    .....
    return HttpResponseRedirect(...)

# What to do here since aaq_confirm is not needed anymore?
return HttpResponseRedirect(reverse('questions.aaq_confirm'))
```
Since a user must be logged in (and active?) to access the AAQ flow, can I just remove the check `request.user.is_active` and always return the `HttpResponseRedirect`?

kitsune/questions/tests/test_views.py
```diff
@@ -27,7 +27,9 @@
 from kitsune.wiki.tests import DocumentFactory, RevisionFactory
 
 
-class AAQTests(ElasticTestCase):
+# Note:
+# Tests using the ElasticTestCase are not being run bc of this line: `-a '!search_tests'`
+class AAQSearchTests(ElasticTestCase):
```
mcranston18: Just an fyi -- I noticed that tests using the `ElasticTestCase` class are excluded when we run `make test`. I also noticed that these tests seem to be failing.

kitsune/questions/tests/test_templates.py
```diff
@@ -1411,96 +1410,6 @@ def test_localized_creation(self):
         question = Question.objects.filter(title='A test question')[0]
         eq_(question.locale, 'pt-BR')
 
-    def test_full_workflow_inactive(self):
```
mcranston18: Removed this because an inactive user cannot enter the AAQ flow

kitsune/questions/views.py
```diff
@@ -743,18 +695,15 @@ def aaq(request, product_key=None, category_key=None, showform=False,
         # Submitting the question counts as a vote
         question_vote(request, question.id)
 
-        if request.user.is_active:
```
akatsoulas: Why are you removing this?

kitsune/questions/tests/test_templates.py
akatsoulas: But what if a user with an active session is marked as inactive? In that case we will have an authenticated user that is not active

mcranston18: The `aaq` flow now uses the `login_required` decorator. By default, this checks if the user is authenticated and `is_active`:
```
def login_required(func, login_url=None, redirect=REDIRECT_FIELD_NAME, only_active=True):
    if only_active:
        def redirect_func(user):
            return not (user.is_authenticated() and user.is_active)
    ...
```

With that in mind, it seems that a user that was authenticated but was not active would not be able to access the AAQ flow at all, and therefore, we would not need to perform `is_active` checks in the view itself. I added a test `test_inactive_user` to confirm that a user that becomes inactive, they cannot see the AAQ flow (`test_inactive_user`).

Let me know if I misunderstood something about this flow.

mcranston18:
@akatsoulas I've added `test_full_workflow_inactive` which asserts that an inactive user cannot add a new question. LMK if you have any questions.

Do I need to rebase into one commit or are you okay just using squash and merge?

akatsoulas:
@mcranston18 rebase please if possible :)

PR #3564, opened 2019-05-01
Features/ffx login disable 3510

mcranston18:
If a user has already migrated to FXA, do not allow them to log in via SUMO:

Screen Shot 2019-05-01 at 2 39 31 PM


kitsune/users/admin.py
```diff
@@ -19,7 +19,8 @@ class ProfileAdmin(admin.ModelAdmin):
     fieldsets = (
         (None, {
             'fields': ['user', 'name', 'public_email',
-                       ('avatar', 'delete_avatar'), 'bio'],
+                       ('avatar', 'delete_avatar'), 'bio', 'is_fxa_migrated',
```
akatsoulas: nice!

kitsune/users/admin.py
```diff
@@ -34,6 +35,7 @@ class ProfileAdmin(admin.ModelAdmin):
     list_display = ['full_user']
     list_select_related = True
     search_fields = ['user__username', 'user__email', 'name']
+    readonly_fields = ['is_fxa_migrated', 'fxa_uid']
```
akatsoulas: It might be useful, initially to be able to edit these fields for testing purposes. How about dropping this change and opening an issue instead just to keep track of this?

kitsune/users/jinja2/users/includes/notification_macros.html
```diff
@@ -0,0 +1,51 @@
...
+<div id="fxa-login-error" class="notification alert">
+  <img class="ff-logo" src="{{ static_url }}sumo/img/firefox-logo.svg" alt="{{ _('Firefox logo') }}"/>
+  <div class="notification-content">
+    <h3 class="notification-title">{{ _('Sorry!') }}</h3>
+    <p>{{ _('You already have a Firefox Accounts. Please use your Firefox Account instead of your Mozilla Support
+    account.') }}</p>
+  </div>
+  <div class="notification-actions">
+    <a href="{{ url('users.fxa_authentication_init') }}" class="button">{{ _('Continue with Firefox Accounts') }}</a>
+  </div>
```
akatsoulas: Probably we don't need this link here since it's right above the one that asks the user to login. Thoughts?

kitsune/users/views.py
```diff
@@ -95,6 +104,13 @@ def login(request, template):
     form = handle_login(request, only_active=only_active)
 
     if request.user.is_authenticated():
+        if request.user.profile.is_fxa_migrated:
+            auth.logout(request)
+            statsd.incr('user.logout')
```
akatsoulas: Let's remove the statsd from here.

kitsune/users/views.py
```diff
@@ -95,6 +104,13 @@ def login(request, template):
     form = handle_login(request, only_active=only_active)
 
     if request.user.is_authenticated():
+        if request.user.profile.is_fxa_migrated:
+            auth.logout(request)
```
akatsoulas: you can call the logout view from users/views.py which does the same thing and deleted the cookie.

kitsune/users/admin.py
mcranston18: Sure. I'll remove this for now

kitsune/users/jinja2/users/includes/notification_macros.html
mcranston18: Agreed. I'll remove

mcranston18:
@akatsoulas made changes per your suggestions. Ready for re-review

PR #3566, opened 2019-05-01
Pass PIPELINE_COLLECTOR_ENABLED env var to Pipeline config

mcranston18:
On each page refresh, `collectstatic` is run. This makes development extremely slow on OSX due to the way OSX handles shared file systems. For me, each page refresh takes between 10s-20s:

questions

We can prevent `collectstatic` from being run on every page refresh by passing `PIPELINE_COLLECTOR_ENABLED` to the Pipeline configuration. If that env var is not present, it defaults to `True` which keeps the behaviour consistent with the current dev.

After disabling `PIPELINE_COLLECTOR_ENABLED`, the same page takes 1s instead of 20s:

question_after

### DIFF
--- a/README.md
+++ b/README.md
@@ -22,7 +22,7 @@ and follow the following steps.
    git clone https://github.com/mozilla/kitsune.git
    ```
 
-2. Pull base Kitsune Docker images, create your database, and install node and bower packages.
+2. Pull base Kitsune Docker images, run `collectstatic`, create your database, and install node and bower packages.
    ```
    make init
    ```
@@ -54,6 +54,14 @@ and act much more like developing without Docker as you may be used to. You shou
 instead of `make shell` as the latter does not bind port 8000 which you need to be able to load the site.
 
 Run `make help` to see other helpful commands.
+
+### Compiling LESS files
+
+By default, the `.less` files do not compile in local development because it impacts load times on OSX. If you are working on `.less` files, you can enable compilation with the following environment variable:
+
+  ```
+  echo "PIPELINE_COLLECTOR_ENABLED=True" >> .env
+  ```
 
 ### The Admin
 
--- a/bin/run-bootstrap.sh
+++ b/bin/run-bootstrap.sh
@@ -8,5 +8,7 @@ yarn
 ./node_modules/.bin/bower install --allow-root
 # ensure the DB server is ready
 urlwait
+# run collectstatic
+python manage.py collectstatic
 # run DB migrations
 python manage.py migrate
--- a/kitsune/settings.py
+++ b/kitsune/settings.py
@@ -767,6 +767,11 @@ PIPELINE = {
 
     'BROWSERIFY_BINARY': path('node_modules/.bin/browserify'),
     'BROWSERIFY_ARGUMENTS': '-t babelify -t debowerify',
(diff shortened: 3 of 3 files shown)


akatsoulas:
@pmac @jpetto could you review this? I don't have access to OSX to test it.

pmac:
No concerns. This should already have been the case! Can't believe we've not done this to this point. My only suggestion would be to default this to `False` when `DEBUG` is on. So change it to `default=not DEBUG` and change the note to indicate how to set it manually if this isn't working for you. It should be fine for everyone being off for dev though.

akatsoulas:
@mcranston18 could you open this PR against the master branch?

mcranston18:
@akatsoulas I've updated the base branch to be `master`.

@pmac I've added your suggestion and default the value to `not DEBUG`. One other note, I believe a new dev will have to manually run `collectstatic` on project set up in order to initially compile the css. Is this correct? If so, maybe I should document that as well

pmac:
> One other note, I believe a new dev will have to manually run `collectstatic` on project set up in order to initially compile the css. Is this correct? If so, maybe I should document that as well

We should test that. I don't think that's the case since Django should find and load resources from where they live in apps and directories without having to have them in a single static folder. But perhaps pipeline does need that to do whatever compiling it needs to do?

mcranston18:
@pmac I rebuilt the project on this branch and the styles did appear to be compiled using the new settings:

Screen Shot 2019-05-07 at 11 58 58 AM

mcranston18:
@pmac I finally got around to testing this. It seems that `PIPELINE_COLLECTOR_ENABLED` must be `TRUE` initially for a new developer to set the project up. When `PIPELINE_COLLECTOR_ENABLED` is turned off, the static files aren’t collected which means there won’t be a `static` folder in the root of the project and then you’ll get:

Screen Shot 2019-05-07 at 12 54 56 PM

I just updated the code so `PIPELINE_COLLECTOR_ENABLED` defaults back to `True`. If you'd rather, we could leave it to `not DEBUG` but then we'd need to mention in the read me that a new dev must run `collectstatic` when starting the project for the first time.

pmac:
I think defaulting to `not DEBUG` and adding the note to the README is the way to go here. Another option is to add a `collectstatic` step to the server startup procedure in the scripts. That might make the dev server take a bit more time to start but it would keep each request snappy.

jpetto:
I like the idea of adding a `collectstatic` call to the server startup for local dev.

README.md
```diff
@@ -55,6 +55,14 @@ instead of `make shell` as the latter does not bind port 8000 which you need to
 
 Run `make help` to see other helpful commands.
 
+### Compiling LESS files
+
+By default, the `.less` files do not compile in local development because it impacts load times on OSX. If you are working on `.less` files, you can enable compilation with the following environment variable:
+
+  ```
+  echo "PIPELINE_COLLECTOR_ENABLED=True" >> .env
```
pmac: 👍

mcranston18: @pmac Happy to help ! I'll leave it to you to merge this in once ready (I don't have write access to the repo)

pmac: Sounds good. I'll let @akatsoulas have the final look and let him merge it since he's primary on the project. But it all looks good to me! 🍰

PR #3570, opened 2019-05-02
Disable sumo account management for FxA users.

akatsoulas:


akatsoulas:
@johngian / @mcranston18 r? I didn't hide the links in the UI as this will be done in the tickets related to the profile pages.

johngian:
Since its throwaway code, please add a comment so we can track it and delete later.

PR #3587, opened 2019-05-06
Redirect new to FxA users to the edit profile page.

akatsoulas:


johngian:
something is broken in the dependencies

Issue #3562, opened 2019-05-01
[FxA] Add information banners for Firefox Accounts.


akatsoulas:
Banners from issue #3554 should be integrated into the project.

* First, success, banner should be displayed either on the first login with Firefox Accounts or when logged in users connects Firefox Account to their profile. Banner should be dismissable 

* Second banner should be displayed only to newly created profiles through Firefox Accounts. Banner should be dismissable 

* Third banner should be displayed to users that used username/password to login. The banner should be displayed in the profile view and edit screens. The banner is *not* dismissable.

PR #3588, opened 2019-05-06
Add FXA notification banners #3562

mcranston18:
* Redirect to profile screen on SUMO login
* Show FXA deprecation notification on SUMO login
* Show FXA update notification on FXA account linking
* Show FXA create notification on FXA account creation
* Fix typos: FFX > FXA


kitsune/users/views.py
```diff
@@ -525,9 +527,13 @@ def edit_profile(request, username=None, template=None):
 
     # TODO: detect timezone automatically from client side, see
     # http://rocketscience.itteco.org/2010/03/13/automatic-users-timezone-determination-with-javascript-and-django-timezones/  # noqa
+    msgs = messages.get_messages(request)
```
akatsoulas: Can't we just add a message in the `auth.py` like here: 
https://github.com/mozilla/kitsune/blob/firefox-accounts-integration/kitsune/users/auth.py#L127

mcranston18: @akatsoulas I did that here: https://github.com/mozilla/kitsune/pull/3588/files#diff-ce68f8e5c4103e3bd04a6d97db5eef2dR129

The purpose of the list comp is to filter out non-FXA related messages.  I could just do that in the template if you'd prefer

PR #3594, opened 2019-05-07
Set Locale.locale unique=True to fix #3593

### LINKED ISSUE #3593: Add unique=True to Locale.locale
- This will prevent issues like https://github.com/mozilla/kitsune/issues/3591 from recurring.
- https://docs.djangoproject.com/en/2.2/ref/models/fields/#django.db.models.Field.unique
- https://github.com/mozilla/kitsune/blob/production/kitsune/wiki/models.py#L1188

jgmize:
- [X] push commit fd8f83a setting Locale.locale unique=True
- [x] wait for https://ci.vpn1.moz.works/blue/organizations/jenkins/Kitsune/detail/unique-locales/1/pipeline triggered by above commit to finish
- [x] pull image and generate migrations
  > $ export GIT_COMMIT_SHORT=fd8f83 
  - (`git rev-parse --short HEAD` gives us fd8f83a which is one character longer, unfortunately)
  > $ docker-compose pull web
  > $ docker-compose run web ./manage.py makemigrations


jgmize:
Note that when I ran `./manage.py makemigrations` several other migrations were generated that I have not included in this PR:
- kitsune/dashboards/migrations/0006_auto_20190506_2016.py
- kitsune/gallery/migrations/0005_auto_20190506_2016.py
- kitsune/kbadge/migrations/0003_auto_20190506_2016.py
- kitsune/questions/migrations/0008_auto_20190506_2016.py

I will follow up on those in a separate issue

jgmize:
On further review, only the `kbadge` migration was unrelated to this change. I've updated this PR to include the additional migrations, and I also removed the [now redundant](https://docs.djangoproject.com/en/2.2/ref/models/fields/#django.db.models.Field.unique) `db_index=True`

PR #3605, opened 2019-05-07
Features/user profile 3555

mcranston18:
@akatsoulas This PR adds styles and functionality for the profile and edit profile screens. A few notes:

- I ended up leaving the FXA notifications on the Edit Profile screen as we initially intended
- The diff in `kitsune/users/jinja2/users/confirm_avatar_delete.html` is just due to indentation. That screen needed to be wrapped in `grid_9` class

### DIFF
--- a/kitsune/users/jinja2/users/profile.html
+++ b/kitsune/users/jinja2/users/profile.html
@@ -7,11 +7,12 @@
 {% set canonical_url = unlocalized_url('users.profile', profile.user.username) %}
 {% set active = 'my-profile' %}
 {% set crumbs = [(None, profile.user.username)] %}
+{% set is_owner = profile and request.user == profile.user %}
 
 {% block content %}
   <div class="grid_9">
     <article id="profile" class="cf">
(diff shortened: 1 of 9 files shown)


kitsune/users/jinja2/users/profile.html
```diff
@@ -11,15 +11,14 @@
...
-      <div class="avatar">
-        <img src="{{ profile_avatar(profile.user) }}" alt="">
-      </div>
-      {% if user.has_perm('users.change_profile') %}
-        <div id="admin-actions">
-          <a class="edit" href="{{ url('users.edit_profile', profile.user.username) }}">
-            {{ _('Edit user profile') }}
-          </a>
-        </div>
+      {% if profile and profile.is_fxa_migrated %}
```
mcranston18: Yes. This should have been:
```
{% if is_owner and profile.is_fxa_migrated %}
```
Because only the owner should see the "account" section

akatsoulas: Isn't `profile` always present here?

PR #3608, opened 2019-05-07
set SKIP_MOBILE_DETECTION=True in prod

metadave:
already deployed to frankfurt, oregon-a, oregon-b

I suppose this branch needs to be created against prod...


jpetto:
Waiting on merging just in case. (These changes are already deployed.)

pmac:
Yes. There might be a better solution which hopefully we can discuss tomorrow. Maybe we can make google go back to primarily using the desktop bot, or we can remove the noindex tag from the mobile pages, or we can modify our mobile detection to always see the google bot as desktop. I'm hoping Raphael will have some advice.

PR #3613, opened 2019-05-08
Remove noindex from mobile templates & fix canonical URLs.

jpetto:


kitsune/wiki/jinja2/wiki/document.html
```diff
@@ -4,7 +4,7 @@
 {% from "wiki/includes/document_macros.html" import contributor_list, document_title, document_messages, document_content, topic_list %}
 {% from "wiki/includes/document_macros.html" import vote_form with context %}
 {% set classes = 'document' %}
-{% set canonical_url = document.get_absolute_url() %}
```
pmac: The canonical URL for a page should have the locale in it because it should be a URL that returns a 200, not a 301. So I think you can just concat `document.get_absolute_url()` with the `CANONICAL_URL` setting and go with that.

kitsune/sumo/templatetags/jinja_helpers.py
```diff
@@ -78,7 +78,8 @@ def unlocalized_url(viewname, *args, **kwargs):
     """Helper for Django's ``reverse`` in templates.
 
     Uses django's default reverse."""
-    return django_reverse(viewname, args=args, kwargs=kwargs)
```
pmac: It looks like this function is only used for canonical URLs, and so now it seems poorly named. How about we go ahead and change it to something like `get_canonical_url()` or `canonicalize()`. Looks like it could just take the URL as an argument too, which would mean usage like `{{ canoncialize(document.get_absolute_url() }}`.

jpetto:
@pmac Ready for review when you've got time tomorrow!

kitsune/landings/jinja2/landings/home.html
```diff
@@ -1,6 +1,7 @@
 {% extends "base.html" %}
 {% from 'products/includes/product_macros.html' import product_cards with context %}
 {% set styles = ('home',) %}
+{% set canonical_url = settings.CANONICAL_URL %}
```
pmac: Canonical URL must contain the locale.

kitsune/questions/jinja2/questions/mobile/question_list.html
```diff
@@ -17,6 +17,13 @@
 {% set headline = title %}
 {% set return_url = url('questions.home') %}
 
+{% set canonical_url = canonicalize(viewname='questions.list', product_slug=product_slug)|urlparams(None, request.GET) %}
+{% if questions.number > 1 %}
+  {% set canonical_url = canonical_url|urlparams(page=questions.number) %}
+{% endif %}
+
+{% set meta = (('robots', 'noindex'),) %}
```
pmac: we do want robots to index, right?

kitsune/landings/jinja2/landings/home.html
jpetto: 🤦‍♂

kitsune/questions/jinja2/questions/mobile/question_list.html
jpetto: The desktop version of this template has `{% set meta = (('robots', 'noindex'),) %}`, so I figured that should be kept for mobile.

(I don't know the history of this decision, but if desktop is not indexing and mobile hasn't been indexing, probably best to keep that going.)

rraue:
First question, do you have any allizom link for me to review?

Reacting to @pmac comment in the doc, that we dont have "real" mobile urls, I dont think that this is necessary:
If we set a canonical on all pages to the non-parameter version, that will be always the one showing up in search results and users will get the page they need.

We do not need the alternate then, just a canonical to the non-parameter url and removing the noindex and we are back. Afterwards we can re-enable the mobile view.

We just need to keep in mind then, no content not available o the mobile view will be indexed, no link in the sidebar will be counted for pagerank, the mobile view is basically then for Google the standard and only view.

Is there any way for a mobile user agent to see the desktop version?

pmac:
> If we set a canonical on all pages to the non-parameter version, that will be always the one showing up in search results and users will get the page they need.
> 
> We do not need the alternate then, just a canonical to the non-parameter url and removing the noindex and we are back. Afterwards we can re-enable the mobile view.
> 
> We just need to keep in mind then, no content not available o the mobile view will be indexed, no link in the sidebar will be counted for pagerank, the mobile view is basically then for Google the standard and only view.

That sounds like a bad thing. We can change it such that a user needs to take action to get to the mobile version. Considering what Patrick said about our mobile traffic perhaps that's okay?

> Is there any way for a mobile user agent to see the desktop version?

No. Not after we enable mobile views again. I think we should change that, right?

pmac:
> First question, do you have any allizom link for me to review?

We'll need to deploy it somewhere. Staging could work but it's currently on the master branch and so we'll need to do the cache clearing dance. Perhaps @jgmize could help with that this morning? Our dev instance is currently in use for other feature testing.

rraue:
> That sounds like a bad thing. We can change it such that a user needs to take action to get to the mobile version. Considering what Patrick said about our mobile traffic perhaps that's okay?

But isn't that exactly how it was before? The canonical makes sure, we get no duplicate content and removing noindex will enable google again to index and crawl everything. I would agree with you if we would show differnet content, but its the same. So google sees it all (besides the sidebar) and for users there will be no change to status before.

jpetto:
This PR has been deployed to stage for testing: https://support.allizom.org/

pmac:
> So google sees it all (besides the sidebar) and for users there will be no change to status before.

I thought you told me before that any change to the HTML or assets was a new page. The desktop and mobile indexers will see very different markup.

rraue:
> I thought you told me before that any change to the HTML or assets was a new page. The desktop and mobile indexers will see very different markup.
Not any change, just changes which could be interpreted as different content from a machine. Thats not the case here I would say. And there will be sort of no desktop indexer anymore, just a fetch and render bot to get a desktop screenshot, but I think thats fine. Responsive websites do not show everything in the same way as well. If you hide some of the content on mobile its not considered for indexing anymore, if you try to trick Google with that it wont work, but nothing on our mobile/desktop page tries to trick anyone, right?

Issue #3614, opened 2019-05-09
[FxA] Add new user to contributors though the Volunteer flow


akatsoulas:
New users who are creating an account through the volunteer flow, should be automatically added to the contributors group.

PR #3619, opened 2019-05-09
Add support for a fxa user to become a contributor on sign up #3614

mcranston18:
* Add tests for auth.py
* Add tests for setting is_contributor in session

### DIFF
--- a/kitsune/users/views.py
+++ b/kitsune/users/views.py
@@ -45,14 +45,13 @@ from kitsune.users.forms import (
     EmailChangeForm, SetPasswordForm, PasswordChangeForm, SettingsForm,
     ForgotUsernameForm, RegisterForm, PasswordResetForm)
 from kitsune.users.templatetags.jinja_helpers import profile_url
-from kitsune.users.models import (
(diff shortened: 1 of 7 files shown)


kitsune/users/views.py
```diff
@@ -835,6 +817,11 @@ def get_settings(attr, *args):
             return val
         return super(FXAAuthenticateView, FXAAuthenticateView).get_settings(attr, *args)
 
+    def get(self, request):
```
akatsoulas: nit: You could override here the `get_extra_params` method to set the session key

mcranston18: I might not be understanding, but doesn't `get_extra_params` just update the query params for the request url? Or are you saying just to update session in that method instead, e.g.
```
    def get_extra_params(self, request):
        is_contributor = request.GET.get('is_contributor') == 'True'
        request.session['is_contributor'] = is_contributor
        return self.get_settings('OIDC_AUTH_REQUEST_EXTRA_PARAMS', {})
```

akatsoulas: I was going for the later but let's keep it as it is.

PR #3623, opened 2019-05-09
 Set SKIP_MOBILE_DETECTION=True in all environments 

jgmize:


jgmize:
Filed against wrong branch

PR #3630, opened 2019-05-13
Fix cron command for ES reindex.

akatsoulas:
Fixes issue #3620

### DIFF
--- a/scripts/cron.py
+++ b/scripts/cron.py
@@ -81,7 +81,7 @@ def job_process_exit_surveys():
 def job_reindex():
     # Look back 90 minutes for new items to avoid racing conditions between
     # cron execution and db updates.
-    call_command('esreindex', '--minutes-ago 90')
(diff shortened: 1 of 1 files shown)


pmac:
This seems like a good change that will fix the scheduled reindex task, but I don't think it fixes #3620 which I thought referred to the command being broken on the es_upgrade branch. I could be wrong about that though.

akatsoulas:
I thought that the indexing command was fixed through #3575 . I tested locally this patch and it looked like the indexing run normally.  Is there another issue related to the indexing?

pmac:
It's broken on the `es_upgrade` branch, not `master`, on which this PR is based. So upgrading to a new version of the ElasticSearch server and moving to `elasticsearch_dsl` is what is blocked, not the Django upgrade.

akatsoulas:
Thanks for the explanation! This PR though should fix the cron command that is failing in master.

PR #3639, opened 2019-05-17
Update project urls to remove the warnings.

akatsoulas:
Apps updated
* announcements
* community
* flagit
* groups
* inproduct
* messages
* products
* search
* uploads
* base urls


akatsoulas:
This is the first PR in order to start tackling the warnings. The reason I am targeting the FxA branch is because it's going to be merged in master early next week.

jpetto:
As you can see from all the test failures, this isn't as straight forward of a fix as you'd hope. 😞

pmac:
Yup. We discussed tackling this before the 1.11 upgrade merged but decided to put it off since it will turn out to potentially be a significant task.

PR #3653, opened 2019-05-20
Add canonical urls to master

akatsoulas:


akatsoulas:
@jpetto @jgmize those commits were in the production but they were not included in the upgrade last week. I cherry-picked those from the `production-django-1.8` branch. Probably are not in the same order but the important thing here is to add them back since we have the same indexing problem as before without those. The results are already visible.

PR #3659, opened 2019-05-20
Improve request path checking in the FxA callback

johngian:


kitsune/users/auth.py
```diff
@@ -204,10 +204,9 @@ def update_user(self, user, claims):
     def authenticate(self, request, **kwargs):
         """Authenticate a user based on the OIDC/oauth2 code flow."""
 
-        # If the request has the /oidc/callback/ path then probably there is a login
-        # attempt in the admin interface. In this case just return None and let
-        # the OIDC backend handle this request.
-        if request and request.path == django_reverse('oidc_authentication_callback'):
+        # If the request is not targeting the FxA backend then return None
+        # and let the rest of the Django auth backends handle the request
+        if request and not request.path == django_reverse('users.fxa_authentication_callback'):
```
akatsoulas: I am not so sure about this. The ideal logic here is to go to the next backend if and only if a user is targeting the admin interface. If it's targeting the Django backend it will go there and all other requests should remain here. By adding this line we basically exempt this Backend if it's not related to FxA. Thoughts?

johngian: I would prefer the backends to be more self contained and not reference each other. This way we only route /admin requests to the admin backend and all the fxa callbacks to the FxA backend.

Both ways are valid. It just felt more explicit to write it this way because it took me some time to figure out things while reviewing.

akatsoulas:
@johngian tests are failing

akatsoulas:
Closing in favor of #3833

PR #3663, opened 2019-05-20
Pull changes for deploying to it-sre infra

ziegeer:


docker/bin/build-docker-images.sh
```diff
@@ -18,7 +18,7 @@ do
                  -t ${DOCKER_REPO}:${image}-${GIT_SHA_SHORT} \
                  --cache-from ${DOCKER_REPO}:${image}-latest \
                  --cache-from kitsune:${image}-latest \
-                 -f docker/dockerfiles/${image} \
```
akatsoulas: @ziegeer could you please elaborate a bit about this change?

docs/kitsune-support.md
```diff
@@ -4,8 +4,8 @@
 
 High level:
 
-- [SUMO Infra home](https://github.com/mozmeao/infra/tree/master/apps/sumo)
-- [Deploying SUMO](https://github.com/mozilla/kitsune/tree/master/k8s#deploying-sumo)
+- [SUMO Infra home](https://github.com/mozilla-it/sumo-infra)
+- [Deploying SUMO](https://github.com/mozilla-it/kitsune/tree/master/k8s#deploying-sumo)
```
akatsoulas: Since we are using this repo, I think it would be easier to have everything related to the app referenced here.

k8s/commander.sh
```diff
@@ -2,8 +2,8 @@
 set -exo pipefail
 GREEN='\033[1;32m'
 NC='\033[0m' # No Color
-SLACK_CHANNEL=sumodev
-DOCKER_HUB="https://hub.docker.com/r/mozmeao/kitsune/tags/"
+SLACK_CHANNEL=itsre-sumo
```
akatsoulas: Shall we use the existing channel? It's already well known by users and it usually doesn't have any noise apart from dev related topics. The new channel has also logs from `yar` which will make it harder to locate the builds.

bin/slack-notify.sh
```diff
@@ -5,9 +5,9 @@ set -eo pipefail
 # BRANCH_NAME, BUILD_NUMBER
 
 # defaults and constants
-CHANNEL="sumodev"
```
akatsoulas: Shall we use the existing channel? It's already well known by users and it usually doesn't have any noise apart from dev related topics. The new channel has also logs from `yar` which will make it harder to locate the builds.

docker/bin/build-docker-images.sh
ziegeer: Yes, this script is for building docker images locally, which I don't think anyone actually does anymore, but I had to and it pointed to some non-existant dockerfiles that look to have been previously consolidated into a single Dockerfile.  So I just pointed the script at that and it worked.

bin/slack-notify.sh
ziegeer: I'd be happy to but in the interim it's going to be noisy while we get everything stood up, and I don't want to confuse a non-production event with a production event.  Could we perhaps switch back post-migration to the current channels?

docs/kitsune-support.md
ziegeer: Pointed it to this repo now, thanks!

metadave:
:+1:

PR #3681, opened 2019-05-23
Use FxA avatars for migrated uses.

akatsoulas:


kitsune/users/templatetags/jinja_helpers.py
```diff
@@ -37,8 +37,13 @@ def profile_avatar(user, size=48):
     except (Profile.DoesNotExist, AttributeError):
         avatar = settings.STATIC_URL + settings.DEFAULT_AVATAR
     else:
-        avatar = (profile.avatar.url if profile and profile.avatar else
-                  settings.STATIC_URL + settings.DEFAULT_AVATAR)
+        if profile:
```
glogiotatidis: if you reach this point profile exists, right?

PR #3687, opened 2019-05-28
Fix avatar fxa

akatsoulas:


kitsune/users/templatetags/jinja_helpers.py
```diff
@@ -37,12 +38,14 @@ def profile_avatar(user, size=48):
     except (Profile.DoesNotExist, AttributeError):
         avatar = settings.STATIC_URL + settings.DEFAULT_AVATAR
     else:
-        if profile.is_fxa_migrated:
+        if (profile.fxa_avatar and
+                requests.head(profile.fxa_avatar).status_code == 200):
```
johngian: Why do you need this ?

johngian: This adds potential i/o overhead to all helper calls.

johngian: Also the default timeout is None which means that the request/response cycle relies on the availability of ``requests.head()``.

akatsoulas: The problem here is that while the user is logged in, the image can be removed from the management page of Firefox Accounts which has as a result a broken image everywhere where it's used.

johngian:
After discussing it with @akatsoulas its a better solution to handle this on the client side because this has the potential of adding a lot of i/o overhead

PR #3709, opened 2019-05-31
Return only valid emails.

akatsoulas:


akatsoulas:
Relevant traceback
https://sentry.prod.mozaws.net/operations/sumo-prod/issues/5859473/?query=is:unresolved

PR #3715, opened 2019-06-04
Use get for 'total_pages' key.

akatsoulas:


akatsoulas:
Fixes 
https://sentry.prod.mozaws.net/operations/sumo-prod/issues/5860095/?query=is:unresolved

PR #3716, opened 2019-06-05
Migrate to the new infra

akatsoulas:


newrelic.ini
```diff
@@ -85,7 +85,7 @@ ssl = true
 # configuration file AND be set to 'true' in the server-side
 # configuration in the New Relic user interface. For details, see
 # https://docs.newrelic.com/docs/subscriptions/high-security
-high_security = true
+high_security = false
```
The-smooth-operator: is this value supposed to change?

akatsoulas: Yes it's intentional as the NR account we are migrating to does not support this currently. This will be switched probably back when the account is set accordingly. Good catch!

Issue #3775, opened 2019-06-28
[fxa] front page items do not organise into grid of 3 when minimized


flekkowich:
Current behavior
menu items take full space of minimized page https://support-dev.allizom.org/en-US/

Expected behavior
Grid of menu items. https://support.mozilla.org/en-US/

kkellydesign:
@flekkowich I'm not sure what you mean by 'minimized' here. We only have one breakpoint on this system, and it's where the navigation breaks from mobile to desktop. On the mobile size, I'm stacking these items because displaying them at 33% doesn't work on mobile. 

When we rework the frontend I can build this to be a bit more flexible, displaying items at 100% wide on mobile, 50% on tablet sizes, and 33% on desktop. But since there were no designs provided for those breakpoints, I opted to keep it simple for this interim step.

rleitan:
Just to confirm, on tablet view, the grid has 2 products displayed 

![2019-07-02_16h32_41](https://user-images.githubusercontent.com/41474886/60516733-04000780-9ce7-11e9-9668-5f70110fdd6c.png)

@flekkowich

kkellydesign:
@rleitan asked that I re-confirm with @flekkowich: I added a fix yesterday to go from 1 column on mobile, to two on tablet, to three on desktop. Please let me know if you had something different in mind.

https://support.allizom.org/en-US/

flekkowich:
@kkellydesign confirmed sounds like a good solution.

Issue #3774, opened 2019-06-28
[fxa] Upon minimizing window of get involved page - image overlaps text. 


flekkowich:
reproduce
visit https://support-dev.allizom.org/en-US/get-involved
Minimize window to half screensize (13 inch)

expected behaviour
image should minimize as per https://support.mozilla.org/en-US/get-involved

viorelaioia:
This still looks bad, when resizing the browser window to a certain size. 
For example, for 963 X 662 px, the page looks like this:
<img width="956" alt="Screen Shot 2019-07-02 at 12 31 59" src="https://user-images.githubusercontent.com/5293100/60502115-d5256980-9cc5-11e9-8ad4-a871ffaa54b5.png">

kkellydesign:
fix is in here: https://github.com/mozilla/kitsune/pull/3788

<img width="1023" alt="Screen Shot 2019-07-02 at 8 25 05 AM" src="https://user-images.githubusercontent.com/202590/60512584-ef018500-9ca2-11e9-8398-71b5148a665f.png">

PR #3779, opened 2019-07-01
landing page revisions, ref #3774. grid exception, ref #3775

kkellydesign:
I just made a revision to the grid on the front page so that on tablet sizes, it's two-wide. I also made some design revisions to the get-involved page so that the images don't overlap. 

Screen Shot 2019-07-01 at 9 23 18 AM

Screen Shot 2019-07-01 at 9 31 13 AM


flekkowich:
Great thanks @kkellydesign

PR #3783, opened 2019-07-01
Enforce the SumoOIDC backend for staff members.

akatsoulas:


kitsune/sumo/middleware.py
```diff
@@ -58,8 +58,10 @@ def __init__(self, *args, **kwargs):
 
     def process_request(self, request):
         """Only allow refresh and enforce OIDC auth on admin URLs"""
-        if request.path.startswith('/admin/') and request.path != '/admin/login/':
-            if 'oidc_id_token_expiration' not in request.session:
+        # If the admin is targeted let's check the backend used, if any
+        if request.path.startswith('/admin/'):
```
johngian: I think you need to add the check for ``/admin/login``

akatsoulas: I would say that it's not needed any more since we are checking the backend. If it's None it means that the user is not authenticated which is going to redirect to the `/admin/login`. Only when the user is logged in the `backend` evaluates to True thus there is a chance to log the user out depending on the backend used. That said, happy to add it back to make it more explicit.

PR #3854, opened 2019-08-02
Get sub information for users.

akatsoulas:


kitsune/products/jinja2/products/includes/topic_macros.html
```diff
@@ -34,15 +34,15 @@ <h3 class="support-callout-title"><span class="highlight">{{ _('Ask the Communit
       </div>
     {% endif %}
 
-    {% if show_contact_form %}
+    {% if show_contact_form and settings.OIDC_ENABLE %}
```
johngian: Why do we need `OIDC_ENABLE` here? Isn't it the default behaviour?

kitsune/users/auth.py
```diff
@@ -175,8 +175,12 @@ def update_user(self, user, claims):
         profile = user.profile
         fxa_uid = claims.get('uid')
         email = claims.get('email')
-
         user_attr_changed = False
+        # Check if the user has active subscriptions
+        subscriptions = claims.get('subscriptions')
+        isSubscriber = False
+        if subscriptions and 'isSubscriber' in subscriptions:
```
johngian: Maybe you can convert this to an oneliner like:
```
isSubscriber = bool(subscriptions and 'isSubscriber' in subscriptions)
```

kitsune/users/models.py
```diff
@@ -86,6 +86,7 @@ class Profile(ModelBase, SearchMixin):
     is_fxa_migrated = models.BooleanField(default=False)
     fxa_uid = models.CharField(blank=True, null=True, unique=True, max_length=128)
     fxa_avatar = models.URLField(max_length=512, blank=True, default='')
+    has_subscription = models.BooleanField(default=False)
```
johngian: Maybe convert this to `has_subscriptions` to keep this similar to the claim name.

johngian:
I added a couple of nit comments. Other than that it looks straightforward.

kitsune/products/jinja2/products/includes/topic_macros.html
akatsoulas: Because it doesn't make sense to have this link enabled in case of a failover to Frankfurt because users will not be able to login (read-only mode).

johngian: :+1:

PR #3868, opened 2019-08-15
Replace bn_IN, bn_BD in favor of bn locale.

akatsoulas:


safwanrahman:
You need to create migration and others for it. Its bit much complex actually!

PR #3869, opened 2019-08-17
[Fix #3846] Merge locales bn-BD and bn-IN to bn

### LINKED ISSUE #3846: Merge locales bn-BD and bn-IN to single locale bn
See https://github.com/mozilla-l10n/sumo-l10n/issues/6

safwanrahman:
This should merge both `bn-BD` and `bn-IN` to `bn`.
I have not deleted the documents of `bn-IN`. I think we can do it with seperate migration after this things go into production.


safwanrahman:
@akatsoulas r?

akatsoulas:
Let's move the migrations into a single commit please.

kitsune/gallery/migrations/0006_change_bn_BD_and_bn_IN_to_bn.py
```diff
@@ -0,0 +1,38 @@
...
+    Image.objects.all().filter(locale='bn-BD').update(locale='bn')
+    # Change bn-IN images to bn
+    Image.objects.all().filter(locale='bn-IN').update(locale='bn')
+
+    Video.objects.all().filter(locale='bn-BD').update(locale='bn')
+    # Change bn-IN videos to bn
+    Video.objects.all().filter(locale='bn-IN').update(locale='bn')
+
+
+def change_locale_bn_to_bn_bd_backwards(apps, schema_editor):
```
akatsoulas: I would go with just `pass`. This is a one way migration and we cannot preserve the initial data here. So instead of moving everything back to `bn-BD` let's keep them in `bn`.

akatsoulas: Let's update the backwards migration to `pass` for all the data migrations that we have.

kitsune/wiki/migrations/0010_change_locale_bn_BD_to_bn.py
```diff
@@ -0,0 +1,37 @@
...
+
+from django.db import migrations
+
+
+def change_locale_bn_bd_to_bn_forwards(apps, schema_editor):
+    Document = apps.get_model('wiki', 'Document')
+    DraftRevision = apps.get_model('wiki', 'DraftRevision')
+    Locale = apps.get_model('wiki', 'Locale')
+
+    Document.objects.all().filter(locale='bn-BD').update(locale='bn')
```
akatsoulas: We need to do the same for all the articles in `bn-IN` that do not exist in `bn-BD`.

akatsoulas: Probably we need to also delete any documents/revisions/locales that may exist under the old locales

kitsune/wiki/tests/test_templates.py
```diff
@@ -389,18 +389,18 @@ def test_document_with_fallback_locale(self):
         """The document template falls back to fallback locale if there is
         custom wiki fallback mapping for the locale and the locale have no translation
         exists."""
-        # Create an English document and a bn-BD translated document
+        # Create an English document and a es translated document
```
akatsoulas: Instead of `es` here, let's use `bn` since we actually added it as a fallback for `bn-IN` and `bn-BD`.

kitsune/wiki/tests/test_views.py
```diff
@@ -902,17 +902,17 @@ def test_custom_locale_mapping(self):
         Test that the article is shown in the defined fallback locale
         """
 
-        # Create a document localized into bn-BD. Attempt to resolve to the bn-IN version
+        # Create a document localized into es. Attempt to resolve to the ca version
```
akatsoulas: Similar let's use either `bn-BD` or `bn-IN` and `bn`.

kitsune/wiki/tests/test_views.py
```diff
@@ -936,30 +936,30 @@ def test_custom_locale_mapping_for_header_locale(self):
         # The user-requested locale has a defined custom wiki fallback locale,
         # but the article is not localized into that fallback locale
 
-        # Define a client with ca, pt-PT, ja as locale choices in the ACCEPT_LANGUAGE header
-        header = 'ca,bn-IN;q=0.7,ja;q=0.3,'
-        # Create a document localized into pt-BR
+        # Define a client with ca, wo, ja as locale choices in the ACCEPT_LANGUAGE header
```
akatsoulas: Similar here. s/bn-IN/bn

safwanrahman:
@akatsoulas Updated the PR as per you commented. r?

kitsune/wiki/tests/test_views.py
safwanrahman: As per our conversation over slack, this feature was implemented for `bn-BD` and `bn-IN` locale fallback. As both locales are merging together, we do not need it anymore.

PR #3870, opened 2019-08-20
Add a uuid as an FxA uid when closing an account.

akatsoulas:
* Remove unique=True from fxa_uid field.

### DIFF
--- a/kitsune/users/views.py
+++ b/kitsune/users/views.py
@@ -1,4 +1,5 @@
 import os
+from uuid import uuid4
 from ast import literal_eval
 from datetime import datetime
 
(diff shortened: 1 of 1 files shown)


johngian:
Instead of doing the checks why don't we fix it in the cleanup user level. Instead of removing unique fxa uid we can add a random fxa uid like ``inactive-<uuid4>`` for inactive users.

kitsune/users/views.py
```diff
@@ -331,13 +332,15 @@ def profile(request, username):
 @require_POST
 def close_account(request):
     # Clear the profile
-    profile = get_object_or_404(Profile, user__id=request.user.id)
+    user_id = request.user.id
+    profile = get_object_or_404(Profile, user__id=user_id)
     profile.clear()
+    profile.fxa_uid = '{user_id}-{uid}'.format(user_id=user_id, uid=str(uuid4()))
```
johngian: I would prefer some sort of information that user is inactive here.

PR #3888, opened 2019-09-11
Use the /subscriptions endpoint to get subscription related info.

akatsoulas:


kitsune/users/auth.py
```diff
@@ -170,6 +171,32 @@ def filter_users_by_claims(self, claims):
...
+                },
+                verify=self.get_settings('OIDC_VERIFY_SSL', True))
+            sub_response.raise_for_status()
+        except requests.exceptions.RequestException:
+            log.error('Failed to fetch subscription status', exc_info=True)
+            # if something went wrong, just return whatever the profile endpoint holds
+            return user_info
+        # This will override whatever the profile endpoint returns
+        # until https://github.com/mozilla/fxa/issues/2463 is fixed
+        user_info['subscriptions'] = sub_response.json().get('subscriptions')
```
akatsoulas: TODO: The default value of get should be changed to an empty string

PR #3895, opened 2019-09-17
Show contact us button on relevant products.

akatsoulas:


kitsune/products/jinja2/products/includes/topic_macros.html
```diff
@@ -34,7 +34,7 @@ <h3 class="support-callout-title"><span class="highlight">{{ _('Ask the Communit
       </div>
     {% endif %}
 
-    {% if show_contact_form and settings.OIDC_ENABLE %}
+    {% if settings.OIDC_ENABLE and product.id in subscribed_products_ids %}
```
johngian: you can always pass the query and check if the query exists.

johngian: Strong suggestion: To avoid list comprehension and the check here you can always implement this query with subqueries which might be a more elegant + performant solution

akatsoulas: The subqueries suggestion is great and I am already working on this to optimize big queries that slow down the side. This is though not one of them. Regarding the `exists`, this is more "expensive" from the list comprehension.

PR #3923, opened 2019-10-11
Set the timezone to UTC.

akatsoulas:


akatsoulas:
The time will remain in PST.

PR #3941, opened 2019-10-18
Remove marketplace and related functionality.

akatsoulas:


akatsoulas:
@johngian could you take a look at this?

PR #3942, opened 2019-10-18
Replace rfc822 module.

akatsoulas:


akatsoulas:
@johngian r?

kitsune/customercare/management/commands/collect_tweets.py
```diff
@@ -2,8 +2,8 @@
 import json
 import logging
 import re
-import rfc822
 from datetime import datetime
+from email import utils
```
johngian: nit: ``from email import utils as email_utils`` just because utils is very common. but no pressure

PR #3951, opened 2019-10-22
Drop deprecated libs assets

akatsoulas:


akatsoulas:
Merging since both of the commits were reviewed before the revert.

PR #3956, opened 2019-10-23
Convert all PST datetime objects to UTC.

johngian:
PR background:

* Currently default timezone in kitsune is PST
* Currently django uses naive `DateTime` objects
* We want to make everything aware and UTC
* Instead of changing each field one by one we interpolate all apps -> models -> fields

Steps:

* Failover to kitsune read-only because data migration is huge
* Run migration first to convert all naive PST objects to naive UTC
* Change USE_TZ=True
* Failover DB to read-write

TODO

- [ ] My assumption was that apps was populated from migration dependencies. Apparently its not, we need to exclude non-kitsune apps


johngian:
Open topics:

* Migration takes too long (it involves going through a lot of tables and all rows) 
* How can we rollback

safwanrahman:
Rathar than doing it in Python end, do there any way to do it in DB end so it will be much faster?

johngian:
@safwanrahman this is more of a PoC so @akatsoulas can take a look of some ideas i had to interpolate fields and make a minimal small migration.

I am also testing locally the approach to do DB operations directly because `save` operation costs a lot of IO.

We are also thinking to leave it as it is (PST) and activate UTC timezone with a middleware.

Thanks for your suggestion :rocket:

safwanrahman:
@johngian I also propose to keep the PST timezone as it is and create seperate type DB field that will convert the PST time to UTC time. What do you think?

akatsoulas:
I am closing this for now. As mentioned in a different issue, different solutions are being currently discussed to solve for this issue. Thanks @johngian for this elegant migration

PR #3972, opened 2019-11-01
🚧 WIP: Frontend tooling updates

kkellydesign:
To use this branch: 

`docker-compose down && make build && docker-compose up`

Open http://localhost:3000

This is a proxy on top of the compiled styleguide, which lives in /static/build/styleguide. 

All of the build and dev scripts now live in package.json in the 'scripts' area. This doesn't touch the javascript system, but at the end of this project will completely replace the current less-based system.


ipmb:
A few notes here:

* We added a new Nodejs-specific container which builds these files and runs the watcher in development.
* I haven't tested the production Docker pipeline with this setup, I could use some help confirming that still works as expected.
* As @kkellydesign mentioned, this doesn't touch the JS system, but it would make sense to migrate that to this setup as well. That would allow the project to drop bower & django-pipeline altogether. This would also simplify the Docker setup somewhat (the base container wouldn't need Node installed).

akatsoulas:
After #4039 is merged everything should be under responsive-redesign branch.

PR #3969, opened 2019-10-31
Receive and store SET tokens.

akatsoulas:

PR #3989, opened 2019-11-08
Receive and store SET tokens.

akatsoulas:


akatsoulas:
This was approved in #3969

PR #4006, opened 2019-11-14
Add ES7 document for `Document` model 

johngian:
* Implement locale aware multi-field
* Add basic config required by ES7
* Implement document definition for `Document` model
* Multiple minor fixes


kitsune/search/v2/documents.py
```diff
@@ -0,0 +1,103 @@
...
+        if instance.current_revision and not (instance.is_template and
+                                              instance.html.startswith(REDIRECT_HTML) and
+                                              instance.category == 50):
+            return instance.recent_helpful_votes
+        return 0
+
+    def prepare_display_order(self, instance):
+        return instance.original.display_order
+
+    def prepare(self, instance):
```
safwanrahman: I think this need to be a class method. So we can get a document instance by providing a model instance.
The API can be something like
```
django_obj = Document.objects.first()
es_document = WikiDocument.prepare(instance=django_obj)
es_document.save()
```

kitsune/search/v2/documents.py
```diff
@@ -0,0 +1,103 @@
...
+            'url', 'indexed_on', 'updated', 'product', 'topic', 'title',
+            'keywords', 'content', 'summary', 'locale', 'current_id', 'parent_id',
+            'category', 'slug', 'is_archived', 'recent_helpful_votes', 'display_order'
+        ]
+
+        for f in fields:
+            try:
+                prepare_method = getattr(self, 'prepare_{}'.format(f))
+                value = prepare_method(instance)
+                setattr(self, f, value)
```
safwanrahman: What about making it a class method and create a new instance of the document object and set value to the new object dynamically?

kitsune/search/v2/documents.py
johngian: Hey @safwanrahman i am planning to do this but we needed to unblock the process so @akatsoulas can also work on the codebase

safwanrahman: @johngian Cool.

PR #4007, opened 2019-11-14
Use classmethod for ES object preparation.

johngian:


johngian:
@akatsoulas r?

kitsune/search/v2/documents.py
```diff
@@ -91,13 +92,17 @@ def prepare(self, instance):
...
-                prepare_method = getattr(self, 'prepare_{}'.format(f))
+                prepare_method = getattr(obj, 'prepare_{}'.format(f))
                 value = prepare_method(instance)
-                setattr(self, f, value)
             except AttributeError:
                 value = getattr(instance, f)
-                setattr(self, f, value)
 
-        self.meta.id = instance.id
+            setattr(obj, f, value)
```
safwanrahman: I think the `id` should be set with the `document.id`. So it keeps unique and does not create new one while update.

PR #4036, opened 2019-11-28
Styleguide

johngian:


kitsune/sumo/static/sumo/scss/config/_variables.scss
```diff
@@ -0,0 +1 @@
+$styleguide_path: "../../../../../../styleguide-examples/";
```
akatsoulas: Can't we sue the full path here?

kss-config.json
```diff
@@ -0,0 +1,15 @@
...
+  "source": "kitsune/sumo/static/sumo/scss",
+  "destination": "static/build/styleguide",
+  "//": "The css and js paths are URLs relative to the style guide destination.",
+  "css": [
+    "static/css/screen.css"
+  ],
+  "//": [
+    "Change the key to js and remove this note when JS is compiled with the frontend. old: js/bundle.js"
+  ],
+  "homepage": "../../../../../styleguide-examples/styleguide-index.md",
```
akatsoulas: Maybe a full path?

kss-config.json
```diff
@@ -0,0 +1,15 @@
+{
```
akatsoulas: Do we need to have this top level?

akatsoulas: I would move it inside the folder with all the kss files.

kitsune/sumo/static/sumo/scss/config/_variables.scss
johngian: full path from where ?

kss-config.json
johngian: same

kss-config.json
johngian: yes because of `package.json`

PR #4039, opened 2019-11-28
Missing frontend updates.

akatsoulas:


akatsoulas:
This PR pulls the latest updates from #3972 . @kkellydesign With this PR there should be parity between the open PR and the responsive-redesign branch one this is merged. The only thing that is missing is the hot reloading which I am going to fix in another PR.

PR #4048, opened 2019-12-04
LL homepage design and styleguide fixes

kkellydesign:


kkellydesign:
@akatsoulas I have merged in my most recent CSS and template work on the homepage, and the styleguide is now fixed. You had the right idea with the sass variable, but kss-node actually doesn't compile SCSS before they parse the comments for documentation, so that sass variable didn't work. I replaced it all with plain text, and it's working correctly now. 

I've also deployed my S3 scripts locally, so I can still publish the styleguide. This should all be good now.

PR #4049, opened 2019-12-04
Instant reload through browser-sync

akatsoulas:
* Proxy requests through the django dev server.


akatsoulas:
@kkellydesign quick r?

PR #4050, opened 2019-12-05
WIP upgrade to Python 3.7

benspaulding:
This is currently based on mozilla/kitsune@responsive-design because I was told that had some of the latest Docker configuration that should be used.

The process will be to upgrade to Python 3, upgrading packages as necessary. Once that is done a subsequent upgrade to Django 2.2 (including other packages as necessary) will be done.


akatsoulas:
@benspaulding maybe we should start the upgrade from the master branch instead of the responsive-redesign. The later removes some front-end tooling (bower, yarn, gulp) in favor of npm and adds an intermediate container to build the CSS. This can be easily added in master when the work here is complete. 

The reason for proposing this is because I would like to decouple the two projects (python/django upgrade and responsive-redesign) and treat them as separate. Both of them are going to change a good portion of the code-base and probably have two separate branches based on master will be easier to manage.

akatsoulas:
We can use the `python3` branch for the duration of this project.

benspaulding:
@akatsoulas That sounds like a good idea. I will move my work to the appropriate branch tomorrow. Thanks!

benspaulding:
Closed by #4128.

PR #4055, opened 2019-12-06
homepage WIP in styleguide

kkellydesign:
There's a commit from yesterday that fixes the build, but there are still a couple of javascript errors left over from (I believe) the differences between the bower and NPM registries. For example, there's one change from `jquery/jquery.min.js` to `jquery/dist/jquery.min.js`. I'm not sure how to fix the modernizr issue though, so that one still needs work. 

Here are the errors that I'm seeing: 

Screen Shot 2019-12-06 at 9 52 25 AM

As far as the docker build issue, I had removed the fonts from the project since they're also included with Protocol, and that made the old CSS/LESS build process break. I added the fonts back in, and that fixed the issue. When I'm ready to spend some time pulling out all of the LESS files, the fonts can be pulled back out again as well.


kkellydesign:
@akatsoulas I just fixed the modernizr issue. I added a custom build to only use the modernizr features that the project is using. 

https://github.com/mozilla/kitsune/pull/4055/commits/dd459c95201b12d572a0f3f1b30154107797968b

akatsoulas:
Closing this in favor of #4061 . The only difference is that the later excludes the Merge commit

PR #4065, opened 2019-12-10
Add scss files into static.

akatsoulas:


Dockerfile
```diff
@@ -5,6 +5,7 @@ FROM node:12 AS frontend-base
 
 WORKDIR /app
 COPY ["./package.json", "./package-lock.json", "prepare_django_assets.js", "/app/"]
+COPY ./kitsune/sumo/static/sumo ./kitsune/sumo/static/sumo
```
johngian: Use absolute paths for references inside docker

PR #4104, opened 2020-01-02
LL fix messages

kkellydesign:
This should fix the test errors that we were getting, plus add support for multi-line text in the protocol notification bars.


akatsoulas:
Closing this in favor of #4107. I cherry-picked the commits from here into that branch apart from the merge commit

PR #4113, opened 2020-01-10
Put period into l10n strings

petercpg:
A tiny fix for putting sentence periods into l10n strings. 

There are some languages do not use dots as periods. We should keep them consist and have them localizable.


akatsoulas:
Thanks @petercpg !

PR #4122, opened 2020-01-16
Temporarily remove user-agent from logging.

akatsoulas:


akatsoulas:
After talking with @ziegeer @kfferrando we probably need to keep this as it provides the logging. We will try to switch to gunicorn workers .

PR #4129, opened 2020-01-17
Update imports for urllib

akatsoulas:
* Black updates.


akatsoulas:
@benspaulding This PR all it does is to update the imports for urllib. The rest is noise from the black formatter which was a recent decision.

benspaulding:
@akatsoulas Thanks for letting me know. I am +1 on formatters, but it will get difficult if this happens to a lot of modules in this process.

PR #4131, opened 2020-01-17
Fix matchMedia.

akatsoulas:


akatsoulas:
Js tests and linting are passing with this PR

PR #4137, opened 2020-01-22
Ll fix js search tests

kkellydesign:
Tests are passing in this branch, and search is back to the previous functionality. As in, it works, but without the couple of new javascript plugins that are required from protocol. This will at least let us deploy code and test on tickets that don't apply to the react-based instant search.


akatsoulas:
Fixed here 
https://github.com/mozilla/kitsune/pull/4131

PR #4155, opened 2020-02-03
Python3 upgrade

benspaulding:
We are getting close to a passing test suite.


benspaulding:
The test suite is down to just 8 errors! All of those will be fixed once jazzband-roadies/issues#183 lands and we can update our version of django-authority.

akatsoulas:
@benspaulding can we squash/drop the commits that revert changes to the initial state?

PR #4171, opened 2020-02-11
Remove django-mobility.

akatsoulas:
* Drop mobile-template decorator from the project.


johngian:
I don't know a lot about the internals but code wise it looks OK after some comments already communicated.

PR #4176, opened 2020-02-13
Remove alternate image field

kkellydesign:
This removes a database field for "alternate image" on the product model. Since all product logos sitewide are now squares that are resized with CSS, only one image is needed. I also changed the help text on the main image to reflect the new design. 

Screen Shot 2020-02-13 at 10 23 40 AM


akatsoulas:
The alternate image field is used in the AAQ products. I am closing this for now. We can revisit it once we migrate those products away from this field.

PR #4180, opened 2020-02-19
Fixing locale sync

safwanrahman:
@akatsoulas @kfferrando As per our conversation over slack, it seems like there are some conflict in pull. This should fix this issue.
r?


akatsoulas:
Thanks @safwanrahman for this PR. @kfferrando already fixed that and the missing translations are already deployed in prod. Let's re-open this if we notice any similar behavior. Thanks again.

PR #4201, opened 2020-02-28
LL updates 2-28

kkellydesign:


akatsoulas:
Closing in favor of #4207 which has more atomic commits (eg test fixes merged into one)

PR #4223, opened 2020-03-12
Hide raw-content in kbforums.

akatsoulas:


akatsoulas:
Without this the raw content in the thread is also displayed and the thread ends up having the content two times. @kkellydesign shall I move this do a different file?

PR #4234, opened 2020-03-17
prevent None timezone in session

LeoMcA:
This prevents the `astimezone() argument 1 must be datetime.tzinfo, not None` exception in two ways:

First, if `session['timezone']` is already `None`, default to the server timezone when converting the value.

Then, never allow a user in future to set their `session['timezone']` value to None by only updating the value in session if the user has set a timezone.

Before these changes the exception could be triggered on a fresh profile by:

1. Create a new profile, with a None timezone.
2. Set the profile's timezone to a value.
3. Through the user's preferences, set the profile's timezone back to None (appears as `-------` in the dropdown).
4. Visit the forum page, see the exception.

This commit should probably be cherry picked on top of master, because this is an issue in prod, and doesn't seem to affect/be affected by the responsive redesign.

### DIFF
--- a/kitsune/sumo/templatetags/jinja_helpers.py
+++ b/kitsune/sumo/templatetags/jinja_helpers.py
@@ -266,7 +266,7 @@ def datetimeformat(context, value, format="shortdatetime"):
                     pass
             request.session["timezone"] = convert_tzinfo
         else:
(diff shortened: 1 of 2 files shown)


akatsoulas:
Cherry-picked in #4264

PR #4238, opened 2020-03-17
style link modal like media modal

LeoMcA:
https://github.com/mozilla/sumo-project/issues/253

This looks a lot better to me:

Screen Shot 2020-03-17 at 17 03 50

### DIFF
--- a/kitsune/sumo/static/sumo/js/markup.js
+++ b/kitsune/sumo/static/sumo/js/markup.js
@@ -499,7 +499,8 @@
         destroy: true,
         modal: true,
         id: 'link-modal',
-        container: $('body')
+        container: $('body'),
+        position: 'none'
       });
       kbox.open();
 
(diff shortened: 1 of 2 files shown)


kkellydesign:
This looks good with one minor tweak – could you add the class `reverse-on-desktop` to the `<div class="sumo-button-wrap">` component inside that modal? The revised pattern is to have the main actions on the right, so I've been changing that as I catch them throughout.

LeoMcA:
You've already done so in a previous commit: https://github.com/mozilla/kitsune/commit/cc76a0156a3de7896f774d52fcc0625496539473#diff-e38939b551282f6bfc59ad8f20c1336cR318

> The revised pattern is to have the main actions on the right, so I've been changing that as I catch them throughout.

Is that on the right on mobile, or on the right on desktop? This is consistent with the media modal in that the main action is on the right on desktop, and on the left on mobile.

akatsoulas:
I am merging this PR in order to start the merge process into master. Please open a new PR for that class if it is still valid.

PR #4244, opened 2020-03-18
LL AAQ design work

kkellydesign:


akatsoulas:
Closing in favor of #4284 which is rebased on current master and included only the missing commits.

PR #4256, opened 2020-03-23
Fxa live events

LeoMcA:


akatsoulas:
with tests passing  :)

akatsoulas:
Please squash the last two commits into one. r+

kitsune/users/migrations/0022_auto_20200402_0429.py
```diff
@@ -0,0 +1,20 @@
+# -*- coding: utf-8 -*-
```
akatsoulas, line 1: Let's squash this migration together with the previous one

kitsune/settings.py
```diff
@@ -125,6 +125,10 @@ def parse_conn_max_age(value):
 # http://www.i18nguy.com/unicode/language-identifiers.html
 LANGUAGE_CODE = 'en-US'
 
+LOCALE_EXCLUDE_URLS = [
```
akatsoulas, line 128: let's use the `OIDC_EXEMPT_URLS` instead

kitsune/users/migrations/0022_auto_20200402_0429.py
akatsoulas: and number `0020` :)

kitsune/users/urls.py
```diff
@@ -98,6 +98,6 @@
             name='users.fxa_authentication_init'),
         url(r'^fxa/logout/$', never_cache(views.FXALogoutView.as_view()),
             name='users.fxa_logout_url'),
-        url(r'^fxa/events/$', never_cache(views.WebhookView.as_view()), name='users.fxa_webhook'),
+        url(r'^fxa/events', never_cache(views.WebhookView.as_view()), name='users.fxa_webhook'),
```
akatsoulas, line 101: why do we need this change?

kitsune/users/views.py
```diff
@@ -822,24 +841,38 @@ def post(self, request, *args, **kwargs):
...
-        # SET event must be of type `application/secevent+jwt`
-        if request.META.get('CONTENT_TYPE', '') != 'application/secevent+jwt':
+        authorization = request.META.get('HTTP_AUTHORIZATION')
+        if not authorization:
             raise Http404
 
+        log.debug(authorization)
+        log.debug(request.META.get('CONTENT_TYPE'))
+
+        # dummy tokens sent by the fxa-event-broker test script don't use this header,
```
akatsoulas, line 851: Let's open an issue for this

kitsune/users/urls.py
LeoMcA: in bug 1595529 you requested `support-dev.allizom.org/fxa/events` as the webhook url

I'm realising now that this was regex... so I should probably change it to something like `^fxa/events/?$` to handle both with and without a trailing slash

kitsune/users/tests/test_views.py
```diff
@@ -399,3 +402,84 @@ def test_authenticate_does_update_session(self):
...
         self.client.get(url)
         assert self.client.session.get('is_contributor')
+
+
+class WebhookViewTests(TestCase):
+
+    def _setup_key(test):
+        @mock.patch('kitsune.users.views.requests')
+        def wrapper(self, mock_requests):
+            pem = """-----BEGIN RSA PRIVATE KEY-----
```
akatsoulas, line 412: let's move the key one line down

kitsune/users/views.py
```diff
@@ -779,31 +786,43 @@ def get_settings(attr, *args):
...
+        key = None
+        for jwk in jwks['keys']:
+            # dummy tokens sent by the fxa-event-broker test script don't have a kid value,
+            # we'll have to see if this is true in dev
+            # if jwk['kid'] != header['kid']:
+            #     continue
+            if 'alg' in jwk and jwk['alg'] != header['alg']:
+                raise SuspiciousOperation('alg values do not match.')
+            key = jwk
+        if key is None:
```
akatsoulas, line 807: you can do `if not key`. This will also catch an empty string for example. I am not sure though if an empty string could be appended to the key though

akatsoulas:
Let's also rebase this PR with current master

PR #4265, opened 2020-03-26
Sumo accounts fixes

akatsoulas:


kitsune/users/views.py
```diff
@@ -311,6 +311,33 @@ def get(self, request):
...
+        messages.add_message(
+            self.request,
+            messages.ERROR,
+            _('This account is not active. '
+              'Please contact an administrator if you believe this is an error')
+        )
+        return HttpResponseRedirect(reverse('home'))
+
+
 class FXALogoutView(OIDCLogoutView):
```
LeoMcA: Is this meant to log the user out from FxA as well?

Looking at the OIDCLogoutView it seems to expect a setting called `OIDC_OP_LOGOUT_URL_METHOD`:
https://github.com/mozilla/mozilla-django-oidc/blob/master/mozilla_django_oidc/views.py#L194

But we don't set such a setting in `get_oidc_fxa_setting`

kitsune/settings.py
```diff
@@ -537,6 +537,9 @@ def parse_conn_max_age(value):
     'commonware.middleware.RobotsTagHeader',
     # 'axes.middleware.FailedLoginMiddleware'
     'django_user_agents.middleware.UserAgentMiddleware',
+
+    'kitsune.users.middleware.LogoutDeactivatedUsersMiddleware',
```
LeoMcA: We still have `ModelBackendAllowInactive` enabled:
https://github.com/akatsoulas/kitsune/blob/61c21004263ef7575c9a6aabe67e974a268d1f45/kitsune/settings.py#L572

kitsune/users/views.py
akatsoulas: If this method doesn't exist it just does a logout in the django site.

PR #4273, opened 2020-03-30
still render show_for logged out users

LeoMcA:
This is a rough fix for https://github.com/mozilla/sumo-project/issues/307#issuecomment-605893728

@kkellydesign I'm not too familiar with the sidebar behaviour, so you might want to neaten this up a little bit.

For instance, we have duplicated headings on mobile:

Screen Shot 2020-03-30 at 13 54 33

### DIFF
--- a/kitsune/wiki/jinja2/wiki/includes/sidebar_modules.html
+++ b/kitsune/wiki/jinja2/wiki/includes/sidebar_modules.html
@@ -4,13 +4,15 @@
   {% if not parent %}
     {% set parent = document %}
   {% endif %}
-  {% if user.is_authenticated() %}
   <nav id="doc-tools">
     <ul class="sidebar-nav sidebar-folding">
       <li id="editing-tools-sidebar">
-        <h3 class="sidebar-subheading large-only">{{ _('Editing Tools') }}</h3>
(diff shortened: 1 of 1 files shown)


kitsune/wiki/jinja2/wiki/includes/sidebar_modules.html
```diff
@@ -64,9 +64,16 @@ <h3 class="sidebar-subheading large-only">{{ _('Editing Tools') }}</h3>
             </li>
         </ul>
       </li>
+      {% else %}
+      <span class="details-heading"></span>
+      <ul class="sidebar-nav--list">
+        <li>
+          {{ show_for((document or parent).get_products()) }}
```
akatsoulas: Shouldn't this be under `{% if include_showfor %}`?

akatsoulas: On second thought instead of adding an else statement, can't we just move the `showfor` section outside of the `is_authenticated` check?

LeoMcA: Ah, good point, I'd overlooked the `include_showfor` condition.

PR #4277, opened 2020-04-01
Show abuse form to authenticated users.

akatsoulas:


kitsune/sumo/tests/test_templates.py
```diff
@@ -1,22 +1,19 @@
 from django.template import engines as template_engines
```
LeoMcA, line 1: Seems like black touched this file for some reason, can we remove these changes from this commit?

kitsune/forums/jinja2/forums/includes/post.html
```diff
@@ -30,26 +31,31 @@
...
-      {% if post.id and user.is_authenticated() and not thread.is_locked and forum.allows_posting_by(user) %}
+      {% if user.is_authenticated() %}
+      {% if post.id and not thread.is_locked and forum.allows_posting_by(user) %}
       <li class="mzp-c-menu-list-item post-action">
         <a class="reply" data-post="{{ post.id }}" href="#thread-reply">{{ _('Quote') }}</a>
       </li>
-    {% endif %}
+      {% endif %}
+
+      {% if post.id %}
```
LeoMcA, line 49: Should we move these common `if post.id` conditions into the `if user.is_authenticated()` conditional?

kitsune/forums/jinja2/forums/includes/post.html
```diff
@@ -30,26 +31,31 @@
...
-    <a href="#post-{{ post.id }}">{{ _('Link to this post') }}</a>
-    </li>
-    {% if post.id %}
-    <li class="mzp-c-menu-list-item post-action">
-      {{ flag_form(url('forums.flag_post', post.thread.forum.slug, post.thread.id, post.id)) }}
-    </li>
-    {% endif %}
+      <li class="mzp-c-menu-list-item post-action">
+        <a href="#post-{{ post.id }}">{{ _('Link to this post') }}</a>
+      </li>
```
LeoMcA, line 58: Although, surely this should also be wrapped in a `if post.id`?

kitsune/sumo/tests/test_templates.py
akatsoulas: this is not from black. Just some (automatic) sorting of the import modules

PR #4290, opened 2020-04-07
do mobile ua detection in util function

LeoMcA:
https://github.com/mozilla/sumo-project/issues/326

### DIFF
--- a/kitsune/questions/views.py
+++ b/kitsune/questions/views.py
@@ -1,7 +1,6 @@
 import json
 import logging
 import random
-import re
 import time
 from datetime import date, datetime, timedelta
 
@@ -23,11 +22,6 @@ from django.views.decorators.http import (require_GET, require_http_methods,
                                           require_POST)
 from django_statsd.clients import statsd
 from django_user_agents.utils import get_user_agent
-from ordereddict import OrderedDict
-from taggit.models import Tag
-from tidings.events import ActivationRequestFailed
-from tidings.models import Watch
-
 from kitsune.access.decorators import login_required, permission_required
 from kitsune.products.models import Product, Topic
 from kitsune.questions import config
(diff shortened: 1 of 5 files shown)


kitsune/questions/views.py
```diff
@@ -514,32 +506,20 @@ def aaq(
 
         return HttpResponseRedirect(path)
 
-    if product_key is None:
+    user_agent = request.META.get("HTTP_USER_AGENT", "")
+    is_mobile_device = parse_ua(user_agent).is_mobile
 
+    if product_key is None:
```
akatsoulas, line 512: You can remove this line and move the `product_key = request.GET.get("product")` to the same if statement below.

kitsune/questions/views.py
```diff
@@ -514,32 +506,20 @@ def aaq(
 
         return HttpResponseRedirect(path)
 
-    if product_key is None:
+    user_agent = request.META.get("HTTP_USER_AGENT", "")
+    is_mobile_device = parse_ua(user_agent).is_mobile
```
akatsoulas, line 510: Let's use `get_user_agent` instead which checks cache first.

LeoMcA: Oh fair enough, I couldn't work out why you were using that module when we weren't using the middleware - it looks like we can use the caching without the middleware though.

kitsune/questions/views.py
LeoMcA: We'll still need another `if product_key is None` somewhere so we don't overwrite a `?product=blah` on mobile. Thought this was cleaner than more nested ifs.

Thinking about it, IDK if `?product=blah` is actually used anywhere... but I don't want to open any more cans of worms before the responsive release.

akatsoulas: it's used only in mobile devices in order to avoid the UA check if the user decided to follow the `change product` link

akatsoulas: Lines 511 and 514 both have the same check. Why don't we add the line below under 514 so that we ca remove one of the two if?

LeoMcA: Because they're not the same check. They both check `product_key`, but line 512 modifies the value of `product_key`. So line 514 is actually `if product_key is None and request.GET.get("product") is None`.

Were we to move line 512 under 514 we'd still need another `if` to check if `request.GET.get("product")` is None.

LeoMcA: > it's used only in mobile devices in order to avoid the UA check if the user decided to follow the `change product` link

That's the `change_product` param used below: https://github.com/mozilla/kitsune/pull/4290/files#diff-f5ee25d927694fd302a3bb72ac3ebb4eR517

akatsoulas: something like `product_key or request.GET.get("product")`

PR #4293, opened 2020-04-08
fix when we show that we've guessed browser/os in aaq

LeoMcA:
https://github.com/mozilla/sumo-project/issues/330

### DIFF
--- a/kitsune/sumo/static/sumo/js/aaq.js
+++ b/kitsune/sumo/static/sumo/js/aaq.js
@@ -41,6 +41,8 @@
         if (!$input.val()) {
           $input.val(self.getPlugins());
         }
+      } else {
+        $form.find('li.system-details-info').hide();
       }
 
       // Expanders.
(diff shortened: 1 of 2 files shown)


kitsune/sumo/static/sumo/js/aaq.js
```diff
@@ -147,15 +149,17 @@
     },
     isDesktopFF: function() {
       // Is the question for FF on the desktop?
-      return document.location.pathname.indexOf("desktop") >= 0;
+      return document.location.pathname.indexOf("desktop") >= 0 ||
+        document.location.pathname.indexOf("firefox-enterprise") >= 0;
```
akatsoulas, line 153: I don't have any objections about the code but I am not sure if we want this here. I would suggest to remove this from here to unblock this PR and open a new one with only this change so that we can discuss it with the rest of the team.

kitsune/sumo/static/sumo/js/questions.js
```diff
@@ -152,10 +152,6 @@
       .closest('ul')
       .toggleClass('show-details');
     });
-
-    if (!aaq.isDesktopFF() && !aaq.isMobileFF() && !aaq.isFirefoxOS()) {
```
akatsoulas, line 156: don't we need this section anymore (without the `isFirefoxOS` of course)?

LeoMcA: We've moved it into the above file: https://github.com/mozilla/kitsune/pull/4293/files#diff-1fb1ffccd814794452284c9186f28d79R44-R46

PR #4294, opened 2020-04-08
fix brittle test_metadata_keys test

LeoMcA:


kitsune/questions/tests/test_forms.py
```diff
@@ -56,7 +54,7 @@ def test_metadata_keys(self):
         expected = ['troubleshooting', 'ff_version', 'os',
                     'plugins', 'useragent']
         actual = form.metadata_field_keys
-        eq_(expected, actual)
+        eq_(sorted(expected), sorted(actual))
```
akatsoulas, line 57: Why are you using sorted here?

LeoMcA: This is what was causing the test to fail - the items were equal, but their order was different. I assume we don't care in which order this metadata appears?

PR #4302, opened 2020-04-14
use capture_message to log

LeoMcA:


kitsune/users/views.py
```diff
@@ -437,13 +438,17 @@ def post(self, request, *args, **kwargs):
         # TODO add docstrings
         # TODO add error handling
 
+        capture_message("""
```
akatsoulas, line 441: Indent please

PR #4307, opened 2020-04-15
Mark events without profiles as IGNORED.

akatsoulas:


akatsoulas:
This is pulled in #4348

PR #4317, opened 2020-04-16
[bug 1630396] Strip unnecessary user info when quoting

philipp-sumo:
attempting to fix https://bugzilla.mozilla.org/show_bug.cgi?id=1630396 - please bear with me if this is the wrong approach :-)


LeoMcA:
It's a little bit esoteric - can we just wrap the display name in the templates with a `<span class="display-name">` and use that to select on?

philipp-sumo:
thanks, you're probably right ;-)
i'll close this pull request and try again

PR #4318, opened 2020-04-16
make <pre> overflow auto

LeoMcA:
@kkellydesign I'm not sure if this is the right place for this

### DIFF
--- a/kitsune/sumo/static/sumo/scss/base/_reset.scss
+++ b/kitsune/sumo/static/sumo/scss/base/_reset.scss
@@ -18,3 +18,7 @@ body {
 img {
   max-width: 100%;
 }
+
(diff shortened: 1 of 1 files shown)


kkellydesign:
@LeoMcA that works for me. It's a toss-up between reset.scss and typography.scss.

LeoMcA:
Cool! I was thinking about typography.scss as well, but from the comments at the head of that file it looks like the intent was to remove it once protocol had updated, so reset.scss is probably the best:

```
// The following mixins were taken and edited from
// protocol/css/includes/mixins/_typography.scss.
// Once protocol is caught up, these can be removed
// and the protocol mixins can be used throughout the system.
```

PR #4319, opened 2020-04-16
[bug 1630396] Strip unnecessary user info when quoting

philipp-sumo:
attempting to fix https://bugzilla.mozilla.org/show_bug.cgi?id=1630396


LeoMcA:
@philipp-sumo apologies, we were in a (ultimately unsuccessful) rush yesterday to get these fixes out the door, so I forgot to say: thanks for this!

This is deployed on [staging](https://support.allizom.org/) and we'll attempt to deploy all the fixes there to prod on Monday.

PR #4320, opened 2020-04-16
only lazyload images in kb preview

LeoMcA:
https://github.com/mozilla/sumo-project/issues/346#issuecomment-614591678

without this fix previewing outside the kb (e.g. in a forum post reply) causes an error﻿


jens1o:
Couldn't this get refactored to use the native browser method? Is this already on the backlog?

akatsoulas:
@jens1o indeed, this is in the backlog. Would you be interesting in working on something like that?

jens1o:
> @jens1o indeed, this is in the backlog. Would you be interested in working on something like that?

Yeah, I couldn't find a public reference though. :/

PR #4326, opened 2020-04-17
Clear settings when an account is deleted or deactivated.

akatsoulas:


akatsoulas:
@LeoMcA pr is updated

PR #4334, opened 2020-04-20
fix downlad link on articles

LeoMcA:
ref: https://github.com/mozilla/sumo-project/issues/376﻿

### DIFF
--- a/kitsune/products/jinja2/products/product.html
+++ b/kitsune/products/jinja2/products/product.html
@@ -133,14 +133,14 @@
       {% if product.slug == 'thunderbird' %}
       <div class="download-buttons">
         <div class="download-thunderbird">
-          <a href="https://www.thunderbird.net/" class="download-button"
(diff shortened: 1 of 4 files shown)


LeoMcA:
@jess-cook03 I wasn't sure how this should appear, but did a best-effort based on how it appears on other KB pages and how it used to appear pre-responsive redesign:

Screenshot_2020-04-20 lorem ipsum Firefox Help - 127 0 0 1

akatsoulas:
Following up on this, there is another button under the `profile_nav` macro which is controlled by the `show_header_fx_download`. That header on purpose hides that button in any product related page. I would suggest to use the button in the header everywhere that we want to show it and remove the additional functionality. 

On the implementation side of things, probably we need to refactor a bit this functionality. Having the macro and the function used in the profile_nav probably is a bit more cleaner than having multiple references or a hardcoded button.

PR #4336, opened 2020-04-20
add prefers-reduced-motion

LeoMcA:
ref: https://github.com/mozilla/sumo-project/issues/371﻿


akatsoulas:
The test failure appears random. Can you amend the commit to trigger a build?

PR #4338, opened 2020-04-21
Enable the banner for FxA.

akatsoulas:


akatsoulas:
I am waiting for the finalized review for the text in the banner. Otherwise this is good for review.

kitsune/wiki/views.py
```diff
@@ -207,6 +210,17 @@ def document(request, document_slug, document=None):
     # The list above was built backwards, so flip this.
     breadcrumbs.reverse()
 
+    # Decide whether the sumo banner should be displayed in the KB.
+    # Base the decision on the English version of the document. If this is True
+    # all translation should display the banner
+    if doc.locale != settings.WIKI_DEFAULT_LANGUAGE:
+        doc_for_banner = doc.parent if doc.parent.locale == settings.WIKI_DEFAULT_LANGUAGE else None
```
LeoMcA, line 217: If the parent document isn't in `settings.WIKI_DEFAULT_LANGUAGE` this will blow up the `re.search` below. Rather than `else None` this should be `else ""`

LeoMcA: Also this line is too long: https://circleci.com/gh/mozilla/kitsune/2605?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link

akatsoulas:
> I think this is a fine implementation of the "search the slug" approach... I wonder if that's the best though. Are we sure that the things we'll we searching for (e.g. `lockwise|private-network` etc) will always be in the slug? Or should we search through the `breadcrumbs` instead?

Since this functionality will be redesigned, slug should be enough here. The articles to target are provided in the forms of urls/slugs as this is how we get the metrics regarding which ones to prioritize.

kitsune/wiki/jinja2/wiki/document.html
```diff
@@ -89,6 +89,10 @@
       </div>
     {% endif %}
 
+    {% if show_cta_banner %}
+      {{ sumo_cta_banner(text=_('Unlock more from the Firefox suite: Sync, Monitor, Lockwise and more.'), modifier_class="hide-on-mobile") }}
```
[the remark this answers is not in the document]
akatsoulas: base will give us flexibility to use this component (almost) anywhere in the site in the future iterations. Happy to move it though if you prefer the more explicit approach.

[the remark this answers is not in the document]
LeoMcA: It feels a bit disconnected? But minor thing, whatever you're more comfortable with

kitsune/wiki/views.py
```diff
@@ -207,6 +210,18 @@ def document(request, document_slug, document=None):
...
+    # Decide whether the sumo banner should be displayed in the KB.
+    # Base the decision on the English version of the document. If this is True
+    # all translation should display the banner
+    if doc.locale != settings.WIKI_DEFAULT_LANGUAGE:
+        doc_for_banner = doc.parent if doc.parent.locale == settings.WIKI_DEFAULT_LANGUAGE else ''
+    else:
+        doc_for_banner = doc
+
+    if (settings.SUMO_BANNER_STRING and not request.user.is_authenticated()
+            and re.search(settings.SUMO_BANNER_STRING, doc_for_banner.slug)):
```
LeoMcA, line 222: Apparently I wasn't reading properly in my previous review: this will still error when `doc_for_banner == ''`, since `''.slug` isn't defined.

kitsune/wiki/views.py
```diff
@@ -207,6 +210,22 @@ def document(request, document_slug, document=None):
     # The list above was built backwards, so flip this.
     breadcrumbs.reverse()
 
+    # Decide whether the sumo banner should be displayed in the KB.
+    # Base the decision on the English version of the document. If this is True
+    # all translation should display the banner
+    if doc.locale != settings.WIKI_DEFAULT_LANGUAGE:
+        doc_for_banner = doc.parent if doc.parent.locale == settings.WIKI_DEFAULT_LANGUAGE else ''
```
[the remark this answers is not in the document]
akatsoulas: Definitely that will work but I kept it to an empty string because locale is a string too and this way we can have the same type.

akatsoulas:
Did you add a string for the regex in the settings.py or your .env file?

kitsune/wiki/views.py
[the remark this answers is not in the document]
LeoMcA: but you're not assigning the locale, you're assigning a doc object

akatsoulas:
Following up - that part was mistakenly excluded by a rebase

kitsune/wiki/jinja2/wiki/document.html
```diff
@@ -70,6 +70,10 @@
       </div>
     </header>
 
+    {% if show_cta_banner %}
+      {{ sumo_cta_banner(text=_('Unlock more from the Firefox suite: Sync, Monitor, Lockwise and <a href="https://www.mozilla.org/firefox/accounts/">more</a>'),
```
LeoMcA, line 74: Do we sanitize this localised string to ensure it doesn't link somewhere we don't want, or include any `onclick` attributes, for instance?

kitsune/sumo/jinja2/includes/common_macros.html
```diff
@@ -12,13 +12,13 @@
...
+{% macro sumo_cta_banner(text, modifier_class, id="id_ff_alert") %}
 <div class="sumo-ff-alert {{ modifier_class }}" {% if id %}id="{{ id }}"{% endif %}>
   <div class="mzp-l-content">
     <div class="sumo-ff-alert--content">
       <img class="sumo-ff-alert--icon" src="{{ STATIC_URL }}sumo/img/FF_Master__icon.svg" />
       <p class="sumo-ff-alert--heading">{{ text }}</p>
-      <a href="{{ url('users.auth') }}" class="sumo-button primary-button button-lg">Sign Up</a>
+      <a href="https://accounts.firefox.com" class="sumo-button primary-button button-lg">Start Now</a>
       <button class="sumo-close-button" data-close-button="#{{ id }}">
         <img src="{{STATIC_URL}}protocol/img/icons/close-white.svg" />
```
LeoMcA, line 23: Here and above on line 19, I have to use the `{{ static('protocol/img/icons/close-white.svg' }}` helper to get the image to load, but I'm not sure if this is just a problem with my local dev environment.

akatsoulas: using static is more correct but the image is loading fine for me. The hardcoded path is the correct one. Will update.

PR #4345, opened 2020-04-23
Fix "Popular Searches" strings missing from l10n

philipp-sumo:


kelimuttu:
Some contributors say that the string is still unable to localize. Has this been deployed?

PR #4348, opened 2020-04-27
handle delete-user SET

LeoMcA:


kitsune/users/tests/test_models.py
```diff
@@ -28,3 +28,53 @@ def test_default_values(self):
         for setting in keys:
             field = SettingsForm.base_fields[setting]
             eq_(field.initial, Setting.get_for_user(self.u, setting))
+
+
+class AccountEventTests(TestCase):
+
+    def test_process_delete_user(self):
```
akatsoulas, line 35: Let's not have these events through cron. We can `apply` the task through celery as they come in. After that we can start exploring solutions (check the celery result, cron tasks) for failed/errored tasks.

kitsune/users/migrations/0021_auto_20200415_0409.py
```diff
@@ -0,0 +1,20 @@
+# -*- coding: utf-8 -*-
+# Generated by Django 1.11.29 on 2020-04-15 04:09
+from __future__ import unicode_literals
+
+from django.db import migrations, models
+
+
+class Migration(migrations.Migration):
```
akatsoulas, line 8: let's squash all the migrations into one

akatsoulas:
Could you also cherry-pick #4307 (squash that migration too)

kitsune/users/models.py
```diff
@@ -458,7 +464,72 @@ class AccountEvent(models.Model):
...
     profile = models.ForeignKey(Profile, related_name='account_events', null=True)
+    error = models.TextField(max_length=4096, blank=True)
+
+    def process(self):
+        if self.status == self.PROCESSED:
+            return
+
+        try:
+            if self.event_type == self.DELETE_USER:
+                self.profile.delete()
```
akatsoulas, line 479: Profiles are not deleted but rather deactivated which purges any PII from the db. Profile deletion would also affect the AAQ, forums, KB etc

kitsune/users/models.py
```diff
@@ -458,7 +464,72 @@ class AccountEvent(models.Model):
...
                                                   null=True,
                                                   blank=False)
-    fxa_uid = models.CharField(blank=True, null=True, unique=True, max_length=128)
+    fxa_uid = models.CharField(blank=True, null=True, max_length=128)
     jwt_id = models.CharField(max_length=256)
     issued_at = models.CharField(max_length=32)
     profile = models.ForeignKey(Profile, related_name='account_events', null=True)
+    error = models.TextField(max_length=4096, blank=True)
+
+    def process(self):
```
akatsoulas, line 473: Personally, I would go for atomic tasks that handle each case. Eg this process method could just be something like:
```python
def process:
   if even_type == self.DELETE_USER:
        process_delete_task.apply_async() # or delay
   elif another_type_here:
         process_other_type_task.apply()
```

The above will keep this method short, easy to read and make it act as a kind of scheduler which is what we want here. Processing should take place in each task.

kitsune/users/models.py
```diff
@@ -458,7 +464,72 @@ class AccountEvent(models.Model):
...
+                    self.events
+                )["https://schemas.accounts.firefox.com/event/subscription-state-change"]
+
+                try:
+                    last_event_obj = AccountEvent.objects.filter(
+                        profile_id=self.profile.pk,
+                        status=self.PROCESSED,
+                        event_type=self.SUBSCRIPTION_STATE_CHANGE
+                    ).order_by("last_modified")[0]
+                except IndexError:
```
akatsoulas, line 494: let's not use an exception for this

kitsune/users/models.py
```diff
@@ -81,6 +82,7 @@ class Profile(ModelBase, SearchMixin):
     fxa_uid = models.CharField(blank=True, null=True, unique=True, max_length=128)
     fxa_avatar = models.URLField(max_length=512, blank=True, default='')
     products = models.ManyToManyField(Product, related_name='subscribed_users')
+    password_change_time = models.DateTimeField(blank=True, null=True)
```
akatsoulas, line 85: probably we need a better naming for this. As is, it can be confused with the password of the local profile and not from FxA.

kitsune/users/middleware.py
```diff
@@ -19,6 +20,25 @@ def process_request(self, request):
             return HttpResponseRedirect(reverse('home'))
 
 
+class LogoutInvalidatedSessionsMiddleware(object):
```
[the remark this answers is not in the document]
LeoMcA: Good point, I just copied the middleware above. Should we open an issue to add MiddlewareMixin to all our custom middleware?

kitsune/users/middleware.py
```diff
@@ -19,6 +20,25 @@ def process_request(self, request):
...
+        user = request.user
+
+        if user.is_authenticated():
+            first_seen = request.session.get("first_seen")
+            if first_seen:
+                change_time = user.profile.fxa_password_change
+                if change_time and change_time > first_seen:
+                    logout(request)
+                    return HttpResponseRedirect(reverse('home'))
+            else:
```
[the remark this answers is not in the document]
LeoMcA: Then we'd be updating session["first_seen"] on every request, which isn't what we want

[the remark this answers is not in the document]
akatsoulas: missed an `if`. Ignore this

kitsune/users/models.py
```diff
@@ -430,35 +431,31 @@ class AccountEvent(models.Model):
...
     created_at = models.DateTimeField(auto_now_add=True)
     last_modified = models.DateTimeField(auto_now=True)
-    events = models.TextField(max_length=4096, blank=False)
-    event_type = models.PositiveSmallIntegerField(choices=EVENT_TYPE,
-                                                  default=None,
-                                                  null=True,
-                                                  blank=False)
-    fxa_uid = models.CharField(blank=True, null=True, unique=True, max_length=128)
+    body = models.TextField(max_length=4096, blank=False)
+    event_type = models.CharField(max_length=256,
```
akatsoulas, line 454: :+1: for the change to charfield. It may be useful though to keep the event_type enum

kitsune/users/tasks.py
```diff
@@ -0,0 +1,65 @@
...
+@timeit
+def process_event_subscription_state_change(event_id):
+    event = AccountEvent.objects.get(id=event_id)
+    body = json.loads(event.body)
+
+    last_event_q = AccountEvent.objects.filter(
+        profile_id=event.profile.pk,
+        status=AccountEvent.PROCESSED,
+        event_type=AccountEvent.SUBSCRIPTION_STATE_CHANGE
+    ).order_by("last_modified")
```
akatsoulas, line 31: You can also add an ordering entry in the model's `Meta` class to sort AccountEvents based on the `last_modified` event. This will apply to all events but here we are lucky because we do care about the last processed event. Then you can modify this query like `AccountEvent.objects.filter(do_some_filtering).last()`

kitsune/users/tasks.py
```diff
@@ -0,0 +1,65 @@
...
+
+@task
+@timeit
+def process_event_password_change(event_id):
+    event = AccountEvent.objects.get(id=event_id)
+    body = json.loads(event.body)
+
+    change_time = datetime.utcfromtimestamp(body["changeTime"] / 1000.0)
+
+    if (isinstance(event.profile.fxa_password_change, datetime) and
```
akatsoulas, line 56: Why do we need the isinstance check here?

kitsune/users/tests/test_models.py
```diff
@@ -1,11 +1,10 @@
 import logging
 
-from nose.tools import eq_
-
 from kitsune.sumo.tests import TestCase
 from kitsune.users.forms import SettingsForm
 from kitsune.users.models import Setting
 from kitsune.users.tests import UserFactory
+from nose.tools import eq_
```
akatsoulas, line 7: this should be reverted to the previous stage. First python libs followed by django libs and external libs and finally our import paths.

kitsune/users/views.py
```diff
@@ -435,6 +422,36 @@ def verify_token(self, token, **kwargs):
         # deserializing https://bugs.python.org/issue17909
         return json.loads(jws.payload.decode('utf-8'))
 
+    def process_events(self, payload):
+        fxa_uid = payload.get('sub')
+        profile_q = Profile.objects.filter(fxa_uid=fxa_uid)
```
akatsoulas, line 427: You can do here:
```python
try:
    profile = Profile.objects.get(fxa_uid=fxa_uid)
except Profile.DoesNotExist:
    profile = None
```

Then you can check if the profile var has a value. If not you can skip the computation and just save the event in the db without any profile

akatsoulas: This can also remove some nesting below.

akatsoulas: You can also bundle this to a handy utility like here (probably not in the jinja helpers):
https://github.com/mozilla/mozillians/blob/master/mozillians/common/templatetags/helpers.py#L267

kitsune/users/tasks.py
LeoMcA: Ah, just an artefact of the test failure without it being `TypeError: can't compare datetime.datetime to NoneType` so I jumped straight to checking its type - I can of course just check if it's truthy.

kitsune/users/tasks.py
LeoMcA: Ah cool, actually I realised my test wasn't resilient enough here and actually the behaviour was wrong (I needed to sort by "-last_modified" if I was going to select `last_event_q[0]`. So I've updated the test and I'll move the sorting to the model's Meta class.

kitsune/users/utils.py
```diff
@@ -71,6 +72,24 @@ def deactivate_user(user, moderator):
     deactivation.save()
 
 
+def anonymize_user(user):
```
akatsoulas, line 75: can you change also the view which deletes the user to use this method?

kitsune/users/models.py
```diff
@@ -430,35 +431,31 @@ class AccountEvent(models.Model):
...
+                                  default=None,
+                                  null=True,
+                                  blank=False)
```
akatsoulas, line 457: Default should be an empty string and blank should be True here.

kitsune/users/models.py
```diff
@@ -430,35 +431,31 @@ class AccountEvent(models.Model):
...
+    fxa_uid = models.CharField(blank=True, null=True, max_length=128)
```
akatsoulas, line 458: default this to an empty string and remove null

kitsune/users/models.py
akatsoulas: also pls remove null

PR #4351, opened 2020-04-28
wire up popular articles in new aaq

LeoMcA:
https://github.com/mozilla/sumo-project/issues/320﻿


kitsune/questions/views.py
```diff
@@ -630,7 +632,10 @@ def aaq(
             statsd.incr("questions.aaq.details-form")
         else:
             form = None
-            if search:
+            if product_config:
```
akatsoulas, line 635: is there any chance that we won't have a product_config? 

This is also referenced in the if statement above so maybe you could use a variable here for both places instead of `get`ting the product in both places.

kitsune/questions/views.py
```diff
@@ -604,6 +605,7 @@ def aaq(
             statsd.incr("questions.aaq.select-product")
 
     if request.method == "GET":
+        featured = []
```
akatsoulas, line 608: Instead of adding an empty list, you could tweak the `get_featured_articles` method to return an empty list if there aren't any articles

LeoMcA:
Closing this without merging, there's just too much we have to change in the backend to be atomic about it. Expect a mega-PR in the next few days.

PR #4353, opened 2020-04-29
remove steps of old aaq and wire up most of the new aaq

LeoMcA:
connect popular articles https://github.com/mozilla/sumo-project/issues/320
link "ask a question" button to form submission page https://github.com/mozilla/sumo-project/issues/322
display popular topics https://github.com/mozilla/sumo-project/issues/323
display progress bar on created question https://github.com/mozilla/sumo-project/issues/288
hook up product title and image on form https://github.com/mozilla/sumo-project/issues/327
rename form labels and add extra fields https://github.com/mozilla/sumo-project/issues/286

### DIFF
--- a/kitsune/products/jinja2/products/includes/topic_macros.html
+++ b/kitsune/products/jinja2/products/includes/topic_macros.html
@@ -1,9 +1,9 @@
-{% macro help_topics(topics, show_community_support=False) -%}
+{% macro help_topics(topics, show_community_support=False, product_slug=None) -%}
 
   <div class="sumo-card-grid" style="--cg-count: {{ topics|length }};">
     <div class="scroll-wrap">
       {% for topic in topics %}
-      {% set topic_url = url('products.documents', product_slug=product.slug, topic_slug=topic.slug) %}
+      {% set topic_url = url('products.documents', product_slug=product_slug or product.slug, topic_slug=topic.slug) %}
       <div class="card card--topic">
         <img
           class="card--icon"
--- a/kitsune/questions/forms.py
+++ b/kitsune/questions/forms.py
@@ -175,7 +175,7 @@ class EditQuestionForm(forms.Form):
         )
         self.fields["content"] = field
 
-        if isinstance(self, NewQuestionForm) and product and product["categories"]:
+        if isinstance(self, NewQuestionForm) and product and product.get("categories"):
             category_choices = [
                 (key, value["name"]) for key, value in product["categories"].items()
             ]
@@ -266,7 +266,7 @@ class EditQuestionForm(forms.Form):
     def metadata_field_keys(self):
         """Returns the keys of the metadata fields for the current
         form instance"""
-        non_metadata_fields = ["title", "content", "email"]
+        non_metadata_fields = ["title", "content", "email", "notifications"]
 
         def metadata_filter(x):
             return x not in non_metadata_fields
--- a/kitsune/questions/forms.py
+++ b/kitsune/questions/forms.py
@@ -36,8 +36,8 @@ FREQUENCY_CHOICES = [
     (u"EVERY_TIME", _lazy(u"Every time Firefox opened")),
 ]
 STARTED_LABEL = _lazy(u"This started when...")
-TITLE_LABEL = _lazy(u"Question")
-CONTENT_LABEL = _lazy(u"Details")
+TITLE_LABEL = _lazy(u"Subject")
+CONTENT_LABEL = _lazy(u"How can we help?")
 EMAIL_LABEL = _lazy(u"Email")
 EMAIL_HELP = _lazy(
     u"A confirmation email will be sent to this address in "
@@ -48,6 +48,8 @@ OS_LABEL = _lazy(u"Operating system")
 PLUGINS_LABEL = _lazy(u"Installed plugins")
 ADDON_LABEL = _lazy(u"Extension/plugin you are having trouble with")
 DEVICE_LABEL = _lazy(u"Mobile device")
+CATEGORY_LABEL = _lazy(u"Which topic best describes your question?")
+NOTIFICATIONS_LABEL = _lazy(u"Email me when someone answers the thread")
 
 # Validation error messages
 MSG_TITLE_REQUIRED = _lazy(u"Please provide a question.")
@@ -130,7 +132,7 @@ DEVELOPER_REQUEST_CATEGORY_CHOICES = [
 class EditQuestionForm(forms.Form):
     """Form to edit an existing question"""
 
-    def __init__(self, product=None, category=None, *args, **kwargs):
+    def __init__(self, product=None, *args, **kwargs):
         """Init the form.
 
         We are adding fields here and not declaratively because the
@@ -143,8 +145,6 @@ class EditQuestionForm(forms.Form):
 
         if product:
             extra_fields += product.get("extra_fields", [])
-        if category:
-            extra_fields += category.get("extra_fields", [])
 
         #  Add the fields to the form
         title_error_messages = {
@@ -174,6 +174,17 @@ class EditQuestionForm(forms.Form):
             error_messages=content_error_messages,
         )
         self.fields["content"] = field
+
+        if isinstance(self, NewQuestionForm) and product and product["categories"]:
+            category_choices = [
+                (key, value["name"]) for key, value in product["categories"].items()
+            ]
+            category_field = forms.ChoiceField(
+                label=CATEGORY_LABEL,
+                choices=[("", "Please select")] + category_choices,
+                required=True
+            )
+            self.fields["category"] = category_field
 
         if "sites_affected" in extra_fields:
             field = forms.CharField(
@@ -247,6 +258,10 @@ class EditQuestionForm(forms.Form):
             )
             self.fields["troubleshooting"] = field
 
+        for field in self.fields.values():
+            if not field.required:
+                field.label_suffix = _lazy(" (optional):")
+
     @property
     def metadata_field_keys(self):
         """Returns the keys of the metadata fields for the current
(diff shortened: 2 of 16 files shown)


kitsune/questions/forms.py
```diff
@@ -309,6 +327,12 @@ def __init__(self, product=None, category=None, *args, **kwargs):
             widget=forms.HiddenInput(), required=False
         )
 
+        self.fields["notifications"] = forms.BooleanField(
```
akatsoulas, line 330: Can you give me some context about this field?

kitsune/questions/forms.py
```diff
@@ -161,6 +162,19 @@ def __init__(self, product=None, *args, **kwargs):
         )
         self.fields["title"] = title_field
 
+        if isinstance(self, NewQuestionForm) and product and product.get("categories"):
+            category_choices = [
+                (key, value["name"]) for key, value in product["categories"].items()
+            ]
+            category_field = forms.ChoiceField(
+                label=CATEGORY_LABEL,
+                choices=[("", "Please select")] + category_choices,
```
akatsoulas, line 171: You can `insert` the `[("", "Please select")]` right above in the `category_choices` list comprehension for better readability

kitsune/questions/forms.py
```diff
@@ -175,7 +175,7 @@ def __init__(self, product=None, *args, **kwargs):
         )
         self.fields["content"] = field
 
-        if isinstance(self, NewQuestionForm) and product and product["categories"]:
+        if isinstance(self, NewQuestionForm) and product and product.get("categories"):
```
akatsoulas: since you are using the categories below too, let's assign a var

kitsune/questions/forms.py
```diff
@@ -36,8 +36,8 @@
     (u"EVERY_TIME", _lazy(u"Every time Firefox opened")),
 ]
 STARTED_LABEL = _lazy(u"This started when...")
-TITLE_LABEL = _lazy(u"Question")
-CONTENT_LABEL = _lazy(u"Details")
+TITLE_LABEL = _lazy(u"Subject")
```
akatsoulas, line 39: Will this affect existing questions?

kitsune/questions/views.py
```diff
@@ -1734,138 +1677,6 @@ def screen_share(request, question_id):
     )
 
 
-def _search_suggestions(request, text, locale, product_slugs):
```
akatsoulas, line 1737: Let's leave this code for now with a comment to adapt it or merge it with featured articles.

kitsune/questions/jinja2/questions/question_details.html
```diff
@@ -30,7 +30,9 @@
 
 {% block breadcrumbs %}
 
-{# {{ progress_bar(4) }} #}
+{% if aaq_final_step %}
+  {{ progress_bar(4) }}
```
akatsoulas, line 34: similar comment for the number as before. It would be nice if we could use something like `QUESTION_SUBMITTED` or similar

kitsune/questions/jinja2/questions/includes/question_editing_frame.html
```diff
@@ -126,21 +116,25 @@ <h2>Do some stuff.</h2>
...
+                        {{ field|safe }}
+                      {% endif %}
                     {% endif %}
 
                     {% if field.help_text and field.name != 'troubleshooting' %}
                       <p>{{ field.help_text|safe }}</p>
                     {% endif %}
 
-                    {% if field.name == 'title' %}
+                    {% if field.name != 'content' %}
```
akatsoulas, line 137: This looks weird, excluding just content for errors. What's the desired behavior here?

kitsune/products/jinja2/products/includes/topic_macros.html
```diff
@@ -1,9 +1,9 @@
-{% macro help_topics(topics, show_community_support=False) -%}
+{% macro help_topics(topics, show_community_support=False, product_slug=None) -%}
```
akatsoulas, line 1: Why are we changing this to accept `product_slug`s too?

kitsune/questions/jinja2/questions/includes/aaq_macros.html
```diff
@@ -38,25 +40,11 @@ <h3 class="card--title">
   </div>
 {%- endmacro %}
 
-{% macro selected_product(product, can_change) -%}
-  <div id="selected-product">
-    <br />
-    <p>{{ _('Product') }}: <strong><span>{{ product.name }}</span></strong>
-    {% if can_change %}
-      <a class="text-body-sm"
-         href="{{ url('questions.aaq_step1') }}{% if is_mobile_device %}?q=change_product{% endif %}">
```
akatsoulas, line 47: If this is not needed anymore let's also remove the check from the view. Without this we had an infinite loop in the old desgin.

kitsune/questions/forms.py
LeoMcA: It doesn't seem to

kitsune/questions/forms.py
```diff
@@ -161,6 +162,19 @@ def __init__(self, product=None, category=None, *args, **kwargs):
         )
         self.fields["title"] = title_field
 
+        if isinstance(self, NewQuestionForm) and product and product.get("categories"):
```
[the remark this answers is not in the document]
LeoMcA: I couldn't find a way to insert this field between the title and content fields without including it in this class

kitsune/questions/views.py
```diff
@@ -477,12 +482,11 @@ def edit_details(request, question_id):
...
-    request, product_key=None, category_key=None, showform=False, template=None, step=1
+    request, product_key=None, category_key=None, step=1
 ):
     """Ask a new question."""
 
-    if not template:
-        template = "questions/new_question.html"
+    template = "questions/new_question.html"
 
     # This tells our LogoutDeactivatedUsersMiddleware not to
```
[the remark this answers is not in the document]
LeoMcA: Yes, in the various `render`s below

kitsune/questions/views.py
```diff
@@ -524,11 +528,14 @@ def aaq(
     if product_key and not product_config:
         raise Http404
 
-    if product_config and "product" in product_config:
+    deadend = product_config.get("deadend", False) if product_config else False
+    html = product_config.get("html") if product_config else None
+
+    if step > 1 and not deadend:
```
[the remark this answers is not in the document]
LeoMcA: I find numbers clearer than names, but that might just be me.

> Shouldn't this more specific, eg step == 2?

No, we need the product object in step 3 as well.

kitsune/questions/views.py
```diff
@@ -548,93 +555,44 @@ def aaq(
...
+            statsd.incr("questions.aaq.explore-solutions")
+            if not deadend:
+                featured = get_featured_articles(product)
+                topics = topics_for(product)
+        elif step == 3:
             form = NewQuestionForm(
                 product=product_config,
-                category=category_config,
-                initial={"title": search},
+                initial={"category": category_key},
```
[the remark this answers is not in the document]
LeoMcA: Yes, to maintain backwards compatibility with the old urls. E.g. `/questions/new/desktop/cookies` will pre-populate "Cookies" in the drop-down field.

kitsune/questions/views.py
```diff
@@ -677,28 +635,29 @@ def aaq(
             # fully IA-driven (sync isn't special case anymore).
             question.add_metadata(product=product_config["key"])
 
-        if category_config:
-            # TODO: This add_metadata call should be removed once we are
-            # fully IA-driven (sync isn't special case anymore).
-            question.add_metadata(category=category_config["key"])
```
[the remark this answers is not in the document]
LeoMcA: Since `category` is now a form field, it's added to the question metadata a few lines above with `question.add_metadata(**form.cleaned_metadata)`

kitsune/questions/views.py
LeoMcA: Why? Featured articles isn't ES driven, is it? And if we really need this code again we can find it in the git history.

kitsune/questions/jinja2/questions/includes/question_editing_frame.html
LeoMcA: content handles its own errors in the `content_editor` macro

kitsune/products/jinja2/products/includes/topic_macros.html
LeoMcA: So we can use it in the aaq flow - in the aaq flow the slug isn't stored as `product.slug` as this macro assumes, but `product.product`, so we need to pass the slug explicitly rather than the whole "product" object.

kitsune/questions/forms.py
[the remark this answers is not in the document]
akatsoulas: That's fine but the population of this field should happen in the `NewQuestionForm` and not in the `EditQuestionForm`.

kitsune/questions/jinja2/questions/includes/aaq_macros.html
LeoMcA: I forgot to test this, this logic should be moved into the progress bar to ensure we don't get users stuck in a loop. But it also seems like the auto-selection on mobile is broken, so I'll have to look into that.

kitsune/questions/models.py
```diff
@@ -154,7 +154,7 @@ def clear_cached_tags(self):
     def clear_cached_contributors(self):
         cache.delete(self.contributors_cache_key % self.id)
 
-    def save(self, update=False, *args, **kwargs):
+    def save(self, update=False, notifications=True, *args, **kwargs):
```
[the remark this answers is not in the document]
LeoMcA: It can't happen in forms.py because the question object doesn't exist yet there, but it can be moved to the aaq view

kitsune/questions/views.py
```diff
@@ -548,93 +555,44 @@ def aaq(
...
-        html = product_config.get("html") if product_config else None
-        articles = None
-        if product_config:
-            # User is on the select category step
-            statsd.incr("questions.aaq.select-category")
-        else:
-            # User is on the select product step
-            statsd.incr("questions.aaq.select-product")
-
     if request.method == "GET":
```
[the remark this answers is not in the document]
LeoMcA: We'd then have to wrap all of that code in a `if request.method == "POST"` conditional, I'm not sure what we gain by reordering this?

I'll happy extract the form post into a separate view, but was trying to make as few changes to the current architecture as possible.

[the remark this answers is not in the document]
akatsoulas: you can just do if form.is_valid() instead of checking for POST which already exists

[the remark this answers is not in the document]
LeoMcA: I don't think I understand. We already do an `if form.is_valid()`, but the implicit else covers the "POSTed form invalid, display error messages" case. So we'd still need to add a check to cover the GET case.

PR #4357, opened 2020-05-05
Allow searching for multiple sub-strings.

akatsoulas:
* Add configuration in k8s.


k8s/regions/frankfurt/prod.yaml
```diff
@@ -136,3 +136,4 @@ app:
     disable_questions_list_all: False
     disable_feeds: False
     sentry_dsn: SECRET
+    sumo_banner_string: "firefox-account,firefox-monitor,account-data,firefox-private-network"
```
LeoMcA, line 139: no newline

PR #4358, opened 2020-05-06
refactor aaq to use modelforms

LeoMcA:


kitsune/questions/forms.py
```diff
@@ -315,36 +299,13 @@ def __init__(self, product=None, *args, **kwargs):
...
-
-        question = Question(
-            creator=user,
-            title=self.cleaned_data["title"],
-            content=self.cleaned_data["content"],
-            locale=locale,
-            product=product
-        )
+    def save(self, user, locale, product, product_config, *args, **kwargs):
+        question = super(NewQuestionForm, self).save(commit=False, *args, **kwargs)
```
akatsoulas, line 305: Why are you using `commit=False` here?

kitsune/questions/views.py
```diff
@@ -556,82 +573,64 @@ def aaq(
...
-                "topics": topics
-            },
+    form = None
+    images = None
+    featured = []
+    topics = []
+
+    if step == 2:
+        featured = get_featured_articles(product)
+        topics = topics_for(product, parent=None)
```
akatsoulas, line 583: Why don't you query here the db to get the topics for this product?

kitsune/questions/forms.py
LeoMcA: Otherwise it errors because the creator_id column is blank

kitsune/questions/views.py
LeoMcA: it seemed easier to use this helper function

kitsune/questions/forms.py
akatsoulas: Wouldn't something like this work:
```python
self.instance.creator = user
super(NewQuestionForm, self).save()
```

LeoMcA: That's clever, update incoming

kitsune/questions/views.py
```diff
@@ -556,93 +573,72 @@ def aaq(
...
         request,
         template,
-        {
-            "form": form,
-            "images": images,
-            "products": config.products,
-            "current_product": product_config,
-            "current_step": step,
-        },
+        context
```
akatsoulas, line 641: this fits now in a single line

PR #4359, opened 2020-05-06
move login to just before step 3 of aaq

LeoMcA:
https://github.com/mozilla/sumo-project/issues/408

### DIFF
--- a/kitsune/users/auth.py
+++ b/kitsune/users/auth.py
@@ -48,7 +48,7 @@ class FXAAuthBackend(OIDCAuthenticationBackend):
         user = super(FXAAuthBackend, self).create_user(claims)
         # Create a user profile for the user and populate it with data from
         # Firefox Accounts
-        profile, _ = Profile.objects.get_or_create(user=user)
+        profile = Profile.objects.get_or_create(user=user)[0]
         profile.is_fxa_migrated = True
         profile.fxa_uid = claims.get('uid')
         profile.fxa_avatar = claims.get('avatar', '')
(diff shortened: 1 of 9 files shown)


kitsune/users/views.py
```diff
@@ -44,9 +45,13 @@ def user_auth(request, notification=None):
...
     next_url = get_next_url(request) or reverse('home')
 
+    aaq = re.match(r"^/[\w\-]+/questions/new/([\w\-]+)/[\w\-]+$", next_url)
+    aaq_product = aaq.group(1) if aaq else None
+
     return render(request, 'users/auth.html', {
         'next_url': next_url,
-        'notification': notification
+        'notification': notification,
+        'aaq_product': aaq_product
```
akatsoulas, line 54: How about removing the regex and adding something like 
```python
{
...
...
'in_aaq_flow': True if request.session['in-aaq'] or False
```
Probably the in_aaq_flow is not a good name but rather an example. Wouldn't that work? It feels to me that's much simpler than the regex and probably more robust since we don't depend on the next_url

kitsune/users/auth.py
```diff
@@ -68,9 +68,22 @@ def create_user(self, claims):
         profile.products.clear()
         profile.products.add(*products)
 
-        # This is a new sumo profile, redirect to the edit profile page
-        self.request.session['oidc_login_next'] = reverse('users.edit_my_profile')
-        messages.info(self.request, 'fxa_notification_created')
+        # This is a new sumo profile, show edit profile message
+        messages.success(
```
akatsoulas, line 72: I am not sure if we need to show this on the form page or it would have been better to amend it when the user has posted the question. If the user clicks any of the links in the message then basically we are getting the user out of the flow. Another solution could be to remove completely this part (if the messaging when posting a question for a new user is too much) for new users when in flow.

kitsune/users/auth.py
```diff
@@ -48,7 +48,7 @@ def create_user(self, claims):
         user = super(FXAAuthBackend, self).create_user(claims)
         # Create a user profile for the user and populate it with data from
         # Firefox Accounts
-        profile, _ = Profile.objects.get_or_create(user=user)
+        profile = Profile.objects.get_or_create(user=user)[0]
```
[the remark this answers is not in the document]
LeoMcA: So we don't overwrite the `import ugettext as _`

kitsune/users/views.py
LeoMcA: That cookie won't get set if a user goes directly to the `/questions/new/desktop/form` url (for instance), which we'll eventually be linking to from around the site with the "ask the community" widget we want to add. Also the cookie doesn't tell us what product the user wants to aaq about (which we need to go direct to the form).

The regex should be robust since it's a near-direct copy of what we're using in urls.py

kitsune/users/auth.py
LeoMcA: We could _blank the links, to make it clearer the user can come back to the AAQ flow.

kitsune/users/auth.py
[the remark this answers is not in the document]
akatsoulas: You can use then `profile, created = Profile.objects....`

kitsune/users/views.py
akatsoulas: The regex might be, the next_url though is not. Ideally the entry point for everything should be only the `aaq` view which will handle all the flow. Having multiple entry points for the aaq is prone to errors

kitsune/users/auth.py
akatsoulas: In that case though you also alter the behavior for new users not in the AAQ flow.

kitsune/users/views.py
LeoMcA: But the `aaq` view isn't called, because step 3 is wrapped in the login_required decorator. I don't see why next_url isn't robust? It's the only way users will get back to the aaq flow after login, if it's not set they'll end up on the homepage.

PR #4365, opened 2020-05-07
Remove other products from the AAQ.

akatsoulas:
Change button copy.


LeoMcA:
Oh actually, you can also remove the `render_deadend` macro which we wouldn't be using anywhere any more.

PR #4366, opened 2020-05-08
Hide Sign In/Up when logging in

akatsoulas:


kitsune/sumo/jinja2/includes/common_macros.html
```diff
@@ -145,72 +145,72 @@ <h4 class="mzp-c-menu-item-title">{{ _('Tools') }}</h4>
...
+                  {{ _('Sign Out') }}
+                </a>
+            </div>
+          </div>
         </div>
       </div>
     </div>
-  </div>
-  {% else %}
+  {% elif not user.is_authenticated and not request.path == url('users.auth') %}
```
LeoMcA, line 213: isn't `not user.is_authenticated` redundant since our original check is `if user.is_authenticated()`

kitsune/sumo/jinja2/includes/common_macros.html
```diff
@@ -145,72 +145,72 @@ <h4 class="mzp-c-menu-item-title">{{ _('Tools') }}</h4>
...
-             </a>
-           </section>
-          </div>
+    <a class="mzp-c-menu-title sumo-nav--link has-avatar hide-on-mobile" href="{{ url('users.auth') }}">
+      <span class="sumo-nav--username">{{ display_name(user) }}</span>
+      {% if user.is_authenticated() %}
+      <img class="avatar" src="{{ profile_avatar(user) }}" alt="Avatar for Username">
+      {% else %}
+      <img class="avatar" src="{{ STATIC_URL }}sumo/img/default-FFA-avatar.png" alt="Avatar for Username">
+      {% endif %}
```
LeoMcA, line 154: While we're here: I don't see why we need this if, it'll always be shown, the else will never we shown

PR #4367, opened 2020-05-08
Conditionally change the login copy.

akatsoulas:


kitsune/users/jinja2/users/includes/login-fxa.html
```diff
@@ -2,10 +2,14 @@
   <div class="sumo-page-section">
     <h1 class="sumo-page-heading">{{ _('Log in / Sign up with Firefox Accounts') }}</h1>
     <p>
-      {{ _('We upgraded our log-in system to replace Mozilla Support accounts with Firefox Accounts. Firefox Accounts
-      offers one log-in for convenient and secure access to Mozilla services.') }}
-      <a href="/kb/firefox-accounts-mozilla-support-faq"> {{ _('See our FAQ') }}</a>
-      {{ _('for more information') }}.
+      {% if request.session.get('in-aaq') %}
```
LeoMcA, line 5: I assume this is all in a separate file to auth.html for historical reasons, but I don't see any need now: auth.html is the only other template to reference this one.

Even without merging it with auth.html, we should still use the same `aaq_product` variable as we use there.

kitsune/sumo/middleware.py
```diff
@@ -278,3 +277,26 @@ def process_request(self, request):
...
+    def process_view(self, request, callback, callback_args, callback_kwargs):
+        # This key is used both in templates and the
+        # LogoutDeactivateUsersMiddleware
+
+        view_name = callback.func_name
+        # If we are authenticating or there is no session, do nothing
+        if (view_name in ['user_auth', 'login', 'serve_cors'] or not
+                request.session):
+            return None
+        elif 'aaq' in view_name:
```
LeoMcA, line 298: nit: `elif` isn't strictly necessary after a return, could just be `if`

kitsune/sumo/middleware.py
```diff
@@ -278,3 +277,30 @@ def process_request(self, request):
...
+    Middleware that updates session's keys based on the view used.
+
+    aaq_* views -> in-aaq = True
+    """
+
+    def process_view(self, request, callback, callback_args, callback_kwargs):
+        # This key is used both in templates and the
+        # LogoutDeactivateUsersMiddleware
+
+        if not request.session or callback:
```
LeoMcA, line 293: I assume this is a typo, fixed here: https://github.com/mozilla/kitsune/pull/4371/files#diff-6f05e342c77cc4322987a6ff3c3edd64R293

PR #4374, opened 2020-05-12
Do not cache for local dev.

akatsoulas:


akatsoulas:
This can wait - let's first release AAQ.

akatsoulas:
@LeoMcA this should be good to go

kitsune/settings.py
```diff
@@ -94,6 +94,11 @@ def parse_conn_max_age(value):
     },
 }
 
+if  DEV and DEBUG:
```
LeoMcA, line 97: I think there are two spaces between `if` and `DEV`

PR #4381, opened 2020-05-15
vscode setup improvements:

LeoMcA:
* add debug config
* don't launch kitsune on docker startup so it can't crash vscode
* add npm script for running browsersync/asset compilation in docker container
* expose browsersync ports


README.md
```diff
@@ -96,3 +96,33 @@ First, make sure you have run the "Create some data" step above.
 1. Enter the web container: `docker-compose exec web bash`
 2. Build the indicies: `./manage.py esreindex` (You may need to pass the `--delete` flag)
 3. Precompile the nunjucks templates: `./manage.py nunjucks_precompile`
+
+## Development in vscode (beta)
```
akatsoulas, line 100: I don't think that we should include any editor specific instructions in the main README file. I would be fine with that if we could provide instructions for more editors. Also this workflow may not be for anyone working with vscode. Let's move this section to a readme in the vscode dir.

requirements/dev.txt
```diff
@@ -227,3 +227,42 @@ pylint-plugin-utils==0.6 \
 parameterized==0.7.1 \
     --hash=sha256:6a94dbea30c6abde99fd4c2f2042c1bf7f980e48908bf92ead62394f93cf57ed \
     --hash=sha256:ea0326ba5bbbe7c427329a27b75003410df07d1173ca254976f8f5a64922c322
+debugpy==1.0.0b9 \
```
akatsoulas, line 230: Let's also remove ptvsd from the file.

README.md
```diff
@@ -96,3 +96,33 @@ First, make sure you have run the "Create some data" step above.
...
+
+Install [vscode](https://code.visualstudio.com/) and the [Remote Development](https://marketplace.visualstudio.com/items?itemName=ms-vscode-remote.vscode-remote-extensionpack) extension.
```
akatsoulas, line 102: Is it possible to add the extension in the vscode dockerfile directly?

README.md
```diff
@@ -96,3 +96,33 @@ First, make sure you have run the "Create some data" step above.
...
+```
+
+In the vscode command palette (Ctrl+Shift+P) build and launch the container:
+```
+>Remote-Containers: Rebuild and Reopen in Container
+```
+
+Once vscode has reopened in the container, open a terminal (View > Terminal) and run the initial configuration:
+```
+./bin/run-bootstrap.sh
```
akatsoulas, line 116: Can we remove this step and automate it?

akatsoulas:
Thinking about this, how would you feel about adding another entry that runs the app through gunicorn?

README.md
LeoMcA: Makes sense :+1:

README.md
LeoMcA: I don't think so, because this extension needs to run on the "host" vscode instance.

PR #4382, opened 2020-05-18
add guide to querying the celery queue length

LeoMcA:


k8s/README.md
```diff
@@ -201,3 +201,52 @@ KUBECTL_BIN=/home/metadave/kubectl1.6.4 invoke -f ./regions/oregon-b/dev.yaml de
...
+In [4]: r.llen('celery')
+Out[4]: 44
+```
+
+The last item in the queue can be fetched with:
+```
+l = r.lrange('celery', -1, -1)
+```
+
+This must then be parsed and de-pickled:
```
akatsoulas, line 248: Let's add a comment here as a TODO to update/remove this entry with the Django/Python milestone that is coming up. We will switch to the default json instead of pickle.

LeoMcA: I'll add a catch all for this whole section, because I think we'll have to change the `./manage.py celery` command above to just `celery` as well?

PR #4386, opened 2020-05-19
Exclude Featured articles section from some products.

akatsoulas:


kitsune/products/jinja2/products/product.html
```diff
@@ -154,7 +154,7 @@ <h2 class="sumo-page-subheading">{{ _('Frequent Topics') }}</h2>
 
   {{ topic_metadata(topics, show_community_support=product.questions_enabled(request.LANGUAGE_CODE)) }}
 
-{% if not product.slug == 'thunderbird' %}
+{% if not product.slug in excluded_products %}
```
LeoMcA, line 157: Why hard code this? Why not `{% if featured %}` and wrap the topics with an `{% if topics %}` above, and do the same in aaq?

PR #4390, opened 2020-05-21
fix geoip-suggestion message bar

LeoMcA:
* update to reflect responsive redesign
* don't flash empty bar
* use bn as suggested locale now that bn-BD and bn-IN have been merged https://discourse.mozilla.org/t/44554
* ensure bar remains hidden once closed

Screenshot_2020-05-21 Firefox Help - 127 0 0 1

This PR should reinstate the prior behaviour of this bar. That behaviour was:

* When a user is visiting the en-US version of the site:
  * if they're in Indonesia, prompt them to switch to the Indonesian version of the site
  * if they're in Bangladesh, prompt them to switch to the Bengali version of the site
* Once a user has acted on, or dismissed the notification, don't show it again

The rationale behind this behaviour is explained here: https://bugzilla.mozilla.org/show_bug.cgi?id=1032825

## My big question:

The reasoning from the bug for just including Indonesia and Bangladesh seems to be that they have a high number of users using en-US Firefox bouncing, and so could be wanting to see localised content. Do we have metrics for the bounce rate per country when using en-US Firefox? Do we know if this is true for any other countries?

What if we just expand this to *all* countries with supported locales in SUMO? It means we don't have to keep up with what countries have a high proportion of en-US users who might want to see localised content, and for those users who do want to see English content - they'll only see the banner once: once it's dismissed it'll never show again (on the same profile).

We already have an endpoint to fetch all supported locales in SUMO, so I feel like expanding to all countries and locales is a bit of a free lunch.

### DIFF
--- a/kitsune/sumo/jinja2/base.html
+++ b/kitsune/sumo/jinja2/base.html
@@ -125,7 +125,7 @@
 {# end TrafficCop #}
 
 <div id="announcements">
-  {% call announcement_bar('geoip-suggestion', 'warning') %}{% endcall %}
+  {% call announcement_bar('geoip-suggestion', 'warning', close_memory="remember", close_type="remove") %}{% endcall %}
 
   {% for announ in get_announcements() %}
       {% call announcement_bar(announ.id, 'warn') %}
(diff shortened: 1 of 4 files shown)


akatsoulas:
After discussing this with @LeoMcA, we decided to initially fix the bug in this PR. All the valid points that are raised here will be tracked in a separate issue. Nice work :100:

PR #4391, opened 2020-05-21
Remove zendesk and marketplace.

akatsoulas:


kitsune/questions/models.py
```diff
@@ -882,15 +865,6 @@ def extract_document(cls, obj_id, obj=None):
 )
 
 
-def _tag_added(sender, question_id, tag_name, **kwargs):
-    """Signal handler for new tag on question."""
-    if tag_name == config.ESCALATE_TAG_NAME:
-        escalate_question.delay(question_id)
-
-
-tag_added.connect(_tag_added, sender=Question, dispatch_uid="tagged_1337")
```
LeoMcA, line 891: From what I can see, this is the only place the `tag_added` signal is processed, so we could also remove that from `kitsune/questions/views.py` and `kitsune/questions/signals.py` if wanted.

PR #4395, opened 2020-05-27
use localised versions of featured articles

LeoMcA:
https://github.com/mozilla/sumo-project/issues/393

### DIFF
--- a/kitsune/wiki/utils.py
+++ b/kitsune/wiki/utils.py
@@ -1,14 +1,14 @@
 import random
-import requests
 
+import requests
 from django.conf import settings
 from django.contrib.auth.models import User
-from django.db.models import Q
+from django.db.models import Q, Prefetch
 from django.utils.http import urlencode
 
 from kitsune.dashboards import LAST_7_DAYS
 from kitsune.dashboards.models import WikiDocumentVisits
-from kitsune.wiki.models import Revision
+from kitsune.wiki.models import Document, Revision
 
 
 class BitlyException(Exception):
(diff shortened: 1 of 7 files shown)


kitsune/wiki/utils.py
```diff
@@ -133,28 +133,44 @@ def _active_contributors_id(from_date, to_date, locale, product):
...
-    if featured.count() < 10:
-        return featured
+    if locale == settings.WIKI_DEFAULT_LANGUAGE:
+        for visit in visits:
+            documents.append(visit.document)
+            if len(documents) == 10:
+                break
+    else:
+        # prefretch localised documents to avoid n+1 problem
+        visits = visits.prefetch_related(
```
akatsoulas, line 162: What happens if there isn't a translation for the selected document? Shall we look for the next article in popularity to check for translations? Probably this should be continued until either we have a list or we exhaust our top ten. In that case we can either hide the articles for that locale or query for the next 10

LeoMcA: As written the `documents` array will be filled with the top 10 documents in any locale, ordered by their popularity in english.

Thinking about his over lunch, however, the prefetch query is going to be an absolute monster on the prod db, so we should limit the original query to 10 items. Testing it against the staging db I see a 100x speedup. I'll make this change and update the PR.

> In that case we can either hide the articles for that locale or query for the next 10

We only end up displaying 4 out of the 10, and I think it's highly unlikely that any locale won't have at least 4 of the 10 most popular articles in English localised. If they don't, then I'm not sure it's any more likely that the next 10 would... and then we're going to end up with a very expensive set of DB queries.

kitsune/wiki/utils.py
```diff
@@ -133,28 +133,41 @@ def _active_contributors_id(from_date, to_date, locale, product):
...
+                continue
+            documents.append(translation)
 
-    # This returns a WikiDocumentVisits object and not the actual Wiki doc.
-    # The wiki doc it's under .document which is prefetched b/c of select_related
-    featured = featured[:10]
-    return random.sample(featured, 4)
+    if len(documents) <= 4:
+        return documents
+    return random.sample(documents, 4)
```
akatsoulas, line 173: Let's add a check in the templates to hide completely featured articles (if it doesn't exist already) in case there aren't any.

PR #4403, opened 2020-06-01
Fxa live events

LeoMcA:
https://github.com/mozilla/kitsune/pull/4402 squashed, rebased against master, and with some small additional commits


kitsune/users/middleware.py
```diff
@@ -17,3 +19,22 @@ def process_request(self, request):
...
+
+class LogoutInvalidatedSessionsMiddleware(MiddlewareMixin):
+    """Logs out any sessions started before a user changed their
+    Firefox Accounts password.
+    """
+    def process_request(self, request):
+
+        user = request.user
+
+        if user.is_authenticated():
```
akatsoulas, line 32: Let's add a TODO here for python3 which will be merged right after this. We already have an issue to track the TODOs

kitsune/sumo/middleware.py
```diff
@@ -78,6 +80,13 @@ class LocaleURLMiddleware(object):
     """
 
     def process_request(self, request):
+        try:
```
akatsoulas, line 83: Can we remove the try/except from here? If ti's not possible let's rewrite this to remove some logic out of try clause.  How about using request.path?

PR #4408, opened 2020-06-03
update pre-commit config

LeoMcA:
We should wait until the py3 upgrade is merged into master to merge this, and then run `pre-commit run --all-files` in a separate commit to update all our files.

A few notes:

* Added prettier so we can have opinionated formatting of pretty much all our files, excluding `*.html` files as prettier doesn't understand jinja2 yet. We might want to add a specific jinja2 linter/formatter.
* Disabled eslint for the time being, as we need to dedupe some of the rules from what prettier is doing, and also fix quite a number of js files manually.
* Made flake8 ignore line length (E501) - black won't split up strings so they can be wrapped, but otherwise will enforce line length.


akatsoulas:
I would love to have a good jinja2 formatter but I am not aware of anything that can work here.

.pre-commit-config.yaml
```diff
@@ -1,29 +1,38 @@
...
+          - id: black
+    - repo: https://gitlab.com/pycqa/flake8.git
+      rev: 3.8.2
+      hooks:
+          - id: flake8
+    - repo: https://github.com/prettier/prettier
+      rev: 2.0.5
+      hooks:
+          - id: prettier
+    # - repo: git://github.com/pre-commit/mirrors-eslint
```
akatsoulas, line 29: Let's drop these lines if are not needed

pyproject.toml
```diff
@@ -0,0 +1,3 @@
+[tool.black]
```
akatsoulas, line 1: Let's exclude migrations for now

.pre-commit-config.yaml
```diff
@@ -1,29 +1,38 @@
...
+      hooks:
+          - id: black
+    - repo: https://gitlab.com/pycqa/flake8.git
+      rev: 3.8.2
+      hooks:
+          - id: flake8
+    - repo: https://github.com/prettier/prettier
+      rev: 2.0.5
+      hooks:
+          - id: prettier
```
akatsoulas, line 28: We should also add a configuration for prettier. Eg two spaces for CSS and JS files, single vs double quotes etc.

.pre-commit-config.yaml
LeoMcA: Ideally we would keep eslint support, but we'll need to fix a whole load of errors first.

.pre-commit-config.yaml
LeoMcA: prettier will obey our .editorconfig file, so I've updated that

pyproject.toml
LeoMcA: I've excluded them globally from being touched by any pre-commit hook

LeoMcA:
This will need to be redone when approved and rebased, but I wanted to add a flavour of what our pre-commit hooks will be doing, so ran the following snippet to add a commit for each stage:

```
for x in $(cat .pre-commit-config.yaml | grep "id: " | sed 's/^.*: //'); pre-commit run --all-files $x || git commit -am "ran $x"
```

PR #4411, opened 2020-06-05
Celery json serializer

akatsoulas:


kitsune/upload/tasks.py
```diff
@@ -1,23 +1,21 @@
...
 from PIL import Image
 
-
 log = logging.getLogger('k.task')
 
 
 @task(rate_limit='15/m')
 def generate_thumbnail(for_obj, from_field, to_field,
-                       max_size=settings.THUMBNAIL_SIZE):
+                       max_size=settings.THUMBNAIL_SIZE, serializer="pickle"):
```
LeoMcA, line 18: Fair enough, could we pass a path to the image instead?

kitsune/upload/tasks.py
```diff
@@ -100,7 +98,7 @@ def _scale_dimensions(width, height, longest_side=settings.THUMBNAIL_SIZE):
 
 
 @task(rate_limit='15/m')
-def compress_image(for_obj, for_field):
+def compress_image(for_obj, for_field, serializer="pickle"):
```
LeoMcA, line 101: Same q as above

kitsune/customercare/tasks.py
```diff
@@ -1,15 +1,22 @@
...
 
 
 @task()
-def maybe_award_badge(badge_template, year, user):
+def maybe_award_badge(badge_template: Dict, year: int, user_id: int):
     """Award the specific badge to the user if they've earned it."""
+    try:
+        user = User.objects.get(id=user_id)
+    except User.DoesNotExist:
+        return
```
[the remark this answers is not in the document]
akatsoulas: Maybe something in between like fail silently but log the error/send it to sentry.

kitsune/upload/tasks.py
akatsoulas: Worth investigating

kitsune/customercare/tasks.py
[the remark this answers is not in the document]
LeoMcA: Yeah sending the error to sentry would be good

PR #4412, opened 2020-06-05
Remove statsd from the project.

akatsoulas:


kitsune/sumo/tasks.py
```diff
@@ -1,18 +1,15 @@
...
-    """A task that measures the time it was sitting in the queue.
-
-    It saves the data to graphite via statsd.
-    """
+    """A task that measures the time it was sitting in the queue."""
     lag = datetime.now() - queued_time
-    lag = (lag.days * 3600 * 24) + lag.seconds
-    statsd.gauge('celery.lag', max(lag, 0))
+    lag = max((lag.days * 3600 * 24) + lag.seconds, 0)
+    log.debug(f'Measuse queue lag task value is {lag}')
```
LeoMcA, line 15: typo

LeoMcA: Also, shouldn't we use `log.info` here so we see it appear in logs?

And we can just use `timedelta.total_seconds()` now instead of doing the maths ourselves: https://docs.python.org/3/library/datetime.html#datetime.timedelta.total_seconds

kitsune/wiki/tasks.py
```diff
@@ -239,8 +227,8 @@ def _rebuild_kb_chunk(data):
         if message:
             log.debug(message)
             messages.append(message)
-    d = time.time() - start
-    statsd.timing('wiki.rebuild_chunk', int(round(d * 1000)))
+    elapsed_time = int(round((time.time() - start) * 1000))
+    log.debug(f'wiki rebuild task chuck time: {elapsed_time}')
```
LeoMcA, line 231: We don't need this, celery times our tasks for us, and the durations can be seen quite easily through flower.

PR #4415, opened 2020-06-10
add priority to celery queue

LeoMcA:
Slightly confusingly `0` is the highest priority, and `9` is the lowest.

`CELERY_WORKER_PREFETCH_MULTIPLIER` is reduced to 1, so a load of lower priority tasks won't be fetched and clog up the queue for higher priority tasks.

I placed some settings in celery.py as opposed to settings.py as those struck me as much more integral to the behaviour of the application (if we change the default, or inheriting from parents, then we need to rewrite the priorities we assign tasks), than that in settings.py which is more of a configurable performance setting.

We should set the priority for the fxa event tasks once py3 is merged into master.

Confusingly, the docs indicate that some of these settings only work for RabbitMQ, when through testing I've determined that they work for Redis too:
https://docs.celeryproject.org/en/stable/userguide/routing.html#rabbitmq-message-priorities

### DIFF
--- a/kitsune/settings.py
+++ b/kitsune/settings.py
@@ -920,6 +920,7 @@ if not CELERY_TASK_ALWAYS_EAGER:
 CELERY_WORKER_CONCURRENCY = config('CELERY_WORKER_CONCURRENCY', default=4, cast=int)
 CELERY_TASK_EAGER_PROPAGATES = config('CELERY_TASK_EAGER_PROPAGATES', default=True, cast=bool)  # Explode loudly during tests.
 CELERY_WORKER_HIJACK_ROOT_LOGGER = config('CELERY_WORKER_HIJACK_ROOT_LOGGER', default=False, cast=bool)
+CELERY_WORKER_PREFETCH_MULTIPLIER = config('CELERY_WORKER_PREFETCH_MULTIPLIER', default=1, cast=int)
 
 # Wiki rebuild settings
 WIKI_REBUILD_TOKEN = 'sumo:wiki:full-rebuild'
--- a/bin/run-celery-worker.sh
+++ b/bin/run-celery-worker.sh
@@ -1,3 +1,15 @@
 #!/bin/bash
 
-exec newrelic-admin run-program celery -A kitsune worker --maxtasksperchild=${CELERY_WORKER_MAX_TASKS_PER_CHILD:-25}
+exec newrelic-admin run-program celery -A kitsune worker -n default@%h --maxtasksperchild=${CELERY_WORKER_MAX_TASKS_PER_CHILD:-25} &
+status=$?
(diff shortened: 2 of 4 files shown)


kitsune/wiki/tasks.py
```diff
@@ -166,7 +166,7 @@ def add_short_links(doc_ids):
         unpin_this_thread()
 
 
-@task(rate_limit='3/h')
+@task(rate_limit='3/h', priority=9)
```
akatsoulas, line 169: This task shouldn't have such a low priority

akatsoulas:
My proposal would be to add a different queue that will be used only by the SET related tasks and leave everything else as is. This way we won't have any tasks flooding the SET queue

kitsune/wiki/tasks.py
LeoMcA: From anecdotal evidence when inspecting the queue on dev, this seems to be the task that clogs everything else up. But I'll remove the priority for now and we can reconsider when we have proper numbers coming through from flower.

LeoMcA:
Cool, converted to draft as we'll need to wait until after master and py3 have been merged.

Also included the test tasks I was running which show me that we're able to have both separate queues and priorities, which I'll obviously remove prior to merging.

kitsune/settings.py
```diff
@@ -920,6 +920,7 @@ def JINJA_CONFIG():
 CELERY_WORKER_CONCURRENCY = config('CELERY_WORKER_CONCURRENCY', default=4, cast=int)
 CELERY_TASK_EAGER_PROPAGATES = config('CELERY_TASK_EAGER_PROPAGATES', default=True, cast=bool)  # Explode loudly during tests.
 CELERY_WORKER_HIJACK_ROOT_LOGGER = config('CELERY_WORKER_HIJACK_ROOT_LOGGER', default=False, cast=bool)
+CELERY_WORKER_PREFETCH_MULTIPLIER = config('CELERY_WORKER_PREFETCH_MULTIPLIER', default=1, cast=int)
```
akatsoulas, line 923: What's the rationale behind this? Usually we don't have tasks that consume too much time. Have some prefetching should help (eg the default value). Especially for this fxa queue where tasks are small, limiting it to `1` probably won't help.

LeoMcA: This was to ensure high priority tasks won't be blocked by long-running prefetched low-priority tasks, maybe less necessary now that we have a separate queue for fxa tasks, however we still have those very long-running rebuild_chunk tasks. Tuning is probably necessary here, which is why I added it as an env var.

akatsoulas: Actually this is just a config. I would suggest to other remove it completely or change the default to the default celery value (that should 4 I believe)

PR #4416, opened 2020-06-10
fix comparison between NoneType and int

LeoMcA:
Fixes https://github.com/mozilla/sumo-project/issues/471, which triggered these sentry alerts: https://sentry.prod.mozaws.net/operations/sumo-dev-sp/issues/8679656/?query=is%3Aunresolved

Also fixes these sentry alerts: https://sentry.prod.mozaws.net/operations/sumo-dev-sp/issues/8694466/?query=is%3Aunresolved

### DIFF
--- a/kitsune/dashboards/readouts.py
+++ b/kitsune/dashboards/readouts.py
@@ -500,7 +500,7 @@ class Readout(object):
             rows = self.rows(max_rows)
 
         # Compute percents for bar widths:
-        max_visits = max(r['visits'] for r in rows) if rows else 0
+        max_visits = max(r['visits'] or 0 for r in rows) if rows else 0
         for r in rows:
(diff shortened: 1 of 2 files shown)


kitsune/dashboards/readouts.py
```diff
@@ -500,7 +500,7 @@ def render(self, max_rows=None, rows=None):
             rows = self.rows(max_rows)
 
         # Compute percents for bar widths:
-        max_visits = max(r['visits'] for r in rows) if rows else 0
```
akatsoulas, line 503: Are we saying here that row['visits'] is None?

LeoMcA: Yeah, some rows are a dict like `{ ... visits: None, ... }`

PR #4417, opened 2020-06-12
add blocklist for questions and replies

LeoMcA:


kitsune/settings.py
```diff
@@ -640,6 +640,11 @@ def _username_algo(email):
 )
 
 USERNAME_BLACKLIST = path('kitsune', 'configs', 'username-blacklist.txt')
+QUESTION_BLOCKLIST = config(
+    'QUESTION_BLOCKLIST',
+    default=None,
```
akatsoulas, line 645: default to an empty string.

kitsune/settings.py
```diff
@@ -640,6 +640,11 @@ def _username_algo(email):
...
+    cast=lambda x: re.compile(x) if x else None
```
akatsoulas, line 646: let's remove this from the configuration

k8s/regions/frankfurt/dev.yaml
```diff
@@ -144,3 +144,4 @@ app:
     oidc_rp_client_id: SECRET
     oidc_rp_client_secret: SECRET
     sumo_banner_string: "firefox-account,firefox-monitor,account-data,firefox-private-network"
+    question_blocklist: "1[^0-9]*8[^0-9]*3[^0-9]*3[^0-9]*6[^0-9]*0[^0-9]*5[^0-9]*1[^0-9]*0[^0-9]*1[^0-9]*0"
```
akatsoulas, line 147: Let's move this into secrets

akatsoulas: Shouldn't blocklist be about specific values?

kitsune/questions/utils.py
```diff
@@ -1,9 +1,12 @@
 import logging
 import re
 
+from django.conf import settings
+
 from kitsune.questions.models import Answer, Question
 
 log = logging.getLogger('k.questions')
+block_regex = re.compile(settings.QUESTION_BLOCK_REGEX) if settings.QUESTION_BLOCK_REGEX else None
```
akatsoulas, line 9: Shouldn't this be evaluated to False if settings is an empty string? Otherwise we return None instead of False in the `in_blocklist`. It will work but probably False is better.

LeoMcA: Ah yes. To fix this I changed the in_blocklist function to be a bit more explicit and return True/False, instead of Match/None

PR #4420, opened 2020-06-16
All toll free numbers

akatsoulas:


kitsune/questions/views.py
```diff
@@ -583,6 +585,9 @@ def aaq(request, product_key=None, category_key=None, step=1):
 
         if form.is_valid() and not is_ratelimited(request, "aaq-day", "5/d"):
             if in_blocklist(form.cleaned_data["content"]):
+                if switch_is_active('spam_deactivate_user'):
```
LeoMcA, line 588: "blocklist_deactivates_users" might be a more specific term

Do we also need to set a default here, or some kind of migration? I'm not familiar with waffle

kitsune/settings.py
```diff
@@ -640,7 +640,6 @@ def _username_algo(email):
 )
 
 USERNAME_BLACKLIST = path('kitsune', 'configs', 'username-blacklist.txt')
-QUESTION_BLOCK_REGEX = config('QUESTION_BLOCK_REGEX', default="")
```
LeoMcA, line 643: Shall we also remove this from the .yaml files?

kitsune/questions/views.py
akatsoulas: If waffle doesn't find that switch returns False.

PR #4422, opened 2020-06-18
Disallow links in the AAQ forum.

akatsoulas:


kitsune/settings.py
```diff
@@ -1196,3 +1196,17 @@ def filter_exceptions(event, hint):
 
 # Substring to match in slug in order to display the SUMO CTA banner
 SUMO_BANNER_STRING = config('SUMO_BANNER_STRING', default='', cast=Csv())
+
+# List of domains that links are allowed
+ALLOW_LINKS_FROM = [
```
akatsoulas, line 1201: This is not used atm

kitsune/sumo/utils.py
```diff
@@ -313,3 +313,22 @@ def get_browser(user_agent):
...
+        return True
+    return False
+
+@timeout(seconds=settings.REGEX_TIMEOUT)
+def match_regex_with_timeout(compiled_regex, data):
+    """Matches the specified regex.
+
+    Adds a timeout to avoid catastrophic backtracking.
+    """
+    return any(compiled_regex.findall(data))
```
LeoMcA, line 334: I think `compiled_regex.search(data)` will be marginally faster since it'll return on the first match

kitsune/sumo/forms.py
```diff
@@ -0,0 +1,51 @@
...
+        if not cdata:
+            return super(KitsuneBaseForumForm, self).clean(*args, **kwargs)
+
+        # At the moment validation doesn't reach this point b/c the logout and deactivation
+        is_spam = is_toll_free_number(cdata)
+
+        if is_spam:
+            self.cleaned_data.update({
+                'is_spam': True
+            })
```
LeoMcA, line 35: Since this won't start working without removing the existing behaviour, should we just comment it out/remove it for now?

kitsune/sumo/utils.py
LeoMcA: I guess we'll need findall anyway for the allowlist, if we implement that

kitsune/questions/views.py
```diff
@@ -1210,16 +1196,17 @@ def edit_answer(request, question_id, answer_id):
             request, "questions/edit_answer.html", {"form": form, "answer": answer}
         )
 
-    form = AnswerForm(request.POST)
+    form = AnswerForm(request.POST, **{'user': request.user})
```
LeoMcA, line 1199: Is there a reason for unpacking the dict instead of using keyword args? I'm genuinely curious

AliceWyman:
> 
> 
> It's not pretty, but I don't think it's going to break anything, and should tide us over the weekend.

Reported bug:
https://bugzilla.mozilla.org/show_bug.cgi?id=1647091  Bug 1647091 Opened 2020-06-20 14:16 EDT Updated 2020-06-21 15:14 EDT
error "Links are not allowed in the forums" when no links are being posted
(quote)  It seems any dot with a non-space character on both sides has been blocked, including:
    preference names (e.g., keyword.enabled)
    file names (e.g. logins.json)
    dotted user name when quoting a previous poster (e.g., jesse.james)

PR #4424, opened 2020-06-22
Tune toll free regex.

akatsoulas:


LeoMcA:
Do we have an authoritative source for this?

akatsoulas:
https://en.wikipedia.org/wiki/North_American_Numbering_Plan

PR #4425, opened 2020-06-22
automatically place suspected spam in moderation queue, and hide it

LeoMcA:


kitsune/questions/views.py
```diff
@@ -580,6 +574,9 @@ def aaq(request, product_key=None, category_key=None, step=1):
                 product_config=product_config,
             )
 
+            if form.cleaned_data.get('is_spam'):
+                _save_and_mark_as_spam(request, question)
```
akatsoulas, line 578: I would remove the `save` part from here and keep only the spam handling. Moreover that should be renamed to something else as it essentially moves the reply to the moderation queue. It doesn't mark the reply as spam.

kitsune/questions/views.py
```diff
@@ -1580,3 +1568,23 @@ def _add_tag(request, question_id):
...
+        object_id=object.id,
+        reason='spam',
+        notes="Automatically flagged at creation and marked as spam. \
+            If it's not spam, please reactivate the user's account \
+            and click 'Unmark as spam' on the post to make it publicly visible.",
+        creator=request.user,
+    )
+    flag.save()
+
+    if switch_is_active('blocklist_deactivates_users'):
```
akatsoulas, line 1588: Let's remove this completely from here. There is not point in deactivating users if we are to move replies into a moderation queue.

kitsune/questions/views.py
```diff
@@ -1580,3 +1568,23 @@ def _add_tag(request, question_id):
...
+
+def _save_and_mark_as_spam(request, object):
+    object.is_spam = True
+    object.save()
+
+    flag = FlaggedObject(
+        content_type=ContentType.objects.get_for_model(object),
+        object_id=object.id,
+        reason='spam',
+        notes="Automatically flagged at creation and marked as spam. \
```
akatsoulas, line 1581: This message should change to reflect the absence of the deactivation.

kitsune/questions/views.py
```diff
@@ -1553,3 +1567,24 @@ def _add_tag(request, question_id):
 def _init_watch_form(request, event_type="solution"):
     initial = {"event_type": event_type}
     return WatchQuestionForm(request.user, initial=initial)
+
+
+def _add_to_moderation_queue(request, object):
+    object.is_spam = True
```
akatsoulas, line 1573: The save operation should not be in this function but part of the view along with the rest options on the object.

LeoMcA: I have it here because the object needs to be saved before we save the `FlaggedObject` instance.

PR #4426, opened 2020-06-22
Add is_spam flag in cleaned data.

akatsoulas:


LeoMcA:
flake8 :)

kitsune/sumo/utils.py
```diff
@@ -315,21 +315,30 @@ def get_browser(user_agent):
...
-    if not digits:
-        return False
-
-    if settings.TOLL_FREE_REGEX.match(digits):
-        return True
-    return False
-
-
-@timeout(seconds=settings.REGEX_TIMEOUT)
+# @timeout(seconds=settings.REGEX_TIMEOUT)
```
LeoMcA, line 318: presumably this wasn't intentionally included

kitsune/sumo/utils.py
```diff
@@ -315,21 +315,30 @@ def get_browser(user_agent):
...
+
+    # keep only the digits in text
+    digits = filter(type(data).isdigit, data)
+    is_toll_free = settings.TOLL_FREE_REGEX.match(digits)
+
+    is_vanity_number = match_regex_with_timeout(settings.VANITY_PHONE_NUMBER_REGEX, data)
+
+    has_links = match_regex_with_timeout(settings.SIMPLE_DOMAIN_IP_REGEX, data)
+
+    return is_toll_free or is_vanity_number or has_links
```
LeoMcA, line 344: It would be great to return a truthy value with more information here, perhaps a dict like `{ 'is_toll_free': is_toll_free, 'is_vanity_number': ...` so that we can then surface that information in the review queue (like "This was flagged for having a toll free number") and do different things to users posting links and those posting numbers.

kitsune/sumo/forms.py
```diff
@@ -20,32 +20,18 @@ def __init__(self, *args, **kwargs):
     def clean(self, *args, **kwargs):
         """Generic clean method used by all forms in the question app.
 
-        Parse content for any toll free numbers.
+        Parse content for any toll free numbers or links.
```
LeoMcA, line 23: We might want to change this to "Parse content for spam"

kitsune/settings.py
```diff
@@ -1209,3 +1209,4 @@ def filter_exceptions(event, hint):
 TOLL_FREE_REGEX = re.compile(r'^.*8(00|33|44|55|66|77|88)[0-9]\d{6,}$')
 SIMPLE_DOMAIN_IP_REGEX = re.compile(r'(?:[^\s/]+\.[^\s0-9/.]{2,})|(?:\d{1,3}\.){3}\d{1,3}')
 REGEX_TIMEOUT = config('REGEX_TIMEOUT', default=5, cast=int)
+VANITY_PHONE_NUMBER_REGEX = re.compile(r'(\+?1?)-?([0-9]{3})-?([a-zA-Z0-9]{7,8})$')
```
LeoMcA, line 1212: Do we have an authoritative source for what form vanity numbers will take? Can we add some tests to tests against things that are and things that aren't vanity numbers?

LeoMcA: From https://en.wikipedia.org/wiki/North_American_Numbering_Plan#Alphabetic_mnemonic_system, I believe the regexp should be `[0-9]{3}-?[a-zA-Z2-9][a-zA-Z0-9]{6}`:
* removed unnecessary capturing groups
* no `(\+?1?)-?` is needed because it's all marked as optional (`?`)
* `[a-zA-Z2-9]` because from what I can see the "first digit in the NXX must be [2-9]" rule still applies
* `{6}` because I can't see any source which allows the xxxx to be longer than 4 digits
* no `$` because it could appear anywhere in a line of text

Also, this will match all vanity numbers, *and* all non-vanity ones in the NANP - should we just rename the regex to something like `NANP_REGEX`?

kitsune/settings.py
```diff
@@ -1209,3 +1209,4 @@ def filter_exceptions(event, hint):
 TOLL_FREE_REGEX = re.compile(r'^.*8(00|33|44|55|66|77|88)[0-9]\d{6,}$')
 SIMPLE_DOMAIN_IP_REGEX = re.compile(r'(?:[^\s/]+\.[^\s0-9/.]{2,})|(?:\d{1,3}\.){3}\d{1,3}')
 REGEX_TIMEOUT = config('REGEX_TIMEOUT', default=5, cast=int)
+VANITY_PHONE_NUMBER_REGEX = re.compile(r'[0-9]{3}-?[a-zA-Z2-9][a-zA-Z0-9]{2}-?[a-zA-Z0-9]{4}')
```
LeoMcA, line 1212: Strictly speaking this regex will match any well-formed number in the NANP, so perhaps this should be called "NANP_REGEX"

kitsune/sumo/forms.py
```diff
@@ -20,32 +17,22 @@ def __init__(self, *args, **kwargs):
     def clean(self, *args, **kwargs):
         """Generic clean method used by all forms in the question app.
 
-        Parse content for any toll free numbers.
+        Parse content for suspicious content.
+        - Toll free numbers
+        - Toll vanity numbers
```
LeoMcA, line 22: Again, strictly speaking we'll match obfuscated toll free numbers and any well formed number in the NANP

kitsune/sumo/forms.py
```diff
@@ -20,32 +17,22 @@ def __init__(self, *args, **kwargs):
...
-        if not self.user.has_perm('flagit.can_moderate'):
-            if switch_is_active('disallow_all_links'):
-                try:
-                    has_links = match_regex_with_timeout(settings.SIMPLE_DOMAIN_IP_REGEX, cdata)
-                except TimeoutError:
-                    has_links = True
-
-        if has_links:
-            raise forms.ValidationError('Links are not allowed in the forums.')
+        if not self.user.has_perm('flagit.can_moderate') and check_for_spam_content(cdata):
```
LeoMcA, line 33: Will we add the exclusion for trusted contributors in a later commit?

akatsoulas: added

PR #4427, opened 2020-06-22
add has_blocked_link util which matches any non-allowed valid domain

LeoMcA:


kelimuttu:
Does this mean that preferences and file extensions will still not be allowed?

LeoMcA:
> Does this mean that preferences and file extensions will still not be allowed?

Anything of the form `something.anything` will be allowed, provided `anything` isn't in the list of valid TLDs.

So for example:

"example.jpg" wouldn't be flagged, neither would "dom.ipc.processCount"

but "layout.display-list.retain.chrome" would be flagged, as ".chrome" is [a valid TLD](https://icannwiki.org/.chrome).

kitsune/sumo/utils.py
```diff
@@ -315,6 +319,42 @@ def get_browser(user_agent):
...
+        tld = match.group(1).upper()
+        if tld in VALID_TLDS:
+            full_domain = match.group(0).lower()
+            in_allowlist = False
+            for allowed_domain in settings.ALLOW_LINKS_FROM:
+                split = full_domain.rsplit(allowed_domain, 1)
+                if len(split) != 2:
+                    # allowed_domain isn't in full_domain, or something went wrong
+                    continue
+                if split[-1] != '':
```
akatsoulas, line 334: you can merge the conditionals which have the `continue` statement together with the `or` statement

kitsune/sumo/utils.py
```diff
@@ -315,6 +319,42 @@ def get_browser(user_agent):
...
+                    # allowed_domain isn't at the end of full_domain
+                    continue
+                if split[0] == '':
+                    # allowed_domain equals full_domain
+                    in_allowlist = True
+                    break
```
akatsoulas, line 340: If you break then you will allow a post that maybe has more links in it that are not whitelisted.

kitsune/sumo/utils.py
```diff
@@ -315,6 +319,42 @@ def get_browser(user_agent):
...
 
 
+def has_blocked_link(data):
+    # TODO: deal with punycode when presented in non-ascii form
+    for match in POTENTIAL_LINK_REGEX.finditer(data):
+        tld = match.group(1).upper()
+        if tld in VALID_TLDS:
+            full_domain = match.group(0).lower()
+            in_allowlist = False
+            for allowed_domain in settings.ALLOW_LINKS_FROM:
```
akatsoulas, line 329: couldn't just test here `if tld in settings.ALLOW_LINKS_FROM`?

kitsune/lib/tlds.py
```diff
@@ -0,0 +1,6 @@
+import os
+
+path = os.path.join(os.path.dirname(__file__), 'tlds-alpha-by-domain.txt')
```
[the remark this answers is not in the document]
LeoMcA: Looks like it's updated every day: today https://data.iana.org/TLD/tlds-alpha-by-domain.txt has "Last Updated Wed Jun 24 07:07:02 2020 UTC" however no new TLDs were added.

kitsune/sumo/utils.py
```diff
@@ -13,9 +13,13 @@
 from ratelimit.utils import is_ratelimited as rl_is_ratelimited
 from timeout_decorator import timeout
 
+from kitsune.lib.tlds import VALID_TLDS
 from kitsune.journal.models import Record
 from kitsune.sumo import paginator
 
+POTENTIAL_LINK_REGEX = re.compile(r'[^\s/]+\.([^\s/.]{2,})')
```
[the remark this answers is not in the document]
LeoMcA: This isn't a setting though: its format is inherently tied to the behaviour of `has_blocked_link` - if the capturing groups change, for instance, the function will behave differently.

kitsune/sumo/utils.py
LeoMcA: This breaks the `for allowed_domain in settings.ALLOW_LINKS_FROM` loop, not the `for match in POTENTIAL_LINK_REGEX.finditer(data)` one

This testcase covers the situation you're describing:
https://github.com/mozilla/kitsune/blob/2e66162243caeb1594db463c215f9c7c27f4351c/kitsune/sumo/tests/test_utils.py#L230

kitsune/sumo/utils.py
LeoMcA: tld will be a value like "COM", so doing that would only allow us to whitelist top level domains, not domains at any level

kitsune/lib/tlds.py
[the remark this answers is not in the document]
akatsoulas: How do we intend to update that file (daily it's too much) but we should update it from time to time

kitsune/sumo/utils.py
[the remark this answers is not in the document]
akatsoulas: but how is the function related to the regex? My proposal is to mainly group all regexes related to spam together. Marking this as nit

kitsune/sumo/utils.py
```diff
@@ -328,18 +328,13 @@ def has_blocked_link(data):
             in_allowlist = False
             for allowed_domain in settings.ALLOW_LINKS_FROM:
                 split = full_domain.rsplit(allowed_domain, 1)
-                if len(split) != 2:
+                if len(split) != 2 or split[-1] != "":
```
akatsoulas, line 331: nit: `or split[-1]`

kitsune/sumo/utils.py
```diff
@@ -328,18 +328,13 @@ def has_blocked_link(data):
...
                     # allowed_domain isn't in full_domain, or something went wrong
+                    # or allowed_domain isn't at the end of full_domain
                     continue
-                if split[-1] != '':
-                    # allowed_domain isn't at the end of full_domain
-                    continue
-                if split[0] == '':
+                if split[0] == "" or split[0][-1] == ".":
```
akatsoulas, line 335: nit `not split[0]`

kitsune/lib/tlds.py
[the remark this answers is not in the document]
LeoMcA: I considered including it as part of the build process, but then we have no way to manually validate it in case the format changes. I guess it'll cause the tests to fail if it does, so that might work.

kitsune/sumo/utils.py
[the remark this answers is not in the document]
LeoMcA: Because the function's behaviour relies on the regex being in this exact form, with this exact capturing group. If the capturing groups were changed, the function would cease to work. In my mind if something's in settings.py then it should be configurable, and the configurable aspect of this function is the allowlist, not the regexp.

PR #4430, opened 2020-06-25
Remove spam flag if moderated.

akatsoulas:


kitsune/questions/views.py
```diff
@@ -10,7 +10,7 @@
 from django.contrib.sites.models import Site
 from django.core.exceptions import PermissionDenied
 from django.core.paginator import EmptyPage, PageNotAnInteger
-from django.db.models import Q
+from django.db.models import Q, Subquery
```
LeoMcA, line 13: flake8 is gonna complain about an unused import

PR #4433, opened 2020-06-25
Check spam status before firing a notification.

akatsoulas:
* Do not save objects in the moderation queue.


LeoMcA:
I don't think this will work - we need the question or answer object to exist in db before we save the `FlaggedObject` object.

PR #4435, opened 2020-06-26
add mailcatcher container and docs

LeoMcA:


docker-compose.yml
```diff
@@ -138,5 +138,10 @@ services:
   redis:
     image: redis:3
 
+  mailcatcher:
```
akatsoulas, line 141: Will this brought up with `make run`? If not let's add it there too

LeoMcA: Do we want it to come up by default? Since using it requires you to customise your environment variables, I'd leave it as optional.

akatsoulas: Let's also add these vars with some default values in the env file

PR #4436, opened 2020-06-29
Py3 upgrade

akatsoulas:


Dockerfile
```diff
@@ -21,8 +21,8 @@ ENV PYTHONDONTWRITEBYTECODE=1
 ENV PYTHONUNBUFFERED=1
 ENV PATH="/venv/bin:$PATH"
 
-RUN pip install --upgrade pip"==20.1.1"
-RUN virtualenv /venv
+RUN pip install --upgrade pip"==20.0.1"
```
LeoMcA, line 24: Was the downgrade intentional?

PR #4441, opened 2020-07-01
update pre-commit and add read-only hooks

LeoMcA:


setup.cfg
```diff
@@ -1,9 +1,19 @@
 [flake8]
-exclude = *migrations*,
-    kitsune/settings*,
-    kitsune/sumo/db_strings.py,
+exclude =
```
akatsoulas, line 2: This break is just ugly :)

LeoMcA: We'll need to find a formatter for our *.cfg files then ;)

PR #4449, opened 2020-07-08
fix empty bio in review queue

LeoMcA:


kitsune/sumo/templatetags/jinja_helpers.py
```diff
@@ -122,6 +122,8 @@ def urlparams(url_, hash=None, query_dict=None, **query):
 @library.filter
 def wiki_to_html(wiki_markup, locale=settings.WIKI_DEFAULT_LANGUAGE, nofollow=True):
     """Wiki Markup -> HTML jinja2.Markup object"""
+    if not wiki_markup:
```
akatsoulas, line 125: How about passing an optional arg in the function definition `wiki_markup=""`?

LeoMcA: Ah, actually, the helper gets a `None` value like if one were to call `wiki_to_html(None)` rather than `wiki_to_html()`, so that won't work.

PR #4453, opened 2020-07-09
Remove traffic cop and FxA CTA experiment.

akatsoulas:


kitsune/bundles.py
```diff
@@ -243,15 +243,6 @@
         "source_filenames": ("d3/d3.js", "kpi/js/kpi.browserify.js",),
         "output_filename": "build/kpi.dashboard-min.js",
     },
-    "experiment_fxa_cta_topbar": {
-        "source_filenames": (
-            "sumo/js/libs/mozilla-dnt-helper.js",
```
LeoMcA, line 248: This library isn't used anywhere else and can be deleted, we use "kitsune/sumo/static/sumo/js/dnt-helper.js" for things like GA.

kitsune/bundles.py
```diff
@@ -243,15 +243,6 @@
...
-            "sumo/js/libs/mozilla-cookie-helper.js",
```
LeoMcA, line 249: Also not used elsewhere and can be removed.

PR #4455, opened 2020-07-10
update docs

LeoMcA:


.vscode/settings.json
```diff
@@ -9,5 +9,9 @@
         "--disable=C0111",
         "--enable=W0614",
         "--load-plugins=pylint_django"
-    ]
+    ],
+    "restructuredtext.confPath": "${workspaceFolder}/docs",
```
akatsoulas, line 13: Let's not mix vscode related configuration with the docs in the same commit.

contribute.json
```diff
@@ -19,7 +21,8 @@
     },
     "urls": {
         "prod": "https://support.mozilla.org/",
-        "stage": "https://support.allizom.org/"
+        "stage": "https://support.allizom.org/",
+        "dev": "https://support-dev.allizom.org/"
```
akatsoulas, line 25: Let's remove dev. I would treat that instance as ephemeral. Even the database is not similar to stage.

docs/sop.rst
```diff
@@ -1,31 +0,0 @@
-============================
-Standard Operating Procedure
```
akatsoulas, line 2: I would probably port a part of this section in the localization document. Eg why locales may not be updated in the site b/c of broken strings in pontoon.

docs/wikidocs.rst
```diff
@@ -1,19 +0,0 @@
-.. _wiki-chapter:
-
-========================
-Important Wiki Documents
-========================
```
akatsoulas, line 5: Why are we dropping this?

docs/wsgi.rst
```diff
@@ -1,193 +0,0 @@
...
-    media/uploads/gallery/images
-    media/uploads/gallery/images/thumbnails
-    media/uploads/gallery/videos
-    media/uploads/gallery/videos/thumbnails
-
-``media/uploads`` and its subdirectories should never be added to version
-control, as they are installation-/content-specific.
-
-
-Product Details JSON
```
akatsoulas, line 116: A section about this probably should stay.

docs/wsgi.rst
```diff
@@ -1,193 +0,0 @@
...
-
-    path/to/virtualenv/lib/python<version>/site=packages/django-mozilla-product-details/product_details/json
-
-This is configurable. If you have multiple web servers, they should share this
-data. You can set the ``PROD_DETAILS_DIR`` variable in
-``kitsune/settings_local.py`` to a different path, for example on NFS.
-
-
-Debugging
-=========
```
akatsoulas, line 137: Maybe we could write something about debugging sumo in your local env. Both with the vscode and through docker without the editor.

docs/wsgi.rst
```diff
@@ -1,193 +0,0 @@
...
-``wsgi/kitsune.wsgi``, tries to run all the initial validation that
-the dev server runs, to catch these errors early.
-
-So where *is* the stack trace? You'll need to look in your Apache error logs.
-Where these are is OS-dependent, but a good place to look is
-``/var/log/httpd``. If you are using SSL, also check the SSL ``VirtualHost``'s
-logs, for example ``/var/log/httpd/ssl_error_log``.
-
-
-With ``DEBUG=True``
```
akatsoulas, line 162: Should be mentioned in the debug section?

docs/wsgi.rst
```diff
@@ -1,193 +0,0 @@
-.. _wsgi-chapter:
-
-=============================
-Running Kitsune with mod_wsgi
-=============================
```
akatsoulas, line 5: Here we can mention the alternative of running sumo with gunicorn instead of the django dev server.

requirements/dev.txt
```diff
@@ -168,6 +169,10 @@ pre-commit==2.4.0 \
     --hash=sha256:5559e09afcac7808933951ffaf4ff9aac524f31efbc3f24d021540b6c579813c \
     --hash=sha256:703e2e34cbe0eedb0d319eff9f7b83e2022bb5a3ab5289a6a8841441076514d0 \
     # via -r dev.in
+prometheus-client==0.8.0 \
```
akatsoulas, line 172: this is an interesting entry.

.adr-dir
```diff
@@ -1 +0,0 @@
-docs/architecture/decisions
```
akatsoulas, line 1: I was thinking about this in the past but I never got to the point to update the documentation. I believe there is some value in adding a few entries here with each milestone that we are working on. For example, the indexing rationale in the new ES architecture could be a great fit. The future work about the API would also be a good candidate. I guess what I am saying is that there is some value in keeping this section and start using it. Thoughts.

docs/wikidocs.rst
LeoMcA: The doc points to (and is about) https://support.mozilla.org/en-US/kb/managing-landing-page-content, none of which applies any more (from what I can see).

.adr-dir
LeoMcA: Sure, if we think we're going to use it, let's leave it in

docs/wsgi.rst
LeoMcA: It sounds like there's actually some snippets in here we might want to keep? I'm not sure I have the brain space to keep+rewrite this page until Friday, so I'm tempted to un-delete and slap the "this may be outdated" warning on it.

requirements/dev.txt
LeoMcA: Looks like it's pulled in from the flower 0.9.4 -> 0.9.5 update

README.md
```diff
@@ -11,88 +11,3 @@ See [what's deployed](https://whatsdeployed.io/s-J18)
 
 ## Code of Conduct
 By participating in this project, you're agreeing to uphold the [Mozilla Community Participation Guidelines](https://www.mozilla.org/en-US/about/governance/policies/participation/). If you need to report a problem, please see our [CODE_OF_CONDUCT.md](./CODE_OF_CONDUCT.md) guide.
-
-## Development
```
akatsoulas, line 15: My suggestion would be to leave this in the README for ease of use. Alternatively we should have a link to the relevant part of the documentation.

contribute.json
```diff
@@ -8,9 +8,11 @@
     },
     "participate": {
         "home": "https://kitsune.readthedocs.io/en/latest/contributors.html",
-        "docs": "https://kitsune.readthedocs.io/en/latest/hacking_howto.html#hacking-howto-chapter",
-        "irc": "irc://irc.mozilla.org/#sumodev",
-        "irc-contacts": ["giorgos", "pmac", "bensternthal"]
+        "docs": "https://kitsune.readthedocs.io/en/latest/hacking_howto.html",
+        "matrix": "https://chat.mozilla.org/#/room/#sumodev:mozilla.org",
+        "matrix-contacts": ["tasos", "leo"],
+        "slack": "https://mozilla.slack.com/messages/sumodev",
```
akatsoulas, line 14: Let's either change this to `support-platform`

docs/contactus.rst
```diff
@@ -15,24 +15,15 @@ for SUMO community discussions.
 Kitsune hackers
 ===============
 
-If you're hacking on the Kitsune code and have questions, ping us on
-IRC.
+If you're hacking on the Kitsune code and have questions, ping us on `Matrix <https://wiki.mozilla.org/Matrix>`_.
 
-We hang out in ``#sumodev`` on irc.mozilla.org. We're usually around
-during the day in Eastern Time.
+We hang out in `#sumodev:mozilla.org <https://chat.mozilla.org/#/room/#sumodev:mozilla.org>`_.
```
akatsoulas, line 20: Similar let's change this to support-platform. I would remove any reference to timezone. We should have 24h coverage to at least acknowledge a request/question.

docs/contactus.rst
```diff
@@ -15,24 +15,15 @@ for SUMO community discussions.
...
+We're usually around during the day in Europe.
 
 If you ask something and all you get is silence, then it's probably
 the case that we're not around. Please try pinging us again.
 
-You can also use the `dev-sumo mailing list
-<https://www.mozilla.org/about/forums/#dev-sumo>`_.
+Current developers:
```
akatsoulas, line 26: Let's link to AUTHORS and update that if needed.

docs/deployments.rst
```diff
@@ -40,23 +34,33 @@ such as changes that require re-indexing. See
...
+Servers
+=======
+
+
+Development
+===========
+
+https://support-dev.allizom.org/
+
+We use dev primarily to develop infrastructure changes, like upgrading to Python 3.8.
```
akatsoulas, line 52: I would remove Python 3.8. It will be obsolete in some time

docs/deployments.rst
```diff
@@ -40,23 +34,33 @@ such as changes that require re-indexing. See
 We deploy to production from ``master``.
 
 
-Dev
-===
+Deploying
+=========
```
akatsoulas, line 38: We can also mention here that we are currently using K8s and maybe something about conventions similar to what we documented in the processes document.

docs/contributors.rst
```diff
@@ -28,19 +28,11 @@ Help with hacking!
 ==================
 
 First step is to set up Kitsune so you can run it and hack on it. For
-that, see the :ref:`hacking-howto-chapter`.
```
akatsoulas, line 31: It would be probably helpful to have a section on where to file bugs for this project and a link to Bugzilla.

docs/sla.rst
```diff
@@ -1,30 +0,0 @@
-=======================
```
akatsoulas, line 1: I would keep this. Let's add an outdated note for now and update it with the current SLA.

docs/users.rst
```diff
@@ -1,15 +0,0 @@
-=====
```
akatsoulas, line 1: Let's keep it and add a few entries about FxA here and links to the KB articles.

requirements/default.txt
```diff
@@ -29,13 +29,13 @@ bleach==2.1.4 \
     --hash=sha256:0ee95f6167129859c5dce9b1ca291ebdb5d8cd7e382ca0e237dfd0dad63f63d8 \
     --hash=sha256:24754b9a7d530bf30ce7cbc805bc6cce785660b4a10ff3a43633728438c105ab \
```
akatsoulas, line 30: Let's move all the changes related to packages to a separate commit

docs/contactus.rst
LeoMcA: AUTHORS got [renamed to CONTRIBUTORS](https://github.com/mozilla/kitsune/commit/2ba4b1573bd97f1e077980ac6f114fbef4e80c1e#diff-98f07bc20cb66328be238119df96c490) which was then later replaced with this: https://github.com/mozilla/kitsune/blob/master/CONTRIBUTORS.rst

So I would say CONTRIBUTORS serves a different purpose from this, CONTRIBUTORS is about attribution, this is about who a contributor with questions should contact.

PR #4461, opened 2020-07-14
Update naming schema in AAQ.

akatsoulas:


rtanglao:
bugzilla bug is here: https://bugzilla.mozilla.org/show_bug.cgi?id=1650231

PR #4462, opened 2020-07-14
Update naming schema in AAQ.

akatsoulas:


kitsune/questions/config.py
```diff
@@ -438,8 +438,8 @@
         ])
     }),
     ('firefox-private-network-vpn', {
-        'name': _lazy('Firefox Private Network Device-level Protection'),
-        'subtitle': _lazy('Firefox Private Network Device-level Protection'),
+        'name': _lazy(u'Mozilla VPN'),
+        'subtitle': _lazy(u'Mozilla VPN'),
```
LeoMcA, line 442: Presumably these should be non-unicode strings?

PR #4470, opened 2020-07-21
Update readonly announcement.

akatsoulas:


kitsune/sumo/jinja2/includes/readonly_annoucement.html
```diff
@@ -1,9 +1,9 @@
 {% trans
-    irc_channel="https://client02.chat.mibbit.com/?server=irc.mozilla.org&channel=%23sumo",
+    matrix_channel="https://chat.mozilla.org/#/room/#support-platform:mozilla.org",
```
LeoMcA, line 2: Given the wording of this announcement (we have our support community waiting to help you [...] live on Matrix), I'm not sure #support-platform is the right channel. #sumo's description is "This is room for Support Mozilla (SUMO) Contributors https://support.mozilla.org/get-involved", so I'm not sure that's the right room either. For this release maybe we should just remove the "or live on" clause, and re-add Matrix if there is an appropriate channel later.

akatsoulas:
@LeoMcA pr updated

PR #4473, opened 2020-07-22
Safeguard against non existing parameters.

akatsoulas:
* Add .ssh in gitignore

### DIFF
--- a/kitsune/questions/views.py
+++ b/kitsune/questions/views.py
@@ -243,7 +243,8 @@ def question_list(request, product_slug):
     question_qs = question_qs.filter(locale_query)
 
     # Set the order.
-    order_by = ORDER_BY[order][0]
+    # Set a default value if a user requested a non existing order parameter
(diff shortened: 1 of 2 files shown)


kitsune/questions/views.py
```diff
@@ -244,7 +244,8 @@ def question_list(request, product_slug):
     question_qs = question_qs.filter(locale_query)
 
     # Set the order.
-    order_by = ORDER_BY[order][0]
+    # Set a default value if a user requested a non existing order parameter
+    order_by = ORDER_BY.get(order, 'updated')
```
LeoMcA, line 248: This needs to be something like `ORDER_BY.get(order, ['updated'])[0]` otherwise it breaks when `order` exists in the dict.

akatsoulas:
@LeoMcA pr updated

LeoMcA:
oh! hold on, black would reformat the single quotes, can you rebase and run black please?

PR #4474, opened 2020-07-22
Use black formatter for access and announcements apps.

akatsoulas:


kitsune/announcements/models.py
```diff
@@ -31,10 +31,7 @@ class Announcement(ModelBase):
     )
     content = models.TextField(
         max_length=10000,
-        help_text=(
-            "Use wiki syntax or HTML. It will display similar to a "
-            "document's content."
-        ),
+        help_text=("Use wiki syntax or HTML. It will display similar to a " "document's content."),
```
LeoMcA, line 34: This doesn't need to be two string literals

akatsoulas:
Added the reviewed and updated commit in pr #4475

PR #4479, opened 2020-07-24
Install pylint in dev requirements.

akatsoulas:
* Update vscode extensions.


Dockerfile
```diff
@@ -49,7 +49,7 @@ ENV GIT_SHA=${GIT_SHA}
 FROM base AS base-dev
 RUN apt-get update && apt-get install apt-transport-https && \
     curl -sL https://deb.nodesource.com/setup_12.x | bash -
-RUN apt-get update && apt-get install -y --no-install-recommends optipng nodejs && \
+RUN apt-get update && apt-get install -y --no-install-recommends optipng nodejs black && \
```
LeoMcA, line 52: Is there a reason we're installing black through apt? We already have it in our dev.txt, so it should get installed by pip already (and will be the version we want instead of whatever horribly outdated version debian carries).

PR #4481, opened 2020-07-24
Remove RabbitMQ functionality.

akatsoulas:
Safeguard against non existing parameters.


kitsune/sumo/views.py
```diff
@@ -283,20 +260,6 @@ def error(request):
     fu  # noqa
 
 
-@require_GET
-@never_cache
-def version_check(request):
```
LeoMcA, line 288: `VersionCheckTests` need to be removed from `kitsune/sumo/tests/test_views.py`

kitsune/sumo/views.py
```diff
@@ -283,20 +260,6 @@ def error(request):
...
-    content_type = "application/x-json"
-    token = settings.VERSION_CHECK_TOKEN
```
LeoMcA, line 290: `VERSION_CHECK_TOKEN` can be removed from settings.py

kitsune/sumo/urls.py
```diff
@@ -1,21 +1,14 @@
...
 
 from watchman import views as watchman_views
 
 from kitsune.sumo import views
 
 
-services_patterns = [
-    url("^/monitor$", views.monitor, name="sumo.monitor"),
-    url("^/version$", views.version_check, name="sumo.version"),
-    url("^/error$", views.error, name="sumo.error"),
```
LeoMcA, line 12: Should we remove `error` from `kitsune/sumo/views.py` then?

Also will need to remove `ForceErrorTests` from `kitsune/sumo/tests/test_views.py`

kitsune/sumo/urls.py
```diff
@@ -1,21 +1,14 @@
...
-]
-
-
 urlpatterns = [
     url(r"^robots.txt$", views.robots, name="robots.txt"),
-    url(r"^services/", include(services_patterns)),
+    url(r"^services/monitor", views.monitor, name="sumo.monitor"),
```
LeoMcA, line 11: We're missing a `$` at the end of the regexp

LeoMcA: Indeed the url isn't consistent with the old one, which was `example.com/services//monitor`.

Do we actually use this? Can we remove it too?

akatsoulas: Hopefully we can augment the functionality and have some more consistent reporting. I will update the `$`. The url is not consistent because of the leading slash. This will remove 3 warnings out of the many we have about leading slashes

kitsune/sumo/views.py
```diff
@@ -120,183 +112,6 @@ def test_memcached(host, port):
...
-    for component, output in list(status.items()):
-        if ERROR in [item[0] for item in output]:
-            status_code = 500
-            status_summary[component] = False
-        else:
-            status_summary[component] = True
-
-    return render(
-        request,
-        "services/monitor.html",
```
LeoMcA, line 272: `kitsune/sumo/jinja2/services/monitor.html` can be removed, and I think `kitsune/sumo/static/sumo/css/monitor.css` can also be removed

PR #4487, opened 2020-07-28
Editable username

akatsoulas:


kitsune/users/jinja2/users/edit_profile.html
```diff
@@ -4,126 +4,130 @@
 {% set title = _('Edit Profile') %}
 {% set classes = 'profile' %}
 {% set active = 'edit-profile' %}
-{% set crumbs = [(profile_url(profile.user), profile.user.username), (None, title)] %}
```
LeoMcA, line 7: Is there a reason behind removing the breadcrumbs?

akatsoulas: just missed that - I need to uncomment the commented out section

kitsune/users/models.py
```diff
@@ -72,14 +73,14 @@ class Profile(ModelBase, SearchMixin):
         validators=[TwitterValidator],
         verbose_name=_lazy("Twitter Username"),
     )
-    facebook = models.URLField(
-        max_length=255, null=True, blank=True, verbose_name=_lazy("Facebook URL")
+    community_mozilla_org = models.CharField(
+        max_length=255, null=True, blank=True, verbose_name=_lazy("Mozilla Community Username")
```
LeoMcA, line 77: Do we want `null=True` here and below?

From https://docs.djangoproject.com/en/2.2/ref/models/fields/#null:
> Avoid using null on string-based fields such as CharField and TextField. If a string-based field has null=True, that means it has two possible values for “no data”: NULL, and the empty string. In most cases, it’s redundant to have two possible values for “no data;” the Django convention is to use the empty string, not NULL.

kitsune/users/models.py
```diff
@@ -72,14 +73,14 @@ class Profile(ModelBase, SearchMixin):
...
     )
-    mozillians = models.CharField(
-        max_length=255, null=True, blank=True, verbose_name=_lazy("Mozillians Username")
+    people_mozilla_org = models.CharField(
+        max_length=255, null=True, blank=True, verbose_name=_lazy("Mozilla Directory Username")
     )
-    irc_handle = models.CharField(
-        max_length=255, null=True, blank=True, verbose_name=_lazy("IRC nickname")
+    matrix_handle = models.CharField(
+        max_length=255, null=True, blank=True, verbose_name=_lazy("Matrix nickname")
```
LeoMcA, line 83: `Matrix username` isn't capitalised like the above verbose names, should we make it `Matrix Nickname`?

kitsune/users/models.py
LeoMcA, line 77: Isn't "Mozilla" redundant here? I'd call this "Community Portal Username"

kitsune/users/models.py
```diff
@@ -72,14 +73,14 @@ class Profile(ModelBase, SearchMixin):
...
     )
-    facebook = models.URLField(
-        max_length=255, null=True, blank=True, verbose_name=_lazy("Facebook URL")
+    community_mozilla_org = models.CharField(
+        max_length=255, null=True, blank=True, verbose_name=_lazy("Mozilla Community Username")
     )
-    mozillians = models.CharField(
-        max_length=255, null=True, blank=True, verbose_name=_lazy("Mozillians Username")
+    people_mozilla_org = models.CharField(
+        max_length=255, null=True, blank=True, verbose_name=_lazy("Mozilla Directory Username")
```
LeoMcA, line 80: Same comment as above, could be "People Directory Username"

LeoMcA (approved):
r+wc (and circle still isn't happy)

LeoMcA (approved):
I've got a commit which "fixes" the failing test (by just removing it, and facebook widget/validation it was attempting to test, none of which we need any more) which you can cherry pick: https://github.com/LeoMcA/kitsune/commit/34e1947b1e316e06cd8ace8c37676eb7915e3fdf

r+ once circle passes

PR #4501, opened 2020-08-09
run collectstatic with noinput

saadmk11:
when running `make init` to setup the project locally the script stops and asks for input. this will run `collectstatic` with `--noinput`


akatsoulas (approved):
Thanks @saadmk11 for this PR!

Issue #4369, opened 2020-05-10
Copyright symbol


finn0:
Isn't it a good idea to use copyright symbol instead of simple `c.` symbol at footer of page (i.e c.1998-2019)?

![copyright](https://user-images.githubusercontent.com/28824347/81497430-01c6ec00-92ae-11ea-9923-602c8d627a69.png)

Also update copyright year from 1998-2020.

finn0:
I thing only replacing `c.` with `&copy;` would be enough at these places, https://github.com/mozilla/kitsune/blob/7c3585c4ff4382e340734419da39527a3ee098b9/kitsune/sumo/jinja2/base.html#L334 and https://github.com/mozilla/kitsune/blob/7c3585c4ff4382e340734419da39527a3ee098b9/styleguide/styleguide-examples/footer.njk#L105

nileshdas:
Taking up this issue

finn0:
@nileshdas If you don't mind, I've kept this issue for myself. I opened this issue to discuss with developers, whether they like a original copyright symbol or no? If you haven't created a pull request, I would like by myself to make a PR.

nileshdas:
Oh sorry! Sure, go ahead!

PR #4503, opened 2020-08-10
Fix footer

finn0:
Fix footer's content present in styleguide example file.

Also closes: #4369


akatsoulas (approved):
Thanks @finn0 :rocket:

finn0:
@akatsoulas like this PR, I want to replace © symbol with HTML's copyright symbol (`&copy`) and put a space between copyright symbol and year in [base.html#L314](https://github.com/mozilla/kitsune/blob/master/kitsune/sumo/jinja2/base.html#L314) file. What do you think, is it a good idea?

PR #4506, opened 2020-08-20
es7-convenience

LeoMcA:


akatsoulas (approved):
squash the last two commits into a single one please. r+

PR #4512, opened 2020-08-24
add all fields in QuestionDocument

LeoMcA:
This could probably do with a little bit of DRYing. Also needs to have a few fields substituted for the MultiLocaleObject field when we have that.

### DIFF
--- a/kitsune/search/v2/base.py
+++ b/kitsune/search/v2/base.py
@@ -1,6 +1,6 @@
 from django.utils import timezone
 
-from elasticsearch_dsl import Document as DSLDocument
+from elasticsearch_dsl import InnerDoc, Document as DSLDocument
 from elasticsearch_dsl import field
 
 
@@ -22,11 +22,16 @@ class SumoDocument(DSLDocument):
 
             # This will allow child classes to have their own methods in the form of prepare_field
             prepare_method = getattr(obj, "prepare_{}".format(f), None)
-            value = cls.get_field_value(f, instance, prepare_method)
+            value = obj.get_field_value(f, instance, prepare_method)
 
-            # Assign values to each field. Assign a dictionary to multi locale Objects
+            # Assign values to each field.
             field_type = doc_mapping.resolve_field(f)
-            if isinstance(field_type, field.Object) and "en-US" in field_type._mapping:
+            if isinstance(field_type, field.Object) and not (
+                isinstance(value, InnerDoc)
+                or (isinstance(value, list) and isinstance((value or [None])[0], InnerDoc))
+            ):
+                # if the field is an Object but the value isn't an InnerDoc
+                # or a list containing an InnerDoc then we're dealing with locales
                 locale = obj.prepare_locale(instance)
                 if locale and value:
                     obj[f] = {locale: value}
@@ -38,13 +43,11 @@ class SumoDocument(DSLDocument):
 
         return obj
 
-    @classmethod
-    def get_field_value(cls, field, instance, prepare_method):
+    def get_field_value(self, field, instance, prepare_method):
         """Allow child classes to define their own logic for getting field values."""
         if prepare_method is not None:
             return prepare_method(instance)
-        else:
-            return getattr(instance, field)
+        return getattr(instance, field)
 
     def prepare_locale(self, instance):
         """Return the locale of an object if exists."""
--- a/kitsune/search/v2/base.py
+++ b/kitsune/search/v2/base.py
@@ -22,11 +22,7 @@ class SumoDocument(DSLDocument):
 
             # This will allow child classes to have their own methods in the form of prepare_field
             prepare_method = getattr(obj, "prepare_{}".format(f), None)
-            # Overwrite the value if there is a specific method implemented for the specific field
-            if prepare_method is not None:
-                value = prepare_method(instance)
-            else:
-                value = cls.get_field_value(instance, f)
+            value = cls.get_field_value(f, instance, prepare_method)
 
             # Assign values to each field. Assign a dictionary to multi locale Objects
             field_type = doc_mapping.resolve_field(f)
@@ -43,12 +39,12 @@ class SumoDocument(DSLDocument):
         return obj
 
     @classmethod
-    def get_field_value(cls, instance, field):
-        """
-        Allow child classes to define their own logic for getting field values when a specific
-        method doesn't exist.
-        """
-        return getattr(instance, field)
+    def get_field_value(cls, field, instance, prepare_method):
+        """Allow child classes to define their own logic for getting field values."""
+        if prepare_method is not None:
+            return prepare_method(instance)
+        else:
+            return getattr(instance, field)
 
     def prepare_locale(self, instance):
         """Return the locale of an object if exists."""
--- a/kitsune/search/v2/base.py
+++ b/kitsune/search/v2/base.py
@@ -22,18 +22,18 @@ class SumoDocument(DSLDocument):
 
             # This will allow child classes to have their own methods in the form of prepare_field
             prepare_method = getattr(obj, "prepare_{}".format(f), None)
-            # Ovewrite the value if there is a specific method implented for the specific field
+            # Overwrite the value if there is a specific method implemented for the specific field
             if prepare_method is not None:
                 value = prepare_method(instance)
             else:
-                value = getattr(instance, f)
+                value = cls.get_field_value(instance, f)
 
-            # Assign values to each field. Assign a dictionary to Objects
-            if isinstance(doc_mapping.resolve_field(f), field.Object):
+            # Assign values to each field. Assign a dictionary to multi locale Objects
+            field_type = doc_mapping.resolve_field(f)
+            if isinstance(field_type, field.Object) and "en-US" in field_type._mapping:
                 locale = obj.prepare_locale(instance)
                 if locale and value:
                     obj[f] = {locale: value}
-                # We  populate non-locale object fields here
             else:
                 setattr(obj, f, value)
 
@@ -41,6 +41,14 @@ class SumoDocument(DSLDocument):
         obj.indexed_on = timezone.now()
 
         return obj
+
+    @classmethod
+    def get_field_value(cls, instance, field):
+        """
+        Allow child classes to define their own logic for getting field values when a specific
+        method doesn't exist.
+        """
+        return getattr(instance, field)
 
     def prepare_locale(self, instance):
         """Return the locale of an object if exists."""
--- a/kitsune/search/v2/documents.py
+++ b/kitsune/search/v2/documents.py
@@ -94,11 +94,19 @@ class UserInnerDoc(InnerDoc):
     id = field.Integer()
     username = field.Keyword()
 
+    @classmethod
+    def prepare(cls, instance):
+        return cls(id=instance.id, username=instance.username)
+
 
 class ProductInnerDoc(InnerDoc):
     id = field.Integer()
     title = field.Keyword()
     slug = field.Keyword()
+
+    @classmethod
+    def prepare(cls, instance):
+        return cls(id=instance.id, title=instance.title, slug=instance.slug)
 
 
 class TopicInnerDoc(InnerDoc):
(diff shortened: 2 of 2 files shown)


kitsune/search/v2/documents.py
```diff
@@ -137,12 +137,37 @@ def get_model(cls):
...
 
 
+class UserInnerDoc(InnerDoc):
+    id = field.Integer()
+    username = field.Keyword()
+
+
+class ProductInnerDoc(InnerDoc):
+    id = field.Integer()
+    title = field.Keyword()
```
LeoMcA, line 147: Our product titles can be localised, should we index these at all, or index them as a multi-locale object?

kitsune/search/v2/documents.py
```diff
@@ -137,12 +137,37 @@ def get_model(cls):
...
+    slug = field.Keyword()
+
+
+class TopicInnerDoc(InnerDoc):
+    id = field.Integer()
+    title = field.Keyword()
```
LeoMcA, line 153: Our topic titles can be localised too, should these also be indexed as a multi-locale Object?

kitsune/search/v2/documents.py
```diff
@@ -137,12 +137,37 @@ def get_model(cls):
...
+    slug = field.Keyword()
+
+
+class TagInnerDoc(InnerDoc):
+    id = field.Integer()
+    name = field.Keyword()
+    slug = field.Keyword()
```
LeoMcA, line 160: Currently tags will get flattened, the ES docs imply nested fields are quite expensive, and being able to filter on multiple fields of the same child object seem unnecessary here when id, name and slug could all be used equally effectively to filter.

Should we use an Object here at all, or should question_tags just be a list of tag slugs?

kitsune/search/v2/documents.py
akatsoulas: probably this makes sense.

kitsune/search/v2/base.py
```diff
@@ -22,18 +22,18 @@ def prepare(cls, instance):
...
                 value = prepare_method(instance)
             else:
-                value = getattr(instance, f)
+                value = cls.get_field_value(instance, f)
 
-            # Assign values to each field. Assign a dictionary to Objects
-            if isinstance(doc_mapping.resolve_field(f), field.Object):
+            # Assign values to each field. Assign a dictionary to multi locale Objects
+            field_type = doc_mapping.resolve_field(f)
+            if isinstance(field_type, field.Object) and "en-US" in field_type._mapping:
```
akatsoulas, line 33: Why do we need the `en-US` check?

kitsune/search/v2/documents.py
akatsoulas: Yes! Why are we treating both title and product as an object and not use them directly in the question doc?

akatsoulas: Shouldn't title be `Text` instead of `Keyword`?

kitsune/search/v2/documents.py
akatsoulas: Shouldn't title be `Text` instead of `Keyword`?

kitsune/search/v2/documents.py
akatsoulas: In case we do need an object, you can squash `TopicInnerDoc` and `ProductInnerDoc` together into a single doc like `SumoTitleInnerDoc` (or something with a nicer name :) )

kitsune/search/v2/documents.py
```diff
@@ -90,19 +89,82 @@ def get_model(cls):
...
+
+    @classmethod
+    def prepare(cls, instance):
+        if not instance:
+            return None
+        return cls(id=instance.id, title=instance.title, slug=instance.slug)
+
+
+class TagInnerDoc(InnerDoc):
```
akatsoulas, line 127: All the comments above apply here too

kitsune/search/v2/documents.py
```diff
@@ -111,43 +173,53 @@ class Index:
         using = config.DEFAULT_ES7_CONNECTION
 
     @classmethod
-    def prepare_locale(cls, instance):
-        return cls.get_question_instance(instance).locale
+    def prepare_question_creator(cls, instance):
+        user = cls.get_question_instance(instance).creator
```
akatsoulas, line 177: What's the value of using the class method `get_question_instance` and not directly use the `instance`? This add some confusion and  makes the code less readable.

akatsoulas (changes requested):

kitsune/search/v2/base.py
LeoMcA: To distinguish a multi-locale object (which will have a "en-US" parameter) from a "normal" object, like a UserInnerDoc

kitsune/search/v2/documents.py
LeoMcA: It's redefined in `AnswerDocument` to `return instance.question`, so these methods will work both when `instance` is of type `Question` and `Answer`. Agreed readability is quite bad, `AnswerDocument` could instead redefine `get_field_value` to call these methods with `instance.question` instead - that would also reduce the change of a later addition to these methods being buggy, because `instance` would always be of type `Question`.

akatsoulas: You can also add the `get_question_instance` as a class method only in the answer document and skip it from here.

kitsune/search/v2/base.py
akatsoulas: This feels a bit hacky.  I would suggest to at least use `locale` here for the check.

```python
 if isinstance(field_type, field.Object) and "en-US" in field_type._mapping:
    if locale := obj.prepare_locale(instance) and locale in field_type._mapping:
       if value:
           obj[f] = {locale:value}
    else:
       other_type_of_objects
```

kitsune/search/v2/documents.py
```diff
@@ -111,47 +173,46 @@ class Index:
...
 
-        return obj
+    @classmethod
+    def prepare_question_has_solution(cls, instance):
+        return instance.solution_id is not None
 
     @classmethod
-    def get_fields(cls):
-        return []
+    def prepare_locale(cls, instance):
```
akatsoulas, line 208: this already exists in the parent class and it can be removed from here since it's the same.

kitsune/search/v2/documents.py
```diff
@@ -111,47 +173,46 @@ class Index:
         using = config.DEFAULT_ES7_CONNECTION
 
     @classmethod
```
akatsoulas, line 175: All the `prepare_*` methods in the WikiDocument are not classmethods which probably makes sense. When we have a prepare_foo method it's going to be in most cases unique to that class. Let's remove all the classmethods for consistency.

kitsune/search/v2/documents.py
akatsoulas: Let's squash `ProductInnerDoc, TagInnerDoc, TopicInnderDoc` into a single document. Those are exactly the same for at least now. No need to redefine them 3 times.

akatsoulas (changes requested):

kitsune/search/v2/base.py
```diff
@@ -42,6 +38,14 @@ def prepare(cls, instance):
 
         return obj
 
+    @classmethod
+    def get_field_value(cls, field, instance, prepare_method):
+        """Allow child classes to define their own logic for getting field values."""
+        if prepare_method is not None:
+            return prepare_method(instance)
+        else:
```
akatsoulas, line 46: No need for the `else` here.

kitsune/search/v2/base.py
LeoMcA: I turned the problem on its head and came up with a less hacky solution - we know we're dealing with a locale field if the ES type is Object but the prepared type is something like a String (that is to say, not an InnerDoc).

LeoMcA:
I decided to remove the InnerDocs completely for now. I think we should only denormalise data where we have a use-case for doing queries on that data, since the overhead of denormalising is quite high (the need for SQL joins when indexing data, the additional signals needed to be set up, reindexing when any denormalised data is changed, etc.).

So some thought needs to go into that, but I don't think it needs to be done before we move forward with everything else. Indeed, building out the signals piece may make us more comfortable erring on the side of denormalising data we don't necessarily have a use-case for yet.

kitsune/search/v2/base.py
```diff
@@ -22,18 +22,19 @@ def prepare(cls, instance):
...
-
-            # Assign values to each field. Assign a dictionary to Objects
-            if isinstance(doc_mapping.resolve_field(f), field.Object):
+            value = obj.get_field_value(f, instance, prepare_method)
+
+            # Assign values to each field.
+            field_type = doc_mapping.resolve_field(f)
+            if isinstance(field_type, field.Object) and not (
+                isinstance(value, InnerDoc)
+                or (isinstance(value, list) and isinstance((value or [None])[0], InnerDoc))
```
akatsoulas, line 31: Could you please elaborate a bit on the different scenarios that you are checking against here?

kitsune/search/v2/documents.py
```diff
@@ -90,68 +89,52 @@ def get_model(cls):
         return wiki_models.Document
 
 
-class QuestionDocument(DSLDocument):
+class QuestionDocument(SumoDocument):
     """
     """
 
-    question_id = field.Integer()
-    question_title = field.Keyword(fields={"text": field.Text()})
```
akatsoulas, line 98: maybe it still makes sense to keep both this as keyword and localized text. We can do this in another PR but let's make sure it's captured in an issue.

kitsune/search/v2/base.py
LeoMcA: assuming we have 5 values, all being inserted into an Object field:

1. `"hello hello"` - this is a string (or, not an `InnerDoc`), so we need to apply the locale logic, the conditional matches here
2. `UserInnerDoc(username="foobar")` - this is an InnerDoc, we don't want to apply any locale logic, the conditional doesn't match
3. `[UserInnerDoc(username="foobar")]` - this is an InnerDoc in a list, we don't want to apply any locale logic, however `value` is not an `InnerDoc` (being a `list`), so we need the logic after the `or` to check the type of the first instance if we have a list, the conditional doesn't match
4. `["hello hello"]` - we do want to apply locale logic here, the conditional matches
5. `[]` - we don't really care what happens here, as long as nothing throws an exception, which is why we need the `(value or [None])[0]`. (I think) the conditional will match this.

kitsune/search/v2/documents.py
LeoMcA: I don't think keyword is a good fit here - it'll require the search term to exactly match the title

kitsune/search/v2/base.py
akatsoulas: I still believe that having something like the snippet (re-adding it here) that I added before should be the way to go for now, especially without any other `InnerDoc`s.

eg
```python
 if isinstance(field_type, field.Object) and "en-US" in field_type._mapping:
    if locale := obj.prepare_locale(instance) and locale in field_type._mapping:
       if value:
           obj[f] = {locale:value}
    else:
       other_type_of_objects
```
For example the `isinstance(value, InnerDoc)` is redundant here. When a field is of type `Object` without and `doc_class` as we have in the `SumoLocaleField` then it's initialized as an InnerDoc. So basically here we are saying if something is of type Object(InnerDoc) and not InnerDoc. The above snippet is not optimum but I would defer dealing with this issue - if it ever becomes one - when we have the need for it.

akatsoulas: Also this way we are treating all `String` in the object as a locale field which should not be the case. Ideally we should get somehow the type of `SumoLocale` field. If this is not possible, probably the second best would be to look at the data. Thoughts?

akatsoulas (approved):
r+wc

kitsune/search/v2/base.py
LeoMcA: > So basically here we are saying if something is of type Object(InnerDoc) and not InnerDoc.

Yes, that's the idea - if we have a mismatch in types between our field and value then we know that we're dealing with a SumoLocale field.

> Also this way we are treating all String in the object as a locale field which should not be the case.

I'm not sure there are any cases other than the SumoLocale field where a prepare_some_object function would return a String (or any other non-InnerDoc).

Regardless this isn't needed here yet, I'll add a non-list-handling version in the community PR. Anyway we may want to change this logic again later once we implement the parent->child indexing of kb articles.

PR #4513, opened 2020-08-25
Set csrf cookie to secure.

akatsoulas:
https://bugzilla.mozilla.org/show_bug.cgi?id=1642736

### DIFF
--- a/kitsune/settings.py
+++ b/kitsune/settings.py
@@ -757,6 +757,8 @@ SESSION_SERIALIZER = config(
     "SESSION_SERIALIZER", default="django.contrib.sessions.serializers.PickleSerializer"
 )
 
+# CSRF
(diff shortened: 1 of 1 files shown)


kitsune/settings.py
```diff
@@ -757,6 +757,8 @@ def JINJA_CONFIG():
     "SESSION_SERIALIZER", default="django.contrib.sessions.serializers.PickleSerializer"
 )
 
+# CSRF
+CSRF_COOKIE_SECURE = config("CSRF_COOKIE_SECURE", default=True, cast=bool)
```
LeoMcA, line 761: Perhaps this should be `default=not DEBUG` like `SESSION_COOKIE_SECURE` above?

Or maybe both should be `default=True` so they fail-safe? Firefox seems to send Secure cookies to localhost anyway.

LeoMcA (approved):
r+

PR #4517, opened 2020-08-28
migrate community search to use ES7

LeoMcA:
@akatsoulas here's where I got with this on Friday, it should all be functional

Some notes:
- It'll conflict with my AAQ PR because I duplicate the `ProductInnerDoc` which I use here too.
- I wonder if there's an elegant abstraction we can create to wire up the signals for us, or if it's easier to just do it manually. I'm thinking a bit ahead to where we might have a couple of layers of denormalisation, e.g. a GroupInnerDoc in a UserInnerDoc in a QuestionDocument. We'd have to add signals handlers to the relevant Groups, Users and Questions. But then creating an abstraction to handle all that is probably more complex than just doing it manually.


kitsune/search/v2/apps.py
```diff
@@ -0,0 +1,8 @@
+from django.apps import AppConfig
+
+
+class SearchV2Config(AppConfig):
```
akatsoulas, line 4: Let's move this to the search dir (a level up).

kitsune/search/v2/base.py
```diff
@@ -22,26 +22,33 @@ def prepare(cls, instance):
...
+                # or a list containing an InnerDoc then we're dealing with locales
                 locale = obj.prepare_locale(instance)
                 if locale and value:
                     obj[f] = {locale: value}
-                # We  populate non-locale object fields here
             else:
                 setattr(obj, f, value)
 
-        obj.meta.id = instance.id
+        obj.meta.id = instance.pk
```
akatsoulas, line 41: I like the attention to detail :P

akatsoulas (approved):
r+wc. Let's merge first the AAQ Doc and then this. I am adding an issue to extract/review any InnerDoc that we have throughout our indices.

kitsune/search/v2/base.py
LeoMcA: It's needed! `Profile`'s pk is its `user_id` field

PR #4519, opened 2020-09-01
remove {% trans %} block from styleguide template

LeoMcA:
it causes npm run start to fail


akatsoulas:
You were faster than me :+1:

akatsoulas (approved):

PR #4520, opened 2020-09-02
Add youtube-nocookie domain

finn0:
Add www.youtube-nocookie.com domain to Youtube's domain list.

### DIFF
--- a/kitsune/sumo/parser.py
+++ b/kitsune/sumo/parser.py
@@ -403,4 +403,4 @@ def _is_youtube_url(url):
     """Returns true if the URL is to youtube."""
     parsed_url = urlparse(url)
     netloc = parsed_url.netloc
-    return netloc in ["youtu.be", "youtube.com", "www.youtube.com"]
+    return netloc in ["youtu.be", "youtube.com", "www.youtube.com", "www.youtube-nocookie.com"]


kitsune/sumo/parser.py
```diff
@@ -403,4 +403,4 @@ def _is_youtube_url(url):
     """Returns true if the URL is to youtube."""
     parsed_url = urlparse(url)
     netloc = parsed_url.netloc
-    return netloc in ["youtu.be", "youtube.com", "www.youtube.com"]
+    return netloc in ["youtu.be", "youtube.com", "www.youtube.com", "www.youtube-nocookie.com"]
```
akatsoulas, line 406: shall we also add another entry without the `www` subdomain?

finn0: I also thought about that, but when you open https://youtube-nocookie.com, you will get a certificate error message, saying that certificate isn't valid for this domain. Once I had reported this to the Google, but never get a reply. If you want, I can add domain without `www` subdomain.

akatsoulas: nice catch. I believe it's fine as is. We can revisit if needed

akatsoulas (approved):

kitsune/sumo/parser.py
finn0: Youtube also uses these domains, https://youtubeeducation.com, https://youtubekids.com, and https://yt.be, but don't know whether these domains are useful for us or not. Should I add these domains also?

LeoMcA: I don't think we'll be embedding youtube videos from those domains (which is the ultimate purpose of this function) so I think we're ok with the urls we've now got there

PR #4521, opened 2020-09-04
[docs] mention complete command

finn0:
New user who is setting-up development environment may get confused,
where they should run `./manage.py generatedata` command. So it would be
nice idea to mention complete command, whenever it could be possible.

### DIFF
--- a/docs/hacking_howto.md
+++ b/docs/hacking_howto.md
@@ -131,7 +131,7 @@ running this command::
 ### Install linting tools
 
 ```eval_rst
-Kitsune uses `Yelps Pre-commit <http://pre-commit.com/>`_ for linting. It is
+Kitsune uses `Yelps Pre-commit <https://pre-commit.com/>`_ for linting. It is
 installed as a part of the dev dependencies in ``requirements/dev.txt``. To
 install it as a Git pre-commit hook, run it::
 
@@ -147,7 +147,7 @@ checks for all files::
 
    $ venv/bin/pre-commit run --all-files
 
-For more details see the `Pre-commit docs <http://pre-commit.com/>`_.
+For more details see the `Pre-commit docs <https://pre-commit.com/>`_.
 ```
 
 ### Product Details Initialization
--- a/docs/hacking_howto.md
+++ b/docs/hacking_howto.md
@@ -96,10 +96,26 @@ After the above you can do some optional steps if you want to use the admin:
 
 First, make sure you have run the "Create some data" step above.
 
-1. Enter the web container: `docker-compose exec web bash`
-2. Build the indicies: `./manage.py esreindex` (You may need to pass the `--delete` flag)
-3. Precompile the nunjucks templates: `./manage.py nunjucks_precompile`
-4. Now, exit from web's bash shell: `exit`
+1. Enter into the web container
+    ```
+    docker-compose exec web bash
+    ```
+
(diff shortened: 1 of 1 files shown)


finn0:
I've also plans for adding some more clarification in [Install liniting tools](https://kitsune.readthedocs.io/en/latest/hacking_howto.html#install-linting-tools), as I discussed it with leo.

docs/hacking_howto.md
```diff
@@ -108,7 +108,7 @@ First, make sure you have run the "Create some data" step above.
 We include some sample data to get you started. You can install it by
 running this command::
 
-    $ ./manage.py generatedata
+    docker-compose exec web ./manage.py generatedata
```
akatsoulas, line 111: I would suggest to move the whole section `Install Sample Data` above the `Get Search working`. The latter leaves the user into a bash shell into the container and running `docker-compose ...` in there won't work. By moving it up we can avoid this scenario

akatsoulas (approved):
Thanks for this PR @finn0! r+wc :rocket:

docs/hacking_howto.md
finn0: But in my humble opinion, `Install Sample Data` subsection is more suitable under `Further Setup` section. Not everyone needs to generate sample data, it's an additional step for those who wants to work on Kitsune instance which contains some data. What's your thought now? If still, wants stick to your proposed changes, I'm more than happy to make those changes.

akatsoulas: Content wise I agree with what you are saying. All I am saying is that the user might get confused by this command b/c the last step left the user inside the container. Another solution here could be to add the `docker-compose exec web` command to the search part too for consistency.

finn0: Actually, `docker-compose exec web` command already mentioned in [Get search working](https://kitsune.readthedocs.io/en/latest/hacking_howto.html#get-search-working) section, see first instruction. I'm thinking about another solution, adding one another step (4th step) to exit from `web`'s bash shell. Something like,

4. Now exit from web's bash shell: `exit`

and adding `docker-compose exec web` command to other remaining sections, also like [Product Details Initialization](https://kitsune.readthedocs.io/en/latest/hacking_howto.html#product-details-initialization), [Pre-compiling JavaScript Templates](https://kitsune.readthedocs.io/en/latest/hacking_howto.html#pre-compiling-javascript-templates). What do you think? Is it a good idea?

akatsoulas: sounds good!

finn0:
@akatsoulas, after going through [hacking-howto](https://kitsune.readthedocs.io/en/latest/hacking_howto.html) page, I realized that `$` sign is being used with command which runs in web's bash shell. I tried to maintain this style in this PR, also. But all commands which are present in [Install linting tools](https://kitsune.readthedocs.io/en/latest/hacking_howto.html#install-linting-tools) section have `$` sign with them, despite their are being running outside the web's bash shell. Should I remove `$` sign from all commands present in [Install linting tools](https://kitsune.readthedocs.io/en/latest/hacking_howto.html#install-linting-tools) section?

akatsoulas:
@finn0 I believe it's fine to leave things as is. Thank you for this PR. This should be good to go!

finn0:
@akatsoulas any idea why build is failing?

PR #4523, opened 2020-09-07
Update functionality to edit username.

akatsoulas:
* Fix avatar regression.
* Fix 404 error when + sign is used in username.
* Update help_text wording to match the one from the form validation.

### DIFF
--- a/kitsune/sumo/static/sumo/scss/layout/_profile.scss
+++ b/kitsune/sumo/static/sumo/scss/layout/_profile.scss
@@ -12,6 +12,6 @@
   }
 }
 
-.avatar {
-  padding-top: p.$spacing-lg;
+#edit-profile .avatar {
+  margin-top: p.$spacing-sm;
 }
--- a/kitsune/users/jinja2/users/profile.html
+++ b/kitsune/users/jinja2/users/profile.html
@@ -39,10 +39,12 @@
           {% endif %}
         </h2>
         <p>
-          {% trans a_open='<a href="'|safe + url("users.edit_profile", profile.user.username) + '">'|safe, a_close='</a>'|safe %}
-            Your privacy is important to us, your username is always visible to the public,
-            if you would like to edit your username {{ a_open }} click here {{ a_close }}.
-          {% endtrans %}
+          {% if is_owner %}
+            {% trans a_open='<a href="'|safe + url("users.edit_my_profile") + '">'|safe, a_close='</a>'|safe %}
+              Your privacy is important to us, your username is always visible to the public,
+              if you would like to edit your username {{ a_open }} click here {{ a_close }}
+            {% endtrans %}
+          {% endif %}
         </p>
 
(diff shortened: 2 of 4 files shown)


kitsune/sumo/static/sumo/scss/layout/_profile.scss
```diff
@@ -13,5 +13,5 @@
 }
 
 .avatar {
-  padding-top: p.$spacing-lg;
+  margin-top: p.$spacing-lg;
 }
```
LeoMcA, line 17: The root of the problem here is this selector is overly broad, something like `#edit-profile .avatar` will only select the avatar on the edit profile page, without selecting the one in the navbar as well.

I'd also suggest using `$spacing-sm` as that's the same spacing we have between labels and fields below.

kitsune/users/jinja2/users/profile.html
```diff
@@ -39,10 +39,12 @@ <h2 class="sumo-callout-heading user">{{ display_name(profile.user) }}
...
         </h2>
         <p>
-          {% trans a_open='<a href="'|safe + url("users.edit_profile", profile.user.username) + '">'|safe, a_close='</a>'|safe %}
-            Your privacy is important to us, your username is always visible to the public,
-            if you would like to edit your username {{ a_open }} click here {{ a_close }}.
-          {% endtrans %}
+          {% if is_owner %}
+            {% trans a_open='<a href="'|safe + url("users.edit_my_profile") + '">'|safe, a_close='</a>'|safe %}
+              Your privacy is important to us, your username is always visible to the public,
+              if you would like to edit your username {{ a_open }} click here {{ a_close }}.
```
LeoMcA, line 45: The spaces in `{{ a_open }} click here {{ a_close }}` make the link look a little weird

kitsune/users/views.py
```diff
@@ -246,16 +247,20 @@ def edit_watch_list(request):
...
+                raise Http404
 
-        # Make sure the auth'd user has permission:
-        if not request.user.has_perm("users.change_profile"):
-            return HttpResponseForbidden()
-    else:
+            # Make sure the auth'd user has permission:
+            if not request.user.has_perm("users.change_profile"):
+                return HttpResponseForbidden()
+    if not user:
```
LeoMcA, line 263: I think this can remain an `else`: the only case in which the above block leaves `user` falsy it will have raised `Http404`. Then L250 can also be removed.

LeoMcA (approved):
r+wc

kitsune/users/views.py
akatsoulas: The problem here is that the username is passed and the `else` clause was activated because the second comparison of the `and` operator was False. With the new change it will get into the username but not in the second if thus it will never reach the `else` clause producing an unbound variable error from line 267

LeoMcA: Oh of course, I've oversimplified, ignore me!

PR #4526, opened 2020-09-09
Multi field es7

akatsoulas:


kitsune/search/v2/base.py
```diff
@@ -0,0 +1,48 @@
...
+        fields.remove("indexed_on")
+
+        # Loop through the fields of each object and set the right values
+        for f in fields:
+
+            try:
+                # This will allow child classes to have own methods in the form of prepare_field
+                prepare_method = getattr(obj, "prepare_{}".format(f))
+                value = prepare_method(instance)
+            except AttributeError:
```
LeoMcA, line 27: I had a few problems before with this `except AttributeError` catching errors within the prepare_method itself, for instance if the method was `return instance.some_attribute_which_doesnt_exist`. Is there some non try/except method for detecting whether a prepare_method exists?

kitsune/search/v2/base.py
```diff
@@ -0,0 +1,48 @@
...
+                value = getattr(instance, f)
+            # Assign values to each field. Assign a dictionary to Objects
+            if isinstance(doc_mapping.resolve_field(f), field.Object):
+                locale = cls.prepare_locale(instance)
+                if locale and value:
+                    obj[f] = {locale: value}
+                # We  populate non-locale object fields here
```
LeoMcA, line 34: I'm not sure it'll be the case that every `field.Object` on an instance with a `.locale` attribute will be an object with multiple locales. I think this will need to be re-jigged a bit for AAQ, but maybe it makes sense for me to do this rather than you writing in the abstract?

kitsune/search/v2/base.py
```diff
@@ -0,0 +1,48 @@
...
+        obj.indexed_on = timezone.now()
+
+        return obj
+
+    @classmethod
+    def prepare_locale(cls, instance):
+        """Return the locale of an object if exists."""
+        if getattr(instance, "locale"):
+            return instance.locale
+        return ""
```
LeoMcA, line 48: This being a classmethod feels funny to me when all the other prepare_methods are instance methods. I think it should possibly also just return None, or even raise NotImplementedError, for subclasses to implement themselves. Again this might become clearer when we have a number of Document classes.

kitsune/search/v2/documents.py
```diff
@@ -185,8 +138,6 @@ def prepare(cls, instance):
             if f in locale_fields:
                 setattr(obj, f, {})
                 setattr(obj[f], cls.prepare_locale(instance), value)
-            else:
-                setattr(obj, f, value)
```
LeoMcA, line 189: not sure what this deletion is doing, leave my QuestionDocument alone :P

kitsune/search/v2/base.py
akatsoulas: the comment is just a placeholder for an `else` statement in case we need to populate an Object without a locale. Although in this case the `value` needs to go away from the `if` statement. Let's tackle that when there is a need for it.

kitsune/search/v2/base.py
akatsoulas: I am removing the `classmethod` for now. Let's see how it's going to behave with the rest of the docs and adapt if needed.

PR #4533, opened 2020-09-17
improve indexing performance even more

LeoMcA:
https://github.com/mozilla/sumo-project/issues/605

```
(time to index 10000 / sql queries executed per 100)

QuestionDocument:
before: 0m26.192s / 201
after:  0m12.801s / 2

AnswerDocument:
before: 0m51.871s / 501
after:  0m14.140s / 3

ProfileDocument:
before: 0m34.018s / 301
after:  0m11.356s / 3

ForumDocument:
before: 0m13.332s / 101
after:  0m6.579s  / 1
```

### DIFF
--- a/kitsune/search/v2/base.py
+++ b/kitsune/search/v2/base.py
@@ -43,6 +43,15 @@ class SumoDocument(DSLDocument):
 
         return obj
 
+    @classmethod
+    def get_queryset(cls):
+        """
+        Return the manager for a document's model.
+        This allows child classes to add optimizations like select_related or prefetch_related
+        to improve indexing performance.
+        """
+        return cls.get_model()._default_manager
+
     def get_field_value(self, field, instance, prepare_method):
         """Allow child classes to define their own logic for getting field values."""
         if prepare_method is not None:
--- a/kitsune/search/v2/documents.py
+++ b/kitsune/search/v2/documents.py
@@ -1,11 +1,12 @@
 from elasticsearch_dsl import connections, field, InnerDoc
+from django.db.models import Prefetch, Count, Q
 from kitsune.search import config
 from kitsune.search.v2.base import SumoDocument
 from kitsune.search.v2.es7_utils import es7_client
 from kitsune.search.v2.fields import SumoLocaleAwareTextField
 from kitsune.wiki import models as wiki_models
 from kitsune.wiki.config import REDIRECT_HTML
-from kitsune.questions import models as question_models
+from kitsune.questions.models import Question, Answer
 from kitsune.users.models import Profile
 from kitsune.forums.models import Post
 
@@ -134,6 +135,11 @@ class QuestionDocument(SumoDocument):
     def prepare_question_has_solution(self, instance):
         return instance.solution_id is not None
 
+    def prepare_question_num_votes(self, instance):
+        if hasattr(instance, "es_question_num_votes"):
+            return instance.es_question_num_votes
+        return instance.num_votes
+
     def get_field_value(self, field, *args):
         if field.startswith("question_"):
             field = field[len("question_") :]
@@ -141,7 +147,17 @@ class QuestionDocument(SumoDocument):
 
     @classmethod
     def get_model(cls):
-        return question_models.Question
+        return Question
+
+    @classmethod
+    def get_queryset(cls):
+        return (
+            Question.objects
+            # prefetch tags to avoid extra queries when iterating over them
+            .prefetch_related("tags")
+            # count votes in db to improve performance
+            .annotate(es_question_num_votes=Count("votes"))
+        )
 
 
 class AnswerDocument(QuestionDocument):
@@ -170,6 +186,16 @@ class AnswerDocument(QuestionDocument):
     def prepare_locale(self, instance):
         return instance.question.locale
 
+    def prepare_num_helpful_votes(self, instance):
+        if hasattr(instance, "es_num_helpful_votes"):
+            return instance.es_num_helpful_votes
+        return instance.num_helpful_votes
+
+    def prepare_num_unhelpful_votes(self, instance):
+        if hasattr(instance, "es_num_unhelpful_votes"):
+            return instance.es_num_unhelpful_votes
+        return instance.num_unhelpful_votes
+
     def get_field_value(self, field, instance, *args):
         if field.startswith("question_"):
             instance = instance.question
@@ -177,7 +203,21 @@ class AnswerDocument(QuestionDocument):
 
     @classmethod
     def get_model(cls):
-        return question_models.Answer
+        return Answer
+
+    @classmethod
+    def get_queryset(cls):
+        return (
+            Answer.objects
+            # prefetch each answer's question,
+            # applying the same optimizations as in the QuestionDocument
+            .prefetch_related(Prefetch("question", queryset=super().get_queryset()))
+            # count votes in db to improve performance
+            .annotate(
+                es_num_helpful_votes=Count("votes", filter=Q(votes__helpful=True)),
+                es_num_unhelpful_votes=Count("votes", filter=Q(votes__helpful=False)),
+            )
+        )
 
 
 class ProductInnerDoc(InnerDoc):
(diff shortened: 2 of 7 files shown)


kitsune/search/v2/es7_utils.py
```diff
@@ -98,9 +98,8 @@ def index_objects_bulk(doc_type_name, obj_ids):
     """Bulk index ORM objects given a list of object ids and a document type name."""
 
     doc_type = next(cls for cls in get_doc_types() if cls.__name__ == doc_type_name)
-    model = doc_type.get_model()
 
-    objects = model.objects.filter(pk__in=obj_ids)
+    objects = doc_type.get_queryset().all().filter(pk__in=obj_ids)
```
akatsoulas, line 102: why do you need `all()` here?

kitsune/search/v2/management/commands/es7_reindex.py
```diff
@@ -32,6 +43,9 @@ def handle(self, *args, **kwargs):
 
         progress_msg = "Indexed {progress} out of {count}"
 
+        if kwargs["print_sql_count"]:
```
akatsoulas, line 46: nope :)

kitsune/search/v2/management/commands/es7_reindex.py
```diff
@@ -41,10 +55,13 @@ def handle(self, *args, **kwargs):
 
             percentage = kwargs["percentage"]
             total = count
-            if percentage < 100:
-                count = int(count * percentage / 100)
-                qs = qs[:count]
-            print("Indexing {}%, so {} documents out of {}".format(percentage, count, total))
+            if count := kwargs["count"]:
```
akatsoulas, line 58: Will `count` always be in kwargs?

kitsune/search/v2/documents.py
```diff
@@ -244,6 +275,10 @@ def prepare_groups(self, instance):
     def get_model(cls):
         return Profile
 
+    @classmethod
```
akatsoulas, line 278: why `classmethod`?

akatsoulas: I guess the same Q holds true for all the other occurrences

kitsune/search/v2/documents.py
```diff
@@ -170,14 +182,33 @@ def prepare_is_solution(self, instance):
...
             instance = instance.question
         return super().get_field_value(field, instance, *args)
 
     @classmethod
     def get_model(cls):
-        return question_models.Answer
+        return Answer
+
+    @classmethod
+    def get_queryset(cls):
```
akatsoulas, line 205: comments pls on what exactly this query does

kitsune/search/v2/documents.py
```diff
@@ -134,14 +135,25 @@ def prepare_question_tag_id(self, instance):
...
         return super().get_field_value(field, *args)
 
     @classmethod
     def get_model(cls):
-        return question_models.Question
+        return Question
+
+    @classmethod
+    def get_queryset(cls):
+        return Question.objects.prefetch_related("tags").annotate(
```
akatsoulas, line 154: tags?

kitsune/search/v2/base.py
```diff
@@ -43,6 +43,10 @@ def prepare(cls, instance):
 
         return obj
 
+    @classmethod
+    def get_queryset(cls):
+        return cls.get_model()._default_manager
```
akatsoulas, line 48: I am curious about this, pls tell me more

kitsune/search/v2/es7_utils.py
LeoMcA: We don't, oversight from me

kitsune/search/v2/management/commands/es7_reindex.py
LeoMcA: I was planning on adding some documentation saying something along the lines of "after you write/edit an ES document, run an index with `--print-sql-count` to ensure you're not throwing hundreds of requests at the DB. If you are, consider adding a `select_` or `prefetch_related` to the document's get_queryset method to avoid that."

`django.db.connection.queries` needs `DEBUG=True` to do anything anyway, so being behind an argument as well, I'm not sure this could do anything negative on prod.

kitsune/search/v2/management/commands/es7_reindex.py
LeoMcA: Yep, the `default=None` on L25 ensures it'll be there (but falsey).

kitsune/search/v2/documents.py
LeoMcA: It needs to be able to refer to the `get_model` classmethod to get the model and then its manager.

kitsune/search/v2/documents.py
LeoMcA: Yep

kitsune/search/v2/base.py
LeoMcA: This is essentially calling `cls.get_model().objects`, so if we created a `FoobarDoc(SumoDocument)` with `get_model()` returning `Foobar`, this would be returning `Foobar.objects`.

`_default_manager` seems to be the more "djangonic" way to refer to `.objects` in case the manager has been renamed: https://docs.djangoproject.com/en/2.2/topics/db/managers/#django.db.models.Model._default_manager

The reason for adding this as a method is to be able to control the queryset used when indexing from the document, so we can add optimisations and aggregations and so on.

kitsune/search/v2/management/commands/es7_reindex.py
LeoMcA: Oh but there is a bug here, even if `kwargs["count"]` is falsey, it'll be assigned to count, overwriting the previous value of with `None`

kitsune/search/v2/documents.py
```diff
@@ -170,14 +186,38 @@ def prepare_is_solution(self, instance):
...
-        return question_models.Answer
+        return Answer
+
+    @classmethod
+    def get_queryset(cls):
+        return (
+            Answer.objects
+            # prefetch each answer's question,
+            # applying the same optimizations as in the QuestionDocument
+            .prefetch_related(Prefetch("question", queryset=super().get_queryset()))
```
akatsoulas, line 214: Since we are using `classmethod` I am wondering if it's better/cleaner to call here `QuestionDocument.get_queryset()` instead of super

kitsune/search/v2/documents.py
```diff
@@ -170,14 +186,38 @@ def prepare_is_solution(self, instance):
...
+            # count votes in db to improve performance
+            .annotate(
+                es_num_helpful_votes=Count("votes", filter=Q(votes__helpful=True)),
```
akatsoulas, line 217: Why do we need the `Q` object here? Same question for the next line

kitsune/search/v2/management/commands/es7_reindex.py
akatsoulas: I am mostly referent to the conditional in order to do the import

akatsoulas (approved):
r+wc

akatsoulas:
@LeoMcA I am merging this PR to continue working on ES. Please do open a new PR to address the comments in here. Thanks!

kitsune/search/v2/documents.py
LeoMcA: This is needed to filter by only votes that are helpful/not helpful when counting. The generated SQL might help:
```
COUNT(CASE WHEN `questions_answervote`.`helpful` = True THEN `questions_answervote`.`id` ELSE NULL END) AS `es_num_helpful_votes`, COUNT(CASE WHEN `questions_answervote`.`helpful` = False THEN `questions_answervote`.`id` ELSE NULL END) AS `es_num_unhelpful_votes`
```

There are a lot of gotchas with annotations, so I want to add some basic unit tests here to ensure we're not double counting votes or anything.

kitsune/search/v2/management/commands/es7_reindex.py
```diff
@@ -53,4 +72,7 @@ def handle(self, *args, **kwargs):
                 start = x * bulk_count
                 end = start + bulk_count
                 index_objects_bulk.delay(dt.__name__, id_list[start:end])
+                if kwargs["print_sql_count"]:
+                    print("{} SQL queries executed".format(len(connection.queries)))
+                    reset_queries()
```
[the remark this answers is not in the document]
LeoMcA: `reset_queries()`? it clears the contents of `connections.queries`, so we're only counting the number of queries executed in that batch, rather than the running total.

PR #4537, opened 2020-09-23
debug circle build

LeoMcA:
please ignore, getting more output from circle to compare with local build


LeoMcA:
see https://github.com/mozilla/kitsune/pull/4538

PR #4538, opened 2020-09-23
fix circle build

LeoMcA:
remove typing (now part of stdlib)
re-order venv and pip install commands to ensure consistent pip version across circle and local dev
update pip, pip-tools and pip-compile-multi to latest


The-smooth-operator:
:clap:

akatsoulas (approved):

PR #4541, opened 2020-09-27
Fixes #4500, collections depreciated

### LINKED ISSUE #4500: Importing ABC directly from collections has been deprecated and will be removed in Python 3.10
Use `collections` for Python 2 and `collections.abc` for Python 3.4+

https://github.com/mozilla/kitsune/blob/db42cf7841b75ff0d4b6d3b21ac2524636110429/kitsune/wiki/widgets.py#L34

Metropass:
AS of Python 3.10, `collections` will be depreciated in terms of abstract classes. Added a check if Python is >= 3.3, and imports `collections.abc`

### DIFF
--- a/kitsune/wiki/widgets.py
+++ b/kitsune/wiki/widgets.py
@@ -1,5 +1,3 @@
-import sys
-
 
 from django import forms
 from django.template.loader import render_to_string
--- a/kitsune/wiki/widgets.py
+++ b/kitsune/wiki/widgets.py
@@ -1,4 +1,9 @@
-import collections
+import sys
+
+if sys.version_info >= (3,3):
+    from collections.abc import Iterable
+else:
+    from collections import Iterable
 
 from django import forms
 from django.template.loader import render_to_string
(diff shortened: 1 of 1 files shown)


akatsoulas (approved):
thanks @Metropass for this fix! :rocket:

kitsune/wiki/widgets.py
```diff
@@ -1,11 +1,17 @@
...
+import sys
+
 
 from django import forms
 from django.template.loader import render_to_string
 
 from kitsune.products.models import Topic
 from kitsune.wiki.models import Document
 
+if sys.version_info >= (3,3):
```
akatsoulas, line 10: Could you replace the `if/else` statement with:

```python
try:
    from collections.abc import Iterable
except ImportError:
    from collections import Iterable
``` 

just to be consistent with our imports?

Metropass: Will do!

Metropass: Updated :)

akatsoulas:
Thanks @Metropass !

PR #4544, opened 2020-10-05
reimplement top_contributors_questions in ES7

LeoMcA:
https://github.com/mozilla/sumo-project/issues/652

### DIFF
--- a/kitsune/community/utils.py
+++ b/kitsune/community/utils.py
@@ -4,56 +4,83 @@ from operator import itemgetter
 
 from django.conf import settings
 from django.core.cache import cache
-from django.db.models import Count, F
+from django.db.models import Count
 
 from kitsune.products.models import Product
-from kitsune.questions.models import Answer
 from kitsune.users.models import User, UserMappingType
 from kitsune.wiki.models import Revision
 
+from elasticsearch_dsl import A
+from kitsune.search.v2.documents import AnswerDocument, ProfileDocument
 
-def top_contributors_questions(
-    start=None, end=None, locale=None, product=None, count=10, page=1, use_cache=True
-):
+
+def top_contributors_questions(start=None, end=None, locale=None, product=None, count=10):
     """Get the top Support Forum contributors."""
-    if use_cache:
-        cache_key = "{}_{}_{}_{}_{}_{}".format(start, end, locale, product, count, page)
-        cache_key = hashlib.sha1(cache_key.encode("utf-8")).hexdigest()
-        cache_key = "top_contributors_questions_{}".format(cache_key)
-        cached = cache.get(cache_key, None)
-        if cached:
-            return cached
 
-    answers = (
-        Answer.objects.exclude(is_spam=True).exclude(question__is_spam=True)
-        # Adding answer to your own question, isn't a contribution.
-        .exclude(creator_id=F("question__creator_id"))
+    search = AnswerDocument.search()
+
+    search = search.filter(
+        # filter out answers by the question author
+        "script",
+        script="doc['creator_id'].value != doc['question_creator_id'].value",
+    ).filter(
+        # filter answers created between `start` and `end`, or within the last 90 days
+        "range",
+        created={"gte": start or datetime.now() - timedelta(days=90), "lte": end},
     )
-
-    if start is None:
-        # By default we go back 90 days.
-        start = date.today() - timedelta(days=90)
-        answers = answers.filter(created__gte=start)
-    if end:
-        # If no end is specified, we don't need to filter by it.
-        answers = answers.filter(created__lt=end)
     if locale:
-        answers = answers.filter(question__locale=locale)
+        search = search.filter("term", locale=locale)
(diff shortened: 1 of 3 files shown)


kitsune/community/utils.py
```diff
@@ -4,56 +4,86 @@
...
+            {
+                "count": bucket.doc_count,
+                "term": bucket.key,
+                "user": {
+                    "id": user["id"],
+                    "username": user["username"],
+                    "display_name": user["profile__name"],
+                    "avatar": user["profile__fxa_avatar"],
+                    "days_since_last_activity": (
+                        datetime.now()
```
akatsoulas, line 77: nit: maybe we calculate this outside of the return statement. My only reason for proposing this is code readability

kitsune/community/utils.py
```diff
@@ -4,56 +4,86 @@
...
+                        - datetime.fromisoformat(bucket.latest.hits.hits[0]._source.created)
+                    ).days,
+                },
+            }
+            for bucket in contribution_buckets[:count]
```
akatsoulas, line 82: let's move all of this outside of the return. It's a bit challenging wrapping your head around this block of code.

akatsoulas (commented):
just a couple of comments for now. I haven't gone through functionality yet

kitsune/community/utils.py
LeoMcA: I was feeling very functional that day - refactored to use a bog standard for loop instead

kitsune/community/utils.py
```diff
@@ -4,56 +4,87 @@
...
-    return counts
+    contribution_buckets = search.execute().aggregations.filters.contributions.buckets
+    total_contributors = len(contribution_buckets)
+
+    user_ids = map(lambda x: int(x.key), contribution_buckets[:count])
+    users = {
+        # fetch the users found in the ES aggregation,
+        # and put them in a dict with their id as the key
+        user["id"]: user
+        for user in User.objects.filter(id__in=user_ids).values(
```
akatsoulas, line 61: can't we get the data directly from ES?

akatsoulas (approved):
r+wc 
Personally I would like the idea to go fully with ES on this instead of hitting the DB too unless there are reasons not too. I would be surprised if ES is slower on this although I don't expect big differences. To the extend that we can we can reduce queries towards the DB leveraging ES. Up to you though

kitsune/community/utils.py
```diff
@@ -4,56 +4,87 @@
...
-            product = product.slug
-        answers = answers.filter(question__product__slug=product)
-
-    users = (
-        User.objects.filter(answers__in=answers)
-        .annotate(query_count=Count("answers"))
-        .order_by("-query_count")
+        filters.append({"term": {"question_product_id": product.id}})
+
+    search.aggs.bucket(
```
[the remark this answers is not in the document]
LeoMcA: they seem to be as performant as each other, but I think using a filtered search makes the code clearer

LeoMcA:
Actually fetching profiles from ES works rather nicely because it returns them in the order the ids were specified, so I don't need to do that list to dict conversion I was doing.

kitsune/community/utils.py
```diff
@@ -4,56 +4,82 @@
...
-        # By default we go back 90 days.
-        start = date.today() - timedelta(days=90)
-        answers = answers.filter(created__gte=start)
-    if end:
-        # If no end is specified, we don't need to filter by it.
-        answers = answers.filter(created__lt=end)
+    # filter out answers by the question author
+    search = search.filter(
+        "script", script="doc['creator_id'].value != doc['question_creator_id'].value"
+    )
```
akatsoulas, line 25: How about chaining them in order to be more similar to Django's syntax.

kitsune/community/utils.py
```diff
@@ -4,56 +4,82 @@
...
-    if use_cache:
-        cache.set(cache_key, counts, 60 * 180)  # 3 hours
-    return counts
+    contribution_buckets = search.execute().aggregations.contributions.buckets
+    total_contributors = len(contribution_buckets)
+
+    if not total_contributors:
+        return [], 0
+
+    user_ids = map(lambda x: int(x.key), contribution_buckets[:count])
```
akatsoulas, line 56: list comprehension maybe?

kitsune/community/utils.py
```diff
@@ -4,56 +4,82 @@
...
+        if bucket.key != user.meta.id:
+            raise RuntimeError("ES returned profiles out of order")
+        last_activity = datetime.fromisoformat(bucket.latest.hits.hits[0]._source.created)
+        days_since_last_activity = (datetime.now() - last_activity).days
+        top_contributors.append(
+            {
+                "count": bucket.doc_count,
+                "term": bucket.key,
+                "user": {
+                    "id": bucket.key,
```
akatsoulas, line 73: maybe use `user` here too?

akatsoulas (approved):
r+wc

PR #4549, opened 2020-10-09
small revisions to indexing performance optimisations

LeoMcA:
add tests to ensure aggregations return the correct result

there's some reformatting black has done in here after merging with master and bringing in the new version, I'll open a separate PR to contain those, which should be merged before this one

https://github.com/mozilla/kitsune/pull/4533

### DIFF
--- a/kitsune/search/v2/signals.py
+++ b/kitsune/search/v2/signals.py
@@ -1,6 +1,7 @@
 from django.db.models.signals import post_save, post_delete, m2m_changed
 from django.dispatch import receiver
 from django.contrib.auth.models import User, Group
+from django.conf import settings
 from kitsune.users.models import Profile
 from kitsune.search.v2.es7_utils import index_object, delete_object
 from kitsune.products.models import Product
@@ -12,43 +13,51 @@ from kitsune.forums.models import Thread, Post
 @receiver(m2m_changed, sender=User.groups.through)
 @receiver(m2m_changed, sender=Profile.products.through)
 def handle_profile_save(instance, **kwargs):
-    index_object.delay("ProfileDocument", instance.pk)
+    if settings.ES_LIVE_INDEXING:
+        index_object.delay("ProfileDocument", instance.pk)
 
 
 @receiver(post_save, sender=Group)
 def handle_group_save(instance, **kwargs):
-    for user in User.objects.filter(groups=instance):
-        index_object.delay("ProfileDocument", user.pk)
+    if settings.ES_LIVE_INDEXING:
+        for user in User.objects.filter(groups=instance):
+            index_object.delay("ProfileDocument", user.pk)
 
 
 @receiver(post_save, sender=Product)
 def handle_product_save(instance, **kwargs):
-    for user in instance.subscribed_users.all():
-        index_object.delay("ProfileDocument", user.pk)
+    if settings.ES_LIVE_INDEXING:
+        for user in instance.subscribed_users.all():
+            index_object.delay("ProfileDocument", user.pk)
 
 
 @receiver(post_delete, sender=Profile)
 def handle_profile_delete(instance, **kwargs):
-    delete_object.delay("ProfileDocument", instance.pk)
+    if settings.ES_LIVE_INDEXING:
+        delete_object.delay("ProfileDocument", instance.pk)
 
 
 @receiver(post_save, sender=Thread)
 def handle_forum_thread_save(instance, **kwargs):
-    for post in instance.post_set.all():
-        index_object.delay("ForumDocument", post.id)
+    if settings.ES_LIVE_INDEXING:
+        for post in instance.post_set.all():
+            index_object.delay("ForumDocument", post.id)
 
 
 @receiver(post_delete, sender=Thread)
 def handle_forum_thread_delete(instance, **kwargs):
-    for post in instance.post_set.all():
-        delete_object.delay("ForumDocument", post.id)
+    if settings.ES_LIVE_INDEXING:
+        for post in instance.post_set.all():
+            delete_object.delay("ForumDocument", post.id)
 
 
 @receiver(post_save, sender=Post)
 def handle_forum_post_save(instance, **kwargs):
-    index_object.delay("ForumDocument", instance.id)
+    if settings.ES_LIVE_INDEXING:
+        index_object.delay("ForumDocument", instance.id)
 
 
 @receiver(post_delete, sender=Post)
 def handle_forum_post_delete(instance, **kwargs):
-    delete_object.delay("ForumDocument", instance.id)
+    if settings.ES_LIVE_INDEXING:
+        delete_object.delay("ForumDocument", instance.id)
--- a/kitsune/search/v2/tests/test_documents.py
+++ b/kitsune/search/v2/tests/test_documents.py
@@ -1,0 +1,42 @@
+from kitsune.sumo.tests import TestCase
+from kitsune.questions.tests import (
+    QuestionFactory,
+    QuestionVoteFactory,
+    AnswerFactory,
+    AnswerVoteFactory,
+)
+from kitsune.search.v2.documents import QuestionDocument, AnswerDocument
+
+
+class QuestionDocumentTests(TestCase):
+    def test_annotation_has_correct_counts(self):
+        question = QuestionFactory()
+        [f(question=question) for f in [QuestionVoteFactory] * 4]
+
+        document = QuestionDocument.prepare(question)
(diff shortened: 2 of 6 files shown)


kitsune/search/v2/signals.py
```diff
@@ -1,54 +1,75 @@
...
 from django.dispatch import receiver
 from django.contrib.auth.models import User, Group
+from django.conf import settings
 from kitsune.users.models import Profile
 from kitsune.search.v2.es7_utils import index_object, delete_object
 from kitsune.products.models import Product
 from kitsune.forums.models import Thread, Post
 
 
+def skip():
```
akatsoulas, line 11: I don't believe this adds much value. On the contrary makes the code a bit more difficult to read. Instead of having `if skip` you can have again a single line `if not setting.ES_LIVE_INDEXING`. Even better you can reduce it to `if settings.ES_LIVE_INDEXING`: index_object.....

akatsoulas: Why is this flag needed?

akatsoulas (commented):
I believe there were a couple of minor comments in the merged PR about the performance that could be added in this one.

kitsune/search/v2/signals.py
LeoMcA: Yeah, I wasn't entirely comfortable with the formatting of this.

It's needed because our tests, by default, set `ES_LIVE_INDEXING` to `False`, so we're not slowing down our tests with unnecessary writes to ES. However, there are some tests where we'll need/want live indexing to happen, where we set `ES_LIVE_INDEXING` to `True`, and so unless we run those separately, we need to check what the value of `ES_LIVE_INDEXING` is every time the function gets executed. If we're alright executing tests like:

```
ES_LIVE_INDEXING=False ./manage.py test --exclude-es-tests
ES_LIVE_INDEXING=True ./manage.py test --exclude-non-es-tests
```

Then we can check the flag once in `kistune/search/apps.py` like:
```
def ready(self):
    if settings.ES_LIVE_INDEXING:
        from kitsune.search.v2 import signals
```

LeoMcA:
I think I covered all the comments from the other PR here. I've rebased, removing the black changes from the diff, to make the changes in this PR clearer.

kitsune/search/v2/signals.py
akatsoulas: How about using the `CELERY_ALWAYS_EAGER` flag?

LeoMcA: `CELERY_TASK_ALWAYS_EAGER`? That's always set to `True` in tests, so it doesn't help, and I wouldn't want to mix separate concerns - we want every combination of it and `ES_LIVE_INDEXING` to be possible for different situations.

akatsoulas: I am not sure if we are going to need a finer control over this in the future but for the current usage `CELERY_ALWAYS_EAGER` is more than enough.

```python
if settings.CELERY_ALWAYS_EAGER: 
    return
index_object.delay
```

Another way of looking at this is that you don't want a task to run during your tests. If there is a specific test that needs to actually index we can override the celery task. This functionality is executed through Celery and triggered by signals. It's not related to ES. You could also stop signal propagation but it's more complex than just controlling Celery.

kitsune/search/v2/tests/test_documents.py
```diff
@@ -0,0 +1,42 @@
+from kitsune.sumo.tests import TestCase
```
akatsoulas, line 1: nit: How about `test_annotations` so that we can categorize tests per operation? Otherwise we will end up with a lot of tests here

kitsune/search/v2/tests/test_documents.py
```diff
@@ -0,0 +1,42 @@
...
+    AnswerFactory,
+    AnswerVoteFactory,
+)
+from kitsune.search.v2.documents import QuestionDocument, AnswerDocument
+
+
+class QuestionDocumentTests(TestCase):
+    def test_annotation_has_correct_counts(self):
+        question = QuestionFactory()
+        [QuestionVoteFactory(question=question) for _ in range(4)]
```
akatsoulas, line 14: super nit but maybe more fun to write:
```python
[f() for f in [QuestionVoteFactory(question=question)]*4]
```

akatsoulas (approved):
r+wc

kitsune/search/v2/signals.py
LeoMcA: Sure, in most cases `ES_LIVE_INDEXING == not CELERY_TASK_ALWAYS_EAGER`, but there are important ones where that doesn't hold true.

For instance, I would've got very frustrated while developing the indices if I couldn't have both `ES_LIVE_INDEXING` and `CELERY_TASK_ALWAYS_EAGER` set to `True`, as I would've had to kill and restart celery every time I made a code change.

We also need both to be `True` when doing tests that do require signal-based indexing - we don't want turning live signalling on in a test to *also* cause that test to start attempting to execute something in celery.

akatsoulas: We can re-import it if/when we need it.

akatsoulas: Live indexing happens through signals. A signal spawns a celery task to do the update. If we want this to happen, we set the appropriate flag to True/False. I don't see why we need a second flag at the moment. Either we will let the task run when the signal fires or not. This has nothing to do with ES for now at least

LeoMcA: But then how do I specify that I want live indexing to happen, but I don't want tasks to be executed locally instead of sent to the celery queue, which is necessary when developing or debugging live indexing. If everything is controlled by `CELERY_TASK_ALWAYS_EAGER` there's no way of doing that.

`CELERY_TASK_ALWAYS_EAGER` should (and currently does) control one thing: whether tasks are executed in celery, or locally. It causing other side-effects is just confusing. If I set it to True to debug some other celery task, I have no expectation that it'll disable ES live indexing. It doesn't disable any other signals, why should it disable the ES ones?

`ES_LIVE_INDEXING` already exists, and does exactly what it says - it controls whether live indexing happens or not. Whether that happens in celery or locally has nothing to do with whether it happens or not.

akatsoulas: Regardless of the ES_LIVE_INDEXING, this is not the place to use this. If you want to use this flag, it needs to go into the index_object. It has nothing to do with either celery or the signal

LeoMcA: > It has nothing to do with either celery or the signal

Perhaps we're understanding different things by "live indexing". In my mind live indexing === signal based indexing (ensuring that the ES state matches the DB state), as opposed to bulk indexing being indexing what we tell it to index when we tell it to index. So in my mind `ES_LIVE_INDEXING` should **only** control the signal and whether it does something or not (or, equivalently, if it ever gets registered/called in the first place or not).

> If you want to use this flag, it needs to go into the index_object.

On a practical level, if we place it in `index_object` and `index_objects_bulk` then, if live indexing were disabled but we wanted to do a bulk re-index, we'd have to run the command like so:
```
ES_LIVE_INDEXING=True ./manage.py es7_reindex
```
Which is nonsensical (to me): live indexing shouldn't have to be enabled for bulk indexing to work.

However, we could DRY things by having `index_object_live` and `index_objects_bulk` functions which are like:
```
def index_object_live(*args, **kwargs):
    if settings.ES_LIVE_INDEXING:
        index_object.delay(*args, **kwargs)
```

With signals becoming like:
```
def handle_foobar_save(instance, **kwargs):
    index_object_live("FooBarDocument", instance.pk)
```

So as to avoid spawning celery tasks which are a no-op.

kitsune/search/v2/tests/test_documents.py
LeoMcA: IMO we should wait until that becomes a problem until we reorganise the tests

kitsune/search/v2/signals.py
LeoMcA: > However, we could DRY things by having index_object_live and index_objects_bulk_live functions

But I'd be inclined to do this in a later PR as part of https://github.com/mozilla/sumo-project/issues/663

akatsoulas: I believe the naming and the context is wrong here but in any case this is the way it's used in the rest of the project, I don't believe it makes any sense to change it now b/c it adds more complexity and the moment. This flag is for the controlling updates from the project to ES and not for ES. Let's get this merged though.

akatsoulas: r+ without the change

PR #4552, opened 2020-10-12
Prepare for using EKS

The-smooth-operator:
This PR is a preparation for moving to EKS. 
We document the differences, promote the API version used for the deployment (backwards compatible) and improve the deployment script to cope with the new configuration.

### DIFF
--- a/k8s/README.md
+++ b/k8s/README.md
@@ -19,7 +19,7 @@ The production branch maintains an image that can be released in production with
 
 ## Deploying SUMO
 
-#### Setup
+### Setup
 
 - Move to the SUMO `k8s` directory
 
@@ -34,18 +34,6 @@ The production branch maintains an image that can be released in production with
   pip install -r ./requirements.txt
   ```
 
-- Create symbolic links to `kubectl` and `kubeconfig` to use in each region
-
-  e.g. `ln -s ~/bin/kubectl ./regions/frankfurt/kubectl`
-
-  Files to be created:
-    - ./regions/frankfurt/kubectl
-    - ./regions/frankfurt/kubeconfig
-    - ./regions/oregon-a/kubectl
-    - ./regions/oregon-a/kubeconfig
-    - ./regions/oregon-b/kubectl
-    - ./regions/oregon-b/kubeconfig
-
 - Create symbolic links to secrets for each namespace and region.
 
   e.g. `ln -s ~/sumo-encrypted/k8s/secrets/frankfurt/sumo-dev-secrets.yaml ./regions/frankfurt/dev-secrets.yaml`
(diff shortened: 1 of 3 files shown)


k8s/README.md
```diff
@@ -34,7 +34,7 @@ The production branch maintains an image that can be released in production with
   pip install -r ./requirements.txt
   ```
 
-- Create symbolic links to `kubectl` and `kubeconfig` to use in each region
+- If you are deploying to a Kops cluster, create symbolic links to `kubectl` and `kubeconfig` to use in each region
```
akatsoulas, line 37: Can we maybe create two sections similar to what you have to the EKS cluster?

Something like:
`Accessing a kops cluster`
the existing content

`Accessing an EKS cluster`
new content

Probably this will require some duplication of the available information but it will be probably easier to understand and to maintain

k8s/README.md
```diff
@@ -252,3 +252,21 @@ This must then be parsed and de-pickled:
 import pickle, base64, json
 pickle.loads(base64.decodestring(json.loads(l[0])['body']))
 ```
+
+## Accessing an EKS cluster
+In order to establish a connection with a cluster running in EKS, you need to first download its KUBECONFIG file. You can do this running:
+```
+aws eks update-kubeconfig --name <cluster-name> --region <region>
```
akatsoulas, line 259: Shaw we add anything about `maws` here?

k8s/commander.sh
```diff
@@ -17,8 +17,16 @@ function deploy {
     DEPLOY_SECRETS=${5:-NO}
     K8S_NAMESPACE="sumo-${REGION_ENV}"
 
-    export KUBECTL_BIN="./regions/${REGION}/kubectl"
-    export KUBECONFIG="./regions/${REGION}/kubeconfig"
+    if [ -f "./regions/${REGION}/kubectl" ] ; then
+        KUBECTL_BIN="./regions/${REGION}/kubectl"
+    else
+        KUBECTL_BIN=$(which kubectl)
```
akatsoulas, line 23: Shouldn't  we have a requirement to use a kubectl version with an offset of 1 from the apiserver?

k8s/README.md
The-smooth-operator: ok

k8s/README.md
The-smooth-operator: yes, good point. I did not write anything, as theoretically you could log into AWS via other methods. But yeah makes sense to document it

k8s/commander.sh
The-smooth-operator: Do you want me to add the logic of checking the versions in the script, or document the requirement?

The-smooth-operator:
@akatsoulas I've addressed you comments regarding docs. Let me know if you want me to change anything in there.
Also, what do you want me to do with the version check requirement. I don't mind doing one thing or the other.

I'm not squashing the commits now, but I'll be doing so once we are done with the review and changes.

k8s/commander.sh
akatsoulas: Documenting the requirement somewhere in this readme should be enough

The-smooth-operator: documented here https://github.com/mozilla/kitsune/pull/4552/files#diff-1d8ece659445546a89a71faddc37a9b3fca6fc8e8870b3bb40cadb707e54e946R51

The-smooth-operator:
merging this changes. Anyway I'll be keeping the docs up to date as I move forward with the migration

k8s/commander.sh
```diff
@@ -17,8 +17,16 @@ function deploy {
...
+    fi
+    export KUBECTL_BIN
+
+    if [ -f "./regions/${REGION}/kubeconfig" ] ; then
+        export KUBECONFIG="./regions/${REGION}/kubeconfig"
```
akatsoulas, line 28: You could move this export in line 22. Ifl line 20 is true then no need for this check.

akatsoulas (approved):
r+wc 
Please merge these commits into one before merging. I would also suggest to use this script to do a deployment to dev for example especially for the kops cluster where current prod is running

k8s/commander.sh
The-smooth-operator: Well, line 20 checks for kubectl, this checks for a configuration file. 
I think it's safer this way because we do an extra check. Also this doesn't add extra complexity or anything like that

The-smooth-operator:
I've deployed to test using the new version of the commander script and the newer Deployment object.
The deployment went well.

PR #4554, opened 2020-10-15
use es7 indices in instant search

LeoMcA:


kitsune/search/v2/views.py
```diff
@@ -0,0 +1,114 @@
...
+        )
+
+    cleaned = search_form.cleaned_data
+
+    # get language
+    language = locale_or_default(cleaned["language"] or request.LANGUAGE_CODE)
+    lang_name = settings.LANGUAGES_DICT.get(language.lower()) or ""
+
+    # get product
+    product = Product.objects.filter(slug__in=cleaned["product"]).first()
```
akatsoulas, line 34: Do we always have a single product here?

kitsune/search/v2/search.py
```diff
@@ -0,0 +1,209 @@
...
+    def make_result(self, hit):
+        """Transform a hit into a result dictionary."""
+        pass
+
+    @staticmethod
+    def extract_highlight(hit):
+        highlight = getattr(hit.meta, "highlight", None)
+        if highlight:
+            return bleach.clean(
+                next(iter(highlight.to_dict().values()))[0],
```
akatsoulas, line 50: could you maybe explain what's going on here?

kitsune/search/v2/search.py
```diff
@@ -0,0 +1,209 @@
...
+                tags=["strong"],
+                strip=True,
+            )
+        return None
+
+
+class AAQConfig(BaseConfig):
```
akatsoulas, line 57: Shall we rename this to `AAQSearch` or something within this context? Having the word config here implies that it's about configuration but there's also a method that builds the result for the AAQ.

kitsune/search/v2/search.py
```diff
@@ -0,0 +1,209 @@
...
+    highlight_fields = ["summary.{}", "content.{}"]
+
+    def product_filter(self, product):
+        return Q("match", product_ids=product.id)
+
+    def make_result(self, hit):
+        locale = hit.locale
+
+        summary = self.extract_highlight(hit)
+        if not summary:
```
akatsoulas, line 116: I would rewrite this focusing first on `hit.summary[locale]` and `hit.content`. You can determine which of the 2 exists and then maybe run it through bleach. That should remove the need for two identical `if`

akatsoulas:
Merely a first pass. I will continue reviewing this.

kitsune/search/v2/search.py
```diff
@@ -0,0 +1,238 @@
...
+        """An array of fields to highlight."""
+        pass
+
+    @property
+    def filter(self):
+        """A filter which returns all documents of this type to search over."""
+        return Q("term", _index=self.index)
+
+    @abstractmethod
+    def product_filter(self, product):
```
akatsoulas, line 62: s/product_filter/match_product

kitsune/search/v2/search.py
```diff
@@ -0,0 +1,238 @@
...
+        """Transform a hit into a result dictionary."""
+        pass
+
+
+class AAQConfig(BaseConfig):
+    """Search config for AAQ."""
+
+    index = QUESTION_INDEX_NAME
+    fields = [
+        "question_title.{}^4",
```
akatsoulas, line 77: comments pls about the boosting part

kitsune/search/v2/search.py
```diff
@@ -0,0 +1,238 @@
...
+    ]
+    highlight_fields = ["question_content.{}", "answer_content.{}"]
+
+    @property
+    def filter(self):
+        return Q(
+            "bool",
+            filter=super().filter,
+            # exclude AnswerDocuments from the search:
+            must_not=Q("exists", field="updated"),
```
akatsoulas, line 89: can we use a type query here?

kitsune/search/v2/views.py
akatsoulas: why first?

kitsune/search/v2/search.py
```diff
@@ -0,0 +1,238 @@
...
+            "title": hit.question_title[locale],
+            "search_summary": summary,
+            "last_updated": datetime.fromisoformat(hit.question_updated),
+            "is_solved": hit.question_has_solution,
+            "num_answers": num_answers,
+            "num_votes": hit.question_num_votes,
+        }
+
+
+class KBConfig(BaseConfig):
```
akatsoulas, line 119: Maybe `KBSearch`? Similarly for the other classes

kitsune/search/v2/search.py
```diff
@@ -0,0 +1,238 @@
...
+        summary = first_highlight(hit)
+        if not summary and hasattr(hit, "summary"):
+            summary = getattr(hit.summary, locale, None)
+        if not summary:
+            summary = hit.content[locale][:500]
+        summary = sanitize(summary)
+
+        return {
+            "type": "document",
+            "url": "/{}/kb/{}".format(locale, hit.slug[locale]),
```
akatsoulas, line 152: can we use reverse here?

kitsune/search/v2/search.py
```diff
@@ -0,0 +1,238 @@
...
+    """Create an ES7 Search."""
+
+    def __init__(self, query, language, product=None, page=1):
+        self.config_classes = []
+        self.query = query
+        self.language = language
+        self.product = product
+        self.page = page
+
+    def add_config(self, config):
```
akatsoulas, line 169: `config` feels wrong here. Maybe search_type ? Feel free to add something better since the proposed one is not great either

kitsune/search/v2/search.py
```diff
@@ -0,0 +1,238 @@
...
+        return {
+            "type": "document",
+            "url": "/{}/kb/{}".format(locale, hit.slug[locale]),
+            "score": hit.meta.score,
+            "title": hit.title[locale],
+            "search_summary": summary,
+        }
+
+
+class Search:
```
akatsoulas, line 159: why not subclass DSL Search add append the methods we need?

akatsoulas: Regardless of the previous comment this should be prefixed with `Sumo`

akatsoulas (commented):
There are already a few comments in here. I will continue with the search logic once we sync on what's already in here

kitsune/search/v2/search.py
```diff
@@ -0,0 +1,238 @@
...
+from kitsune.search.config import WIKI_DOCUMENT_INDEX_NAME, QUESTION_INDEX_NAME
+from kitsune.wiki.config import CANNED_RESPONSES_CATEGORY, TEMPLATES_CATEGORY
+
+HIGHLIGHT_TAG = "strong"
+
+
+def first_highlight(hit):
+    highlight = getattr(hit.meta, "highlight", None)
+    if highlight:
+        return bleach.clean(
```
akatsoulas, line 16: let's leverage the `sanitize` method below

kitsune/search/v2/search.py
```diff
@@ -0,0 +1,238 @@
...
+            "bool",
+            filter=super().filter,
+            # exclude AnswerDocuments from the search:
+            must_not=Q("exists", field="updated"),
+        )
+
+    def product_filter(self, product):
+        return Q("term", question_product_id=product.id)
+
+    def make_result(self, hit, locale):
```
akatsoulas, line 95: comments pls

kitsune/search/v2/search.py
```diff
@@ -0,0 +1,238 @@
...
+        summary = first_highlight(hit)
+        if not summary:
+            summary = hit.question_content[locale][:500]
+        summary = sanitize(summary)
+
+        num_answers = 0
+        answer_content = getattr(hit, "answer_content", None)
+        if answer_content:
+            num_answers = len(answer_content[locale])
```
akatsoulas, line 104: you can assign this directly to the return statement

kitsune/search/v2/search.py
```diff
@@ -0,0 +1,238 @@
...
+        return Q("term", question_product_id=product.id)
+
+    def make_result(self, hit, locale):
+        summary = first_highlight(hit)
+        if not summary:
+            summary = hit.question_content[locale][:500]
+        summary = sanitize(summary)
+
+        num_answers = 0
+        answer_content = getattr(hit, "answer_content", None)
```
akatsoulas, line 102: is this check for the case that we have a question without answers?

kitsune/search/v2/search.py
```diff
@@ -0,0 +1,238 @@
...
+        if answer_content:
+            num_answers = len(answer_content[locale])
+
+        return {
+            "type": "question",
+            "url": "/questions/{}".format(hit.question_id),
```
akatsoulas, line 108: `reverse()`?

kitsune/search/v2/search.py
```diff
@@ -0,0 +1,238 @@
...
+        "content.{}^1",
+    ]
+    highlight_fields = ["summary.{}", "content.{}"]
+
+    @property
+    def filter(self):
+        return Q(
+            "bool",
+            filter=super().filter,
+            must_not=Q("terms", category=[TEMPLATES_CATEGORY, CANNED_RESPONSES_CATEGORY]),
```
akatsoulas, line 136: do we need this at all? Currently I am not sure if these categories are used at all

kitsune/search/v2/search.py
akatsoulas: Similar comments with the `AAQConfig` apply here too

kitsune/search/v2/search.py
```diff
@@ -0,0 +1,238 @@
...
+    def add_config(self, config):
+        self.config_classes.append(config())
+
+    def execute(self):
+        # determine indices to search over, and create DSL search object
+        index = ",".join(self._getattr("index"))
+        search = DSLSearch(using=es7_client(), index=index)
+
+        # add filters
+        filters = self._getattr("filter")
```
akatsoulas, line 178: let's use something more obvious and less cryptic

kitsune/search/v2/search.py
```diff
@@ -0,0 +1,238 @@
...
+        search = search.query("bool", should=filters)
```
akatsoulas, line 179: I would start the other way around, construct all the queries into a compound one and the call the search. IMO it's easier to follow up how the **queries are being built as you proceed.

kitsune/search/v2/search.py
```diff
@@ -0,0 +1,238 @@
...
+
+        # add product filters
+        if self.product:
+            product_filters = self._getattr("product_filter", self.product)
+            search = search.query("bool", should=product_filters)
+
+        # add query
+        fields = self._localize_fields(self._getattr("fields"))
```
akatsoulas, line 187: this does not localize any field but rather retrieving the localized version of it

akatsoulas: this is clean but it took me a moment to figure out what `_getattr('fields')` does here. I am not particularly suggesting something here - just a general remark

kitsune/search/v2/search.py
```diff
@@ -0,0 +1,238 @@
...
+        search = search.query(
+            "simple_query_string", query=self.query, default_operator="AND", fields=fields
+        )
+
+        # add highlights
+        highlight_fields = self._localize_fields(self._getattr("highlight_fields"))
+        search = search.highlight(
+            *highlight_fields,
+            type="fvh",
```
akatsoulas, line 196: Could you unpack your thinking here?

kitsune/search/v2/views.py
```diff
@@ -0,0 +1,114 @@
...
+from kitsune import search as constants
+from kitsune.search.forms import SimpleSearchForm
+from kitsune.sumo.api_utils import JSONRenderer
+from kitsune.sumo.utils import build_paged_url
+from kitsune.products.models import Product
+from kitsune.search.views import _fallback_results
+from kitsune.search.utils import locale_or_default
+from kitsune.search.v2.search import Search, KBConfig, AAQConfig
+
+
```
akatsoulas, line 16: pls add comments where it's not that straightforward why something it's the way it is

kitsune/search/v2/views.py
```diff
@@ -0,0 +1,114 @@
...
+    if product:
+        data["product"] = product.slug
+    if not results:
+        data["message"] = _("No pages matched the search criteria")
+
+    json_data = JSONRenderer().render(data)
+    return HttpResponse(json_data, content_type="application/json")
+
+
+def _make_pagination(request, page, total):
```
akatsoulas, line 89: why do we need this?

kitsune/search/v2/views.py
```diff
@@ -0,0 +1,114 @@
...
+        "lang_name": lang_name,
+        "products": [
+            {"slug": p.slug, "title": p.title} for p in Product.objects.filter(visible=True)
+        ],
+        "pagination": _make_pagination(request, page, total),
+    }
+    if product:
+        data["product"] = product.slug
+    if not results:
+        data["message"] = _("No pages matched the search criteria")
```
akatsoulas, line 83: let's move this into a constant

kitsune/sumo/static/sumo/js/instant_search.js
```diff
@@ -3,7 +3,12 @@
   var searchTimeout;
   var locale = $('html').attr('lang');
 
-  var search = new k.Search('/' + locale + '/search');
+  if (window.localStorage.getItem("enable_search_v2") === "true") {
```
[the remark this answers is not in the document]
LeoMcA: I think sessionStorage might be quite confusing, as it's bound to the page session, not the browser one - so opening a new tab to test something else, for instance, would require re-toggling v2 being enabled.

Also an advantage of using localStorage is, at least to begin with, we don't need to build a UI to toggle this and can instruct the team to add the value through Firefox devtools.

kitsune/search/v2/search.py
LeoMcA: Yeah, results from these categories were popping up locally without it - perhaps we should filter by the categories we do want to search over, rather than those which we don't

kitsune/search/v2/views.py
LeoMcA: It's needed to render the pagination section at the bottom of the instant search

kitsune/search/v2/search.py
```diff
@@ -0,0 +1,209 @@
...
+
+from elasticsearch_dsl import Search as DSLSearch, Q
+from kitsune.search.v2.es7_utils import es7_client
+from kitsune.search.config import WIKI_DOCUMENT_INDEX_NAME, QUESTION_INDEX_NAME
+
+
+class BaseConfig(ABC):
+    """Semi-abstract base class for Search configs."""
+
+    @property
```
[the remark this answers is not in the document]
LeoMcA: To define these as abstract properties - I think having them as properties is important to clarify that they don't perform an action, they just return a result

kitsune/search/v2/search.py
```diff
@@ -0,0 +1,209 @@
...
+        pass
+
+    @property
+    @abstractmethod
+    def highlight_fields(self):
+        """Array of fields to highlight."""
+        pass
+
+    @property
+    def filter(self):
```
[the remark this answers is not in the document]
LeoMcA: I think filter is the most appropriate word here - we're using this to filter the documents to search over. In my mind, match refers more to the process of searching itself: first we filter by the documents to search, then we match on those documents.

kitsune/search/v2/views.py
LeoMcA: I think the API allows multiple products, however the way its currently used in instant search only allows for a single product (or all products) to be searched.

kitsune/search/v2/search.py
LeoMcA: This has changed a little bit to handle the case where a document has no summary attribute, but imo the double `if not summary` is the clearest way of expressing this: if we haven't got a summary yet, do this, if we still haven't got a summary after that, do this other thing.

kitsune/search/v2/search.py
```diff
@@ -0,0 +1,238 @@
...
+    @property
+    @abstractmethod
+    def highlight_fields(self):
+        """An array of fields to highlight."""
+        pass
+
+    @property
+    def filter(self):
+        """A filter which returns all documents of this type to search over."""
+        return Q("term", _index=self.index)
```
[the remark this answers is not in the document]
LeoMcA: I prefer using `DSLQ` for everything to avoid a bazillion different imports, but this is purely style driven, can change if necessary.

kitsune/search/v2/search.py
LeoMcA: Elastic itself doesn't know the types of QuestionDocuments and AnswerDocuments, just some have documents in question index have certain fields, whereas others don't.

kitsune/search/v2/search.py
```diff
@@ -0,0 +1,238 @@
...
+
+def first_highlight(hit):
+    highlight = getattr(hit.meta, "highlight", None)
+    if highlight:
+        return bleach.clean(
+            # `highlight` is of type AttrDict, which is internal to elasticsearch_dsl
+            # when converted to a dict, it's like:
+            # `{ 'es_field_name' : ['highlight1', 'highlight2'], 'field2': ... }`
+            # so here we're getting the first item in the first value in that dict:
+            next(iter(highlight.to_dict().values()))[0],
```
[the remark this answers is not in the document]
LeoMcA: Elastic doesn't seem to have a way to specify "highlight any one of these fields" - just "highlight all these fields and only return one snippet for each" - however it can still fail to highlight a field at all, if nothing matches, so it's left to us to get the first matching snippet ourselves (with this monstrous line of code).

kitsune/search/v2/search.py
LeoMcA: I'm not totally clear what you mean here, however things will change a little with my latest changes, are things still hard to follow?

kitsune/search/v2/search.py
LeoMcA: I added some comments which should hopefully clarify.

kitsune/search/v2/base.py
```diff
@@ -126,3 +129,91 @@ def prepare_locale(self, instance):
...
+    Implements the run() function, which will perform the search.
+
+    Child classes should define values for the various abstract properties this
+    class has, relevant to the documents the child class is searching over.
+    """
+
+    def __init__(self, locale, product=None):
+        self.locale = locale
+        self.product = product
+        self.results_per_page = 10
```
akatsoulas, line 146: There is the `SEARCH_RESULTS_PER_PAGE` setting available

kitsune/search/v2/base.py
```diff
@@ -126,3 +129,91 @@ def prepare_locale(self, instance):
...
+
+        # perform search
+        self.hits = search.execute().hits
+
+        self.total = self.hits.total.value if self.hits else 0
+        self.results = [self.make_result(hit) for hit in self.hits]
+
+        return self
+
+    @property
```
akatsoulas, line 192: I am not sure about the `@property` in all the abstractmethods in here. Any particular reason you r going down this road?

akatsoulas (commented):
I haven't gone through the details but this is much simpler and cleaner. Really nice work!

akatsoulas (approved):
r+wc with the changes we discussed in our meeting

PR #4555, opened 2020-10-15
Make fields more flat

akatsoulas:
* Replace categories with keyword.

### DIFF
--- a/kitsune/search/v2/documents.py
+++ b/kitsune/search/v2/documents.py
@@ -1,15 +1,15 @@
-from elasticsearch_dsl import connections, field, InnerDoc
-from django.db.models import Prefetch, Count, Q
+from django.db.models import Count, Prefetch, Q
+from elasticsearch_dsl import InnerDoc, connections, field
+
+from kitsune.forums.models import Post
+from kitsune.questions.models import Answer, Question
 from kitsune.search import config
 from kitsune.search.v2.base import SumoDocument
 from kitsune.search.v2.es7_utils import es7_client
 from kitsune.search.v2.fields import SumoLocaleAwareTextField
+from kitsune.users.models import Profile
 from kitsune.wiki import models as wiki_models
 from kitsune.wiki.config import REDIRECT_HTML
-from kitsune.questions.models import Question, Answer
-from kitsune.users.models import Profile
-from kitsune.forums.models import Post
-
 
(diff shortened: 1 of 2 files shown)


LeoMcA:
As an abstraction, I think this is good - but I'm not sure if we want the abstraction. My thinking has been we probably want to move everything towards the implementation used in `QuestionDocument` which doesn't use any `InnerDoc`s, and just uses (lists of) ids. We don't need the title of a product or group to filter on them or aggregate them, their SQL id is enough.

However, storing titles for products would be useful if we want to use product titles (and even descriptions) in search, and store them properly in a `SumoLocaleAwareTextField` as well. So if we matched across `content.en-US` and `product.title.en-US` then a search for "enable do not track in firefox for android" (without any filtering) could cause an answer like "here's how to enable do not track..." in the Firefox for Android AAQ to match higher than "here's how to enable do not track in firefox..." in the Firefox AAQ.

But then again, doing this means we have to reindex everything if a product name changes, or a localisation of a product name changes, and it might be better to detect in python if someone has placed a product name in their query and then suggest they filter by that, or boost the scores of documents with that product's id instead.

akatsoulas:
@LeoMcA I updated the PR to change a few fields and make others more flat by using only IDs.

kitsune/search/v2/documents.py
```diff
@@ -129,7 +137,7 @@ class Index:
         using = config.DEFAULT_ES7_CONNECTION
 
     def prepare_question_tag_id(self, instance):
```
LeoMcA, line 139: I'd also rename the field to `question_tag_ids` to be consistent with our other naming

kitsune/search/v2/documents.py
```diff
@@ -272,11 +252,11 @@ def prepare_avatar(self, instance):
...
 
-    def prepare_products(self, instance):
-        return [ProductInnerDoc.prepare(product) for product in instance.products.all()]
+    def prepare_product_ids(self, instance):
+        return [product.id for product in instance.products.all()]
 
-    def prepare_groups(self, instance):
-        return [GroupInnerDoc.prepare(groups) for groups in instance.user.groups.all()]
+    def prepare_group_ids(self, instance):
+        return [group.id for group in instance.groups.all()]
```
LeoMcA, line 259: This needs to be `instance.user.groups.all()`

LeoMcA (approved):
r+wc

PR #4556, opened 2020-10-20
add before/after arguments to es7_reindex command 

LeoMcA:
WIP solution to https://github.com/mozilla/sumo-project/issues/667

While pushing this I realised we might instead be able to solve the same "celery went down while we were live indexing" problem by re-indexing based on the `indexed_on` field we're adding in `base.py`. It would get around having to implement `get_queryset_between` methods for each document class.

Also, looks like `scripts/cron.py` never got run through black. That should be done on master and merged into the es branch to avoid horrendous merge conflicts.

Will pick this up tomorrow.


LeoMcA:
> While pushing this I realised we might instead be able to solve the same "celery went down while we were live indexing" problem by re-indexing based on the indexed_on field we're adding in base.py.

Quickly realised this wouldn't work well, as it would require comparisons between what exists in DB and what exists in ES to cover documents which were never indexed in the first place.

> It would get around having to implement get_queryset_between methods for each document class.

I've got an alternative solution for this by just getting classes to define an `updated_field_name` attribute.

kitsune/sumo/models.py
```diff
@@ -50,3 +50,29 @@ def __init__(
         return super(LocaleField, self).__init__(
             max_length=max_length, default=default, choices=choices, *args, **kwargs
         )
+
+
+class BetweenQuerySet(models.QuerySet):
```
akatsoulas, line 55: Let's rename this to `SumoRangeQuerySet`

kitsune/search/v2/management/commands/es7_reindex.py
```diff
@@ -55,7 +68,26 @@ def handle(self, *args, **kwargs):
             self.stdout.write("Reindexing: {}".format(dt.__name__))
 
             model = dt.get_model()
-            qs = model.objects.all()
+
+            before = kwargs["updated_before"]
+            after = kwargs["updated_after"]
+            if before or after:
```
akatsoulas, line 74: It looks like this can be omitted

kitsune/sumo/models.py
```diff
@@ -50,3 +50,29 @@ def __init__(
...
+    """
+    QuerySet adding `before(date)` and `after(date)` methods,
+    to filter for rows before or after a datetime.
+
+    The `updated_column_name` property must be defined on a model using this,
+    as that will be used as the column to filter on in the before and after methods.
+    """
+
+    def before(self, date):
```
akatsoulas, line 64: Why not combining these into a `range(self, start, stop)` with stop defaulting to `now` if none is provided?

akatsoulas: And if we go down this road, we probably don't need the manager at all. We could use directly into the management command and if we need it in more places (in the future or now) we could add a method/property in the base model.

akatsoulas (changes requested):

kitsune/sumo/models.py
LeoMcA: I'd prefer to go straight for the generic option, but yes something like `BaseModel.object_range(start, stop)` might be better - involving no fiddling around with existing managers on our models.

akatsoulas (approved):

kitsune/settings.py
```diff
@@ -1154,6 +1154,7 @@ def filter_exceptions(event, hint):
 DMS_UPDATE_CONTRIBUTOR_METRICS = config("DMS_UPDATE_CONTRIBUTOR_METRICS", default=None)
 DMS_AUTO_ARCHIVE_OLD_QUESTIONS = config("DMS_AUTO_ARCHIVE_OLD_QUESTIONS", default=None)
 DMS_REINDEX = config("DMS_REINDEX", default=None)
+DMS_REINDEX_ES7 = config("DMS_REINDEX", default=None)
```
LeoMcA, line 1157: Oops, typo here

PR #4560, opened 2020-10-21
Merge child KB documents into parents.

akatsoulas:
* Add Keyword locale aware field
* Use update with doc_as_upsert instead of save

### DIFF
--- a/kitsune/search/v2/base.py
+++ b/kitsune/search/v2/base.py
@@ -1,25 +1,33 @@
 from datetime import datetime
+
 from django.utils import timezone
 from elasticsearch_dsl import Document as DSLDocument
-from elasticsearch_dsl import InnerDoc, field, MetaField
+from elasticsearch_dsl import InnerDoc, MetaField, field
 
 
 class SumoDocument(DSLDocument):
     """Base class with common methods for all the different documents."""
 
     indexed_on = field.Date()
+    updated_on = field.Date()
 
     class Meta:
         # ignore fields if they don't exist in the mapping
         dynamic = MetaField("false")
 
     @classmethod
-    def prepare(cls, instance):
-        """Prepare an object given a model instance"""
+    def prepare(cls, instance, **kwargs):
+        """Prepare an object given a model instance.
+
+        merge_docs: Controls whether multiple docements will be merged into the parent one.
+        Currently used by the Knowledge Base
+        """
         obj = cls()
         doc_mapping = obj._doc_type.mapping
         fields = [f for f in doc_mapping]
         fields.remove("indexed_on")
+        fields.remove("updated_on")
+        merge_docs = kwargs.pop("merge_docs", False)
 
         # Loop through the fields of each object and set the right values
         for f in fields:
@@ -37,8 +45,10 @@ class SumoDocument(DSLDocument):
                 # if the field is an Object but the value isn't an InnerDoc
                 # or a list containing an InnerDoc then we're dealing with locales
                 locale = obj.prepare_locale(instance)
-                if locale and value:
+                # Check specifically against None, False is a valid value
+                if locale and (value is not None):
                     obj[f] = {locale: value}
+
             else:
                 if (
                     isinstance(field_type, field.Date)
@@ -48,10 +58,18 @@ class SumoDocument(DSLDocument):
                     # set is_dst=False to avoid errors when an ambiguous time is sent:
                     # https://docs.djangoproject.com/en/2.2/ref/utils/#django.utils.timezone.make_aware
                     value = timezone.make_aware(value, is_dst=False).astimezone(timezone.utc)
+
+                # if merge_docs is True and we don't have a SumoLocaleAware field
+                # then We need to get the value from the parent
+                if merge_docs and instance.parent:
+                    value = obj.get_field_value(f, instance.parent, prepare_method)
                 setattr(obj, f, value)
 
-        obj.meta.id = instance.pk
         obj.indexed_on = datetime.now(timezone.utc)
+        obj.updated_on = datetime.now(timezone.utc)
+        obj.meta.id = instance.pk
+        if merge_docs and instance.parent:
+            obj.meta.id = instance.parent.pk
 
         return obj
 
--- a/kitsune/search/v2/documents.py
+++ b/kitsune/search/v2/documents.py
@@ -6,7 +6,11 @@ from kitsune.questions.models import Answer, Question
 from kitsune.search import config
 from kitsune.search.v2.base import SumoDocument
 from kitsune.search.v2.es7_utils import es7_client
-from kitsune.search.v2.fields import SumoLocaleAwareTextField
+from kitsune.search.v2.fields import (
+    SumoLocaleAwareBooleanField,
+    SumoLocaleAwareKeywordField,
+    SumoLocaleAwareTextField,
+)
 from kitsune.users.models import Profile
 from kitsune.wiki import models as wiki_models
 from kitsune.wiki.config import REDIRECT_HTML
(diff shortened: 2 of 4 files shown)


kitsune/search/v2/base.py
```diff
@@ -9,8 +9,12 @@ class SumoDocument(DSLDocument):
     indexed_on = field.Date()
 
     @classmethod
-    def prepare(cls, instance):
-        """Prepare an object given a model instance"""
+    def prepare(cls, instance, merge_docs=False):
```
LeoMcA, line 17: To make this a bit more generic, I think we should add a `parent=None` argument, so as to not tie this to the KB's model layout. Indeed, we don't need `merge_docs` any more then, since all the conditionals that are `if merge_docs and instance.parent` can just be `if parent`.

kitsune/search/v2/base.py
```diff
@@ -32,13 +36,21 @@ def prepare(cls, instance):
...
-                if locale and value:
+                # Check specifically against None, False is a valid value
+                if locale and (value is not None):
                     obj[f] = {locale: value}
+
             else:
+                # if merge_docs is True, then we know that we don't have
+                # a SumoLocaleAware field here. We need to get the value from the parent
+                if merge_docs and instance.parent:
+                    value = obj.get_field_value(f, instance.parent, prepare_method)
```
LeoMcA, line 61: This is nice, I think the comment should read more like: "if merge_docs is True, **and** we know that we don't have..."

kitsune/search/v2/documents.py
```diff
@@ -15,30 +19,41 @@
...
 class WikiDocument(SumoDocument):
-    url = field.Keyword()
-    updated = field.Date()
+    """ES Document for Knowledge Base.
+
+    Each document in ES is built on top of a parent KB article.
+    All the children are included in the same document.
+    """
 
+    updated = field.Date()
```
LeoMcA, line 28: Is it all that useful having the updated date of the parent, without those of the children?

kitsune/search/v2/documents.py
```diff
@@ -15,30 +19,41 @@
...
     product_ids = field.Keyword(multi=True)
     topic_ids = field.Keyword(multi=True)
```
LeoMcA, line 30: Are we sure products and topics are the same across parent and children?

kitsune/search/v2/documents.py
```diff
@@ -15,30 +19,41 @@
...
+    locale = field.Keyword()
+    category = field.Keyword()
```
LeoMcA, line 32: Are we sure categories are the same across parents and children?

kitsune/search/v2/documents.py
```diff
@@ -15,30 +19,41 @@
...
+    recent_helpful_votes = field.Integer()
```
LeoMcA, line 33: I'm not sure it's all that useful having recent helpful votes of just the parent.

kitsune/search/v2/documents.py
```diff
@@ -15,30 +19,41 @@
...
 
     # Document specific fields (locale aware)
+    url = SumoLocaleAwareKeywordField(store=True)
+    slug = SumoLocaleAwareKeywordField(store=True)
```
LeoMcA, line 37: Do we need the slug since we have the url?

kitsune/search/v2/documents.py
```diff
@@ -15,30 +19,41 @@
...
-    keywords = SumoLocaleAwareTextField(multi=True)
-
-    locale = field.Keyword()
-    current_id = field.Integer()
-    parent_id = field.Integer()
-    category = field.Keyword()
-    slug = field.Keyword()
-    is_archived = field.Boolean()
-    recent_helpful_votes = field.Integer()
+    keywords = SumoLocaleAwareKeywordField()
```
LeoMcA, line 41: This should either be a LocaleAwareTextField, or we should set `multi=True` here and split the string of keywords at whitespace in a `prepare_keywords` method

kitsune/search/v2/documents.py
```diff
@@ -15,30 +19,41 @@
...
 
+    updated = field.Date()
     product_ids = field.Keyword(multi=True)
     topic_ids = field.Keyword(multi=True)
+    locale = field.Keyword()
+    category = field.Keyword()
+    recent_helpful_votes = field.Integer()
 
     # Document specific fields (locale aware)
+    url = SumoLocaleAwareKeywordField(store=True)
```
LeoMcA, line 36: What does `store=True` do here (and in all the other places)? I can't find it in the docs.

kitsune/search/v2/documents.py
```diff
@@ -15,30 +19,41 @@
...
-
-    locale = field.Keyword()
-    current_id = field.Integer()
-    parent_id = field.Integer()
-    category = field.Keyword()
-    slug = field.Keyword()
-    is_archived = field.Boolean()
-    recent_helpful_votes = field.Integer()
+    keywords = SumoLocaleAwareKeywordField()
+    doc_ids = SumoLocaleAwareKeywordField(store=True)
```
LeoMcA, line 42: Keeping this consistent with our other naming, this should be `doc_id`.

kitsune/search/v2/es7_utils.py
```diff
@@ -92,8 +92,8 @@ def index_object(doc_type_name, obj_id):
     model = doc_type.get_model()
 
     obj = model.objects.get(pk=obj_id)
-    doc = doc_type.prepare(obj)
-    doc.save()
+    payload = doc_type.prepare(obj)
+    doc_type(_id=obj_id).update(payload, doc_as_upsert=True)
```
LeoMcA, line 96: I don't think we want every kind of doc to be sent as an update, I imagine it impacts on performance, and could lead to data persisting in ES if - for example - a user sets their display name to an empty value.

We could add a class attribute here like `merge_docs` and use that here as well as the base `prepare` method - this would also avoid having to add a `prepare` method on the wiki doc.

kitsune/search/v2/es7_utils.py
```diff
@@ -103,7 +103,14 @@ def index_objects_bulk(doc_type_name, obj_ids):
     doc_type = next(cls for cls in get_doc_types() if cls.__name__ == doc_type_name)
 
     objects = doc_type.get_queryset().filter(pk__in=obj_ids)
+    # prepare the docs for indexing
     docs = [doc_type.prepare(obj).to_dict(include_meta=True) for obj in objects]
+    # Replace _source with doc in order to do a bulk update
+    for doc in docs:
+        doc["doc"] = doc["_source"]
+        del doc["_source"]
+        doc.update({"_op_type": "update", "doc_as_upsert": True})
```
LeoMcA, line 112: Same comment re updating here, and we should merge the list comprehension into the loop.

kitsune/search/v2/fields.py
```diff
@@ -16,8 +17,11 @@ def _get_fields(field, locales, **params):
     data = {}
 
     for locale in locales:
-        analyzer = es_analyzer_for_locale(locale)
-        field_obj = field(analyzer=analyzer, **params)
+        if field == Text:
```
LeoMcA, line 20: Am I wrong or would it be more pythonic to do `field is Text` here?

LeoMcA (commented):
First pass:

kitsune/search/v2/base.py
akatsoulas: This would mean though that you will be merging docs every time you have a parent object which may not be what we want I don't see any other use case to be honest but I prefer a toggle for the functionality and not to base the outcome of the prepare method in the existence of an object.

kitsune/search/v2/documents.py
akatsoulas: slug != url b/c the url contains also the locale. I guess it depends on the usage. We can surely construct one from the other. Slug allows another way to extract the document. I don't have a good answer here. It can go both ways

kitsune/search/v2/documents.py
akatsoulas: https://www.elastic.co/guide/en/elasticsearch/reference/current/mapping-store.html

kitsune/search/v2/es7_utils.py
akatsoulas: This would lead to some docs using the update option and others the create one. I feel that whether is an update or not it's not part of the Document. The document needs to go into the index but it should not decide if it's an update or a create operation. I will try to update the code here to create a document if it's not present or update if exists. 

Regarding data persistency, I am not sure I understand the comment

kitsune/search/v2/documents.py
akatsoulas: The idea (for now) it's to retrieve the helpful_votes of the children if we want to aggregate on them. If this is still the case we need to store the helpful votes of the parent document too. Otherwise we will add all of them as we did for the other fields

akatsoulas:
Closing in favor of #4574

PR #4561, opened 2020-10-22
enable django debug toolbar when USE_DEBUG_TOOLBAR=True

LeoMcA:
Set this up while investigating https://github.com/mozilla/sumo-project/issues/670 and thought I might as well commit it rather than have it languish in a branch someplace where I forget it (which I think is what happened last time I set this up).


akatsoulas (approved):
nice!

PR #4567, opened 2020-10-23
remove deactivated accounts from user autocomplete

LeoMcA:
also optimise queries, mostly by removing ordering by length

I decided against using django-autocomplete-light for a few reasons:
* we still have to supply it with a queryset, so it's not going to do any magical optimisations for us
* it depends on jquery, which we may or may not want to remove
* this dropdown is used not only in messages, but also groups and kb revisions

Optimising this was pretty fun, measurements from django debug toolbar with the mariadb query cache disabled:

With a query of "z", for which there are >10 users which match in db:

Before:

before with "z"

After:

after with "z"

With a query of "leo m", for which there are <10 users which match in db:

Before:

before with "leo m"

After:

after with "leo m"

### DIFF
--- a/kitsune/users/api.py
+++ b/kitsune/users/api.py
@@ -1,7 +1,6 @@
 from datetime import datetime, timedelta
 
 import pytz
-import waffle
 from django.contrib.auth.models import User
 from django.core.exceptions import ValidationError
 from django.db.models import Count, Q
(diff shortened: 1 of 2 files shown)


kitsune/users/api.py
```diff
@@ -54,26 +53,48 @@ def usernames(request):
     if not request.user.is_authenticated:
         return []
 
-    profiles = Profile.objects.filter(Q(name__istartswith=pre)).values_list("user_id", flat=True)
+    profile_ids = list(
+        Profile.objects.filter(Q(name__istartswith=pre)).values_list("user_id", flat=True)[:10]
```
akatsoulas, line 57: let's use `icontains` here

kitsune/users/api.py
```diff
@@ -54,26 +53,48 @@ def usernames(request):
...
+    )
     users = (
-        User.objects.filter(Q(username__istartswith=pre) | Q(id__in=profiles))
-        .extra(select={"length": "Length(username)"})
-        .order_by("length")
+        User.objects.filter(Q(username__istartswith=pre) | Q(id__in=profile_ids))
+        .filter(is_active=True)
         .select_related("profile")
```
akatsoulas, line 62: You can tweak this a bit more by having:
```python
profile_names = list(Profile.objects.filter(Q(name__icontains=pre).values_list('name')[:10])
users = (
    User.objects.filter(Q(username__icontains=pre) | Q(profile__name__in=profile_names))
    .filter(is_active=True)
    .select_related('profile')
)
```

kitsune/users/api.py
```diff
@@ -54,26 +53,48 @@ def usernames(request):
...
 
-    return [
-        {
-            "username": u.username,
-            "display_name": display_name_or_none(u),
-            "avatar": profile_avatar(u, 24),
-        }
-        for u in users[:10]
-    ]
+    if not exact_match_in_list:
```
akatsoulas, line 79: Using `icontains` should be enough. No need for this extra step at all

akatsoulas (changes requested):

kitsune/users/api.py
akatsoulas: This reduces the query 30% more for me

akatsoulas: Another interesting approach would be to use `Subquery` and fetch everything in one go. Unfortunately mysql does not support `limit` in subqueries

LeoMcA: > Another interesting approach would be to use Subquery and fetch everything in one go. Unfortunately mysql does not support limit in subqueries

Indeed that was my first approach. Performance is also :hankey: in subqueries in mysql - the django docs actually suggest this approach: https://docs.djangoproject.com/en/2.2/ref/models/querysets/#nested-queries-performance

kitsune/users/api.py
LeoMcA: I presume `icontains` will work the same as `istartswith` in this regard (but I'll test it). mysql will return as soon as it finds any 10 rows which match the query, so if the exact match is "slower" to access than any other 10 rows it won't end up in the result, which breaks the functionality on the frontend. This is why the previous code was sorting by length, so the exact match would always appear first (but massively at the expense of performance).

I'll test with `icontains` and if it doesn't always return the exact match add a comment here clarifying why it's necessary.

akatsoulas (approved):
We further discussed this in a meeting. r+wc

LeoMcA:
I've ended up sticking with `istartswith` rather than `icontains` because the latter can be a bit over 10x slower in the worst-cases I've tried (I presume because it has to do a full table scan and a full scan of each username in each row, rather than being able to discard rows very early on when they don't start with the correct letter).

The current behaviour uses `istartswith` rather than `icontains` so I'm comfortable keeping that for the sake of performance. If we want really good and effective and fast autocomplete I think we should look at moving this to ES.

PR #4573, opened 2020-10-27
add answer_content to question document

LeoMcA:


docs/architecture/decisions/0003-es-aaq-documents.md
```diff
@@ -0,0 +1,72 @@
+# 3. AAQ structure in Search v2
+
+Date: 2020-10-27
+
+## Status
+
+Accepted
+
+## Context
```
akatsoulas, line 9: I would also add something about the architecture itself which is different from what exists now (denormalization)

kitsune/search/v2/documents.py
```diff
@@ -93,7 +93,17 @@ def get_model(cls):
...
 class QuestionDocument(SumoDocument):
-    """"""
+    """
+    ES document for Questions. Every Question in DB gets a QuestionDocument in ES.
+
+    Parent class to AnswerDocument, so most fields are prefixed with "question_".
+    QuestionDocument and AnswerDocument are stored in the same index, so QuestionDocument
+    cannot define a field with the same name as AnswerDocument.
+    Distinguish QuestionDocuments from AnswerDocuments by filtering on whether the
+    document has a null "created" field.
```
akatsoulas, line 103: I wouldn't add how to distinguish between the two. I would talk more about the arch - denormalized version of the AAQ for ES and about the fact that `QuestionDocument` is the parent class of AnswerDocument but in the end it's a single document.

kitsune/search/v2/documents.py
```diff
@@ -159,7 +175,18 @@ def get_queryset(cls):
...
 class AnswerDocument(QuestionDocument):
-    """"""
+    """
+    ES document for Answers. Every Answer in DB gets an AnswerDocument in ES.
+
+    Child class to QuestionDocument and stored in the same index, so AnswerDocument
+    cannot define a field with the same name as one in QuestionDocument.
+    Most QuestionDocument fields are prefixed by "question_".
+    Distingish AnswerDocuments from QuestionDocuments by filtering on whether the document
+    has a non-null "created" field.
```
akatsoulas, line 185: Similar to the comment above. Let's mostly focus on the arch here and the class itself. Not on how to use it

akatsoulas (approved):
We can optimize the documentation later. This is a good first step

docs/architecture/decisions/0003-es-aaq-documents.md
LeoMcA: Do we use nested/joins currently? I'm not sure we do, we just needed to be cautious not to introduce it while re-modelling our data. I'll add a note about that.

PR #4574, opened 2020-10-27
Add locale aware Keyword field.

akatsoulas:
* Properly populate keywords
* Add to_action method in the base document

### DIFF
--- a/kitsune/search/v2/base.py
+++ b/kitsune/search/v2/base.py
@@ -1,7 +1,8 @@
 from datetime import datetime
+
 from django.utils import timezone
 from elasticsearch_dsl import Document as DSLDocument
-from elasticsearch_dsl import InnerDoc, field, MetaField
+from elasticsearch_dsl import InnerDoc, MetaField, field
 
 
 class SumoDocument(DSLDocument):
@@ -15,7 +16,8 @@ class SumoDocument(DSLDocument):
 
     @classmethod
     def prepare(cls, instance):
-        """Prepare an object given a model instance"""
+        """Prepare an object given a model instance."""
+
         obj = cls()
         doc_mapping = obj._doc_type.mapping
         fields = [f for f in doc_mapping]
@@ -37,8 +39,10 @@ class SumoDocument(DSLDocument):
                 # if the field is an Object but the value isn't an InnerDoc
                 # or a list containing an InnerDoc then we're dealing with locales
                 locale = obj.prepare_locale(instance)
-                if locale and value:
+                # Check specifically against None, False is a valid value
+                if locale and (value is not None):
                     obj[f] = {locale: value}
+
             else:
                 if (
                     isinstance(field_type, field.Date)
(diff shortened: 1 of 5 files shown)


kitsune/search/v2/documents.py
```diff
@@ -15,33 +15,27 @@
...
     category = field.Keyword()
     slug = field.Keyword()
     is_archived = field.Boolean()
     recent_helpful_votes = field.Integer()
 
+    # Document specific fields (locale aware)
+    title = SumoLocaleAwareTextField()
+    content = SumoLocaleAwareTextField(store=True, term_vector="with_positions_offsets")
+    summary = SumoLocaleAwareTextField(store=True, term_vector="with_positions_offsets")
+    keywords = SumoLocaleAwareKeywordField(multi=True)
```
LeoMcA, line 33: do we want to normalize this? https://www.elastic.co/guide/en/elasticsearch/reference/current/normalizer.html

I think in our `prepare_keywords` method we want to split on white-space, so `keyword-1 keyword-2 keyword-3` becomes [`keyword-1`, `keyword-2`, `keyword-3`], otherwise I don't think we can match on `keyword-2` for instance without using wildcards either side.

akatsoulas: Missed that bit from the other PR. Will update. This is the prepare method that I will be using here:
https://github.com/mozilla/kitsune/pull/4560/files#diff-1701974c94f568a36fc78cde0bb414faa0963f61133dc8b4c9627c69071fd560R61

kitsune/search/v2/base.py
```diff
@@ -1,25 +1,27 @@
...
 from elasticsearch_dsl import Document as DSLDocument
-from elasticsearch_dsl import InnerDoc, field, MetaField
+from elasticsearch_dsl import InnerDoc, MetaField, field
 
 
 class SumoDocument(DSLDocument):
     """Base class with common methods for all the different documents."""
 
-    indexed_on = field.Date()
+    updated_on = field.Date()
```
LeoMcA, line 11: IMO this feels too similarly named to the `updated` field we have on most documents, coming from db. I liked `indexed_on` because it put it clearly in the ES domain. I think, even when we're calling the update API, we still are indexing data. Perhaps `indexed_or_updated_on`?

kitsune/search/v2/base.py
```diff
@@ -48,13 +52,28 @@ def prepare(cls, instance):
...
+        Useful for bulk operations.
+        """
+
+        # Default to index if no action is defined or if it's `save`
+        if bulk:
+            return self.to_dict(include_meta=True)
+        if not action or action == "index":
+            self.save(**kwargs)
+        elif action == "delete":
+            self.delete()
```
LeoMcA, line 75: My gut feels a little uncomfortable about this doing two very different things on the basis of the `bulk` argument: either returning a dictionary, or performing operations in ES. Since we only call this in the `index_object(s_bulk)` tasks - and I can't see us wanting or needing to index stuff outside of calling those tasks - do we need this abstraction?

kitsune/search/v2/documents.py
```diff
@@ -15,33 +15,27 @@
...
-    current_id = field.Integer()
-    parent_id = field.Integer()
+    parent_id = field.Keyword()
     category = field.Keyword()
     slug = field.Keyword()
     is_archived = field.Boolean()
     recent_helpful_votes = field.Integer()
 
+    # Document specific fields (locale aware)
+    title = SumoLocaleAwareTextField()
```
LeoMcA, line 30: I think title should also be `store=True`d, so we don't have to parse the whole `_source` field when rending search results.

LeoMcA: I guess the same goes for `slug` and `locale`... maybe we can make this change later once we're actually rending search results

akatsoulas:
@LeoMcA this includes the changes we discussed in a meeting last week. I am merging this to continue working on ES7 and we can open a new PR with any changes needed when you are back

PR #4577, opened 2020-10-28
Reflect development environments migration

The-smooth-operator:


The-smooth-operator:
@akatsoulas please can you have a look?

k8s/commander.sh
```diff
@@ -1,5 +1,5 @@
 #!/bin/bash
-set -exo pipefail
+set -eo pipefail
```
akatsoulas, line 2: Let's keep the `x` here

k8s/commander.sh
```diff
@@ -55,8 +55,17 @@ function post-deploy {
     REGION_ENV=${3}
     K8S_NAMESPACE="sumo-${REGION_ENV}"
 
-    export KUBECTL_BIN="./regions/${REGION}/kubectl"
-    export KUBECONFIG="./regions/${REGION}/kubeconfig"
+    if [ -f "./regions/${REGION}/kubectl" ] ; then
+        KUBECTL_BIN="./regions/${REGION}/kubectl"
+    else
+        KUBECTL_BIN=$(which kubectl)
```
akatsoulas, line 61: My concern here is that we don't ensure a difference of a single minor version. I would prefer to explicit ask in the documentation either to use the same version or anything between 1.(16..18)

k8s/commander.sh
The-smooth-operator: the `x` here is just polluting the output and not giving any useful information. Bash -x is used for debugging scripts.

Anyway, I'm taking the suggestion because don't want to fight this battle :)

k8s/commander.sh
The-smooth-operator: We are explicitly asking in the documentation to do so.
Anyway I get that this is a concern of yours and I understand it. So the best option for me is to enforce this in the script, as I've done.

I've tested this in several scenarios (old kops clusters, EKS, matching versions, versions one step away and more). Results are good. In case of a mismatch by more than 1 step version, we are getting
```
Version mismatch; Client and server versions must be equal or have 1 minor version step. Exiting here
Client version: 19, Server version: 17 
```

akatsoulas (approved):
this is really nice work!

The-smooth-operator:
> this is really nice work!

Thank you!!!

PR #4579, opened 2020-10-29
don't include default requirements in docs requirements

LeoMcA:
this is causing read the docs to fail to build


akatsoulas (approved):
r+ - squash before merging

PR #4580, opened 2020-10-30
Fixed sort article discussion posts by Author, Replies or Last Post

gauravghati:
@LeoMcA or @akatsoulas Can you look into this PR ...?

There was an extra line, which was causing this error, now it's working.
This PR fixes [issue#529](https://github.com/mozilla/sumo-project/issues/529)
sort

### DIFF
--- a/kitsune/kbforums/views.py
+++ b/kitsune/kbforums/views.py
@@ -48,7 +48,7 @@ def sort_threads(threads_, sort=0, desc=0):
         return threads_.order_by(prefix + "last_post__created").all()
 
     # If nothing matches, use default sorting
-    return threads_.order_by("-last_post__created")
+    return threads_.all()
 
 
 def threads(request, document_slug):
@@ -433,19 +433,17 @@ def locale_discussions(request):
         document__locale=request.LANGUAGE_CODE, document__allow_discussion=True
     )
     try:
-        sort = int(request.GET.get("sort", 0))
+        sort = int(request.GET.get("sort", 5))
     except ValueError:
-        sort = 0
+        sort = 5
 
     try:
-        desc = int(request.GET.get("desc", 0))
+        desc = int(request.GET.get("desc", 1))
     except ValueError:
-        desc = 0
+        desc = 1
     desc_toggle = 0 if desc else 1
 
     threads_ = sort_threads(threads, sort, desc)
-
-    # Ignore sticky-ness:
     threads_ = paginate(request, threads_, per_page=kbforums.THREADS_PER_PAGE)
     is_watching_locale = request.user.is_authenticated and NewThreadInLocaleEvent.is_notifying(
         request.user, locale=request.LANGUAGE_CODE
--- a/kitsune/kbforums/views.py
+++ b/kitsune/kbforums/views.py
@@ -48,7 +48,7 @@ def sort_threads(threads_, sort=0, desc=0):
         return threads_.order_by(prefix + "last_post__created").all()
 
     # If nothing matches, use default sorting
-    return threads_.all()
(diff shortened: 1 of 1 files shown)


LeoMcA:
Nearly there, but this is causing a test to fail (run it locally with `DEBUG=False ./manage.py test kitsune/kbforums/tests/test_templates.py:ThreadsTemplateTests.test_locale_discussions_ignores_sticky`):

```
FAIL: kitsune/kbforums/tests/test_templates.py:ThreadsTemplateTests.test_locale_discussions_ignores_sticky
  File "/app/kitsune/kbforums/tests/test_templates.py", line 341, in test_locale_discussions_ignores_sticky
    assert title.startswith("A thread with a very very long")
AssertionError

```

I think we still need this `order_by("-last_post__created")` but as the default in `sort_threads()` rather than overwriting the ordering that function applies.

gauravghati:
Done! all tests are passing locally now, I'm kinda new with CI-testing. :sweat_smile:

LeoMcA:
No worries! And apologies, I led you down slightly the wrong path here:

It's true that in the `locale_discussions` view, we want `sort_threads` to use `order_by("-last_post__created")` if nothing has been selected, but in the `threads` view (on [L68](https://github.com/mozilla/kitsune/blob/85445d0fcdb62f124a20b21beefb31385da69b24/kitsune/kbforums/views.py#L68)) we still want it to use the default order which puts "sticky" threads at the top (by calling `all()`).

Also the `locale_discussions` view now has the slightly odd behaviour that clicking on "Last Post" the first time does nothing at all (because it's already sorting most recent to oldest).

I reckon both of the above can be solved by going with your original approach of just removing the line entirely, and modifying the `except ValueError` values of `sort` and `desc`.

kitsune/kbforums/views.py
```diff
@@ -446,7 +446,6 @@ def locale_discussions(request):
     threads_ = sort_threads(threads, sort, desc)
 
     # Ignore sticky-ness:
```
LeoMcA, line 448: This comment refers to the line which was removed, so should be removed too

gauravghati:
> It's true that in the `locale_discussions` view, we want `sort_threads` to use `order_by("-last_post__created")` if nothing has been selected, but in the `threads` view (on [L68](https://github.com/mozilla/kitsune/blob/85445d0fcdb62f124a20b21beefb31385da69b24/kitsune/kbforums/views.py#L68)) we still want it to use the default order which puts "sticky" threads at the top (by calling `all()`).

Now defaults for `thread` view are sort=0 and desc=0 which will call `all()`, and for `locale_discussions` defaults are sort=5 and desc=1 which will call `order_by("-last_post__created")`.

> Also the `locale_discussions` view now has the slightly odd behaviour that clicking on "Last Post" the first time does nothing at all (because it's already sorting most recent to oldest).
 
This also get resolved by changing defaults to `sort=5 and desc=1`. 
Please review and let me know if there are any other changes.

kitsune/kbforums/views.py
gauravghati: Removed!

kitsune/kbforums/views.py
```diff
@@ -433,20 +433,17 @@ def locale_discussions(request):
         document__locale=request.LANGUAGE_CODE, document__allow_discussion=True
     )
     try:
```
akatsoulas, line 435: Do we need the `try/except` clause here at all?

LeoMcA: I guess it's in case one of the arguments isn't castable to an int (like `&desc=foobar`)

LeoMcA (approved):
Great! :rocket:

PR #4582, opened 2020-10-30
edit question page fixes

LeoMcA:
fixes https://github.com/mozilla/sumo-project/issues/669 and https://github.com/mozilla/sumo-project/issues/304

### DIFF
--- a/kitsune/questions/api.py
+++ b/kitsune/questions/api.py
@@ -5,7 +5,6 @@ import django_filters
 import json
 from django_filters.rest_framework import DjangoFilterBackend
 from django import forms
-from django.contrib.contenttypes.models import ContentType
 from django.db.models import Q
 from rest_framework import serializers, viewsets, permissions, filters, status, pagination
 from rest_framework.decorators import action
@@ -30,7 +29,6 @@ from kitsune.sumo.api_utils import (
     SplitSourceField,
 )
 from kitsune.tags.utils import add_existing_tag
-from kitsune.upload.models import ImageAttachment
 from kitsune.users.api import ProfileFKSerializer
 from kitsune.users.models import Profile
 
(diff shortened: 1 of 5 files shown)


kitsune/upload/views.py
```diff
@@ -75,6 +75,11 @@ def del_image_async(request, image_id):
         message = _("You do not have permission to do that.")
         return HttpResponseForbidden(json.dumps({"status": "error", "message": message}))
 
+    try:
```
akatsoulas, line 78: Maybe you could check against the `content_type` you are interested in and call this method instead of checking for an exception. It would be clearer on which type you are operating on

kitsune/questions/api.py
```diff
@@ -389,18 +387,6 @@ def auto_tag(self, request, pk=None):
         question.auto_tag()
         return Response(status=status.HTTP_204_NO_CONTENT)
 
-    @action(detail=True, methods=["post"])
-    def attach_images(self, request, pk=None):
```
akatsoulas, line 393: why are dropping this completely? Not used anywhere?

kitsune/sumo/static/sumo/js/questions.js
```diff
@@ -104,6 +108,12 @@
     hideDetails($questionForm, aaq);
   }
 
+  function initEditQuestion() {
```
akatsoulas, line 111: nit: Can we maybe combine this with `initNewQuestion` based on whether it's a new or an existing question since they are sharing some code?

akatsoulas (approved):
r+wc

kitsune/questions/api.py
LeoMcA: Yep, and somewhat confusing as it dupes the code that happens in the question save method

kitsune/upload/views.py
LeoMcA: I quite like the duck typing here, but could add a comment explaining where this is currently used

akatsoulas (approved):

PR #4583, opened 2020-10-30
fix contributor sign-up

LeoMcA:
https://github.com/mozilla/sumo-project/issues/563

oidc routes don't have a LANGUAGE_CODE, so set it in the session before the auth flow starts

### DIFF
--- a/kitsune/users/views.py
+++ b/kitsune/users/views.py
@@ -330,6 +330,9 @@ class FXAAuthenticateView(OIDCAuthenticationRequestView):
     def get(self, request):
         is_contributor = request.GET.get("is_contributor") == "True"
         request.session["is_contributor"] = is_contributor
+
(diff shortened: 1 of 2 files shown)


kitsune/users/views.py
```diff
@@ -330,6 +330,11 @@ def get_settings(attr, *args):
     def get(self, request):
         is_contributor = request.GET.get("is_contributor") == "True"
         request.session["is_contributor"] = is_contributor
+
+        locale = getattr(request, "LANGUAGE_CODE", None)
```
akatsoulas, line 334: how about request.get('LANGUAGE_CODE', '')?

akatsoulas: or maybe something like request.get('LANGUAGE_CODE', 'en-US') and set it directly to the session

akatsoulas (approved):
r+wc

PR #4586, opened 2020-11-03
Add Hubs AAQ config.

akatsoulas:


kitsune/questions/config.py
```diff
@@ -724,6 +724,52 @@
                 ),
             },
         ),
+        (
+            "hubs",
+            {
+                "name": _lazy("Hubs"),
+                "subtitle": _lazy("Mozilla Hubs support"),
```
LeoMcA, line 731: should this be "A social 3D platform for collaborating with friends, family and colleagues" (the description in staging)?

akatsoulas: This is the reason I have this as draft. I am waiting for confirmation from Joni

PR #4587, opened 2020-11-04
Delete Inbox messages when an account is deactivated.

akatsoulas:


akatsoulas:
@LeoMcA bump

kitsune/users/utils.py
```diff
@@ -87,8 +88,9 @@ def anonymize_user(user):
     # Deactivate the user and change key information
     user.username = "user%s" % user.id
     user.email = "%s@example.com" % user.id
+    # Delete the inbox of the user
+    InboxMessage.objects.filter(to=user).delete()
```
LeoMcA, line 92: Does the user's outbox also need to be deleted?

kitsune/users/tests/test_views.py
```diff
@@ -121,6 +120,18 @@ def test_deactivate_and_flag_spam(self):
         eq_(1, Answer.objects.filter(creator=u, is_spam=True).count())
         eq_(0, Answer.objects.filter(creator=u, is_spam=False).count())
 
+    def test_inbox_deletion_on_deactivation(self):
+        """Test message deletion on deactivation"""
+        recipient = UserFactory()
+        sender = UserFactory()
+        InboxMessage.objects.create(
+            to=recipient, sender=sender, message="testing message deletion"
+        )
```
LeoMcA, line 129: Can we use a factory here, or even better, log in as the sender and create the message through a post request. That way we can also test that we're not deleting messages from the sender's outbox.

kitsune/users/utils.py
akatsoulas: No, deleting your own account shouldn't affect the recipients inbox.

PR #4588, opened 2020-11-06
removed dashboard overview-link 404 page

gauravghati:
It will solve the [issue#640](https://github.com/mozilla/sumo-project/issues/640).

### DIFF
--- a/kitsune/dashboards/jinja2/dashboards/contributors.html
+++ b/kitsune/dashboards/jinja2/dashboards/contributors.html
@@ -64,7 +64,7 @@
     {% include 'dashboards/includes/kb_overview.html' %}
   </table>
   <div class="table-footer">
-    <a href="{{ url('dashboards.contributors_overview')|urlparams(product=product.slug, category=category) }}">{{ _('Complete overview...') }}</a>
+    <a href="{{ url('dashboards.contributors_overview')|urlparams(category=category) }}">{{ _('Complete overview...') }}</a>
   </div>
 {% endblock %}
 
--- a/kitsune/dashboards/jinja2/dashboards/contributors.html
+++ b/kitsune/dashboards/jinja2/dashboards/contributors.html
@@ -64,7 +64,7 @@
     {% include 'dashboards/includes/kb_overview.html' %}
   </table>
   <div class="table-footer">
-    {% set pslug=product.slug if product!=None else None%}
+    {% set pslug=product.slug if product!=None else None %}
     <a href="{{ url('dashboards.contributors_overview')|urlparams(product=pslug, category=category) }}">{{ _('Complete overview...') }}</a>
   </div>
 {% endblock %}
--- a/kitsune/dashboards/jinja2/dashboards/contributors.html
+++ b/kitsune/dashboards/jinja2/dashboards/contributors.html
@@ -64,10 +64,7 @@
     {% include 'dashboards/includes/kb_overview.html' %}
   </table>
   <div class="table-footer">
-    {% set pslug=product.slug %}
-    {% if product==None %}
-      {% set pslug=None %}
-    {% endif %}
+    {% set pslug=product.slug if product!=None else None%}
     <a href="{{ url('dashboards.contributors_overview')|urlparams(product=pslug, category=category) }}">{{ _('Complete overview...') }}</a>
   </div>
 {% endblock %}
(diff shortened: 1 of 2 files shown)


kitsune/dashboards/views.py
```diff
@@ -252,10 +252,10 @@ def aggregated_metrics(request):
 
 def _get_product(request):
     product_slug = request.GET.get("product")
-    if product_slug:
```
akatsoulas, line 255: Hi @gauravghati and thanks for this PR. May I ask why are proposing this change?

LeoMcA:
Nice approach, but I think a better one would be to prevent the `/contributors` page from having a link to `/contributors/kb-overview?product=[]` in the first place.

kitsune/dashboards/views.py
gauravghati: Yes, the `request.GET.get('product')` is returning `[]` as a string, when the "all product" option is selected from the dropdown menu.

So, that's why, funtion `get_object_or_404` is returing page 404 when `product_slug="[]"`

So, putting it in try-except will return `product=None` if `product_slug="[]"`, that will help us to remove the filter condition from [line#222](https://github.com/mozilla/kitsune/blob/eb32f145c5b9095fcea75fd7f1341604bca35b69/kitsune/dashboards/readouts.py#L222) in readouts.py. Removing the filter condition will return all products from the database.

gauravghati:
> Nice approach, but I think a better one would be to prevent the `/contributors` page from having a link to `/contributors/kb-overview?product=[]` in the first place.

Yes, right, what value should it pass as parameter if "all product" option is select? or we should just not pass any value.

LeoMcA:
Yeah either no value or no argument at all

gauravghati:
I don't think there's a need for try-except now, `_get_product` function will return `None` for "all product" option, and for request like  `/contributors/kb-overview?product=foobar` it will render 404 page. I've undone the changes as it was before.

LeoMcA:
@gauravghati thanks for sticking with this bug, it's a bit of a weird one.

If you haven't yet, now would be a fantastic time to set up a debugger to closely inspect what's going on. I use the in-built vscode one: https://github.com/mozilla/kitsune/tree/master/.vscode/README.md

Changing `product.slug` to `product` sorta avoids the problem, and introduces a large one of its own: on prod, our product names don't always match up with our slugs. So, for instance, Firefox for Android's slug is `mobile`. In that case `str(product) != product.slug`.

Let's consider what's happening here, though. For a start, when "All products" is selected `product` is `None`. With your current change, which works in this case, calling `urlparams(product=product)` (essentially `urlparams(product=None)`) gives us our desired behaviour. However, the current prod code uses `product.slug`, but how can this be since we know in this case `product` is `None`? In python calling `None.slug` would give us an attribute error, but this is happening in our templates, which use Jinja, so something different must be happening here. `urlparams` must be receiving whatever the result of calling `None.slug` is in Jinja.

So, this could be solved by checking if `product` is `None` in the template, and then not passing the `product.slug` argument to `urlparams`. However, since `urlparams` already gracefully handles the `None` case, I feel it should also handle the `None.slug` case.

Some hints:
* Here's where you'll find the `urlparams` code: https://github.com/mozilla/kitsune/blob/4cf90d03f9543dcfad31240183e0eda5f95aca6f/kitsune/sumo/templatetags/jinja_helpers.py#L83 If I were you I'd look closely at L105.
* [This page section of documentation](https://jinja.palletsprojects.com/en/2.11.x/api/#jinja2.Undefined) may be useful if you get stumped by what `None.slug` could become in Jinja.

gauravghati:
> So, this could be solved by checking if `product` is `None` in the template, and then not passing the `product.slug` argument to `urlparams`. However, since `urlparams` already gracefully handles the `None` case, I feel it should also handle the `None.slug` case.

okay, got it, I'll look at it.

kitsune/dashboards/jinja2/dashboards/contributors.html
```diff
@@ -64,7 +64,11 @@ <h1 class="sumo-page-heading">{{ title }}</h1>
     {% include 'dashboards/includes/kb_overview.html' %}
   </table>
   <div class="table-footer">
-    <a href="{{ url('dashboards.contributors_overview')|urlparams(product=product.slug, category=category) }}">{{ _('Complete overview...') }}</a>
+    {% set pslug=product.slug %}
+    {% if product==None %}
+      {% set pslug=None %}
+    {% endif %}
```
LeoMcA, line 70: We can reduce this to one line with a `{ set a = b if c else e }` statement.

gauravghati: correct, done changes.

LeoMcA:
Thanks for this!

PR #4590, opened 2020-11-10
add signals for all search v2 document types

LeoMcA:
Don't worry, most of the code here is the unit tests!

I'm pretty sure there's a few signals I can un-register, and still have all the tests pass. Ran out of time to do that today, will tackle it on Thursday when I'm back.


kitsune/search/v2/base.py
```diff
@@ -87,7 +87,7 @@ def to_action(self, action=None, **kwargs):
         # if we have a bulk update, we need to include the meta info
         # and return the data by calling the to_dict() method of DSL
         if is_bulk := kwargs.pop("is_bulk", False):
-            payload = self.to_dict(include_meta=is_bulk)
+            payload = self.to_dict(include_meta=is_bulk, skip_empty=False)
```
akatsoulas, line 90: Why do we need these empty values?

kitsune/search/v2/documents.py
```diff
@@ -67,18 +67,19 @@ def prepare_topic_ids(self, instance):
...
     def prepare_display_order(self, instance):
         return instance.original.display_order
 
     @classmethod
     def get_model(cls):
         return wiki_models.Document
 
+    @classmethod
+    def get_queryset(cls):
+        return wiki_models.Document.objects.select_related(
```
akatsoulas, line 79: comments/docstrings pls

kitsune/search/v2/es7_utils.py
```diff
@@ -91,8 +91,11 @@ def index_object(doc_type_name, obj_id):
     doc_type = next(cls for cls in get_doc_types() if cls.__name__ == doc_type_name)
     model = doc_type.get_model()
 
-    obj = model.objects.get(pk=obj_id)
-    doc_type.prepare(obj).to_action("index")
+    try:
```
akatsoulas, line 94: I am curious as to why this is needed.

kitsune/search/v2/signals/questions.py
```diff
@@ -0,0 +1,62 @@
...
+    reindex_by_query,
+    index_objects_bulk,
+)
+from kitsune.questions.models import Question, QuestionVote, Answer, AnswerVote
+from taggit.models import Tag
+
+
+@receiver(post_save, sender=Question)
+@receiver(m2m_changed, sender=Question.tags.through)
+def handle_question_save(instance, **kwargs):
```
akatsoulas, line 16: docstrings pls

akatsoulas: I won't add comments to the rest of the signals but let's do this for all the signals

akatsoulas (commented):
merely a first pass. just a few comments by skimming through it

kitsune/search/v2/base.py
LeoMcA: Previously the tests would fail without this, after some changes that's no longer the case, so I added some more explicit tests in https://github.com/mozilla/kitsune/pull/4602

kitsune/search/v2/es7_utils.py
LeoMcA: Specifically here: users can exist in DB without a profile, however in elastic we index on profiles (it made more sense to me to do that, because the majority of the indexable information is in a user's profile, and I don't think we want to be indexing users without a profile). We have signals on User post_save, but there's no guarantee that a profile will exist at that point, so we may trigger a profile index for a profile which doesn't exist.

More generally: it's entirely possible that celery jobs will be executed after a row has been deleted from the DB which the task references, and I don't think that should be treated as a job failure.

kitsune/search/v2/signals/questions.py
LeoMcA: IMO `handle_question_save` is pretty descriptive, what extra detail would you like to see in the docstrings?

kitsune/search/v2/documents.py
LeoMcA: Actually I'm going to move this into a separate PR, as I'm pretty sure we can remove the calls to `parent` since we're updating stuff, which will require some further discussion - anyhow, it's not related to signals, and only here because I wanted to quickly optimise it a bit for when I was doing some tests with a fully indexed elastic instance.

kitsune/search/v2/signals/__init__.py
```diff
@@ -0,0 +1 @@
+from . import questions, users, wiki, forums  # noqa
```
akatsoulas, line 1: why is this needed?

kitsune/search/v2/signals/forums.py
```diff
@@ -0,0 +1,25 @@
+from django.db.models.signals import post_save, post_delete
+from django.dispatch import receiver
+from django.conf import settings
+from kitsune.search.v2.es7_utils import index_object, delete_object, index_objects_bulk
+from kitsune.forums.models import Thread, Post
+
+
+@receiver(post_save, sender=Thread)
+def handle_forum_thread_save(instance, **kwargs):
+    if settings.ES_LIVE_INDEXING:
```
akatsoulas, line 10: nit: shall we add a decorator in the sumo app to use instead of adding this `if` everywhere?

kitsune/search/v2/signals/forums.py
```diff
@@ -0,0 +1,25 @@
...
+        index_objects_bulk.delay(
+            "ForumDocument", list(instance.post_set.values_list("pk", flat=True))
```
akatsoulas, line 12: why the list here?

akatsoulas: this has other occurrences too.

kitsune/search/v2/es7_utils.py
```diff
@@ -117,6 +121,27 @@ def index_objects_bulk(doc_type_name, obj_ids):
...
+    update = update.script(source=script, params={"value": field_value}, conflicts="proceed")
+
+    # refresh index to ensure search fetches all matches
+    doc_type._index.refresh()
+
+    update.execute()
+
+
 @task
 def delete_object(doc_type_name, obj_id):
```
akatsoulas, line 146: could we use the `to_action('delete')` here instead of the `delete()` method?

kitsune/search/v2/signals/questions.py
```diff
@@ -0,0 +1,64 @@
...
+    remove_from_field,
+)
+from kitsune.questions.models import Question, QuestionVote, Answer, AnswerVote
+from taggit.models import Tag
+
+
+@receiver(post_save, sender=Question)
+@receiver(m2m_changed, sender=Question.tags.through)
+def handle_question_save(instance, **kwargs):
+    if getattr(kwargs, "action", "").startswith("pre_"):
```
akatsoulas, line 17: why are we checking for `pre` signals here?

kitsune/search/v2/signals/questions.py
```diff
@@ -0,0 +1,64 @@
...
+    index_object,
+    delete_object,
+    index_objects_bulk,
+    remove_from_field,
+)
+from kitsune.questions.models import Question, QuestionVote, Answer, AnswerVote
+from taggit.models import Tag
+
+
+@receiver(post_save, sender=Question)
```
akatsoulas, line 14: Let's add to all the signals a UID 
https://docs.djangoproject.com/en/3.1/topics/signals/#preventing-duplicate-signals

kitsune/search/v2/signals/tests/test_forums.py
```diff
@@ -0,0 +1,40 @@
...
+from kitsune.forums.tests import PostFactory
+from elasticsearch7.exceptions import NotFoundError
+
+from kitsune.search.v2.documents import ForumDocument
+
+
+class ForumDocumentSignalsTests(Elastic7TestCase):
+    def setUp(self):
+        self.post = PostFactory()
+        self.post_id = self.post.id
```
akatsoulas, line 11: we can skip this assignment

kitsune/search/v2/signals/tests/test_forums.py
```diff
@@ -0,0 +1,40 @@
...
+
+    @property
+    def _doc(self):
```
akatsoulas, line 14: I would remove the underscore

akatsoulas: nit: I would also remove the property here. My preference would be for the tests to be as atomic as possible even if that means boilerplate code

kitsune/search/v2/es7_utils.py
akatsoulas: let's add a couple of lines with comments about what you added here

akatsoulas (changes requested):
I did a first pass. A few of the comments apply to multiple files but I did not mark them in order to wait for your feedback

kitsune/search/v2/signals/__init__.py
LeoMcA: So `kitsune/search/apps.py:SearchV2Config` can set up all the signals at once with just `from kitsune.search.v2 import signals`

kitsune/search/v2/signals/forums.py
LeoMcA: Absolutely!

kitsune/search/v2/signals/forums.py
LeoMcA: Celery won't serialize a `values_list` or cast it to a list itself, it needs to be done manually. (You'll see the error if you remove the `list()` and run the tests with `CELERY_TASK_ALWAYS_EAGER=False`.)

kitsune/search/v2/signals/questions.py
LeoMcA: `m2m_changed` sends a `pre_foobar` signal followed by a `post_foobar` signal (with the changes now made in DB). We want the latter, and I can't see a way to filter them out in the `@receiver` decorator - we have to subscribe to the whole lot.

kitsune/search/v2/signals/tests/test_forums.py
LeoMcA: I could change this to a non-property `get_doc()` to make its usage in tests a bit clearer ("oh it gets the doc from ES fresh every time")

kitsune/search/v2/signals/tests/test_forums.py
LeoMcA: It's needed when testing getting a deleted document raises the correct error, otherwise we'll attempt to fetch with id `self.post.id` which will error out itself (`self.post being None`).

akatsoulas (approved):
nice :+1:

akatsoulas:
pls squash bfr merging

PR #4596, opened 2020-11-13
Add a class property for updating objects.

akatsoulas:


akatsoulas:
Depends on #4600

kitsune/search/v2/base.py
```diff
@@ -83,11 +93,10 @@ def to_action(self, action=None, **kwargs):
         """
 
         # Default to index if no action is defined or if it's `save`
-        payload = {}
         # if we have a bulk update, we need to include the meta info
         # and return the data by calling the to_dict() method of DSL
-        if is_bulk := kwargs.pop("is_bulk", False):
-            payload = self.to_dict(include_meta=is_bulk)
+        is_bulk = kwargs.pop("is_bulk", False)
```
LeoMcA, line 98: This could be moved to the function parameters definition, like `def prepare(cls, instance, is_bulk=False)`

akatsoulas: That would ideal but unfortunately even if the payload is handled in the prepare method we still need to do specific operations here in the bulk case like updated the `doc` attr and deleting `_source`. Because of that, my preference  would be to handle all the bulk related operations in a single place.

LeoMcA: Ah crap, the git diff had me confused, I meant like `def to_action(self, action=None, is_bulk=False, **kwargs)`

LeoMcA (approved):
Rebased against the now-py3.9 elasticsearch-v7 branch and works great!

r+wc (that previous, now corrected, one about moving the `is_bulk` out of `kwargs`)

PR #4600, opened 2020-11-16
Bump Python to version 3.9.

akatsoulas:


LeoMcA:
We should change `default_language_version.python` in `.pre-commit-config.yaml` to `python3.9` as well

Dockerfile
```diff
@@ -23,15 +23,15 @@ ENV PYTHONUNBUFFERED=1
 ENV PATH="/venv/bin:$PATH"
 
 RUN python -m venv /venv
-RUN pip install --upgrade "pip==20.2.3"
+RUN pip install --upgrade "pip==20.2.4"
```
LeoMcA, line 26: `pip` is still `==20.2.3` in dev.in

pyproject.toml
```diff
@@ -1,4 +1,4 @@
 [tool.black]
 line-length = 99
-target-version = ['py38']
+target-version = ['py39']
```
LeoMcA, line 3: Nice catch!

LeoMcA (approved):
Looks like we're still getting that libjpeg error in circle :(
r+wc

Dockerfile
```diff
@@ -23,22 +23,21 @@ ENV PYTHONUNBUFFERED=1
...
+    libxml2-dev libxslt1-dev zlib1g-dev git \
+    libjpeg-dev libjpeg62-turbo-dev libffi-dev libssl-dev libxslt1.1 \
+    libmariadb3 mariadb-client && \
     rm -rf /var/lib/apt/lists/*
 
 COPY ./requirements/*.txt /app/requirements/
 
 RUN pip install --no-cache-dir --require-hashes -r requirements/default.txt && \
-    pip install --no-cache-dir --require-hashes -r requirements/dev.txt && \
-    pip install --no-cache-dir --require-hashes -r requirements/test.txt
```
akatsoulas, line 41: @LeoMcA fyi, I removed this line because dev.txt already includes the test libs

PR #4602, opened 2020-11-17
test indexing/updating empty values

LeoMcA:
add skip_empty=False to to_dict in SumoDocument.to_action to make tests pass

### DIFF
--- a/kitsune/search/v2/base.py
+++ b/kitsune/search/v2/base.py
@@ -95,7 +95,7 @@ class SumoDocument(DSLDocument):
         # Default to index if no action is defined or if it's `save`
         # if we have a bulk update, we need to include the meta info
         # and return the data by calling the to_dict() method of DSL
-        payload = self.to_dict(include_meta=is_bulk)
+        payload = self.to_dict(include_meta=is_bulk, skip_empty=False)
 
         if not action or action == "index":
             return payload if is_bulk else self.save(**kwargs)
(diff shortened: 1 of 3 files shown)


kitsune/search/v2/base.py
```diff
@@ -86,8 +86,7 @@ def to_action(self, action=None, **kwargs):
         payload = {}
         # if we have a bulk update, we need to include the meta info
         # and return the data by calling the to_dict() method of DSL
-        if is_bulk := kwargs.pop("is_bulk", False):
-            payload = self.to_dict(include_meta=is_bulk)
+        payload = self.to_dict(include_meta=is_bulk, skip_empty=False)
```
akatsoulas, line 89: Why do we need to keep empty values?

LeoMcA: Otherwise these tests will fail: https://github.com/mozilla/kitsune/pull/4602/files#diff-f270bdc2c6b7be81ab0afaa0979895167bb9506dbfd5bbacde9b3f9fda3be529R34-R41

A concrete example would be if a kb article was removed from all topics, without `skip_empty=False` when it was updated it would remain in its previous set of topics in ES.

bin/run-search-v2-tests.sh
```diff
@@ -0,0 +1,15 @@
+#!/bin/bash
```
akatsoulas, line 1: Do we need this?

akatsoulas (approved):
r+wc

LeoMcA:
Will squash, rebase, resolve conflicts and merge after https://github.com/mozilla/kitsune/pull/4596 is merged

PR #4603, opened 2020-11-18
Remove check for firefox-preview AAQ.

akatsoulas:


kitsune/questions/utils.py
```diff
@@ -45,12 +44,4 @@ def get_mobile_product_from_ua(user_agent):
         return "ios"
 
     # android
-    try:
-        # We are using firefox instead of Firefox as lower() has been applied to the UA
-        mobile_client = re.search(r"firefox/(?P<version>\d+)\.\d+", ua).groupdict()
-    except AttributeError:
-        return None
```
LeoMcA, line 52: We still need something like this try except (I guess just a search for "firefox" in the ua string) to ensure this test doesn't fail: https://github.com/mozilla/kitsune/blob/db42cf7/kitsune/questions/tests/test_utils.py#L119-L120

kitsune/questions/utils.py
```diff
@@ -45,12 +44,4 @@ def get_mobile_product_from_ua(user_agent):
...
-    else:
-        if int(mobile_client["version"]) >= 69:
-            return "firefox-preview"
```
LeoMcA, line 55: We need to update https://github.com/mozilla/kitsune/blob/db42cf7/kitsune/questions/tests/test_utils.py#L111-L112 to reflect the new intended behaviour.

PR #4610, opened 2020-11-30
reduce memory usage of fix_current_revisions job

LeoMcA:
locally reduces peak memory usage from ~650MB to ~140MB


akatsoulas (approved):
:rocket:

PR #4612, opened 2020-12-01
optimise WikiDocument indexing

LeoMcA:
https://github.com/mozilla/sumo-project/issues/695

### DIFF
--- a/kitsune/search/v2/base.py
+++ b/kitsune/search/v2/base.py
@@ -4,8 +4,6 @@ from django.utils import timezone
 from elasticsearch_dsl import Document as DSLDocument
 from elasticsearch_dsl import InnerDoc, MetaField, field
 
-from kitsune.wiki.models import Document
-
 
 class SumoDocument(DSLDocument):
     """Base class with common methods for all the different documents."""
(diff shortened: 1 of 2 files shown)


kitsune/search/v2/base.py
```diff
@@ -74,15 +67,13 @@ def prepare(cls, instance, **kwargs):
                     # https://docs.djangoproject.com/en/2.2/ref/utils/#django.utils.timezone.make_aware
                     value = timezone.make_aware(value, is_dst=False).astimezone(timezone.utc)
 
-                if use_parent_obj:
-                    value = obj.get_field_value(f, instance.parent, prepare_method)
-
-                setattr(obj, f, value)
+                if not parent_id:
```
akatsoulas, line 70: when we want to merge the docs, we are not ignoring the value of the object but rather we are overwriting with the one of the parent where applicable. The removed snippet is using attrs from both parent and child but the new one does not

akatsoulas (changes requested):

kitsune/search/v2/base.py
LeoMcA: I'm not sure I understand.

If we have parent P, child C, and non-locale-field foobar, previously:

```
WikiDocument.prepare(P).foobar == P.foobar
WikiDocument.prepare(C).foobar == P.foobar
```

now:

```
WikiDocument.preapre(P).foobar == P.foobar
WikiDocument.prepare(C).foobar raises AttributeError
```

akatsoulas: exactly :+1: 

The only data that we want to keep from the child are the locale aware fields, Everything else should default to the attrs of the parent.

akatsoulas: This was messing with fields like `locale` which was pointing to the one of the child when it should have pointed to the one of the parent. The fields though that was causing problems were either removed or updated in #4574 so we could actually proceed with this and avoid preparing field values twice.

akatsoulas (approved):

PR #4618, opened 2020-12-09
Kill containers in failures.

akatsoulas:


The-smooth-operator (approved):
This makes sense. 
We could flat those 2 `try` blocks, but is not needed

PR #4625, opened 2020-12-14
add timeout option to es7_reindex command

LeoMcA:
https://github.com/mozilla/sumo-project/issues/704

We probably don't want to completely eliminate the possibility of connection timeouts, as that risks clogging up the celery queue, I think this strikes a sensible compromise: allow us to configure the timeout length, allow a connection to timeout once, sleeping for the length of the timeout, before trying again once, then giving up and raising an exception in sentry if that second attempt fails.

### DIFF
--- a/kitsune/search/v2/es7_utils.py
+++ b/kitsune/search/v2/es7_utils.py
@@ -59,15 +59,14 @@ def es_analyzer_for_locale(locale):
     )
 
 
-def es7_client():
+def es7_client(**kwargs):
     """Return an ES7 Elasticsearch client"""
-    init_args = {}
     # prefer a cloud_id if available
     if es7_cloud_id := settings.ES7_CLOUD_ID:
-        init_args.update({"cloud_id": es7_cloud_id, "http_auth": settings.ES7_HTTP_AUTH})
+        kwargs.update({"cloud_id": es7_cloud_id, "http_auth": settings.ES7_HTTP_AUTH})
     else:
-        init_args.update({"hosts": settings.ES7_URLS})
-    return elasticsearch7.Elasticsearch(**init_args)
+        kwargs.update({"hosts": settings.ES7_URLS})
+    return elasticsearch7.Elasticsearch(**kwargs)
 
 
 def get_doc_types(paths=["kitsune.search.v2.documents"]):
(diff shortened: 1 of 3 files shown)


kitsune/search/v2/es7_utils.py
```diff
@@ -59,9 +59,9 @@ def es_analyzer_for_locale(locale):
     )
 
 
-def es7_client():
+def es7_client(**kwargs):
     """Return an ES7 Elasticsearch client"""
-    init_args = {}
+    init_args = kwargs
```
akatsoulas, line 64: we could just update kwargs here and remove the `init_args` completely.

kitsune/search/v2/es7_utils.py
```diff
@@ -111,7 +111,7 @@ def index_object(doc_type_name, obj_id):
 
 
 @task
-def index_objects_bulk(doc_type_name, obj_ids):
+def index_objects_bulk(doc_type_name, obj_ids, timeout=10):
```
akatsoulas, line 114: add a var in settings

kitsune/search/v2/es7_utils.py
```diff
@@ -126,7 +126,13 @@ def index_objects_bulk(doc_type_name, obj_ids):
     if doc_type.update_document:
         action = "update"
         kwargs.update({"doc_as_upsert": True})
-    es7_bulk(es7_client(), (doc.to_action(action=action, is_bulk=True, **kwargs) for doc in docs))
+
+    es7_bulk(
+        # if the request doesn't resolve within `timeout`, sleep for `timeout` then try again once,
```
akatsoulas, line 131: let's add the comments a line above, before the `es_buld(`

kitsune/search/v2/es7_utils.py
```diff
@@ -126,7 +126,13 @@ def index_objects_bulk(doc_type_name, obj_ids):
...
+        # before raising an exception:
+        es7_client(timeout=timeout, retry_on_timeout=True, initial_backoff=timeout, max_retries=1),
```
akatsoulas, line 133: let's add a var for max_retries

akatsoulas (approved):
r+wc

kitsune/search/v2/es7_utils.py
LeoMcA: should we remove the command line argument, then, because it could just be specified as an environment variable instead, like:
```
ES_BULK_TIMEOUT=10 ./manage.py es7_reindex
```

LeoMcA: Oh, actually, that won't work with celery - because the workers won't have that specified value in their environment

akatsoulas (approved):

PR #4626, opened 2020-12-14
fix: searches containing stop words returned no results in KB

LeoMcA:
https://github.com/mozilla/sumo-project/issues/710

This was happening because we're searching over a keyword field in the KB, and this seemed to make simple_query_string revert to the default analyser within the wiki_documents index when that field was included. The query is analysed once for all fields, so elastic must have been reverting to the *lowest common denominator*, as it were.

This is the simplest fix, but comes with the downside that now things may not match on the keywords field where they would before. For instance, if a keywords field contains `customize` it now won't match a query for `customize`, because it will become `custom` post-analysis. We could do any one of three things here:

1. The simplest, just make the keywords field a text field.
2. Implement our locale-specific analysers as normalizers for keyword fields as well, though we'd only want to apply this to the KB's keywords field, not all keyword fields.
3. Do a bigger change here, and place the logic constructing the simple_query_string queries into the SumoSearch child classes - this way, the WikiSearch class could specify one simple_query_string with analysis for the text fields, and one without for the keywords field. Something like this be necessary in the future if we wanted to implement multi-locale searches.

### DIFF
--- a/kitsune/search/v2/base.py
+++ b/kitsune/search/v2/base.py
@@ -158,6 +158,11 @@ class SumoSearch(ABC):
         pass
 
     @abstractmethod
+    def get_analyzer(self):
+        """The analyzer which should be used on the search query string."""
+        pass
+
+    @abstractmethod
     def get_highlight_fields(self):
         """An array of fields to highlight."""
         pass
@@ -187,6 +192,7 @@ class SumoSearch(ABC):
             query=query,
             default_operator=default_operator,
             fields=self.get_fields(),
+            analyzer=self.get_analyzer(),
         )
 
         # add highlights for the search class' highlight_fields
--- a/kitsune/search/v2/base.py
+++ b/kitsune/search/v2/base.py
@@ -158,11 +158,6 @@ class SumoSearch(ABC):
         pass
 
     @abstractmethod
-    def get_analyzer(self):
-        """The analyzer which should be used on the search query string."""
-        pass
-
-    @abstractmethod
     def get_highlight_fields(self):
         """An array of fields to highlight."""
         pass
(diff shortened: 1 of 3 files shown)


LeoMcA:
@akatsoulas pr updated with storing keywords in a text field.

I've left the original commit specifying which analyser to use in the search because, at least according to the ES docs, this should be necessary:

> Defaults to the index-time analyzer mapped for the default_field. If no analyzer is mapped, the index’s default analyzer is used. 

Although this isn't the behaviour I've observed - which is the analyser of the fields being searched over is the one used - a future update may change that, so I think we should be explicit here.

akatsoulas (approved):
r+wc. Let's remove the definition of the analyzers and leave ES handle this per our discussion

LeoMcA:
For future reference: without an analyser specified, elastic analyses the query separately for each field being searched over according to the analyser mapped for that field. This actually may be reflected correctly in the docs, since the `default_field` by default is `*`.

For instance:
```
GET /sumo_wiki_document/_search
{
  "query": {
    "bool": {
      "must": [
        {
          "simple_query_string": {
            "query": "customize",
            "default_operator": "AND",
            "fields": [
              "content.en-US",
              "content.de"
            ]
          }
        }
      ]
    }
  },
  "size": 0,
  "profile": true
}
```
returns:
```
"profile" : {
    "shards" : [
      {
        "id" : "[EpfG8EpYR4avryqfRD-wng][sumo_wiki_document][0]",
        "searches" : [
          {
            "query" : [
              {
                "type" : "BooleanQuery",
                "description" : "content.de:customiz content.en-US:custom",
```

Where we see `customize` analysed differently for the `content.en-US` field and the `content.de` field.

PR #4627, opened 2020-12-17
use default elasticsearch 7 locale analyzers with html_char filter

LeoMcA:
elasticsearch doesn't allow analyzers to be extended, so the analyzers
we want to use must be re-defined as custom analyzers

https://github.com/mozilla/sumo-project/issues/712

### DIFF
--- a/kitsune/search/config.py
+++ b/kitsune/search/config.py
@@ -16,40 +16,343 @@ QUESTION_INDEX_NAME = get_index_name("question")
 USER_INDEX_NAME = get_index_name("user")
 FORUM_INDEX_NAME = get_index_name("forum_document")
 
-ES_LOCALE_ANALYZERS = {
-    "ar": "arabic",
-    "bg": "bulgarian",
-    "cs": "czech",
-    "fa": "persian",
-    "hi-IN": "hindi",
-    "id": "indonesian",
-    "ja": "cjk",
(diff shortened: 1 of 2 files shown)


kitsune/search/config.py
```diff
@@ -16,40 +16,365 @@ def get_index_name(index_name):
...
+#     "it": "Italian",
+#     "no": "Norwegian",
+#     "pt-BR": "Portuguese",
+#     "pt-PT": "Portuguese",
+#     "ro": "Romanian",
+#     "ru": "Russian",
+#     "es": "Spanish",
+#     "sv": "Swedish",
+#     "tr": "Turkish",
+# }
```
LeoMcA, line 39: I've used the default [`stemmer`](https://www.elastic.co/guide/en/elasticsearch/reference/current/analysis-stemmer-tokenfilter.html) for each language below, because that was the default in elastic.

We could use these snowball locales instead, with the [`snowball`](https://www.elastic.co/guide/en/elasticsearch/reference/current/analysis-snowball-tokenfilter.html) tokenfilter. In fact some of these snowball stemmers are available through the `stemmer` tokenfilter, and some are even the default.

I haven't been able to find much in the way of documentation about which stemmers are better than others.

akatsoulas (approved):
r+wc

PR #4634, opened 2021-01-04
prevent naming collisions between filters of the same type in a locale

LeoMcA:
https://github.com/mozilla/sumo-project/issues/722

### DIFF
--- a/kitsune/search/v2/es7_utils.py
+++ b/kitsune/search/v2/es7_utils.py
@@ -1,5 +1,7 @@
 import importlib
 import inspect
+from hashlib import shake_128
+import json
 
 from celery import task
 from django.conf import settings
@@ -19,7 +21,11 @@ def _insert_custom_filters(locale, filter_list, char=False):
 
     def mapping_func(filter):
         if type(filter) is dict:
-            name = f'{locale}_{filter["type"]}'
+            # sort filter dict by keys, so reordering doesn't change hash
+            sorted_filter = {key: filter[key] for key in sorted(filter)}
+            # hash to prevent collisions between filters of the same type in a locale
+            hashed_filter = shake_128(json.dumps(sorted_filter).encode("utf-8")).hexdigest(2)
+            name = f'{locale}_{filter["type"]}_{hashed_filter}'
             if char:
                 return char_filter(name, **filter)
             return token_filter(name, **filter)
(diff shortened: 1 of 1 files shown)


kitsune/search/v2/es7_utils.py
```diff
@@ -19,7 +21,11 @@ def _insert_custom_filters(locale, filter_list, char=False):
 
     def mapping_func(filter):
         if type(filter) is dict:
-            name = f'{locale}_{filter["type"]}'
+            # sort filter dict by keys, so reordering doesn't change hash
+            sorted_filter = {key: filter[key] for key in sorted(filter)}
```
akatsoulas, line 25: wouldn't be simpler to tune the naming schema instead?

LeoMcA: The only common key across all filters is `"type"`. Although, I guess naming them based on their position in the locale's filter list would work. I'll try that.

kitsune/search/v2/es7_utils.py
```diff
@@ -17,15 +17,16 @@ def _insert_custom_filters(locale, filter_list, char=False):
     on the value of the `char` argument.
     """
 
-    def mapping_func(filter):
+    def mapping_func(index_filter):
+        index, filter = index_filter
```
akatsoulas, line 21: since this is related to search in general, I would avoid to use the `index` word as it causes confusion with the search index. Anything like `ctr`, `positional_index` or anything that makes it clearer that we are not talking about the search index should do. Maybe a plain comment is the simplest. Nice workaround

akatsoulas: and maybe replace index_filter with something else? filter_args?

akatsoulas (approved):
r+wc

kitsune/search/v2/es7_utils.py
LeoMcA: I went with `index_filter` as that was the approach [used in the PEP which removed tuple parameter unpacking](https://www.python.org/dev/peps/pep-3113/#no-loss-of-abilities-if-removed) (which would have been useful here):

```python
def fxn(a, b_c, d):
    b, c = b_c
    pass
```

I'll add a `_tuple` suffix to make things clearer.

PR #4641, opened 2021-01-08
add defaults for es7_reindex --updated-after per doctype

LeoMcA:
https://github.com/mozilla/sumo-project/issues/724

### DIFF
--- a/kitsune/search/v2/management/commands/es7_reindex.py
+++ b/kitsune/search/v2/management/commands/es7_reindex.py
@@ -6,6 +6,7 @@ from django.db import connection, reset_queries
 from dateutil.parser import parse as dateutil_parse
 
 from kitsune.search.v2.es7_utils import get_doc_types, index_objects_bulk
+from kitsune.search import config
 
 
 class Command(BaseCommand):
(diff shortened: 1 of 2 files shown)


kitsune/search/v2/management/commands/es7_reindex.py
```diff
@@ -78,12 +79,20 @@ def handle(self, *args, **kwargs):
 
             before = kwargs["updated_before"]
             after = kwargs["updated_after"]
+            if not before and not after:
+                after = config.DEFAULT_REINDEX_UPDATED_AFTER.get(dt.__name__)
```
akatsoulas, line 83: why do we care about `before` here?

akatsoulas: depending on the answer about `before` this could be also written like:
```python
after = kwargs.get('updated_after', get_default_timeperiod())
```

kitsune/search/v2/management/commands/es7_reindex.py
```diff
@@ -78,12 +79,20 @@ def handle(self, *args, **kwargs):
...
             if before or after:
```
akatsoulas, line 84: not part of the pr but I would love if we could remove all these `if` statements for `before` and `after` or structure them differently by chaining the query.
```python
if before:
    qs = qs.filter(before)
if after:
   qs = qs.filter(after)
```
We could also move the `try/except` into the manager.

probably this comment falls under the nit category

kitsune/search/v2/management/commands/es7_reindex.py
```diff
@@ -78,12 +79,20 @@ def handle(self, *args, **kwargs):
...
+                    if before and after:
+                        print(f"Indexing objects updated between {after} and {before}")
+                    elif before:
+                        print(f"Indexing objects updated before {before}")
+                    else:
+                        print(f"Indexing objects updated after {after}")
                 except NotImplementedError:
                     print(
-                        f"{model} hasn't implemeneted an `updated_column_name` property."
+                        f"{model} hasn't implemeneted an `updated_column_name` property. "
```
akatsoulas, line 95: was this intentional?

akatsoulas (changes requested):

PR #4642, opened 2021-01-11
Exclude archived wiki articles from results.

akatsoulas:


castroed47 (approved):

kitsune/search/v2/search.py
```diff
@@ -114,7 +115,13 @@ def get_highlight_fields(self):
         ]
 
     def get_filter(self):
-        filters = [DSLQ("term", _index=self.get_index())]
+        # Add default filters:
+        # 1. limit scope to the Wiki index
+        # 2. exclude archived articles
+        filters = [
+            DSLQ("term", _index=self.get_index()),
+            DSLQ("term", is_archived=False),
```
LeoMcA, line 123: `is_archived` is a multi-locale field, so as written this will filter out all wiki documents. The field name needs to be `f"is_archived.{self.locale}"`.

kitsune/search/v2/search.py
```diff
@@ -114,7 +115,13 @@ def get_highlight_fields(self):
         ]
 
     def get_filter(self):
-        filters = [DSLQ("term", _index=self.get_index())]
+        # Add default filters:
+        # 1. limit scope to the Wiki index
+        # 2. exclude archived articles
```
LeoMcA, line 120: nit: I'd put each of these comments directly above the line which does the commented-on filtering

LeoMcA (changes requested):

kitsune/search/v2/search.py
```diff
@@ -114,7 +115,13 @@ def get_highlight_fields(self):
         ]
 
     def get_filter(self):
-        filters = [DSLQ("term", _index=self.get_index())]
+        # Add default filters:
+        filters = [
+            # limit scope to the Wiki index
+            DSLQ("term", _index=self.get_index()),
+            # exclude archived articles
+            DSLQ("term", **{"is_archived.{self.locale}": False}),
```
LeoMcA, line 123: The `f` is missing off the start of this f-string, it should be `f"is_archived.{self.locale}"`

LeoMcA (approved):
r+wc

PR #4644, opened 2021-01-13
Exclude documents without revisions.

akatsoulas:


LeoMcA (commented):
Looks good, tests need to be updated.

I think we also want to add a guard within WikiDocument's `prepare()` or individual `prepare_<field>()` methods, so we (or future devs) never accidentally index unreviewed articles.

kitsune/search/v2/tests/signals/test_wiki.py
```diff
@@ -48,27 +49,39 @@ def test_topics_change(self):
...
 
     def test_topic_delete(self):
+        RevisionFactory(document=self.document, is_approved=True)
         topic = TopicFactory()
         self.document.topics.add(topic)
         topic.delete()
 
         self.assertEqual(self.get_doc().topic_ids, [])
+
+    def test_non_approved_revision_update(self):
```
LeoMcA, line 83: I might make this a little more specific, maybe `test_document_without_approved_revision`?

LeoMcA (approved):
r+wc

PR #4646, opened 2021-01-15
use aliases to simplify deploying changes to indices

LeoMcA:
add es7_init to run-post-deploy script

make es7_init print and skip indices it cannot change the configuration on

https://github.com/mozilla/sumo-project/issues/627

### DIFF
--- a/kitsune/search/v2/base.py
+++ b/kitsune/search/v2/base.py
@@ -7,9 +7,10 @@ from elasticsearch_dsl import Document as DSLDocument
 from elasticsearch_dsl import InnerDoc, MetaField
 from elasticsearch_dsl import Search as DSLSearch
 from elasticsearch_dsl import field
+from elasticsearch7.exceptions import NotFoundError
 
 from kitsune.search import HIGHLIGHT_TAG, SNIPPET_LENGTH
-from kitsune.search.config import UPDATE_RETRY_ON_CONFLICT
+from kitsune.search.config import UPDATE_RETRY_ON_CONFLICT, DEFAULT_ES7_CONNECTION
 from kitsune.search.v2.es7_utils import es7_client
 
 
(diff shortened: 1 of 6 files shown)


docs/search-v2-change-indices.md
```diff
@@ -0,0 +1,121 @@
...
+configure the Documents using those indices to now use `sumo_wiki_document_1` and `sumo_question_2`.
+
+#### Write operations
+
+These will be run against the new index immediately,
+as the new index will be initialized at deploy time by the `bin/run-post-deploy.sh` script.
+
+#### Read operations
+
+Place the changes behind a waffle switch,
```
akatsoulas, line 93: I would prefer not to use waffle and add another dependency in our indexing policies. This must be an autonomous process that is not dependent on any external library.

One approach that we could follow is to use two aliases. One for read and one for write. The read alias though can point to both the new and the old index. This will help as avoid stale data but will create duplicate results. For that case we could conditionally collapse on the ID when the aliases do not point to the exact same index. 

If this is to big of an overhead, we could either use two aliases, one for read and one for write and tolerate stale data while the indexing process is active or use a single one. In any case, this will give us an autonomous indexing process that has no external dependencies. I fear that waffle is over complicating things without any benefits, at least at a first glance. Please let me know if there is anything around waffle though though that can weight on our decision.

akatsoulas (changes requested):

docs/search-v2-change-indices.md
```diff
@@ -0,0 +1,128 @@
...
+```
+
+## Changes requiring a new index
+
+### Development process
+
+#### Config changes
+
+Make any changes to Documents, analyzers or index configuration which you need to do as usual,
+and increment the `Document.Index.version` attribute.
```
akatsoulas, line 82: Since we are going ahead with aliases we can use them everywhere and automate the naming index process by using timestamps

docs/search-v2-change-indices.md
```diff
@@ -0,0 +1,128 @@
...
+
+#### Write operations
```
akatsoulas, line 84: Write operations should also use aliases and update both new and old indices. For the wiki index we could have the following:
* `new_wiki_index_alias -> wiki_index_timestamp_now`
* `current_wiki_index_alias  -> wiki_index_timestamp_old`

We always read and write from/to the `current_wiki_index_alias`.
When there is a write operation and the `new` index is defined we write there too.
When the reindex is complete we point current to `wiki_index_timestamp_new` and delete the reference to the new one.

akatsoulas: We can do this regardless if we are using timestamps or versions. My goal here is twofold: minimize user interaction in the process as this can lead to errors and also provide both fresh data to the old index or avoid stale data to the new one.

docs/search-v2-change-indices.md
```diff
@@ -0,0 +1,128 @@
...
+and writes don't go through the alias.
+
+#### Read operations
+
+Place the changes behind a call to the `Document.alias_points_at_latest_version` property,
+where the prior behavior is what's run when the property returns False,
+and the new behavior is what's run when the property returns True.
+
+For instance,
+if you were adding an `all_names` field to `ProfileDocument` to search over only one field,
```
akatsoulas, line 97: We discussed this in the past too but the more I am thinking about this the more uncomfortable I get with such an approach. If we must have some kind of control on the queries because of the reindexing I strongly prefer a two step release for this.

Having said that, with the `simple_query_string` we get fault tolerance for and worst case if we try to query for a field while our aliases have diverged we will get no results. In most cases during a re-index the aliases should allow searching and indexing to work simultaneously just fine. If there is a case that it's not enough, I would prefer to have a SOP on how to approach that problem than dumping this to the code and/or feature flippers.

akatsoulas (changes requested):

kitsune/search/v2/base.py
```diff
@@ -22,6 +23,86 @@ class Meta:
...
+
+    @classmethod
+    def migrate_reads(cls):
+        """Point the read alias at the same index as the write alias."""
+        cls._update_alias(cls.Index.read_alias, cls.write_alias_points_at)
+
+    @classmethod
+    def _update_alias(cls, alias, new_index):
+        client = es7_client()
+        old_index = cls._alias_points_at(alias)
```
akatsoulas, line 74: nit: I would use read/write here

kitsune/search/v2/documents.py
```diff
@@ -31,8 +31,7 @@ class WikiDocument(SumoDocument):
     doc_id = SumoLocaleAwareKeywordField(store=True)
 
     class Index:
-        name = config.WIKI_DOCUMENT_INDEX_NAME
-        using = config.DEFAULT_ES7_CONNECTION
+        pass
```
akatsoulas, line 34: interesting, do we need to keep it?

kitsune/search/v2/search.py
```diff
@@ -198,5 +198,5 @@ def get_filter(self):
     def make_result(self, hit):
         index = hit.meta.index
         for child in self._children:
-            if index == child.get_index():
+            if index.split("_")[:-1] == child.get_index().split("_")[:-1]:
```
akatsoulas, line 201: nit: why not split("_")[0]? Just to make it more readable

kitsune/search/v2/base.py
```diff
@@ -22,6 +23,86 @@ class Meta:
...
+        if len(aliased_indices) > 1:
+            raise RuntimeError(
+                f"{alias} points at more than one index, something has gone very wrong"
+            )
+
+        return aliased_indices[0] if aliased_indices else None
+
+    @classmethod
+    @property
+    def write_alias_points_at(cls):
```
akatsoulas, line 103: we could remove this wrapper and use directly the `_alias_points_at`

kitsune/search/v2/base.py
```diff
@@ -22,6 +23,86 @@ class Meta:
...
         dynamic = MetaField("false")
 
+    def __init_subclass__(cls, **kwargs):
+        """Automatically set up each subclass' Index attribute."""
+        super().__init_subclass__(**kwargs)
+
+        cls.Index.using = DEFAULT_ES7_CONNECTION
+
+        immediate_parent = cls.__mro__[1]
+        if immediate_parent is SumoDocument:
```
akatsoulas, line 33: let's add a comment on why we need to introspect that

kitsune/search/v2/management/commands/es7_init.py
```diff
@@ -29,11 +35,32 @@ def handle(self, *args, **kwargs):
...
+                index = dt.write_alias_points_at
+                if not index:
+                    print("First time running, creating index and aliases:")
+                    migrate_writes = True
+                    migrate_reads = True
+                else:
+                    print("Updating index")
+                    dt.init(index=index)
+
+            try:
```
akatsoulas, line 56: I would wrap each operation in their own `try/except` clause

akatsoulas (approved):
the comments are mostly nits. This turned out very nice! Fantastic work! :rocket:

kitsune/search/v2/search.py
LeoMcA: With the prefix included (`sumo_wikidocument_20210210`) that wouldn't work - I considered creating a function for this, it'll use a few more chars but probably worth it for the readability, I'll see.

kitsune/search/v2/management/commands/es7_init.py
LeoMcA: I wasn't sure if we'd want to migrate reads if writes failed for whatever reason, that said I'm not sure we'll be migrating both at once in any situation other than local dev.

PR #4647, opened 2021-01-15
exclude article redirects from search results

LeoMcA:
https://github.com/mozilla/sumo-project/issues/705

This is the trivial approach - leaving the documents in ES and filtering them out at query time. Tackling the more complex approach, which we may want to use here, here: https://github.com/mozilla/sumo-project/issues/736

### DIFF
--- a/kitsune/search/v2/documents.py
+++ b/kitsune/search/v2/documents.py
@@ -13,6 +13,7 @@ from kitsune.search.v2.fields import (
 )
 from kitsune.users.models import Profile
 from kitsune.wiki import models as wiki_models
+from kitsune.wiki.config import REDIRECT_HTML
 
 connections.add_connection(config.DEFAULT_ES7_CONNECTION, es7_client())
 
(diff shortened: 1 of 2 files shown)


kitsune/search/v2/documents.py
```diff
@@ -33,6 +34,7 @@ class WikiDocument(SumoDocument):
     slug = SumoLocaleAwareKeywordField(store=True)
     doc_id = SumoLocaleAwareKeywordField(store=True)
     is_archived = SumoLocaleAwareBooleanField()
+    is_redirect = SumoLocaleAwareBooleanField()
```
akatsoulas, line 37: This will add many fields without any apparent benefit. These articles should not have been indexed in the first place. My suggestion is to close this PR and not tackle this during query time. These data won't even give us any useful aggs.

LeoMcA:
Closed in favour of https://github.com/mozilla/kitsune/pull/4649

PR #4649, opened 2021-01-18
Define indexing policies per document.

akatsoulas:


kitsune/search/v2/documents.py
```diff
@@ -47,6 +48,19 @@ def update_document(cls):
...
     def prepare(cls, instance):
         """Override super method to merge docs for KB."""
+        # Add a discard field in the document if the following conditions are met
+        # Wiki document is a redirect
+        # Wiki document is archived
+        # Wiki document is a template
+        if any(
+            [
+                instance.html.startswith(REDIRECT_HTML),
+                instance.is_archived,
```
LeoMcA, line 58: This is currently a localeaware field, which means at least at one point we thought this could differ among locales. If we don't think that's the case any more, we should remove the `is_archived` field from the mapping - though we should check this with Joni first.

kitsune/search/v2/documents.py
```diff
@@ -47,6 +48,19 @@ def update_document(cls):
...
+                instance.category == TEMPLATES_CATEGORY,
```
LeoMcA, line 59: I notice canned responses isn't excluded here (which we currently exclude alongside from templates in wiki search). Was that intentional? I can see us wanting canned responses to allow contributors to search through them to find the one they want - but I can see the same use-case for templates.

Since we don't have either feature built I guess we can remove them from ES for now and we can always start indexing them in future if we want.

kitsune/search/v2/documents.py
```diff
@@ -140,6 +151,17 @@ class Index:
...
+    @classmethod
+    def prepare(cls, instance):
+        """Override super method to exclude certain docs."""
+        # Add a discard field in the document if the following conditions are met
+        # Question document is spam
+        # TODO: figure out if archived questions are excluded too
+        if instance.is_spam:
+            instance.es_discard_doc = True
+
+        return super(QuestionDocument, cls).prepare(instance)
```
LeoMcA, line 163: `super()` should work, why go to all the trouble of upgrading to python 3 if we don't make use of it :stuck_out_tongue:

kitsune/search/v2/base.py
```diff
@@ -37,10 +41,10 @@ def prepare(cls, instance, parent_id=None, **kwargs):
         """
 
         obj = cls()
+
         doc_mapping = obj._doc_type.mapping
         fields = [f for f in doc_mapping]
         fields.remove("indexed_on")
-
         # Loop through the fields of each object and set the right values
         for f in fields:
```
LeoMcA, line 49: Can we put another `hasattr(instance, "es_discard_doc")` check here to conditionally skip the `for f in fields` loop? I can't think of a need to go over the fields preparing the document if we already know we want to delete it. The id is prepared outside the loop, and I think that's all we need.

kitsune/search/v2/base.py
```diff
@@ -79,6 +83,12 @@ def prepare(cls, instance, parent_id=None, **kwargs):
         if parent_id:
             obj.meta.id = parent_id
 
+        # check if the instance is suitable for indexing
+        # the prepare method of each Document Type can mark an object
+        # not suitable for indexing based on criteria defined on each said method
+        if hasattr(instance, "es_discard_doc"):
+            obj.es_discard_doc = True
```
LeoMcA, line 90: The use of `hasattr` and `True` throughout makes me a bit uncomfortable - IMO `True` implies setting `instance.es_discard_doc = False` will keep the doc (when it'll actually delete it) - and the same applies to the `es_discard_doc` attribute on the document obj.

Could we change to either using `getattr(x, "es_discard_doc", False)` or `x.es_discard_doc = None` (I believe that's the pythonic way of initializing an empty attribute)?

kitsune/search/v2/es7_utils.py
```diff
@@ -148,15 +151,19 @@ def index_objects_bulk(doc_type_name, obj_ids, timeout=settings.ES_BULK_DEFAULT_
...
+                timeout=timeout,
+                retry_on_timeout=True,
+                initial_backoff=timeout,
+                max_retries=settings.ES_BULK_MAX_RETRIES,
+            ),
+            (doc.to_action(action=action, is_bulk=True, **kwargs) for doc in docs),
+        )
+    except BulkIndexError as e:
+        # This could be removed and completely silence this errors - need to re-evaluate
+        capture_exception(e)
```
LeoMcA, line 166: What's the advantage of using `sentry_sdk.capture_exception` over just re-raising it?

I don't think there's ever a situation where we want to completely silence `BulkIndexError` - I remember conflict errors are raised this way, and timeouts might also be raised with this error - both need reporting in sentry so we know to reindex the affected documents.

Is this for the case where we're deleting a document which doesn't exist in ES? It should be possible to parse the error to ignore those errors - they're raised in `/vendor/elasticsearch/helpers/actions.py#L188`

LeoMcA (commented):
First pass, haven't tested yet, but looking good!

kitsune/search/v2/documents.py
```diff
@@ -47,6 +48,19 @@ def update_document(cls):
     @classmethod
     def prepare(cls, instance):
         """Override super method to merge docs for KB."""
+        # Add a discard field in the document if the following conditions are met
+        # Wiki document is a redirect
+        # Wiki document is archived
+        # Wiki document is a template
+        if any(
+            [
+                instance.html.startswith(REDIRECT_HTML),
```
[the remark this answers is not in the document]
akatsoulas: https://github.com/mozilla/sumo-project/issues/738

kitsune/search/v2/documents.py
akatsoulas: regardless, I will update this PR to check the flag for each specific locale. If this turns out to be the same for all children we can make the field non locale aware in our index.

kitsune/search/v2/documents.py
akatsoulas: I guess it's D.R.Y vs explicit here. In any case it's a matter of preference. After all super() does the same thing behind the scenes.

Fun-fact: supper = super and then call supper()

kitsune/search/v2/base.py
akatsoulas: I get what you are saying but I am trying to find an nice way to update this but the `hasattr` has nothing to do with the value of the attr. It just checks whether an attr exists. This is more suitable compared to getattr which will return the default value eg False if the attr is not found but it will return the value of the attr if found which is more prone to errors.

kitsune/search/v2/base.py
akatsoulas: I thought of that but elasticsearch does a copy of the object and returns a 500 error because it cannot find the data it expects.

kitsune/search/v2/documents.py
LeoMcA: :roll_eyes:

kitsune/search/v2/base.py
LeoMcA: When does that copy happen? I've wrapped the for loop in a `if not hasattr(instance, "es_discard_doc"):` block, and it works for me with `ProfileDocument`

kitsune/search/v2/base.py
LeoMcA: We could add a helper function like:
```python
def es_discard_doc(instance_or_obj):
    return bool(getattr(instance_or_obj, "es_discard_doc", False))
```

akatsoulas: I still believe that `hasattr` is a better fit. We care about the existence of an attribute and not it's value. Some documents will have it, others will not.

With the snippet above, the `getattr` will return either False b/c the attr is missing or the actual value of the attribute. The function will end up casting to bool and check either for the existence of an attribute or return the actual value. These are different things. It's going to work just fine because of the cast and the use of booleans though I don't see the reason to exchange the hasattr with getattr and force a comparison of different things.

LeoMcA: > We care about the existence of an attribute and not it's value. Some documents will have it, others will not.

If this is the case, then we shouldn't be doing:
```
instance.es_discard_doc = True
```
Because that then implies that setting:
```
instance.es_discard_doc = False
```
won't discard the doc, when it still will.

Happy to keep `hasattr`, but then we should do:
```
instance.es_discard_doc = None
```
to make it clear that it's the existence of the attribute, not it's value, which controls whether a doc is discarded.

LeoMcA: Oh, I see you've set it to `"unindex_me"`, that'll work.

kitsune/search/v2/es7_utils.py
```diff
@@ -148,15 +150,24 @@ def index_objects_bulk(doc_type_name, obj_ids, timeout=settings.ES_BULK_DEFAULT_
...
+            (doc.to_action(action=action, is_bulk=True, **kwargs) for doc in docs),
+        )
+    except BulkIndexError as exc:
+        exc.errors[:] = [
+            error
+            for error in exc.errors
+            if (error.get("delete") and error["delete"]["status"] not in [400, 404])
+        ]
+        if exc.errors:
+            raise exc
```
LeoMcA, line 170: This is :chefs-kiss-emoji:, though a comment explaining this is modifying in place might be nice (since it's pretty obscure syntax).

I'm not sure the if statement is quite right, though:
if `error = { "update": {} }`, then the statement returns `None` and that error would be excluded from the array - when we want it to be included and raised.
I think the if statement needs to be:
```
if not (error.get("delete") and error["delete"]["status"] in [400, 404])
```

kitsune/search/v2/base.py
```diff
@@ -37,11 +40,15 @@ def prepare(cls, instance, parent_id=None, **kwargs):
...
+
         doc_mapping = obj._doc_type.mapping
         fields = [f for f in doc_mapping]
         fields.remove("indexed_on")
-
         # Loop through the fields of each object and set the right values
+
+        # check if the instance is suitable for indexing
+        # the prepare method of each Document Type can mark an object
+        # not suitable for indexing based on criteria defined on each said method
```
LeoMcA, line 51: What's this comment doing here?

kitsune/search/v2/documents.py
```diff
@@ -196,15 +214,24 @@ class AnswerDocument(QuestionDocument):
...
     num_unhelpful_votes = field.Integer()
 
     is_solution = field.Boolean()
 
+    @classmethod
+    def prepare(cls, instance, **kwargs):
+        """Override super method to exclude certain docs."""
+        # Add a discard field in the document if the following conditions are met
+        # Answer document is spam
+        if instance.is_spam:
```
LeoMcA, line 227: I think we want `if instance.is_spam or instance.question.is_spam` here

kitsune/search/v2/documents.py
```diff
@@ -292,6 +312,16 @@ class Index:
         name = config.USER_INDEX_NAME
         using = config.DEFAULT_ES7_CONNECTION
 
+    @classmethod
+    def prepare(cls, instance):
+        """Override super method to merge docs for KB."""
```
LeoMcA, line 317: Got a misplaced docstring here

kitsune/search/v2/documents.py
```diff
@@ -140,6 +147,17 @@ class Index:
         name = config.QUESTION_INDEX_NAME
         using = config.DEFAULT_ES7_CONNECTION
 
+    @classmethod
+    def prepare(cls, instance):
+        """Override super method to exclude certain docs."""
+        # Add a discard field in the document if the following conditions are met
+        # Question document is spam
+        # TODO: figure out if archived questions are excluded too
```
LeoMcA, line 155: I think we can remove this comment

LeoMcA (approved):
r+wc

PR #4652, opened 2021-01-25
Do not ignore 'python-version' files for pyenv

akatsoulas:
* Add slack-cli dependency

### DIFF
--- a/k8s/requirements.txt
+++ b/k8s/requirements.txt
@@ -1,3 +1,4 @@
 invoke
 jinja2
 pyyaml
+slack-cli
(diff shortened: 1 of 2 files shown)


k8s/requirements.txt
```diff
@@ -1,3 +1,4 @@
 invoke
 jinja2
 pyyaml
+slack-cli
```
The-smooth-operator, line 4: Thanks for adding this, I also noticed it was missing

The-smooth-operator (approved):

PR #4655, opened 2021-01-28
exclude spam from QuestionDocument.answer_content

LeoMcA:
https://github.com/mozilla/sumo-project/issues/737

### DIFF
--- a/kitsune/search/v2/documents.py
+++ b/kitsune/search/v2/documents.py
@@ -168,7 +168,17 @@ class QuestionDocument(SumoDocument):
         return instance.num_votes
 
     def prepare_answer_content(self, instance):
-        return [answer.content for answer in instance.answers.all()]
+        return [
+            answer.content
+            for answer in (
+                # when bulk indexing use answer queryset prefetched in `get_queryset` method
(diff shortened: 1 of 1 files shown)


kitsune/search/v2/documents.py
```diff
@@ -168,7 +168,15 @@ def prepare_question_num_votes(self, instance):
         return instance.num_votes
 
     def prepare_answer_content(self, instance):
-        return [answer.content for answer in instance.answers.all()]
+        return [
+            answer.content
+            for answer in (
+                instance.es_question_answers
```
akatsoulas, line 174: Why are you adding this?

akatsoulas: this == `es_question_answers`

LeoMcA: To avoid N+1ing when doing a bulk question index, while also allowing the query to fallback to another query if a prefetch hasn't been done (like in an individual, not bulk, indexing task)

akatsoulas (approved):
nice work. Pls add a comments so that we haven't to think over it in a couple of months. squash+merge

PR #4656, opened 2021-01-28
restrict QuestionSearch to past 2 years

LeoMcA:
https://github.com/mozilla/sumo-project/issues/740

### DIFF
--- a/kitsune/search/v2/search.py
+++ b/kitsune/search/v2/search.py
@@ -1,4 +1,4 @@
-from datetime import datetime
+from datetime import datetime, timezone, timedelta
 
 import bleach
 from elasticsearch_dsl import Q as DSLQ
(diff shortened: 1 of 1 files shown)


kitsune/search/v2/search.py
```diff
@@ -55,7 +55,13 @@ def get_highlight_fields(self):
 
     def get_filter(self):
         filters = [
+            # restrict to the question index
             DSLQ("term", _index=self.get_index()),
+            # only return questions created in the last two years
+            DSLQ(
+                "range",
+                question_created={"gte": datetime.now(timezone.utc) - timedelta(days=365.25 * 2)},
```
akatsoulas, line 63: let's use a variable for the two years. also why float?

LeoMcA: It's the mean number of days in a year :P

akatsoulas: cool, let's make it an integer. floor or ceil totally up to you

kitsune/search/v2/search.py
```diff
@@ -8,6 +8,8 @@
 from kitsune.search.v2.base import SumoSearch
 from kitsune.sumo.urlresolvers import reverse
 
+QUESTION_SEARCH_DELTA = timedelta(days=365.25 * 2)
```
akatsoulas, line 11: Let's keep only the days here and move the timedelta function into DSLQ query

akatsoulas (approved):
r+wc (squash bfr merging)

PR #4657, opened 2021-01-29
seperate bulk-count es7_reindex option into sql and elastic ones

LeoMcA:
set defaults which shouldn't cause timeouts on prod

https://github.com/mozilla/sumo-project/issues/742


akatsoulas (approved):
This can be a bit confusing if you revisit this code in a few months. Specifically why the DB slice is needed. I wonder if there is any room to add a couple of notes either in the code or the documentation. 

Apart from that, can we bump the SQL query slice to 1K docs for example? Is there any reason not to? 

Looks good r+wc

PR #4659, opened 2021-01-29
Enable ES7 console logging for local dev.

akatsoulas:


LeoMcA (approved):
This is very cool.

Can we add a note in the docs about it? Perhaps under the search v2 development tips section, with a subtitle like "Print the queries being executed in Elastic"

PR #4663, opened 2021-02-02
protect 'add-on' and 'add on' from tokenization

LeoMcA:
https://github.com/mozilla/sumo-project/issues/748

a nice side-effect is this causes "add-ons" to be highlighted properly in search results, before:

Screenshot_2021-02-02 Mozilla Support - 127 0 0 1

after:

Screenshot_2021-02-02 Mozilla Support - 127 0 0 1(2)

### DIFF
--- a/kitsune/search/v2/es7_utils.py
+++ b/kitsune/search/v2/es7_utils.py
@@ -21,7 +21,15 @@ def _insert_custom_filters(locale, filter_list, char=False):
     def mapping_func(position_filter_tuple):
         position, filter = position_filter_tuple
         if type(filter) is dict:
-            name = f'{locale}_{position}_{filter["type"]}'
+            prefix = locale
+            default_filters = config.ES_DEFAULT_ANALYZER["char_filter" if char else "filter"]
+            if filter in default_filters:
+                # detect if this filter exists in the default analyzer
+                # if it does use the same name as the default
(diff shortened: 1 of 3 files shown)


kitsune/search/v2/es7_utils.py
```diff
@@ -21,7 +21,15 @@ def _insert_custom_filters(locale, filter_list, char=False):
     def mapping_func(position_filter_tuple):
         position, filter = position_filter_tuple
         if type(filter) is dict:
-            name = f'{locale}_{position}_{filter["type"]}'
+            prefix = locale
```
akatsoulas, line 24: The first argument of this function is accepting a locale. With this optimization you basically adapt a new naming schema by passing the name of the default analyzer as a locale. Personally I believe that the complexity introduced counteracts the benefits. In any case though we should not treat the analyzers name as a locale

akatsoulas: To be clear, I am not against having a default name where applicable for our filters but how we are doing it should be more straightforward

akatsoulas (changes requested):

LeoMcA:
Closing, as code has been merged into https://github.com/mozilla/kitsune/pull/4666, and comments here should be addressed there.

PR #4670, opened 2021-02-10
Bump requirements.

akatsoulas:

PR #4671, opened 2021-02-10
upgrade pip-compile-multi

LeoMcA:


akatsoulas (approved):
Looks good. Let's merge it after #4670 in order to get the latter out to staging

PR #4679, opened 2021-02-15
instant search: show all results for filtered second query

LeoMcA:
data-href was duplicating functionality with data-instant-search-(un)set-params,
and was the cause of the bug, so we can remove it, DRY the code, and fix the bug

https://github.com/mozilla/sumo-project/issues/131


LeoMcA:
don't forget to `./manage.py nunjucks_precompile`!

akatsoulas (approved):
r+wc
Nice find! There are a couple of references of the data('href') in the instant search tests.

PR #4685, opened 2021-02-18
[bug 1671761] Properly expose FxA help_text to Pontoon

akatsoulas:


kitsune/users/views.py
```diff
@@ -68,8 +68,26 @@ def user_auth(request, notification=None):
...
+        + "{a_open}"
+        + terms_url
+        + "'>"
+        + "Term and Conditions {a_close} and "
+        + "{a_open}"
+        + privacy_url
+        + "'>"
+        + "Privacy Policy</a>."
+        + "{a_close}"
+    ).format(a_open="<a href='", a_close="</a>")
```
LeoMcA, line 84: This is gonna be a very confusing string to localise: `Continuing with Firefox Accounts means you have agreed with the {a_open}https://www.mozilla.org/en-US/about/legal/terms/services/'>Term and Conditions {a_close} and...`

This should be something like this, so we don't pass any urls or html to localisers:

```python
help_text = _(
    "Continuing with Firefox Accounts means you have agreed with the "
    + "{terms_open}Terms and Conditions{terms_close} and "
    + "{privacy_open}Privacy Policy{privacy_close}."
).format(
    terms_open="<a href='https://www.mozilla.org/about/legal/terms/services/'>",
    terms_close="</a>",
    privacy_open="<a href='https://www.mozilla.org/privacy/firefox/'>",
    privacy_close="</a>",
)
```

LeoMcA (approved):
r+wc

PR #4688, opened 2021-02-19
instant search: support using back/forward/refresh to leave and enter searches

LeoMcA:
DRY and simplify the search show/hide code, so we can call it when the history state changes

https://github.com/mozilla/sumo-project/issues/776

### DIFF
--- a/kitsune/sumo/static/sumo/js/instant_search.es6
+++ b/kitsune/sumo/static/sumo/js/instant_search.es6
@@ -2,6 +2,7 @@
 (function($) {
   var searchTimeout;
   var locale = $('html').attr('lang');
+  const searchTitle = "Search | Mozilla Support";
 
   if (window.localStorage.getItem("enable_search_v2") === "true" || $("body").data("readonly")) {
     var search = new k.Search("/" + locale + "/search/v2/");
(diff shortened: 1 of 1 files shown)


kitsune/sumo/static/sumo/js/instant_search.es6
```diff
@@ -245,4 +246,21 @@
...
+    let state = e.type == "popstate" ? e.state : history.state;
+    if (state?.query) {
+      window.k.InstantSearchSettings.hideContent();
+      $('[data-instant-search="form"] input[name="q"]').each(function () {
+        $(this).val(state.query);
+      });
+      search.params = state.params;
+      search.query(state.query, k.InstantSearchSettings.render);
+    } else {
+      window.k.InstantSearchSettings.showContent();
```
akatsoulas, line 260: We should also clear the input field or make it respond to user input (by pressing enter or the search icon)

akatsoulas (approved):
r+wc - this is much nicer!

kitsune/sumo/static/sumo/js/instant_search.es6
LeoMcA: The input was already being cleared in showContent for a couple of fields, but not all of them - it should be now.

Enter or search icon should work anyways, I've opened an issue for that: https://github.com/mozilla/sumo-project/issues/785

PR #4691, opened 2021-02-22
instant search: enable v2 with waffle flag

LeoMcA:
upgrade django-waffle

https://github.com/mozilla/sumo-project/issues/753


kitsune/sumo/jinja2/base.html
```diff
@@ -327,6 +327,8 @@ <h5 class="mzp-c-footer-heading">{{ _('Firefox Accounts') }}</h5>
   <script src="{{ static('jsi18n/{lang}/djangojs.js'|f(lang=request.LANGUAGE_CODE|lower)) }}"></script>
 {% endif %}
 
+<script src="{{ url("wafflejs") }}"></script>
```
akatsoulas, line 330: How about using waffle to control whether we are going to load this?

akatsoulas (approved):
r+wc

kitsune/sumo/jinja2/base.html
akatsoulas: This is a nit feel free to ignore. This will be cached but for the use case that we have it shouldn't matter.

LeoMcA: I'll follow up with some ideas I had to avoid the caching problem in another PR

PR #4692, opened 2021-02-23
put wafflejs script tag behind a waffle switch

LeoMcA:
always load the script on kb articles which will get cached
avoid blowing up JS when waffle object doesn't exist

https://github.com/mozilla/kitsune/pull/4691#discussion_r580872162

### DIFF
--- a/kitsune/sumo/static/sumo/js/instant_search.es6
+++ b/kitsune/sumo/static/sumo/js/instant_search.es6
@@ -12,7 +12,7 @@
   // the local storage flag overrides everything else:
   if (
     localStorageUseV2 ??
-    ($("body").data("readonly") || waffle.flag_is_active("instant_search_v2"))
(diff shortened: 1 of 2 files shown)


kitsune/sumo/jinja2/base.html
```diff
@@ -327,7 +327,9 @@ <h5 class="mzp-c-footer-heading">{{ _('Firefox Accounts') }}</h5>
   {% endif %}
 {% endif %}
 
-<script src="{{ url("wafflejs") }}"></script>
+{% if cached_page or waffle.switch("wafflejs") %}
```
akatsoulas, line 330: Enabling it for all cached pages won't benefit us much. We will end up nullifying this check for a big percentage of our traffic. Let's remove it completely and load it for everyone

akatsoulas (approved):
Let's keep only the JS fix. r+wc

PR #4695, opened 2021-02-24
Normalize usernames before account creation.

akatsoulas:


kitsune/users/utils.py
```diff
@@ -34,8 +34,20 @@ def _make_mail(locale):
     email_utils.send_messages([_make_mail(language_code)])
 
 
+def normalize_username(username):
+    """Removes any invalid characters from the username"""
+
+    regex = compile(r"^[\w.@+\-]+$")
```
LeoMcA, line 40: Should we import this directly from `django.contrib.auth.validators.UnicodeUsernameValidator.regex`?

LeoMcA (approved):
r+wc

It would be nice to have a migration for users whose usernames are already invalid, maybe we want to do that in a separate PR.

akatsoulas:
Let's figure out how to handle existing users in a different issue.

PR #4697, opened 2021-03-02
Exempt KB and Locale groups from spam fitler.

akatsoulas:


akatsoulas:
https://bugzilla.mozilla.org/show_bug.cgi?id=1690914

Adonai907 (changes requested):
---

LeoMcA (approved):
It's not pretty but it'll work for the time being, hopefully we can invest in a proper spam/moderation system soon and tear this out.

PR #4700, opened 2021-03-03
Send a notification after queue moderation.

akatsoulas:


kitsune/flagit/views.py
```diff
@@ -69,6 +71,13 @@ def update(request, flagged_object_id):
     flagged = get_object_or_404(FlaggedObject, pk=flagged_object_id)
     new_status = request.POST.get("status")
     if new_status:
+        ct = flagged.content_type
+        # If the object is an Answer let's fire a notification
+        # if the flag is invalid
+        if new_status == str(FlaggedObject.FLAG_REJECTED) and ct.model_class() == Answer:
+            answer = ct.get_object_for_this_type(pk=flagged_object_id)
```
LeoMcA, line 78: `flagged_object_id` is the id of the `FlaggedObject`, which is why the test is failing, so I think this needs to be `answer = flagged.content_object`.

kitsune/flagit/views.py
```diff
@@ -69,6 +71,13 @@ def update(request, flagged_object_id):
     flagged = get_object_or_404(FlaggedObject, pk=flagged_object_id)
     new_status = request.POST.get("status")
     if new_status:
+        ct = flagged.content_type
+        # If the object is an Answer let's fire a notification
+        # if the flag is invalid
+        if new_status == str(FlaggedObject.FLAG_REJECTED) and ct.model_class() == Answer:
```
LeoMcA, line 77: Will `new_status` always be a string? Might it be better to do `str(new_status) == str(FlaggedObject.FLAG_REJECTED)`?

LeoMcA (approved):
r+wc

PR #4702, opened 2021-03-03
filter the aaq top contributors page by the contribution group

LeoMcA:
also fix the count, and add pagination
fix pagination missing the last partial page

https://github.com/mozilla/sumo-project/issues/770
https://github.com/mozilla/sumo-project/issues/787

### DIFF
--- a/kitsune/community/utils.py
+++ b/kitsune/community/utils.py
@@ -3,40 +3,48 @@ from datetime import date, datetime, timedelta, timezone
 from operator import itemgetter
 
 from django.conf import settings
+from django.contrib.auth.models import Group
 from django.core.cache import cache
 from django.db.models import Count
 
 from kitsune.products.models import Product
-from kitsune.users.models import User, UserMappingType
+from kitsune.users.models import CONTRIBUTOR_GROUP, User, UserMappingType
 from kitsune.wiki.models import Revision
 
 from elasticsearch_dsl import A
 from kitsune.search.v2.documents import AnswerDocument, ProfileDocument
 
 
-def top_contributors_questions(start=None, end=None, locale=None, product=None, count=10):
+def top_contributors_questions(start=None, end=None, locale=None, product=None, count=10, page=1):
     """Get the top Support Forum contributors."""
 
     search = AnswerDocument.search()
 
-    search = search.filter(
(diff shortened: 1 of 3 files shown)


kitsune/community/utils.py
```diff
@@ -3,40 +3,48 @@
...
 
+    # our filters above aren't perfect, and don't only return answers from contributors
+    # so we need to collect more buckets than `count`, so we can hopefully find `count`
+    # number of contributors within
     search.aggs.bucket(
-        # create buckets for the `count + 1` most active contributors
+        # create buckets for the `count * 10` most active users
         "contributions",
-        A("terms", field="creator_id", size=count + 1),
+        A("terms", field="creator_id", size=count * 10),
```
akatsoulas, line 47: I believe it would also make sense to filter based on the doc count. Eg adding a `min_doc_count` should improve a bit the efficiency. We can start with something small like `10` or `20` and iterate

akatsoulas (approved):
r+wc

kitsune/community/utils.py
```diff
@@ -3,40 +3,48 @@
...
         search = search.filter("term", locale=locale)
     if product:
         search = search.filter("term", question_product_id=product.id)
 
+    # our filters above aren't perfect, and don't only return answers from contributors
+    # so we need to collect more buckets than `count`, so we can hopefully find `count`
+    # number of contributors within
     search.aggs.bucket(
-        # create buckets for the `count + 1` most active contributors
+        # create buckets for the `count * 10` most active users
```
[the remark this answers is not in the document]
LeoMcA: I mostly plucked it out of the air, but did a bit of playing around in Kibana to confirm that it would return almost all contributors over a 90 day period.

kitsune/community/utils.py
LeoMcA: However, this would mean we were no longer counting "users with at least one contribution in the last 90 days", but instead users with 10 or 20 contributions in the last 90 days.

akatsoulas: That's true but since this is about the "top contributors" that should be fine. We can discuss this with the team though if this is the case. My comment focused solely on the efficiency here

PR #4705, opened 2021-03-08
Exclude replies from owner from spam fitlering.

akatsoulas:


LeoMcA (approved):

akatsoulas:
I will capture an issue to convert this to a modelform. It will simplify a few things for use moving forwrard

PR #4708, opened 2021-03-09
Allow search params through settings

akatsoulas:
* Remove dfs_query_then_fetch param


akatsoulas:
with the synonym expansion this yields a 50% speedup. Given the amount of data that we have it should not affect results but we should definitely verify this hypothesis

kitsune/settings.py
```diff
@@ -795,6 +795,9 @@ def JINJA_CONFIG():
 ES7_USE_SSL = config("ES7_USE_SSL", default=False, cast=bool)
 ES7_HTTP_AUTH = config("ES7_HTTP_AUTH", default="", cast=Csv())
 ES7_ENABLE_CONSOLE_LOGGING = config("ES7_ENABLE_CONSOLE_LOGGING", default=False, cast=bool)
+# can include configuration parameters like
+# "search_typ": "dfs_query_then_fetch"
```
LeoMcA, line 799: typo of `"search_type"`

LeoMcA (approved):
r+wc

PR #4711, opened 2021-03-10
Split search query per index.

akatsoulas:


kitsune/search/config.py
```diff
@@ -372,3 +372,6 @@
 }
 
 DEFAULT_ES7_CONNECTION = "es7_default"
+
+# Minimum should match value
+ES7_MINIMUM_SHOULD_MATCH = 1
```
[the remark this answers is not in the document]
akatsoulas: This doesn't mean that it cannot be a variable. imo it's much readable to use this instead of having a number. I will revert though

kitsune/search/v2/base.py
```diff
@@ -274,10 +280,25 @@ def get_highlight_fields(self):
...
+        query = kwargs.get("query", "")
+        operator = kwargs.get("default_operator", "AND")
+        return DSLQ(
+            "simple_query_string",
+            query=query,
+            default_operator=operator,
+            fields=self.get_fields(),
+            # everything apart from WHITESPACE as that interferes with char mappings
+            # and synonyms with whitespace in them by breaking up the phrase into tokens,
+            # before they have a chance to go through the filter:
```
LeoMcA, line 298: I'd remove the comment here, because we have it above the constant - we could also remove the constant, since the simple_query_string is now DRY

kitsune/search/v2/search.py
```diff
@@ -79,6 +79,7 @@ def get_filter(self):
             filter=filters,
             # exclude AnswerDocuments from the search:
             must_not=DSLQ("exists", field="updated"),
+            must=super().build_query(**kwargs),
```
LeoMcA, line 82: Would it make more sense for this to be `self.build_query`?

kitsune/search/v2/search.py
```diff
@@ -130,15 +131,15 @@ def get_highlight_fields(self):
...
+    def get_filter(self, **kwargs):
         # Add default filters:
         filters = [
             # limit scope to the Wiki index
             DSLQ("term", _index=self.get_index()),
         ]
         if self.product:
             filters.append(DSLQ("term", product_ids=self.product.id))
-        return DSLQ("bool", filter=filters)
+        return DSLQ("bool", filter=filters, must=super().build_query(**kwargs))
```
LeoMcA, line 142: same here

LeoMcA (approved):
r+, couple of nits

PR #4712, opened 2021-03-11
Use SumoSearch class for Profiles

akatsoulas:


kitsune/search/v2/search.py
```diff
@@ -203,6 +239,23 @@ def get_filter(self, **kwargs):
...
+            "number_of_fragments": 1,
+            # split fragments at the end of sentences:
+            "boundary_scanner": "sentence",
+            # return fragments roughly this size:
+            "fragment_size": SNIPPET_LENGTH,
+            # add these tags before/after the highlighted sections:
+            "pre_tags": [f"<{HIGHLIGHT_TAG}>"],
+            "post_tags": [f"</{HIGHLIGHT_TAG}>"],
+        }
+
```
LeoMcA, line 258: This means that bare `QuestionSearch` or `WikiSearch` don't get any highlighting. Can we just conditionally add this in `SumoSearch` if `len(self.get_highlight_fields) > 0`?

kitsune/search/v2/search.py
```diff
@@ -159,6 +159,42 @@ def make_result(self, hit):
...
+    def get_highlight_fields(self):
+        return []
+
+    def get_filter(self, **kwargs):
+        return DSLQ(
+            "boosting",
+            positive=self.build_query(**kwargs),
+            negative=DSLQ(
+                "bool",
+                must_not=DSLQ("terms", group_ids=kwargs.get("group_ids", [])),
```
LeoMcA, line 180: The `group_ids` kwarg ought to be named `boost_group_ids` or similar

akatsoulas: imo `boost_group_ids` is a bit misleading. Basically we are matching against specific group IDs here and we are not boosting anything related to IDs but rather the results of the filtering

kitsune/search/v2/search.py
```diff
@@ -159,6 +159,42 @@ def make_result(self, hit):
...
+
+    def get_highlight_options(self):
+        return {}
+
+    def make_result(self, hit):
+        return {
+            "type": "document",
+            "avatar": getattr(hit, "avatar", None),
+            "username": hit.username,
+            "name": hit.name,
```
LeoMcA, line 193: this needs to be `getattr(hit, "name", None)`, as blank names won't appear in ES

jinja2 then renders `None` as `"None"`, so it might be worth expanding that to `getattr(hit, "name", None) or ""`

kitsune/search/v2/search.py
LeoMcA: `group_ids` alone implies to me that we're filtering on that list - users not in those groups won't appear at all - which isn't what's happening, but consider this a nit

kitsune/search/v2/search.py
```diff
@@ -159,6 +159,42 @@ def make_result(self, hit):
...
+            ),
+            negative_boost=0.5,
+        )
+
+    def get_highlight_options(self):
+        return {}
+
+    def make_result(self, hit):
+        return {
+            "type": "document",
```
LeoMcA, line 190: This should be type: user

LeoMcA (approved):
r+wc

PR #4714, opened 2021-03-12
Encode username for contributed documents

akatsoulas:


kitsune/users/views.py
```diff
@@ -189,6 +190,8 @@ def deactivation_log(request):
 
 @require_GET
 def documents_contributed(request, username):
+    # plus sign (+) is converted to space
+    username = parse.quote_plus(username)
```
LeoMcA, line 194: This should simply be `replace`, like here: https://github.com/mozilla/kitsune/blob/a34482ddbf59302af00db2f473d84f2224f5b99e/kitsune/users/views.py#L126

LeoMcA (approved):
r+wc

kitsune/users/views.py
akatsoulas: That's what quote_plus is basically doing in a nicer way. I will be updating the `replace` with this one

akatsoulas: https://github.com/python/cpython/blob/6086ae7fd4aeb4089282189673f9bd0cc33abf9b/Lib/urllib/parse.py#L864

LeoMcA: Ah, no, we need to revert this - `quote_plus` does a lot more, since it encodes unsafe URL characters.

For instance, `parser.quote_plus("test+user") == test%2Buser`. The root of this problem is our other views not properly encoding usernames when creating links, strictly speaking `+` in a uri component should be encoded as `%2B` (which is what `quote_plus` is doing).

So, if we fix the root of this problem elsewhere, and `username` becomes `test+user`, quote_plus is gonna turn that into `test%2Buser`, which will then make it not possible to find the user.

Similarly,`@` is a valid character in django usernames, running that through quote_plus results in `%40`.

akatsoulas: https://github.com/mozilla/kitsune/pull/4718

PR #4716, opened 2021-03-12
Drop customercare app

akatsoulas:


kitsune/sumo/static/sumo/scss/components/_editor-tools.scss
```diff
@@ -208,16 +201,16 @@ section.marky {
...
     height: 65%;
-    height:  calc(100% - 225px);
+    height: calc(100% - 225px);
   }
 
   &.processing {
     div.placeholder {
-      background: transparent url(/static/sumo/img/customercare/spinner.gif) no-repeat center center;
+      background: transparent url(/static/sumo/img/spinner.gif) no-repeat center
+        center;
```
LeoMcA, line 213: how did this become two lines

akatsoulas: that would be prettier. Looking into it

kitsune/customercare/models.py
```diff
@@ -1,145 +1,80 @@
...
-                "id": {"type": "long"},
-                "model": {"type": "string", "index": "not_analyzed"},
-                "url": {"type": "string", "index": "not_analyzed"},
-                "indexed_on": {"type": "integer"},
-                "created": {"type": "date"},
-                "locale": {"type": "string", "index": "not_analyzed"},
-                "creator_id": {"type": "long"},
-            }
-        }
+# import json
```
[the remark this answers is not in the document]
akatsoulas: Yes, this will produce a migration that will drop the tables. I could delete them but since we need one more step after the release to remove the app from the `INSTALLED_APPS`, I figured that we can remove the tables after that step in case we need them

LeoMcA (approved):
Needs a rebase - feel free to squash as well. r+wc:

Think we can also drop this:
https://github.com/mozilla/kitsune/blob/f21124ee7e247f0662a9a874e9ad2d053abbd01b/kitsune/settings.py#L1178

Pretty sure tests are failing because of these references to `obj.tiwtter_usernames`:
https://github.com/mozilla/kitsune/blob/f21124ee7e247f0662a9a874e9ad2d053abbd01b/kitsune/users/models.py#L267-L273

mirunacurtean:
@LeoMcA  Shouldn't the AoA reference be removed as well from the Social Media support headline? 
image

LeoMcA:
@mirunacurtean ah, that's just outdated data in our staging DB, it doesn't appear on prod: https://support.mozilla.org/en-US/forums

PR #4721, opened 2021-03-18
remove elastic 2.x from kitsune.wiki.facets.documents_for

LeoMcA:
fetch recent helpful votes from sql, cache them for a day in redis

https://github.com/mozilla/sumo-project/issues/806

### DIFF
--- a/kitsune/wiki/facets.py
+++ b/kitsune/wiki/facets.py
@@ -1,13 +1,12 @@
 import hashlib
+from datetime import datetime, timedelta
 
 from django.conf import settings
 from django.core.cache import cache
-from django.db.models import Count
-
-from elasticsearch.exceptions import TransportError
+from django.db.models import Count, Q
 
 from kitsune.products.models import Topic
-from kitsune.wiki.models import Document, DocumentMappingType
+from kitsune.wiki.models import Document, HelpfulVote
 
 
 def topics_for(product, parent=False):
@@ -73,64 +72,49 @@ def documents_for(locale, topics=None, products=None):
 def _documents_for(locale, topics=None, products=None):
     """Returns a list of articles that apply to passed in topics and products."""
     # First try to get the results from the cache
-    documents = cache.get(_documents_for_cache_key(locale, topics, products))
-    if documents:
+    cache_key = _cache_key(locale, topics, products)
+    documents_cache_key = f"documents_for:{cache_key}"
+    documents = cache.get(documents_cache_key)
+    if documents is not None:
         return documents
 
-    try:
-        # Then try ES
-        documents = _es_documents_for(locale, topics, products)
-        cache.add(_documents_for_cache_key(locale, topics, products), documents)
-    except TransportError:
-        # Finally, hit the database (through cache machine)
-        # NOTE: The documents will be the same ones returned by ES
-        # but they won't be in the correct sort (by votes in the last
-        # 30 days). It is better to return them in the wrong order
-        # than not to return them at all.
-        documents = _db_documents_for(locale, topics, products)
-        cache.add(_documents_for_cache_key(locale, topics, products), documents)
-
-    return documents
-
-
-def _es_documents_for(locale, topics=None, products=None):
-    """ES implementation of documents_for."""
-    s = (
-        DocumentMappingType.search()
-        .values_dict("id", "document_title", "url", "document_parent_id", "document_summary")
-        .filter(
-            document_locale=locale,
-            document_is_archived=False,
-            document_category__in=settings.IA_DEFAULT_CATEGORIES,
-        )
+    qs = (
+        Document.objects.filter(
+            locale=locale,
+            is_archived=False,
+            current_revision__isnull=False,
+            category__in=settings.IA_DEFAULT_CATEGORIES,
+        ).select_related("current_revision", "parent")
+        # speed up query by removing any ordering, since we're doing it in python:
+        .order_by()
     )
-
     for topic in topics or []:
-        s = s.filter(topic=topic.slug)
+        # we need to filter against parent topics for localized articles
+        qs = qs.filter(Q(topics=topic) | Q(parent__topics=topic))
     for product in products or []:
-        s = s.filter(product=product.slug)
+        # we need to filter against parent products for localized articles
+        qs = qs.filter(Q(products=product) | Q(parent__products=product))
 
-    results = s.order_by("document_display_order", "-document_recent_helpful_votes")[:100]
-    results = DocumentMappingType.reshape(results)
-    return results
+    votes_cache_key = f"votes_for:{cache_key}"
+    votes_dict = cache.get(votes_cache_key)
+    if votes_dict is None:
+        votes_query = (
+            HelpfulVote.objects.filter(
+                revision_id__in=qs.values_list("current_revision_id", flat=True),
+                created__gt=datetime.now() - timedelta(days=30),
+                helpful=True,
+            )
+            .values("revision_id")
+            .annotate(count=Count("*"))
+            .values("revision_id", "count")
+        )
+        votes_dict = {row["revision_id"]: row["count"] for row in votes_query}
+        # the votes query is rather expensive, and only used for ordering,
+        # so we can cache it rather aggressively
+        cache.set(votes_cache_key, votes_dict, timeout=settings.CACHE_LONG_TIMEOUT)
 
-
-def _db_documents_for(locale, topics=None, products=None):
-    """DB implementation of documents_for."""
-    qs = Document.objects.filter(
-        locale=locale,
-        is_archived=False,
-        current_revision__isnull=False,
-        category__in=settings.IA_DEFAULT_CATEGORIES,
-    ).prefetch_related("current_revision")
-    for topic in topics or []:
-        qs = qs.filter(topics=topic)
-    for product in products or []:
-        qs = qs.filter(products=product)
-
-    # Convert the results to a dicts to look like the ES results.
     doc_dicts = []
-    for d in qs.distinct():
+    for d in qs:
         doc_dicts.append(
             dict(
                 id=d.id,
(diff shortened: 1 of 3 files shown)


kitsune/wiki/facets.py
```diff
@@ -73,62 +75,43 @@ def documents_for(locale, topics=None, products=None):
...
-            document_category__in=settings.IA_DEFAULT_CATEGORIES,
+    qs = (
+        Document.objects.filter(
+            locale=locale,
+            is_archived=False,
+            current_revision__isnull=False,
+            category__in=settings.IA_DEFAULT_CATEGORIES,
         )
+        .prefetch_related("current_revision")
+        .order_by()  # we're sorting in python, no need to do so in sql
```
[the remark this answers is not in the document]
akatsoulas: This looks like a good candidate to use `Subquery` to get to the desired result with the minimum amount of queries. Also it should be much faster than doing all this queries and sort in python.

kitsune/wiki/facets.py
```diff
@@ -73,62 +75,43 @@ def documents_for(locale, topics=None, products=None):
...
+    votes_cache_key = f"votes_for:{cache_key}"
+    votes_dict = cache.get(votes_cache_key)
+    if not votes_dict:
+        votes_query = (
+            HelpfulVote.objects.filter(
+                revision_id__in=qs.values_list("current_revision_id", flat=True),
+                created__gt=datetime.now() - timedelta(days=30),
+                helpful=True,
+            )
+            .order_by()
```
akatsoulas, line 107: no need for this here

kitsune/wiki/facets.py
```diff
@@ -138,13 +121,21 @@ def _db_documents_for(locale, topics=None, products=None):
...
                 document_parent_id=d.parent_id,
                 document_summary=d.current_revision.summary,
+                display_order=d.display_order,
+                helpful_votes=votes_dict.get(d.current_revision_id, 0),
             )
         )
 
+    # sort the results by ascending display_order and descending votes
+    # sort is stable in python, so we sort by votes first, then display_order
+    doc_dicts.sort(key=itemgetter("helpful_votes"), reverse=True)
```
akatsoulas, line 131: nit: you can use `sorted` with two keys

akatsoulas (changes requested):

kitsune/wiki/facets.py
```diff
@@ -73,62 +75,43 @@ def documents_for(locale, topics=None, products=None):
...
-        .values_dict("id", "document_title", "url", "document_parent_id", "document_summary")
-        .filter(
-            document_locale=locale,
-            document_is_archived=False,
-            document_category__in=settings.IA_DEFAULT_CATEGORIES,
+    qs = (
+        Document.objects.filter(
+            locale=locale,
+            is_archived=False,
+            current_revision__isnull=False,
```
akatsoulas, line 87: shouldn't you also check for `is_approved` here?

kitsune/wiki/facets.py
```diff
@@ -138,13 +121,21 @@ def _db_documents_for(locale, topics=None, products=None):
...
+    # sort the results by ascending display_order and descending votes
+    # sort is stable in python, so we sort by votes first, then display_order
+    doc_dicts.sort(key=itemgetter("helpful_votes"), reverse=True)
+    doc_dicts.sort(key=itemgetter("display_order"))
+
+    cache.set(f"documents_for:{cache_key}", doc_dicts)
     return doc_dicts
 
 
-def _documents_for_cache_key(locale, topics, products):
```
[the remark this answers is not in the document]
LeoMcA: It's creating part of the cache key for the votes now as well, so it's no longer specific to documents.

kitsune/wiki/facets.py
```diff
@@ -73,62 +75,43 @@ def documents_for(locale, topics=None, products=None):
 def _documents_for(locale, topics=None, products=None):
     """Returns a list of articles that apply to passed in topics and products."""
     # First try to get the results from the cache
-    documents = cache.get(_documents_for_cache_key(locale, topics, products))
+    cache_key = _cache_key(locale, topics, products)
+    documents = cache.get(f"documents_for:{cache_key}")
```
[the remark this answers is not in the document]
LeoMcA: It seems inefficient to create the same md5 hash multiple times when we can just do it once and store it in a variable. I'll add a `documents_cache_key` variable as well to only construct that formatted string once, as well.

kitsune/wiki/facets.py
```diff
@@ -1,13 +1,15 @@
...
 from django.core.cache import cache
 from django.db.models import Count
 
-from elasticsearch.exceptions import TransportError
-
 from kitsune.products.models import Topic
-from kitsune.wiki.models import Document, DocumentMappingType
+from kitsune.wiki.models import Document, HelpfulVote
+
+VOTES_CACHE_TIMEOUT = 24 * 60 * 60  # 24 hours
```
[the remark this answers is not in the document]
LeoMcA: It's a rather expensive query, and it doesn't need to be 100% up-to-date, since it just affects ordering - indeed in the previous implementation where the votes were stored in ES it wasn't. I'll add a comment explaining this.

kitsune/wiki/facets.py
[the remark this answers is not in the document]
LeoMcA: The `.order_by()` is to remove any ordering in the sql statement to slightly speed up the query, I'll clarify this in the comment.

Using a subquery was hideously slow, far slower than this approach, even without the ability to cache the votes more aggressively and get an even greater speed-up there.

kitsune/wiki/facets.py
LeoMcA: I don't see this in the docs, do you have to use a lambda? https://docs.python.org/3/library/functions.html#sorted

kitsune/wiki/facets.py
LeoMcA: I believe the `current_revision` is always approved:

https://github.com/mozilla/kitsune/blob/3370b8ef85676df11b1dd04b961cb53fe60c1b2f/kitsune/wiki/models.py#L84-L87

kitsune/wiki/facets.py
```diff
@@ -73,62 +75,43 @@ def documents_for(locale, topics=None, products=None):
 def _documents_for(locale, topics=None, products=None):
```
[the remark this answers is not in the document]
LeoMcA: I can't actually think of a better name for this - although it's not faceted search, it's still faceted navigation. Suggestions?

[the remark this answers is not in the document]
LeoMcA: Also I'm not sure there's much be can do to optimise fallback_documents. If we start doing per-doc locale filtering in the query, we'll lose the speedup from having a cached set of all the english documents.

kitsune/wiki/facets.py
```diff
@@ -73,62 +77,43 @@ def documents_for(locale, topics=None, products=None):
...
-    """DB implementation of documents_for."""
-    qs = Document.objects.filter(
-        locale=locale,
-        is_archived=False,
-        current_revision__isnull=False,
-        category__in=settings.IA_DEFAULT_CATEGORIES,
-    ).prefetch_related("current_revision")
     for topic in topics or []:
-        qs = qs.filter(topics=topic)
+        qs = qs.filter(Q(topics=topic) | Q(parent__topics=topic))
```
akatsoulas, line 97: Let's rewrite this query without a `for` loop

kitsune/wiki/facets.py
```diff
@@ -73,62 +77,43 @@ def documents_for(locale, topics=None, products=None):
...
     for product in products or []:
-        qs = qs.filter(products=product)
+        qs = qs.filter(Q(products=product) | Q(parent__products=product))
```
akatsoulas, line 99: same here

kitsune/wiki/facets.py
akatsoulas: or pass a func to the `key`

kitsune/wiki/facets.py
LeoMcA: I've replaced this with `__in`, but of course, the result changes there (from filtering by documents in *all* topics, to filtering by documents in *any* topic), and the tests reflect that.

However, looking through this function's usage, I can't see any situation where it would make more sense to have *all* than *any* (especially when it comes to filtering by product). Shall I revert this change, or change the tests?

kitsune/wiki/facets.py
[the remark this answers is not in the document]
akatsoulas: You can now use the new settings for cache here.

kitsune/wiki/facets.py
```diff
@@ -73,62 +77,43 @@ def documents_for(locale, topics=None, products=None):
...
-        category__in=settings.IA_DEFAULT_CATEGORIES,
-    ).prefetch_related("current_revision")
-    for topic in topics or []:
-        qs = qs.filter(topics=topic)
-    for product in products or []:
-        qs = qs.filter(products=product)
-
-    # Convert the results to a dicts to look like the ES results.
     doc_dicts = []
     for d in qs.distinct():
```
akatsoulas, line 118: Do we need this here? Removing it will benefit performance. I also wonder if we can use `set` in case we do need it. Testing this locally, `set` was faster by 5ms on average

akatsoulas: Thinking on this, the query should be always of a certain size and not return lots of results. Thoughts?

kitsune/wiki/facets.py
akatsoulas: Something worth noting, by using `sorted` I had anything between 6-8x better results when I was timing it.
I used something like
```python
doc_dicts = sorted(doc_dicts, key=lambda item: (-item('helpful_votes'], item['display_order']))
```

kitsune/wiki/facets.py
akatsoulas: I would say let's revert then or update this to filter on all elements. Do we need the filtering on parent topics here?

akatsoulas (approved):
r+wc for the latest comments

kitsune/wiki/facets.py
LeoMcA: I didn't realize you could do `<` comparisons between tuples like that in python, neat.

I don't think the speedup has anything to do with sort/sorted, indeed sort is slightly faster according to the docs, I guess because it's doing the sort in place (which we're ok with here):

> You can also use the list.sort() method. It modifies the list in-place (and returns None to avoid confusion). Usually it’s less convenient than sorted() - but if you don’t need the original list, it’s slightly more efficient.

*from https://docs.python.org/3/howto/sorting.html#sorting-basics*

I also imagine you got a speedup from your lambda sorting the list differently to the two sorts after each other. The tuple should be `(item['display_order'], -item['helpful_votes'])`.

Regardless of any performance improvements, the lambda is a lot more readable than two sorts and a comment explaining that sort is stable, so I'll switch to that.

kitsune/wiki/facets.py
LeoMcA: Yeah I don't think we need the distinct how we're now using this query

PR #4724, opened 2021-03-19
Check if locale exists for documents

akatsoulas:


kitsune/search/v2/search.py
```diff
@@ -86,11 +86,11 @@ def make_result(self, hit):
         # generate a summary for search:
         summary = first_highlight(hit)
         if not summary:
-            summary = hit.question_content[self.locale][:SNIPPET_LENGTH]
+            summary = getattr(hit.question_content, self.locale, "")[:SNIPPET_LENGTH]
```
LeoMcA, line 89: If a question hit lacks a `question_content` in the appropriate locale, it should also lack a title. Wiki docs are also susceptible to this.

I don't think we should make a change like this, because it fundamentally doesn't make sense in the context of search: we only search over `question_content.<locale>`, so if a result lacks content in `question_content.<locale>` it means the user has done a unconventional search, which won't return them anything useful anyway, like `*` or `-foobar`.

I think we should just catch a keyerror like this and return an empty result set, something like (in `base.py`):

```
try:
    self.results = [self.make_result(hit) for hit in self.hits]
except KeyError as e:
    if e.message == self.locale:
        self.results = []
        self.total = 0
    else:
        raise e
```

kitsune/search/v2/base.py
```diff
@@ -243,40 +246,68 @@ def prepare_locale(self, instance):
...
+    Implements the run() function, which will perform the search.
+
+    Child classes should define values for the various abstract properties this
+    class inherits, relevant to the documents the child class is searching over.
+    """
+
+    results_per_page: int = settings.SEARCH_RESULTS_PER_PAGE
+    total: int = 0
+    hits: List[Dict] = dfield(default_factory=list)
+    results: List[Dict] = dfield(default_factory=list)
```
LeoMcA, line 306: Since 3.9 it seems to be preferable to use `dict[]` and `list[]` directly: https://docs.python.org/3/library/typing.html?highlight=typing#typing.Dict
> Deprecated since version 3.9: builtins.dict now supports []. See PEP 585 and Generic Alias Type.

kitsune/search/v2/base.py
```diff
@@ -243,40 +246,68 @@ def prepare_locale(self, instance):
...
+
+    Implements the run() function, which will perform the search.
+
+    Child classes should define values for the various abstract properties this
+    class inherits, relevant to the documents the child class is searching over.
+    """
+
+    results_per_page: int = settings.SEARCH_RESULTS_PER_PAGE
+    total: int = 0
+    hits: List[Dict] = dfield(default_factory=list)
```
LeoMcA, line 305: I don't think each hit is a dict, it's some internal elasticsearch-dsl type which is similar.

kitsune/search/v2/base.py
```diff
@@ -243,40 +246,68 @@ def prepare_locale(self, instance):
...
+    results: List[Dict] = dfield(default_factory=list)
+
+    def get_filter(self, **kwargs):
+        """Apply common filters to all Documents."""
+        return [DSLQ("exists", field=item.split("^")[0]) for item in self.get_fields()]
```
LeoMcA, line 310: A problem with this approach is it excludes documents with a null value in one of those search fields (e.g. questions with no answer, because `answer_content` is then empty [which I've tested], and a wiki document with an empty keywords field [haven't tested but presume to be the case]).

Because of this, and because SumoSearch doesn't concern itself with locale-related stuff otherwise, I think this should be implemented in the child classes, and just on one field we know has to exist for each child, like:

`DSLQ("exists", field=f"question_title.{self.locale}")` and `DSLQ("exists", field=f"title.{self.locale}")`

LeoMcA (approved):
nice! r+wc

PR #4728, opened 2021-03-23
add user questions and user answers views, fetching from the db

LeoMcA:
this removes the dependency on advanced search for these links in
the profile
the user questions view doubles up as the my questions view
bring format of user documents page in line with these new pages
bring pages style in line with other profile pages

https://github.com/mozilla/sumo-project/issues/670


kitsune/users/jinja2/users/base_contributions.html
```diff
@@ -0,0 +1,9 @@
+{% extends "users/base.html" %}
```
akatsoulas, line 1: Let's move this under the `users/base.html`

kitsune/users/jinja2/users/questions.html
```diff
@@ -0,0 +1,25 @@
+{% extends "users/base_contributions.html" %}
+{% set title = _("{name} | Questions")|f(name=display_name(profile.user)) %}
+{% set canonical_url = canonicalize(viewname="users.questions", username=profile.user.username) %}
+{% set active = "user-questions" %}
+
+{% block contributions %}
+  <h2 class="sumo-callout-heading">
```
akatsoulas, line 7: I would go for `sumo-page-subheading` but in any case documents, questions and answers should use the same style

kitsune/users/jinja2/users/answers.html
```diff
@@ -0,0 +1,14 @@
+{% extends "users/base_contributions.html" %}
+{% set title = _("{name} | Recent Answers")|f(name=display_name(profile.user)) %}
+{% set canonical_url = canonicalize(viewname="users.answers", username=profile.user.username) %}
+
+{% block contributions %}
+  <h2 class="sumo-callout-heading">{{ _("Recent Answers") }}</h2>
+  <ul class="mzp-u-list-styled">
+    {% for answer in answers %}
```
akatsoulas, line 8: let's add something here when there aren't any answers (similar to what you have done with the questions)

kitsune/users/jinja2/users/answers.html
```diff
@@ -0,0 +1,14 @@
+{% extends "users/base_contributions.html" %}
+{% set title = _("{name} | Recent Answers")|f(name=display_name(profile.user)) %}
+{% set canonical_url = canonicalize(viewname="users.answers", username=profile.user.username) %}
+
+{% block contributions %}
+  <h2 class="sumo-callout-heading">{{ _("Recent Answers") }}</h2>
```
akatsoulas, line 6: The answers are not the recent but rather the latest 100

kitsune/users/jinja2/users/documents.html
```diff
@@ -0,0 +1,46 @@
+{% extends "users/base_contributions.html" %}
+{% set title = _("{user} | Documents")|f(user=display_name(profile.user)) %}
+{% set canonical_url = canonicalize(viewname="users.documents", username=profile.user.username) %}
+
+{% block contributions %}
+  <h2 class="sumo-page-subheading">{{ _("Documents edited or translated") }}</h2>
+  {% for group in documents|groupby("locale") %}
```
akatsoulas, line 7: similar here, we should have a message if no documents are translated or edited

kitsune/users/views.py
```diff
@@ -187,15 +187,51 @@ def deactivation_log(request):
...
+        {
+            "profile": profile,
+            "answers": answers,
+        },
+    )
+
+
 @require_GET
-def documents_contributed(request, username):
+def documents(request, username):
```
akatsoulas, line 227: imo the previous name was better. `documents` is too generic and it's not clear where is referring to

kitsune/users/views.py
```diff
@@ -187,15 +187,51 @@ def deactivation_log(request):
     return render(request, "users/deactivation_log.html", {"deactivations": deactivations})
 
 
+def questions(request, username):
```
akatsoulas, line 190: similar to the comment in the `documents` view, this should become `questions_contributed`

kitsune/users/views.py
```diff
@@ -187,15 +187,51 @@ def deactivation_log(request):
...
+        request,
+        "users/questions.html",
+        {
+            "profile": profile,
+            "questions": questions,
+        },
+    )
+
+
+def answers(request, username):
```
akatsoulas, line 207: see naming for the other two views

kitsune/users/views.py
```diff
@@ -187,15 +187,51 @@ def deactivation_log(request):
     return render(request, "users/deactivation_log.html", {"deactivations": deactivations})
 
 
+def questions(request, username):
+    # plus sign (+) is converted to space
```
akatsoulas, line 191: The 3 views here basically they are doing the same thing. Personally I would look into combining them into a single one and maybe add a url with optional params

akatsoulas (approved):
r+wc

If you change this PR extensively, I am happy to have another go

akatsoulas:
CirclCI is not very happy

kitsune/users/views.py
LeoMcA: It's inconsistently named with its url which is `/user/:username/documents`, I also can't think of anything else which `kitsune.users.views.{ documents, questions, answers }` could be referring to, other than the instances of those which the user authored.

kitsune/users/views.py
LeoMcA: I think that would add needless complexity: the template logic for documents is very different, and if we ever invest in rendering questions/answers in a better way than just a list of titles, their template logic is likely to be different too.

Additionally, the documents/answers views cater towards users who are contributors, whereas the questions view is for users who aren't, and I imagine combining them all into one contributions route will confuse non-contributors.

kitsune/users/views.py
akatsoulas: True but you also need the context of the project and/or the `users` app. In this case consistently naming all of these endpoints like `{documents, questions, answers}_contributed` should make things clearer. I would go with explicit is better than implicit

kitsune/users/views.py
LeoMcA: Actually a decorator DRYing the user_profile fetching logic works nicely here

kitsune/users/decorators.py
```diff
@@ -0,0 +1,38 @@
...
+
+from django.contrib.auth.models import User
+from django.http import Http404
+from django.shortcuts import get_object_or_404, redirect
+
+from kitsune.sumo.urlresolvers import reverse
+from kitsune.users.models import Profile
+
+
+def get_profile_from_url(func):
```
akatsoulas, line 11: This decorator does not return a profile

kitsune/users/decorators.py
```diff
@@ -0,0 +1,38 @@
...
+    @wraps(func)
+    def wrapped_func(request, *args, **kwargs):
+        # The browser replaces '+' in URL's with ' ' but since we never have ' ' in
+        # URL's we can assume everytime we see ' ' it was a '+' that was replaced.
+        # We do this to deal with legacy usernames that have a '+' in them.
+
+        username = kwargs.pop("username")
```
akatsoulas, line 18: Add a default value

kitsune/users/decorators.py
```diff
@@ -0,0 +1,38 @@
...
+        username = username.replace(" ", "+")
+
+        user = User.objects.filter(username=username).first()
```
akatsoulas, line 21: It looks like this is doing two different things. Either redirects to a specific profile or adds the profile in the kwargs. Can we rethink this so that it will do only something specific? I am a bit hesitant on the idea of having a decorator doing redirects

LeoMcA: I'll remove this change from this PR so we can merge the ES 2.x removal, and file a new issue for this. The redirect it's doing is rather clever (allowing us to use IDs in user routes, rather than usernames which are subject to change), however I'm not sure if we actually make any use of this?

LeoMcA: https://github.com/mozilla/sumo-project/issues/812

PR #4729, opened 2021-03-23
establish guideline for use of html in {% trans %} blocks

LeoMcA:


kitsune/users/jinja2/users/auth.html
```diff
@@ -44,10 +44,14 @@ <h1 class="sumo-page-heading">{{ _('Log in / Sign up with Firefox Accounts') }}<
                   {% if request.session.get('in-aaq') %}
                     {{ _('Never miss an answer. A free Firefox Account lets you view your question history and track or post responses.')}}
                   {% else %}
-                    {{ _('We upgraded our log-in system to replace Mozilla Support accounts with Firefox Accounts. Firefox Accounts
-                    offers one log-in for convenient and secure access to Mozilla services.') }}
-                    <a href="{{ url('wiki.document', 'firefox-accounts-mozilla-support-faq') }}"> {{ _('See our FAQ') }}</a>
-                    {{ _('for more information') }}.
+                    {% trans
+                      link_open='<a href="{}">'.format(url('wiki.document', 'firefox-accounts-mozilla-support-faq'))|safe,
```
akatsoulas, line 48: Le's move this all this code inside the view. I am not particularly fond of having all this logic inside the template. We also have the `auth_page_help_text` inside the view that is used a few lines below.

akatsoulas (approved):
r+wc

kitsune/users/jinja2/users/auth.html
LeoMcA: I disagree, the template is the proper place for this - it's template logic, not view logic. It ends up being truly horrendous if placed in the view
```python
aaq_description = _(
    "Never miss an answer. A free Firefox Account lets you view your question history and "
    "track or post responses."
)
sumo_accounts_description = _(
    "We've upgraded our log-in system to replace Mozilla Support accounts with Firefox "
    "Accounts. Firefox Accounts offers one log-in for convenient and secure access to "
    "Mozilla services. {link_open}See our FAQ{link_close} for more information."
).format(
    link_open='<a href="{}">'.format(
        reverse("wiki.document", args=["firefox-accounts-mozilla-support-faq"])
    ),
    link_close="</a>",
)
description = aaq_description if request.session.get("in-aaq") else sumo_accounts_description
```
`auth_page_help_text` should be constructed within the template as well.

akatsoulas: I did not mean to move all the test in the view but the formatting part. With this PR, this practice is in views, models and now templates. We should come up with a better way to handle this even if it's a couple of lines for a filter/function

LeoMcA:
As it turns out, html tags are formatted very nicely in pontoon, so I don't see a reason to remove them from the strings. This makes these `{% trans %}` blocks much cleaner IMO. I've added a note in the docs, and I think this is the style we should use going forward.

akatsoulas (approved):
Nice! Please update the PR title before merging

PR #4741, opened 2021-04-02
Remove e2e tests.

akatsoulas:


LeoMcA (approved):
r+wc:

The https://github.com/akatsoulas/kitsune/blob/cleanup/docs/tests.rst#functional-ui-tests section can be removed from docs as well

PR #4743, opened 2021-04-05
Update user forums to ES7.

akatsoulas:
* Add a search input in the threads listing
* Migrate forum urls to path
* Small fixes in search v2 views


kitsune/search/v2/base.py
```diff
@@ -329,8 +329,9 @@ def run(self, query, page=1, default_operator="AND", **kwargs):
             self.get_filter(query=query, default_operator=default_operator, **kwargs)
         )
         # add highlights for the search class' highlight_fields
-        search = search.highlight(*self.get_highlight_fields(), **self.get_highlight_options())
-
+        if highlight_options := self.get_highlight_options():
+            highlight_options = highlight_options[0]
```
LeoMcA, line 333: This'll exhibit quite weird behaviour if we set different highlight options across child Searches, but let's fix it in a separate PR: https://github.com/mozilla/sumo-project/issues/819

kitsune/search/v2/documents.py
```diff
@@ -371,10 +371,13 @@ class ForumDocument(SumoDocument):
 
     thread_title = field.Text()
     thread_forum_id = field.Keyword()
+    forum_slug = field.Keyword()
+    thread_id = field.Keyword()
     thread_created = field.Date()
     thread_creator_id = field.Keyword()
     thread_is_locked = field.Boolean()
     thread_is_sticky = field.Boolean()
+    post_id = field.Keyword()
```
LeoMcA, line 380: This shouldn't be necessary, `meta.id` or `_id` is the post id

kitsune/search/v2/search.py
```diff
@@ -195,6 +202,46 @@ def make_result(self, hit):
...
+    def get_fields(self):
+        return ["thread_title", "content"]
+
+    def get_highlight_fields(self):
+        return ["thread_title", "content"]
+
+    def get_filter(self, **kwargs):
+        # Add default filters:
+        filters = [
+            # limit scope to the Wiki index
```
LeoMcA, line 220: s/Wiki/Forum/

kitsune/search/v2/search.py
```diff
@@ -195,6 +202,46 @@ def make_result(self, hit):
...
+        return DSLQ("bool", filter=filters, must=self.build_query(**kwargs))
+
+    def get_highlight_options(self):
+        return {}
+
+    def make_result(self, hit):
+        return {
+            "type": "thread",
+            "title": hit.thread_title,
+            "search_summary": bleach.clean(wiki_to_html(hit.content[:1000]), strip=True),
```
LeoMcA, line 235: Ah, I was wrong, our bleach for other results is a little different, we can just use the `strip_html` function. Additionally we ought to slice after, I saw oddities with a few results because we cut partway through wiki syntax:

```python
strip_html(wiki_to_html(hit.content))[:1000]
```

kitsune/search/v2/views.py
```diff
@@ -1,17 +1,28 @@
...
 from kitsune.products.models import Product
-from kitsune.search.views import _fallback_results
+from kitsune.search.forms import SimpleSearchForm
 from kitsune.search.utils import locale_or_default
 from kitsune.search.v2.search import CompoundSearch, QuestionSearch, WikiSearch
+from kitsune.search.views import _fallback_results
+from kitsune.sumo.api_utils import JSONRenderer
+
+
+def _get_product_title(product_title: str) -> List[str]:
```
LeoMcA, line 19: Shouldn't this annotation be `-> tuple[Product, list[str]]`?

LeoMcA (approved):
r+wc

kitsune/search/v2/views.py
akatsoulas: good point. I added that before switching to tuple. We need to set mypy in pre-commit hook for starters :)

kitsune/search/jinja2/search/includes/result.html
```diff
@@ -1,53 +1,62 @@
...
-          {{ ngettext('1 new this week', '{n} new this week',
-                  result.num_votes_past_week)|f(n=result.num_votes_past_week) }}
-        </li>
-      </ul>{# .thread-meta #}
+  {% set url =  result.url %}
+  <div class="topic-article--text">
+    <h3 class="sumo-card-heading">
+      <a class="title" href="{{ url }}" {% if as == 'aaq' %}target="_blank" {% endif %}>{{ result.title }}</a>
+    </h3>
+    {# the if below can be removed once we have a server side rendered template for all docs #}
```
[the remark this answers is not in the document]
akatsoulas: ideally we can use a single server side template for all `types`. In that case the first level of `if` statement, right under the comment can be removed and use `if`s only for specific fields. Right now we only exclude from this block the wiki type. When the es2 is removed, we could potentially eliminate the if there. I am removing it though as redundant

PR #4744, opened 2021-04-06
Update code settings.

akatsoulas:


.vscode/settings.json
```diff
@@ -16,5 +18,23 @@
...
+    "beautify.language": {
+        "html": [
+            "jinja-html"
+        ]
+    },
+    "[jinja-html]": {
+        "editor.formatOnSave": true,
+        "editor.defaultFormatter": "HookyQR.beautify"
+    }
+}
```
LeoMcA, line 40: Is there a vscode setting we can add to get it to add newlines to the ends of files (like this one).

.vscode/settings.json
```diff
@@ -16,5 +18,23 @@
     ],
     "python.analysis.extraPaths": [
         "/vendor"
-    ]
-}
+    ],
+    "files.associations": {
+        "*.html": "jinja-html"
```
LeoMcA, line 23: Do we need to add this extension to the devcontainer config?

.vscode/settings.json
```diff
@@ -16,5 +18,23 @@
...
+        "jinja-html"
+    ],
+    "beautify.language": {
+        "html": [
+            "jinja-html"
+        ]
+    },
+    "[jinja-html]": {
+        "editor.formatOnSave": true,
+        "editor.defaultFormatter": "HookyQR.beautify"
```
LeoMcA, line 38: Do we need to add this extension to the devcontainer config as well?

LeoMcA (approved):
Nice! r+wc

.vscode/settings.json
akatsoulas: Added it in the settings

.vscode/settings.json
akatsoulas: Added in devcontainer

.vscode/settings.json
akatsoulas: Added in devcontainer

PR #4745, opened 2021-04-06
Update forums urls to use path

akatsoulas:


LeoMcA (approved):
Great! I think it's safe to ignore the test failure (it's a JS test timing out, I've had it happen with previous circle runs, the python tests all passed) and merge

LeoMcA:
Oh, silly me, the mocha tests run before the python tests... looks like there are some failures because of the new behaviour with the trailing slash on some routes.

PR #4750, opened 2021-04-08
deep link to instant search results pages

LeoMcA:
update browser url to deep link while searching with v2
serve stub that launches instant search from /search/v2/
stop showfor from interfering with instant search

https://github.com/mozilla/sumo-project/issues/818

### DIFF
--- a/kitsune/search/jinja2/search/instant-search.html
+++ b/kitsune/search/jinja2/search/instant-search.html
@@ -1,0 +1,21 @@
+{% extends "base.html" %}
+
+{% block breadcrumbs %}
+{% endblock %}
+
+{% block hidden_search_masthead %}
+{% endblock %}
+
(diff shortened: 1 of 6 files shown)


akatsoulas:
Nice work! I tested this locally and it works smoothly, offering a more consistent search experience.

However, this alters the initial scope and completely removes server side rendering for search. This is not a bad thing at all and we should aim for a better, unified user experience. With that said though, there is currently UX work in progress to determine how exactly this unified flow should be. Removing server side rendered `/search` endpoint limits the options of that work.
I discussed this with @madasan and we both agree that we should keep the current functionality as is, including server side templates, until we have a well defined and thought path forward.

LeoMcA:
I'll close, so we can refer back to this when doing the search enhancement chunk of work.

PR #4755, opened 2021-04-09
add SumoSearchPaginator, and make use of it

LeoMcA:
adjust SumoSearch to support its use
make use in forum and community search
adjust generation of pagination json in instant search to make use of it

https://github.com/mozilla/sumo-project/issues/825

### DIFF
--- a/kitsune/search/v2/base.py
+++ b/kitsune/search/v2/base.py
@@ -333,6 +333,8 @@ class SumoSearch(SumoSearchInterface):
         if self.last_key is None or self.last_key != key:
             self.run(key=key)
         if isinstance(key, int):
+            # if key is an int, then self.results will be a list containing a single result
+            # return the result, rather than a 1-length list
             return self.results[0]
         return self.results
 
--- a/kitsune/search/v2/base.py
+++ b/kitsune/search/v2/base.py
@@ -2,9 +2,13 @@ from abc import ABC, abstractmethod
 from dataclasses import dataclass
 from dataclasses import field as dfield
 from datetime import datetime
+from typing import Union, overload
 
 from django.conf import settings
+from django.core.paginator import Paginator as DjPaginator, PageNotAnInteger, EmptyPage
 from django.utils import timezone
+from django.utils.translation import ugettext as _
+
 from elasticsearch7.exceptions import NotFoundError
 from elasticsearch_dsl import Document as DSLDocument
 from elasticsearch_dsl import InnerDoc, MetaField
@@ -275,12 +279,12 @@ class SumoSearchInterface(ABC):
         ...
 
     @abstractmethod
-    def get_filter(self, **kwargs):
+    def get_filter(self):
         """A query which filters for all documents to be searched over."""
         ...
 
     @abstractmethod
-    def build_query(self, **kwargs):
+    def build_query(self):
         """Build a query to search over a specific set of documents."""
         ...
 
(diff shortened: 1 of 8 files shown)


akatsoulas (approved):
nice!

LeoMcA:
@akatsoulas can you take a look at the changes I made in base.py and search.py to dataclass-ify things there when rebasing?

kitsune/search/v2/base.py
```diff
@@ -306,27 +310,46 @@ class SumoSearch(SumoSearchInterface):
     class inherits, relevant to the documents the child class is searching over.
     """
 
-    results_per_page: int = settings.SEARCH_RESULTS_PER_PAGE
-    total: int = 0
-    hits: list[AttrDict] = dfield(default_factory=list)
-    results: list[dict] = dfield(default_factory=list)
+    total: int = dfield(default=0, init=False)
```
akatsoulas, line 313: Nicely done with the `init=False`!

kitsune/search/v2/base.py
```diff
@@ -306,27 +310,46 @@ class SumoSearch(SumoSearchInterface):
...
+
+    @overload
+    def __getitem__(self, key: int) -> dict:
+        ...
+
+    @overload
+    def __getitem__(self, key: slice) -> list[dict]:
+        ...
+
+    def __getitem__(self, key):
```
akatsoulas, line 332: This is clever but it feels too hacky at the same time. You pass  a Compound search object to the paginator and when it tries to get a page you execute the query here and then it returns the result list. The paginator should get the results in the first place and not the query. It works now but it also couples unrelated things together. Moreover, `__getitem__` should not be the method that executes the actual query.

akatsoulas (changes requested):

kitsune/search/v2/base.py
```diff
@@ -306,27 +310,46 @@ class SumoSearch(SumoSearchInterface):
...
+        if self.last_key is None or self.last_key != key:
+            self.run(key=key)
+        if isinstance(key, int):
+            return self.results[0]
```
akatsoulas, line 336: Should this be `0`?

akatsoulas (approved):
mostly nits r+wc

kitsune/sumo/utils.py
```diff
@@ -21,9 +21,9 @@
 POTENTIAL_IP_REGEX = re.compile(r"(?:[0-9]{1,3}\.){3}[0-9]{1,3}")
 
 
-def paginate(request, queryset, per_page=20, count=None):
+def paginate(request, queryset, per_page=20, paginator_cls=paginator.Paginator, **kwargs):
```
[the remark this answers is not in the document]
LeoMcA: `paginate` is mostly used with QuerySets, not ES2 stuff, so we can't really do this

kitsune/search/v2/views.py
```diff
@@ -10,9 +9,12 @@
 from kitsune.products.models import Product
 from kitsune.search.forms import SimpleSearchForm
 from kitsune.search.utils import locale_or_default
+from kitsune.search.v2.base import SumoSearchPaginator
 from kitsune.search.v2.search import CompoundSearch, QuestionSearch, WikiSearch
 from kitsune.search.views import _fallback_results
 from kitsune.sumo.api_utils import JSONRenderer
+from kitsune.sumo.templatetags.jinja_helpers import Paginator as JinjaPaginator
```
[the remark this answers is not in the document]
LeoMcA: This isn't a paginator in the django sense, but rather a jinja helper to render a paginator in a template (which we have to use to get the same behaviour in instant search). I'll rename to something like `PaginatorRenderer` to clarify that

kitsune/search/v2/base.py
LeoMcA: `self.run(key=key)` has already done the offset in the ES query, so `self.results` will be a list with only one item (if the key was an int), agreed this is a little confusing, will add a comment clarifying

PR #4756, opened 2021-04-12
Bump "refresh_interval" to 60s

akatsoulas:


kitsune/search/config.py
```diff
@@ -372,3 +372,6 @@
 }
 
 DEFAULT_ES7_CONNECTION = "es7_default"
+
+# default refresh_interval for all indices
+DEFAULT_ES7_REFRESH_INTERVAL = 60
```
LeoMcA, line 377: This needs to be "60s"

LeoMcA (approved):
r+wc

PR #4758, opened 2021-04-13
add html rendering to search v2 route

LeoMcA:
https://github.com/mozilla/sumo-project/issues/818


akatsoulas (changes requested):
You need to hook up `simple_search` from ES7 to the `search` endpoint in the search/urls.py

akatsoulas (approved):
Ignore my last comment. Looks good r+!

PR #4760, opened 2021-04-13
increase the timeout between sending take question api calls to a minute

LeoMcA:
this was bombarding the backend with calls every second as a user was typing, which in turn would update the database, and reindex the question in elastic

### DIFF
--- a/kitsune/sumo/static/sumo/js/questions.js
+++ b/kitsune/sumo/static/sumo/js/questions.js
@@ -61,7 +61,7 @@
         }
       }
 
-      $('#id_content').on('keyup', _.throttle(takeQuestion, 1000));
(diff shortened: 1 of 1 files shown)


akatsoulas (approved):

LeoMcA:
This is visible in the firefox network panel while typing a reply to a question.

mirunacurtean:
@LeoMcA  I can see it, but it shows with a 400 Bad Request status. This does not affect the posting of the reply. 
image

Verified that a new request is made after erasing previous text and typing in new one or when a sufficient pause is made between typing another line. 
image

mirunacurtean:
To note that the 400 Bad Request shows when rteplying to one own's post. 
A 204 Request status is made when replying to another user's post. 
image

PR #4761, opened 2021-04-14
add snippet of content to users' latest answers pages

LeoMcA:
to distinguish different answers on the same question from each other
use the unparsed content to avoid hitting the cache up to 100 times

### DIFF
--- a/kitsune/users/jinja2/users/answers_contributed.html
+++ b/kitsune/users/jinja2/users/answers_contributed.html
@@ -14,9 +14,10 @@
     {% endif %}
     <ul class="mzp-u-list-styled">
       {% for answer in answers %}
-        <a href="{{ answer.get_absolute_url() }}">
-          <li>{{ answer.question.title }}</li>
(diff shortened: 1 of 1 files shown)


akatsoulas (approved):

mirunacurtean:
Verified:

1. Number of answer in profile page is correctly updated when adding a new one. 
2. New answer is added at the top of the list along with snippet taken from the beginning of the answer. 

image


3. Old answers have all been updated to contain a snippet of the answer to distinguish them from one another. 

image

4. Answers and snippets are displayed with no issues on both desktop and mobile browsers. 

5. Images in snippets are represented through code, for example -  [[Image:Fx47DefaultSearchEngine|width=500]] 

6. Languages that use symbols are also displayed with no issues in snippets.
image

7. When updating an answer the snippet is also updated. 
8. A link will also be represented through code in snippets. 
image

9. The answer links redirect to that particular answer in the question thread.

PR #4763, opened 2021-04-15
Cleanup twitter app.

akatsoulas:


LeoMcA (approved):
r+wc: can't we completely `rm -r kitsune/twitter`, because there's no migrations that'll need to be applied, and then also remove `"kitsune.twitter"` from `INSTALLED_APPS`?

Pretty sure the `kitsune/questions/tests/test_views.py:TestQuestionUpdates.test_no_update_edit` test failure comes from https://github.com/mozilla/kitsune/pull/4759, no idea why it didn't happen there. Can you open a separate PR to remove that test, and rebase this against that?

PR #4773, opened 2021-04-26
re-allow empty query in advanced search

LeoMcA:
5472d99edf8320e20eab38f7b35aa1d9a14928da accidentally made the q parameter required

https://bugzilla.mozilla.org/show_bug.cgi?id=1707552


akatsoulas (approved):

mirunacurtean:
Verified fixed. 
Search with "Post Contains" field empty:

- and no other filter and only filtered by date and relevance are showing as expected. 

- using Created before filter are showing results as expected
- using Created After filter are sowing as expected

Searches using a test keyword in the "Post contains" filter also show results. 
Testes for English and Spanish locales.

PR #4777, opened 2021-04-28
Use /search for ES7

akatsoulas:
* Remove waffle flag.

### DIFF
--- a/kitsune/search/urls.py
+++ b/kitsune/search/urls.py
@@ -4,8 +4,7 @@ from kitsune.search import views
 from kitsune.search.v2 import views as v2_views
 
 urlpatterns = [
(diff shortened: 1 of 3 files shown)


kitsune/search/urls.py
```diff
@@ -4,8 +4,7 @@
 from kitsune.search.v2 import views as v2_views
 
 urlpatterns = [
-    url(r"^$", views.simple_search, name="search"),
+    url(r"^$", v2_views.simple_search, name="search.v2"),
```
LeoMcA: My initial thought was to keep the view name as `name="search"` to avoid having to update it elsewhere, but this works

LeoMcA (approved):
:rocket:

PR #4781, opened 2021-05-04
Remove ES2 from wiki.

akatsoulas:


akatsoulas:
The search API should be updated to use ES7 or dropped completely if it's not used anywhere *within*  the project only

akatsoulas:
closing in favor of #4790

PR #4782, opened 2021-05-05
add advanced search syntax

LeoMcA:
https://github.com/mozilla/sumo-project/issues/843

First draft at this point, should be enough to get the gist of my approach, without overwhelming with a full implementation. Quite a bit still to do, in no particular order:

- [x] DRY `get_advanced_settings()` methods
- [x] add more fields to field_mappings and exact_mappings
- [x] add tests
- [x] error handling
- [ ] add external context modifiers, such as `order:desc:created` and `include:answers`
- [ ] automatically generate documentation (stretch)


akatsoulas:
I believe it would be helpful to have the BNF definitions of our Parser as documentation and for reference in a docstring

kitsune/search/v2/parser/tokens.py
```diff
@@ -0,0 +1,92 @@
...
+        value = self.tokens.value
+        values = None
+
+        mappings = context["settings"]["exact_mappings"]
+        if field in mappings:
+            # map field
+            mapping = mappings[field]
+            field = mapping["field"]
+            # map value
+            if "model" in mapping:
```
akatsoulas, line 85: I am not sure if we want to support this. Curious to hear your thoughts

kitsune/search/v2/parser/__init__.py
```diff
@@ -63,6 +63,6 @@ def __init__(self, query):
     def __repr__(self):
         return repr(self.parsed)
 
-    def elastic_query(self, **context):
+    def elastic_query(self, context={}):
```
akatsoulas, line 66: s/{}/None

kitsune/search/v2/parser/__init__.py
```diff
@@ -0,0 +1,68 @@
...
+
+
+class Parser(object):
+    def __init__(self, query):
+        self.parsed = search_expression.parseString(query)[0]
+
+    def __repr__(self):
+        return repr(self.parsed)
+
+    def elastic_query(self, context={}):
```
akatsoulas, line 66: s/{}/None

andrewhager9:
Still hacking me

kitsune/search/v2/parser/tokens.py
```diff
@@ -0,0 +1,92 @@
...
+
+
+class BaseToken(ABC):
+    def __init__(self, tokens):
+        self.tokens = tokens
+
+    def __str__(self):
+        return str("".join(self.tokens))
+
+    def __repr__(self):
```
akatsoulas, line 14: make this abstractmethod?

andrewhager9:
I don't know how to do any of this.. he's on my wifi

kitsune/search/v2/parser/__init__.py
```diff
@@ -0,0 +1,68 @@
...
+_value = (Regex(r"\"[^\"]+\"") | Regex(r"\([^\(\)]+\)")).setParseAction(removeQuotes) | _token
+
+# operators:
+# a special kind of token which can be nested with any other token (including operators)
+# e.g. NOT a AND field:b:(c OR d)
+_field = Literal("field:") + _arg("field") + _colon
+_not = Literal("NOT")
+_and = Literal("AND")
+_or = Literal("OR")
+_space = White()
```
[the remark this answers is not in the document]
LeoMcA: `infixNotation` seems to assume that there'll be an operator between every term (which is true in arithmetic for instance: `2 + 3 - 1` makes sense `2 + 3 1` does not). However, while `foo AND bar NOT foobar` is valid, `foo bar NOT foobar` is also valid for us. So we need to specify that spaces are our lowest precedence operator.

kitsune/search/v2/parser/tokens.py
```diff
@@ -0,0 +1,92 @@
...
+        self.term += " " + other.term
+        return self
+
+    def elastic_query(self, context):
+        return Q(
+            "simple_query_string",
+            query=self.term,
+            default_operator="AND",
+            fields=context["fields"],
+            flags="PHRASE",
```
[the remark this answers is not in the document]
LeoMcA: Yes, I think so. My understanding is that the `flags` option specifies which operators can be used in the query by the end user (so `AND` enables usage of `+` as an operator), whereas `default_operator` is happening on a lower level to specify whether `foo bar` should turn into `field:foo field:bar` or `+(field:foo field:bar)` (or whatever the the actual lucene being created is).

I haven't verified the above yet, but it seemd to be what was happening when trying things out while developing.

kitsune/search/v2/parser/tokens.py
LeoMcA: It allows us to support queries like `exact:author:leo_` or `exact:product:firefox` without having to index the usernames or slugs in elastic.

kitsune/search/v2/parser/__init__.py
```diff
@@ -0,0 +1,68 @@
...
+
+from .operators import FieldOperator, NotOperator, AndOperator, OrOperator, SpaceOperator
+from .tokens import TermToken, RangeToken, ExactToken
+
+# convenience:
+# DRY things up
+_colon = Literal(":")
+_token = Regex(r"[^\(\)\s]+")  # everything but chars which conflict with the below operators
+_arg = Word(alphas + "_.")
+_value = (Regex(r"\"[^\"]+\"") | Regex(r"\([^\(\)]+\)")).setParseAction(removeQuotes) | _token
```
LeoMcA, line 22: add comment here

kitsune/search/v2/parser/tokens.py
```diff
@@ -0,0 +1,92 @@
...
+
+from django.apps import apps
+from elasticsearch_dsl import Q
+
+
+class BaseToken(ABC):
+    def __init__(self, tokens):
+        self.tokens = tokens
+
+    def __str__(self):
```
akatsoulas, line 11: Let's remove this since if it's not used.

kitsune/search/v2/parser/__init__.py
```diff
@@ -0,0 +1,68 @@
...
+    + stringEnd
+)
+
+
+class Parser(object):
+    def __init__(self, query):
+        self.parsed = search_expression.parseString(query)[0]
+
+    def __repr__(self):
+        return repr(self.parsed)
```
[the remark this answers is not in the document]
LeoMcA: Using raw strings in all the children should take care of this.

kitsune/search/v2/search.py
```diff
@@ -78,6 +109,49 @@ def get_fields(self):
             f"answer_content.{self.locale}",
         ]
 
+    def get_advanced_settings(self):
+        return {
+            "field_mappings": {
+                "title": f"question_title.{self.locale}",
+                "content": f"question_content.{self.locale}",
```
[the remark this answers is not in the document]
LeoMcA: I've updated this so `content` maps to both `question_content` and `answer_content`. `question` maps to just the former, `answer` maps to just the latter.

kitsune/search/v2/parser/tokens.py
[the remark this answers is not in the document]
LeoMcA: Yep this works as I thought, querying with:
```
{
  "query": {
    "simple_query_string": {
      "default_operator": "AND",
      "fields": [
        "question_title.en-US",
        "question_content.en-US"
      ],
      "flags": "PHRASE",
      "query": "foo bar"
    }
  },
  "profile": true,
  "size": 0
}
```
returns:
```
(+question_content.en-US:foo +question_content.en-US:bar) (+question_title.en-US:foo +question_title.en-US:bar)
```
changing `AND` to `OR` returns:
```
question_content.en-US:foo question_content.en-US:bar question_title.en-US:foo question_title.en-US:bar
```

PR #4784, opened 2021-05-11
Enable autoreload in vscode.

akatsoulas:


.vscode/launch.json
```diff
@@ -2,35 +2,34 @@
...
     // For more information, visit: https://go.microsoft.com/fwlink/?linkid=830387
-
     "version": "0.2.0",
     "configurations": [
-
         {
             "name": "Kitsune",
             "type": "python",
-            "request": "launch",
+            "request": "",
```
LeoMcA, line 10: This is an invalid value, vscode throws an error: "Attribute 'request' is missing from the chosen debug configuration".

.vscode/launch.json
```diff
@@ -2,35 +2,34 @@
...
             "program": "/venv/bin/gunicorn",
             "args": [
                 "wsgi.app",
                 "--config",
                 "wsgi/config.py"
-            ]
+            ],
+            "autoreload": {
+                "enable": true
+            }
```
LeoMcA, line 31: This isn't a valid key from what I can see:

https://code.visualstudio.com/docs/python/debugging#_set-configuration-options

.vscode/launch.json
```diff
@@ -40,6 +39,5 @@
                 "port": 5678
             }
         }
-
     ]
-}
+}
```
LeoMcA, line 43: newline

LeoMcA (commented):
What are you trying to do here?

PR #4790, opened 2021-05-19
Remove ES2 throughout the project.

akatsoulas:


requirements/default.txt
```diff
@@ -368,9 +370,9 @@ idna==2.10 \
     # via
     #   -r requirements/default.in
     #   requests
-jinja2==2.11.3 \
-    --hash=sha256:03e47ad063331dd6a3f04a43eddca8a966a26ba0c5b7207a9a9e4e08f1b29419 \
-    --hash=sha256:a6d58433de0ae800347cab1fa3043cebbabe8baa9d29e668f1c768cb87a333c6
+jinja2==3.0.1 \
```
LeoMcA, line 373: Doesn't seem to be anything breaking here, but a couple of deprecations which affect us:

https://jinja.palletsprojects.com/en/3.0.x/changes/#version-3-0-0

> Deprecate the autoescape and with extensions, they are built-in to the compiler.

> pass_context replaces contextfunction and contextfilter.

kitsune/search/urls.py
```diff
@@ -1,10 +1,7 @@
 from django.conf.urls import url
 
-from kitsune.search import views
 from kitsune.search.v2 import views as v2_views
 
 urlpatterns = [
     url(r"^$", v2_views.simple_search, name="search"),
-    url(r"^/xml$", views.opensearch_plugin, name="search.plugin"),
```
LeoMcA, line 8: I think it may be worth keeping the opensearch xml, just without the suggestions piece.

I think all we need to do in `kitsune/search/jinja2/search/plugin.html` is remove the `<Url type="application/x-suggestions+json" />` element, and it'll be compatible with search v2.

LeoMcA (approved):
r+ with tests passing

All the `ElasticTestCase` -> `TestCase` changes are gonna need to become `Elastic7TestCase`s ultimately. However, with them being disabled for so long, there's going to be a number of template and logic changes causing these failures. For this PR I think it's best if we just disable these tests (perhaps keeping the `ElasticTestCase` just for that purpose, or adding a `DisabledTestCase` class or something). IMO removing everything in one PR, then adapting the existing tests in another, will be a cleaner and clearer commit history.

If it's easy enough to do, I think it's worth keeping the opensearchdescription.

PR #4792, opened 2021-05-20
Remove Travis scripts.

akatsoulas:


LeoMcA (approved):
`scripts/create_user_and_superuser.py` also needs to be deleted, since it references `scripts/travis/variables.json`

PR #4800, opened 2021-05-28
add advanced search documentation

LeoMcA:


docs/advanced-search.md
```diff
@@ -0,0 +1,339 @@
...
+
+### AND
+
+`x AND y`
+
+Specifies that both tokens must exist in the document.
+
+```eval_rst
+.. note::
+    A quirk of our current implementation is that ``x AND y`` can return more results than ``x y``.
```
akatsoulas, line 103: Maybe it's worth mentioning that `x y` is also `AND` search but the difference is the requirement of existence in the same field for the latter. I am proposing this b/c usually users expect `OR` operator as the default. It wouldn't matter without the advanced syntax but now it could help avoiding confusion.

docs/advanced-search.md
```diff
@@ -0,0 +1,339 @@
...
+        }
+
+    ``x y`` wouldn't match, but ``x AND y`` would.
+```
+
+### OR
+
+`x OR y`
+
+Specifies that either token must exist in the document.
```
akatsoulas, line 121: Shall we add that by default, this will be applied to both KB and AAQ?

docs/advanced-search.md
```diff
@@ -0,0 +1,339 @@
...
+
+## Exact Token
```
akatsoulas, line 123: rename to something like `Match exact values` or something along these lines? `Exact` defaults to me to the exact phrase and I always need a second to realize it's not the case (can be just me though :) )

docs/advanced-search.md
```diff
@@ -0,0 +1,339 @@
+# Advanced Search
+
+```eval_rst
+.. contents:: :local:
+```
+
+Kitsune supports an advanced search syntax in all its search boxes.
+
+A number of search tokens,
```
akatsoulas, line 9: Maybe add a couple of lines defining what token is since it's referenced in the rest of the text?

akatsoulas (approved):
Nice work! I added a few comments but overall it's looking good. Feel free to merge when you are ready

docs/advanced-search.md
LeoMcA: Not sure I understand this.

Atom1985 (commented):
docs/index.rst

PR #4805, opened 2021-06-02
by default, fallback to matching parent slug for all wiki document routes

LeoMcA:
…utes

fixes what links here and discussion links on show translations page

https://github.com/mozilla/sumo-project/issues/477


akatsoulas:
@escattone since you already went down the rabbit hole with the non-reviewed KB articles, would you like to take this on? I suspect that we only need to adapt Leo's work with the new manager.

akatsoulas:
Fixed in #5193

PR #4806, opened 2021-06-03
fix: 404 error displays in different languages

LeoMcA:
rearrange sumo.urlresolvers.Prefixer to make intent clearer, and ensure it always sets a valid locale code

https://github.com/mozilla/sumo-project/issues/94

### DIFF
--- a/kitsune/sumo/urlresolvers.py
+++ b/kitsune/sumo/urlresolvers.py
@@ -76,7 +76,10 @@ def get_non_supported(lang):
 
 
 def get_best_language(accept_lang):
-    """Given an Accept-Language header, return the best-matching language."""
+    """
+    Given an Accept-Language header, return the best-matching language.
+    This mostly follows the RFCs but read bug 439568 for details.
+    """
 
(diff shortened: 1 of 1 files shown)


LeoMcA:
A bit more explanation:

What was happening was that in routes like `/1/firefox/` (with no country code in the url) for a logged out user (I'm guessing this is why it only happened in nightly for Joni), `Prefixer().locale` was set to `""`. This then meant that `translation.activate("")` was called, which I'm guessing persisted whatever locale was being used in the previous request handled by that thread.

I was able to test the buggy behaviour locally by visiting `http://127.0.0.1:8000/1` and observing the language used in Django Debug Toolbar. If I visited `http://127.0.0.1:8000/de/` then refreshed the error page at `/1`, Django Debug Toolbar then appeared in German. With this fix, it no longer does that.

Atom1985 (commented):
👍

escattone (approved):
Much better. Nicely done, thanks @LeoMcA!

PR #4813, opened 2021-06-11
Remove ES2 DMS entry

akatsoulas:


docs/hacking_howto.md
```diff
@@ -120,71 +129,36 @@ running this command::
...
-    ```
-    docker-compose exec web bash
-    ```
+2.  Add a topic matching a category from that product config,
+    and associate it with the product you just created.
+    You can do this through the admin interface at `/admin/products/topic/add/`.
 
-2. Build the indicies
-    ```
-    $ ./manage.py esreindex
```
LeoMcA, line 167: This is the only bit no longer relevant, I'd suggest updating it to be the es7 command - both the nunjucks precompile and analytics.js note are still very relevant.

LeoMcA (approved):
r+wc

What Markdown formatter are you using? it would be nice to integrate it with pre-commit, and run all our docs through it, to avoid messy diffs like this in the future

akatsoulas:
I missed this PR. I am closing in favor of #4819

PR #4818, opened 2021-06-14
Record deployments in k8s.

akatsoulas:
There are some changes coming from the linter. The important part is the `revisionHistoryLimit` and the `--record` flag for the web pods in the kubectl

### DIFF
--- a/k8s/deploy_utils.py
+++ b/k8s/deploy_utils.py
@@ -7,7 +7,7 @@ from invoke.config import DataProxy
 
 
 def get_kubectl():
-    return os.environ.get('KUBECTL_BIN', 'kubectl')
+    return os.environ.get("KUBECTL_BIN", "kubectl")
 
 
 def todict(obj, classkey=None):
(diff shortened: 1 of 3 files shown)


k8s/deploy_utils.py
```diff
@@ -30,19 +30,22 @@ def render_template(config, template_name, app):
...
-        f.write("\n".encode('utf-8'))
+    namespace = ctx.config["kubernetes"]["namespace"]
+    with tempfile.NamedTemporaryFile(prefix="k8s", suffix=".yaml", delete=False) as f:
+        f.write(template_text.encode("utf-8"))
+        f.write("\n".encode("utf-8"))
         f.flush()
 
     print("Rendering template to:", f.name)
-    cmd = '{} apply -n {} -f {}'.format(get_kubectl(), namespace, f.name)
+    if record:
```
smarnach, line 45: Where does this variable come from? Did you mean to add it as a parameter to this funciton?

smarnach (commented):
Is there a reason we can't add `--record` unconditionally for all deployments?

k8s/deploy_utils.py
akatsoulas: Hm, my bad. Thanks for catching this!

akatsoulas:
No reason at all! Celery and cron run the same image so we do have the image deployed from the web pods. But we should be able to rollback everything if needed. I am amending the PR to include it everywhere.

akatsoulas:
@smarnach PR updated

smarnach (approved):
Looks good to me.

PR #4819, opened 2021-06-14
Cleanup DMS and secrets.

akatsoulas:


k8s/regions/frankfurt/test.yaml
```diff
@@ -62,14 +62,10 @@ app:
...
     dms_cohort_analysis: SECRET
     dms_enqueue_lag_monitor_task: SECRET
     dms_fix_current_revisions: SECRET
     dms_generate_missing_share_links: SECRET
-    dms_process_exit_surveys: SECRET
-    dms_rebuild_kb: SECRET
-    dms_reindex: SECRET
     dms_reindex_es7: SECRET
     dms_reload_question_traffic_stats: SECRET
     dms_reload_wiki_traffic_stats: SECRET
```
LeoMcA, line 71: Did you miss removing `dms_survey_recent_askers` from this file?

k8s/templates/sumo-secrets.yaml
```diff
@@ -1,54 +1,46 @@
-
 apiVersion: v1
 kind: Secret
 metadata:
```
LeoMcA, line 3: Should `dms-calculate-csat-metrics`, `dms-process-exit-surveys`, `es-http-auth`, and `es-urls` be removed from here too?

Since there's a bunch of other secrets missing from here (like the `es7_` settings) do we need this file at all?

scripts/cron.py
```diff
@@ -146,22 +123,6 @@ def job_generate_missing_share_links():
...
-    day="*",
-    hour="00",
-    minute="00",
-    max_instances=1,
-    coalesce=True,
-    skip=settings.READ_ONLY,
-)
-@babis.decorator(ping_after=settings.DMS_REBUILD_KB)
-def job_rebuild_kb():
-    call_command("rebuild_kb")
```
LeoMcA, line 162: Are we confident enough in our signals to rebuild kb articles when needed?

LeoMcA (approved):
Handful of queries, but otherwise lgtm!

PR #4825, opened 2021-06-21
add js and css webpack config

LeoMcA:
add jinja function for loading files from the webpack-build-output.json file
load all javascript through webpack - it's mostly broken because we haven't exposed globals yet - that's our next step


kitsune/sumo/templatetags/jinja_helpers.py
```diff
@@ -453,6 +453,26 @@ def static(path):
         return ""
 
 
+@library.global_function
```
akatsoulas, line 456: Instead of using a Jinja function for loading we could either use the django-webpack-loader or an even better approach could be to use the `HtmlWebpackPlugin` to inject a HTML snippet inside the base template. For the latter, in order to accommodate bundling we could use a single JS file that will dynamically load whatever each page needs. It requires some tweaking in the templates and our JS code to use exports but I believe this way the front-end it's more loosely coupled and will probably be beneficial for the search UI rewrite if we choose to proceed with a framework.

webpack/entrypoints.js
```diff
@@ -1,10 +1,11 @@
 module.exports = {
+  screen: ["sumo/scss/screen.scss"],
   common: [
     "sumo/js/i18n.js",
     "underscore/underscore.js",
     "moment/moment.js",
-    "jquery/dist/jquery.min.js",
-    "jquery/jquery-migrate.js",
+    "jquery",
+    // "jquery/jquery-migrate.js",
```
akatsoulas, line 8: is this intended? shall we remove instead?

webpack/protocol-compat.js
```diff
@@ -0,0 +1,2 @@
+// Necessary until we upgrade protocol to include this fix: https://github.com/mozilla/protocol/issues/687
+window.Mzp = {};
```
akatsoulas, line 2: Could you file an issue to update Protocol that will also reference this file/issue?

kitsune/settings.py
```diff
@@ -459,7 +460,9 @@ def parse_conn_max_age(value):
 TEMPLATES = [
     {
         "BACKEND": "django_jinja.backend.Jinja2",
-        "DIRS": [],
+        "DIRS": [
+            path("dist"),
```
akatsoulas, line 464: why is this needed?

kitsune/settings.py
```diff
@@ -411,6 +411,7 @@ def parse_conn_max_age(value):
 STATICFILES_DIRS = (
     path("assets"),  # emulate bower and css cache busting
     path("jsi18n"),  # Collect jsi18n so that it is cache-busted
+    path("dist"),
```
akatsoulas, line 414: super nit: shall we use something different as a name here? Or even use the `static` folder as the webpack output? Something like static/webpack or similar

kitsune/sumo/static/sumo/js/upload.js
```diff
@@ -1,3 +1,15 @@
+function dialogSet(inner, title) {
+  var kbox = new KBox($('<p/>').text(inner), {
```
akatsoulas, line 2: why did you move this out?

akatsoulas (commented):
Just a few comments as I was skimming through it. I will do a more thorough review asap

akatsoulas:
This is coming along nicely! :+1:

webpack/entrypoints.js
LeoMcA: For now, there's an issue to figure out what happened with it: https://github.com/mozilla/sumo-project/issues/881

webpack/protocol-compat.js
LeoMcA: https://github.com/mozilla/sumo-project/issues/892

kitsune/settings.py
LeoMcA: so we can load the html files created by the webpack html plugin with `{% include ... %}`

kitsune/sumo/static/sumo/js/upload.js
LeoMcA: It needs to be outside the closure so I can expose it globally in webpack - there's a couple of other files which call it

akatsoulas (approved):
r+ nice work!
There are some minor comments. Feel free to adapt and merge

kitsune/sumo/static/sumo/scss/components/_editor-tools.scss
```diff
@@ -59,7 +59,7 @@ div.editor-tools {
 button {
   &.markup-toolbar-button {
     border: 0;
-    background: url(/static/sumo/img/markup.svg) no-repeat 11px 6px;
+    background: url("../img/markup.svg") no-repeat 11px 6px;
```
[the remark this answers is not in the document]
LeoMcA: Yeah, but whatever we're doing in scss is somehow making `.` point to `kitsune/sumo/static/sumo/scss`, I guess `@use`.

Anyway, it'll be nicer to use absolute paths here anyway (through our alias), so I'll change these to `url("sumo/blah")`

PR #4829, opened 2021-07-06
remove unused images from repo

LeoMcA:
my WIP approach for https://github.com/mozilla/sumo-project/issues/875 involves ingesting all images under `kitsune/**/static/**/img`, so any images which exist there which we're not using in the front end increases build times - especially when optimising images as well

this should validate none of these files are used anywhere, when run in the root of the repo:
```sh
git diff --name-only HEAD HEAD~1 | grep -o '[^/]*$' | while read line ; do rg -i "$line" . ; done
```


akatsoulas (approved):
nice!

PR #4830, opened 2021-07-07
Helm chart scaffolding

akatsoulas:


akatsoulas:
@cmharlow this is not ready yet but I wanted to get your input first before spending too much time on it.
Basically I only tuned the deployments and added an extra configmap. Nothing else is configured atm (service for example is not working). 

My idea regarding the different environments is to amend value files. For example we can use `-f values.yaml` for prod and `-f values.yaml -f values.local.yaml` for local development. 

Finally regarding the configmap, it's not hooked into the deployment. I wanted to get your thoughts first whether we should go with env vars or use a volume.

cc @LeoMcA

akatsoulas:
If/when this is looking good I will add commits on top of it to ease the review process until is fully functional.

k8s/sumo/templates/deployment.yaml
```diff
@@ -0,0 +1,206 @@
...
+  name: {{ include "sumo.fullname" . }}-web
+  namespace: {{ .Values.namespace }}
+  labels:
+    app.kubernetes.io/name: {{ include "sumo.name" . }}-web
+    {{- include "sumo.labels" . | nindent 4 }}
+spec:
+  {{- if not .Values.autoscaling.enabled }}
+  replicas: {{ .Values.replicaCount }}
+  {{- end }}
+  revisionHistoryLimit: {{ .Values.config.web.revisionHistoryLimit }}
```
akatsoulas, line 15: I completely skipped the `strategy` section from here as I believe what we have currently it's the default

```yaml
  strategy:
    type: RollingUpdate
    rollingUpdate:
      maxSurge: 25%
      maxUnavailable: 25%
```

cmharlow: yeah, that's fine

cmharlow:
> @cmharlow this is not ready yet but I wanted to get your input first before spending too much time on it.
> Basically I only tuned the deployments and added an extra configmap. Nothing else is configured atm (service for example is not working).
> 
> My idea regarding the different environments is to amend value files. For example we can use `-f values.yaml` for prod and `-f values.yaml -f values.local.yaml` for local development.

this makes sense for now. What we may do in the future, if we use a future version of this helm chart, is to have it installable from this repo, and have the values.yaml per environment specified where we want deploys triggered from. I'll see if I can write up some notes about what i'm thinking here, also as a learning opportunity to walk through the sumo deployments.

> Regarding the configmap, it's not hooked into the deployment. I wanted to get your thoughts first whether we should go with env vars or use a volume.

What we've pivoted to for some web sre services (a lot of the ones I inherited from Alberto, so I'm guessing a fair amount of your codebases) is to aim for:
* configMap template with Values list or map to populate, storing any values that aren't secrets & we want loaded into the runtime as envvars (e.g. DOMAIN_NAME or DEBUG or whatever)
* [externalSecrets template](https://github.com/external-secrets/kubernetes-external-secrets/blob/master/examples/aws-secretsmanager.yaml) with Values list or map to populate metadata about the expected secrets (but not the values). What this does is it allows us to connect the pods via roles to AWS Secrets Manager (or later, GSM) values via the key field. We then expect SRE & some devs (so yall) to be able to handle those secret generation, updates, rotation, etc. manually (or maybe at some point we'll try to automate that side as well). This allows you to indicate what secrets you want to be used as envvars in the runtime as well, without storing any version of the secrets in the repo itself. Open to other ideas though - and going to think about it myself, since i know many k8s folks indicate that secrets via mounts are more secure & can allow pod-specific security policies in a way that envvars can't, and maybe this is an overlap with the secrets management tool secops & services sre uses atm (which works in their workflows where, often, secrets are decrypted then dumped onto s3 buckets & mounts for runtime consumption, afaict).

> Finally all the services for local development are not configured at all.


Sounds good, going to try running on a local cluster myself and see what i find.

akatsoulas:
Closing in favor of #4843 which branches out of the main branch

PR #4841, opened 2021-07-13
Remove bower like asset preparation.

akatsoulas:


prepare_django_assets.js
```diff
@@ -1,21 +0,0 @@
-const util = require('util');
-const cpx = require("cpx");
```
LeoMcA, line 2: We can remove cpx from our deps now: `npm uninstall --save-dev cpx`

LeoMcA (approved):
r+wc

Carmen675 (changes requested):
Remove this please, it’s connected to my phone (has been hacked and taken over)

PR #4854, opened 2021-07-26
Limit GA data fetch to 1 year

akatsoulas:


duallain (approved):
makes a ton of sense to move from forever to 1 year, especially in the context of it sometimes failing, sometimes working.  I'm not the best person to review the djangoy bits of this pr, but the idea seems great from the pov of snitches firing fairly regularly.

kitsune/questions/models.py
```diff
@@ -708,9 +709,8 @@ def reload_from_analytics(cls, verbose=False):
         """Update the stats from Google Analytics."""
         from kitsune.sumo import googleanalytics
 
-        counts = googleanalytics.pageviews_by_question(
-            settings.GA_START_DATE, date.today(), verbose=verbose
-        )
+        one_year_ago = datetime.now() - timedelta(days=365)
```
LeoMcA, line 712: This'll be a `datetime`, whereas previously the first argument to `pageviews_by_question` was a `date`, could this pose a problem?

LeoMcA (approved):
looks good, one query

kitsune/questions/models.py
akatsoulas: it will just add the time portion too but that's a good flag. I am updating this back to date for consistency.

PR #4866, opened 2021-08-04
Aaq step2

akatsoulas:


LeoMcA:
The automatic reformatting of the templates makes it almost impossible to see what's actually changed, can you turn this off in your editor?

akatsoulas:
That's why I separated the PR into 2 commits. The second commits holds the changes and the first one is just formatting.

LeoMcA:
> That's why I separated the PR into 2 commits. The second commit holds the changes and the first one is just formatting.

ah! I hadn't clocked that, apologies

kitsune/products/jinja2/products/includes/topic_macros.html
```diff
@@ -42,11 +42,9 @@ <h3 class="card--title">
         {{ _('Still need help?') }}
       </h3>
       <p class="card--desc">{{ _('We’re here for you. Post a question to our support forums and get answers from our community of experts.') }}</p>
-      <a
-        class="sumo-button primary-button button-lg"
-        href="{{ url('questions.aaq_step1') }}"
-        data-event-category="link click"
-        data-event-action="topic"
+      <a class="sumo-button primary-button button-lg" {% if product %}
+        href="{{ url('questions.aaq_step2', product_key=product_key)}}" {%else %}
```
LeoMcA, line 46: Should be `{% else %}`

nit: might be cleaner to do the if/else above, and `set` a variable which we then use as the value of the href attribute, like:
```
{% if product %}
  {% set aaq_url = ... %}
{% else %}
  {% set aaq_url = ... %}
{% endif %}
<a href="{{ aaq_url }}" ...

File: kitsune/products/views.py [hunk 2]
```diff
@@ -18,12 +19,20 @@ def product_list(request):
     return render(request, template, {"products": products})
 
 
+def _get_aaq_product_key(products, slug):
```
LeoMcA, line 22: nit: I'd just take a single slug argument here, and directly reference `aaq_config.products` in the function

kitsune/products/views.py
```diff
@@ -18,12 +19,20 @@ def product_list(request):
...
+    product_key = ""
```
LeoMcA, line 23: nit: IMO `None` would make more sense here

LeoMcA (approved):
One small formatting typo, and a few optional/opinionated nits. Nice!

PR #4867, opened 2021-08-06
elegantly handle 4-byte utf-8 characters in django

LeoMcA:
currently our mysql db is set up to use utf8mb3 which cannot store 4-byte utf-8 characters. it'll truncate the field just before the offending character, which can lead to some pretty major data loss

this change adds a couple of fields which will strip out those characters, or in the case of utf8mb3TextField can be configured to store them as html numeric character references

all end-user modifiable fields have been changed to use these utf8mb3 fields, and in a few carefully tested cases will store 4-byte characters as html numeric references

https://github.com/mozilla/sumo-project/issues/435


escattone:
Closing this since we're moving to Postgres, which will resolve the `utf8mb3` issue we currently have with our MySQL database.

PR #4871, opened 2021-08-12
handle external aaq better in preparation for premium aaq

LeoMcA:
- aaq: move en-US redirect to step 3
- aaq: step 2 ask now link goes to kb for non-premium
- aaq: remove get-community-support kb links from everywhere but step 2
- show aaq banner on all products regardless of locale

https://github.com/mozilla/sumo-project/issues/915

### DIFF
--- a/kitsune/questions/jinja2/questions/includes/aaq_macros.html
+++ b/kitsune/questions/jinja2/questions/includes/aaq_macros.html
@@ -84,23 +84,31 @@
   </ul>
 {%- endmacro %}
 
-{% macro aaq_widget(key, user, location="aaq") %}
-  <div class="aaq-widget card elevation-01 text-center radius-md">
-    <h2 class="card--title has-bottom-margin">{{ _('Ask the Community') }}</h2>
-    {% if user.is_authenticated %}
-      <p>{{ _('Continue to post your question and get help.') }}</p>
-    {% else %}
-      <p>{{ _('Sign in/up to post your question and get help.') }}</p>
-    {% endif %}
-    <a class="sumo-button primary-button feature-box"
-       href="{{ url('questions.aaq_step3', product_key=key) }}"
-       data-event-category="link click"
-       data-event-action="{{ location }}"
-       data-event-label="aaq widget">{{ _('Ask Now') }}</a>
-  </div>
+{% macro aaq_widget(request, location="aaq") %}
+  {% set aaq_context = request.session.get("aaq_context") %}
+  {% if aaq_context %}
+    <div class="aaq-widget card elevation-01 text-center radius-md">
+      <h2 class="card--title has-bottom-margin">{{ _('Ask the Community') }}</h2>
+      {% if request.user.is_authenticated %}
+        <p>{{ _('Continue to post your question and get help.') }}</p>
+      {% else %}
+        <p>{{ _('Sign in/up to post your question and get help.') }}</p>
+      {% endif %}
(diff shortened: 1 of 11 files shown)


kitsune/questions/views.py
```diff
@@ -462,25 +462,6 @@ def aaq(request, product_key=None, category_key=None, step=1):
 
     template = "questions/new_question.html"
 
-    # Check if any product forum has a locale in the user's current locale
```
akatsoulas, line 465: This change is working fine when the user starts the AAQ flow from `/questions/new`. However if you click the `Get Help` link and there isn't an AAQ flow then the KB is shown directly. Is this the intention here?

kitsune/questions/views.py
```diff
@@ -511,21 +492,13 @@ def aaq(request, product_key=None, category_key=None, step=1):
...
-                ).format(product=product.title, old_lang=old_lang, new_lang=new_lang)
-                messages.add_message(request, messages.WARNING, msg)
-
-                return HttpResponseRedirect(path)
+        locale_has_form = product.questions_enabled(locale=request.LANGUAGE_CODE)
+        premium = product.has_subscriptions
+        request.session["aaq-context"] = {
+            "key": product_key,
+            "locale_has_form": locale_has_form,
+            "premium": premium,
```
akatsoulas, line 500: s/premium/has_subscriptions

akatsoulas: Also instead of assigning the var you can directly call the property
 ` "has_subscriptions": product.has_subscriptions`

kitsune/questions/jinja2/questions/includes/aaq_macros.html
```diff
@@ -84,23 +84,31 @@ <h3 class="card--title">
...
-      <p>{{ _('Sign in/up to post your question and get help.') }}</p>
-    {% endif %}
-    <a class="sumo-button primary-button feature-box"
-       href="{{ url('questions.aaq_step3', product_key=key) }}"
-       data-event-category="link click"
-       data-event-action="{{ location }}"
-       data-event-label="aaq widget">{{ _('Ask Now') }}</a>
-  </div>
+{% macro aaq_widget(request, location="aaq") %}
+  {% set context = request.session.get("aaq-context") %}
```
akatsoulas, line 88: lets rename context to aaq_context

kitsune/questions/views.py
```diff
@@ -511,21 +492,13 @@ def aaq(request, product_key=None, category_key=None, step=1):
...
-                    "have redirected you to the {new_lang} questions forum."
-                ).format(product=product.title, old_lang=old_lang, new_lang=new_lang)
-                messages.add_message(request, messages.WARNING, msg)
-
-                return HttpResponseRedirect(path)
+        locale_has_form = product.questions_enabled(locale=request.LANGUAGE_CODE)
+        premium = product.has_subscriptions
+        request.session["aaq-context"] = {
+            "key": product_key,
+            "locale_has_form": locale_has_form,
```
akatsoulas, line 499: s/locale_has_form/has_public_forum to avoid confusion with the zendesk form that we are going to add

akatsoulas (approved):
Looks good! r+wc

kitsune/questions/views.py
LeoMcA: My thinking was this would be how we enable a zendesk form per locale as well - can change the name now, and revise if it turns out that is the logic for zendesk as well though.

PR #4876, opened 2021-08-17
zendesk

LeoMcA:


kitsune/customercare/forms.py
```diff
@@ -0,0 +1,19 @@
+from django import forms
+
+from django.utils.translation import ugettext_lazy as _lazy
+
+TOPIC_CHOICES = [
+    ("", _lazy("Please select one")),
+    ("id1", _lazy("Topic 1")),
```
akatsoulas, line 7: let's use numbers here

kitsune/questions/views.py
```diff
@@ -25,11 +25,15 @@
...
 from django.views.decorators.http import require_GET, require_http_methods, require_POST
 from django_user_agents.utils import get_user_agent
+from sentry_sdk import capture_exception
 from taggit.models import Tag
 from tidings.events import ActivationRequestFailed
 from tidings.models import Watch
+from zenpy.lib.exception import APIException
 
 from kitsune.access.decorators import login_required, permission_required
+from kitsune.customercare.forms import TicketForm as ZendeskForm
```
akatsoulas, line 35: You can either use directly the `TicketForm` here or rename the actual form to ZendeskForm

kitsune/questions/views.py
```diff
@@ -526,6 +531,43 @@ def aaq(request, product_key=None, category_key=None, step=1):
 
             return HttpResponseRedirect(path)
 
+        if has_subscriptions:
+            form = ZendeskForm(data=request.POST or None)
+            context["form"] = form
```
akatsoulas, line 536: let's use something more descriptive here eg `zendesk_form`

kitsune/customercare/zendesk.py
```diff
@@ -0,0 +1,54 @@
...
+
+class ZendeskClient(object):
+    """Client to connect to Zendesk API."""
+
+    def __init__(self, **kwargs):
+        """Initialize Zendesk API client."""
+        creds = {
+            "email": settings.ZENDESK_USER_EMAIL,
+            "token": settings.ZENDESK_API_TOKEN,
+            "subdomain": settings.ZENDESK_SUBDOMAIN.removesuffix(".zendesk.com"),
```
akatsoulas, line 18: let's update the subdomain and drop the `removesuffix`

akatsoulas (commented):
just a first pass. Ping me when the pr is ready for a proper review

kitsune/questions/views.py
```diff
@@ -509,6 +514,7 @@ def aaq(request, product_key=None, category_key=None, step=1):
     if step == 2:
         context["featured"] = get_featured_articles(product, locale=request.LANGUAGE_CODE)
         context["topics"] = topics_for(product, parent=None)
+        context["zendesk_form"] = has_subscriptions
```
akatsoulas, line 517: s/zendesk_form/zendesk_form_exists or even better a name related to the status of the product. Eg produc_has_subscriptions/has_subscriptions

kitsune/customercare/forms.py
```diff
@@ -0,0 +1,75 @@
...
+TICKET_FORM_ID = 360000417171
+
+# See docs/zendesk.md for details about getting the valid choice values for each field:
+
+PRODUCT_FIELD_ID = 360047198211
+PRODUCT_OTHER_VALUE = "product_other"
+
+CATEGORY_FIELD_ID = 360047206172
+CATEGORY_CHOICES = [
+    ("payment___billing", _lazy("Payment & Billing")),
```
akatsoulas, line 15: I have mixed feelings about all of these choices here. The feel disconnected from the rest of the platform.

kitsune/customercare/zendesk.py
```diff
@@ -0,0 +1,59 @@
...
+    def update_user(self, user):
+        """Given a Django user, update a user in Zendesk."""
+        zendesk_user = self._user_to_zendesk_user(user)
+        zendesk_user = self.client.users.update(zendesk_user)
+        # TODO: if we're updating a user's email, zendesk will add it as a secondary identity
+        # we need to call a separate api (not implemented by zenpy) to update it to primary:
+        # https://developer.zendesk.com/api-reference/ticketing/users/user_identities/#make-identity-primary
+        return zendesk_user
+
+    def create_ticket(self, user, form):
```
akatsoulas, line 51: `form` should not be passed to the client but the other way around.  You can call the API either for a `save` like method or event better by adding a `send` method to the form which will call the appropriate API method

kitsune/questions/jinja2/questions/includes/aaq_macros.html
```diff
@@ -103,19 +116,27 @@ <h2 class="card--title has-bottom-margin">{{ _('Ask the Community') }}</h2>
...
+        {% else %}
+          {{ _('Ask Now') }}
+        {% endif %}
+      </a>
     </div>
   {% endif %}
 {%- endmacro %}
 
-{% macro explore_solutions(product, search_box, featured, topics, request) -%}
+{% macro explore_solutions(product, search_box, featured, topics, request, zendesk_form=True) -%}
```
akatsoulas, line 130: instead of passing another flag here, can't we use `product.has_subscriptions` in the `if` statement?

kitsune/questions/jinja2/questions/includes/question_editing_frame.html
```diff
@@ -24,7 +24,9 @@
       {{ progress_bar(current_step, product_key=current_product.key) }}
     {% endif %}
 
-    {{ scam_banner() }}
+    {% if current_step > 1 and not zendesk_form %}
```
akatsoulas, line 27: Similar to the other comment, maybe it makes sense to pass the product object here and extract the info from there

kitsune/questions/jinja2/questions/new_question.html
```diff
@@ -20,15 +20,27 @@
...
+    {% else %}
+      {{ _('Ask your question') }}
+    {% endif %}
+  </h2>
 {% endblock %}
 
-{% block submit_button_value %}{{ _('Post Question') }}{% endblock %}
+{% block submit_button_value %}
+  {% if zendesk_form %}
+    {{ _('Send') }}
```
akatsoulas, line 41: Can't we keep the `Post Question` for both options?

kitsune/questions/views.py
```diff
@@ -526,6 +532,38 @@ def aaq(request, product_key=None, category_key=None, step=1):
 
             return HttpResponseRedirect(path)
 
+        if has_subscriptions:
+            zendesk_form = TicketForm(data=request.POST or None, product=product)
+            context["form"] = zendesk_form
+            context["zendesk_form"] = True
```
akatsoulas, line 538: These three lines are very confusing :)

akatsoulas (commented):
Just did a first pass. I will thoroughly review once the PR is updated

kitsune/products/models.py
```diff
@@ -13,6 +13,7 @@
 class Product(ModelBase):
     title = models.CharField(max_length=255, db_index=True)
     codename = models.CharField(max_length=255, blank=True, default="")
+    zendesk_id = models.CharField(max_length=255, blank=True, default="")
```
akatsoulas, line 16: let's drop this for now. We can use the slug of the product here

kitsune/customercare/forms.py
```diff
@@ -0,0 +1,75 @@
+from django import forms
+from django.utils.translation import ugettext_lazy as _lazy
+
+from zenpy.lib.api_objects import Ticket
+
+TICKET_FORM_ID = 360000417171
+
+# See docs/zendesk.md for details about getting the valid choice values for each field:
+
+PRODUCT_FIELD_ID = 360047198211
```
akatsoulas, line 10: This refers to VPN which has now the slug that Kitsune uses as value in zendesk. Can we remove this and use directly our slug?

kitsune/customercare/signals.py
```diff
@@ -0,0 +1,29 @@
+from django.contrib.auth.models import User
+from django.db.models.signals import post_save
+from django.dispatch import receiver
+from kitsune.customercare.tasks import update_user
+from kitsune.users.models import Profile
+
+
+@receiver(post_save, sender=User, dispatch_uid="customercare.signals.handle_user_save.User")
```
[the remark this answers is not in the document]
LeoMcA: I think we need to listen to User updates in case the email is updated without the rest of the profile

kitsune/questions/jinja2/questions/new_question.html
akatsoulas: Let's update that to `Submit` for both

kitsune/customercare/forms.py
```diff
@@ -0,0 +1,28 @@
...
+
+
+class ZendeskForm(forms.Form):
+    """Form for submitting a ticket to Zendesk."""
+
+    product = forms.CharField(widget=forms.HiddenInput)
+    category = forms.ChoiceField(
+        label=_lazy("What do you need help with?"), choices=CATEGORY_CHOICES
+    )
+    os = forms.ChoiceField(
```
akatsoulas, line 14: Let's make this an optional field and add an empty default choice

kitsune/customercare/forms.py
```diff
@@ -0,0 +1,28 @@
...
+        label=_lazy("What operating system does your device use?"), choices=OS_CHOICES
+    )
+    subject = forms.CharField(label=_lazy("Subject"), required=False)
```
akatsoulas, line 17: shouldn't this be required?

kitsune/customercare/forms.py
```diff
@@ -0,0 +1,28 @@
+from django import forms
+from django.utils.translation import ugettext_lazy as _lazy
+
+from kitsune.customercare.zendesk import ZendeskClient, CATEGORY_CHOICES, OS_CHOICES
+
+
+class ZendeskForm(forms.Form):
+    """Form for submitting a ticket to Zendesk."""
+
+    product = forms.CharField(widget=forms.HiddenInput)
```
akatsoulas, line 10: Product is not validated at all. I managed to change the field in the form and successfully post to Zendesk.

akatsoulas (approved):
r+wc

kitsune/customercare/forms.py
LeoMcA: Oddly enough, zendesk treats subjects as optional:

> Takes a ticket object that specifies the ticket properties. The only required property is `comment`.

from https://developer.zendesk.com/api-reference/ticketing/tickets/tickets/#create-ticket

docs/zendesk.md
```diff
@@ -0,0 +1,42 @@
...
+import requests
+base = f"https://{settings.ZENDESK_SUBDOMAIN}.zendesk.com/api/v2/"
+auth = requests.auth.HTTPBasicAuth(settings.ZENDESK_USER_EMAIL+"/token", settings.ZENDESK_API_TOKEN)
+
+requests.get(base+"foobar", auth=auth).json()
+
+requests.post(base+"barfoo", auth=auth, json={}).json()
+```
+
+### Getting valid choice values for a custom ticket field
```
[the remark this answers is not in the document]
LeoMcA: This is less relevant now we're using the same values across kitsune and zendesk

PR #4877, opened 2021-08-18
act upon FxA profile-change events

LeoMcA:
get and store refresh_tokens for users to enable this

https://github.com/mozilla/sumo-project/issues/934

depends on https://github.com/mozilla/mozilla-django-oidc/pull/432


kitsune/users/models.py
```diff
@@ -101,6 +101,7 @@ class Profile(ModelBase):
     fxa_avatar = models.URLField(max_length=512, blank=True, default="")
     products = models.ManyToManyField(Product, related_name="subscribed_users")
     fxa_password_change = models.DateTimeField(blank=True, null=True)
+    fxa_refresh_token = models.CharField(blank=True, default="", max_length=128)
```
akatsoulas, line 104: why do we need to save this in the model? The OIDC lib has a flag that will store the token if needed in the user's session. Can't we use that?

LeoMcA: the OIDC lib only stores the `access_token`, which expires after a few hours - we need the long-lived `refresh_token`
we also can't access the user's session when processing an incoming FxA event token

akatsoulas:
closing in favor of 4878

PR #4878, opened 2021-09-01
Fxa profile change

akatsoulas:


kitsune/users/auth.py
```diff
@@ -90,6 +96,10 @@ def create_user(self, claims):
             add_to_contributors(user, profile.locale)
             del self.request.session["is_contributor"]
 
+        # If there is a refresh token, store it
+        if self.refresh_token:
```
LeoMcA, line 100: We either need to define this in`__init__` or use `getattr` to avoid the tests failing

LeoMcA (approved):
r+wc - can you squash before merging

PR #4881, opened 2021-09-03
deduplicate update_zendesk_user tasks sent during a single request

LeoMcA:
https://github.com/mozilla/sumo-project/issues/945

### DIFF
--- a/kitsune/customercare/signals.py
+++ b/kitsune/customercare/signals.py
@@ -1,12 +1,14 @@
 from django.contrib.auth.models import User
 from django.db.models.signals import post_save
 from django.dispatch import receiver
-from kitsune.customercare.tasks import update_zendesk_user
+from kitsune.sumo.utils import dedupe_task
 from kitsune.users.models import Profile
 
 
 @receiver(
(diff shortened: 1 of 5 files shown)


akatsoulas (commented):
I have mixed feelings about this functionality. I understand its purpose but it is also adding complexity. For example one task can have different functionality for user and profile related data. Moreover a simpler approach and more DRY approach could be to have two tasks. One that will responsible for the `User` data and one for the `Profile`. Thoughts?

LeoMcA:
> I understand its purpose but it is also adding complexity.

Task de-duplicating is something we ought to do anyway for search indexing, considering the number of hooks we have set up there, and the high likelihood of modifying multiple indexed things in one request, which was done in the ES2 implementation: https://github.com/mozilla/sumo-project/issues/847

> For example one task can have different functionality for user and profile related data.

If a task did do different things based on whether it was passed a user, or a profile, it would need to take that as an argument and therefore wouldn't be de-duplicated.

> Moreover a simpler approach and more DRY approach could be to have two tasks. One that will responsible for the User data and one for the Profile.

This could lead to twice as many Zendesk API requests, and I can't see how it would be any simpler.

akatsoulas:
If relevant let's reopen this against main

PR #4885, opened 2021-09-06
Add method to update primary email

akatsoulas:


kitsune/users/auth.py
```diff
@@ -210,6 +211,13 @@ def update_user(self, user, claims):
             if user_attr_changed:
                 user.save()
             profile.save()
+
+        # If we have an updated email, let's update Zendesk too
+        # the check is repeated for now but it will save a few
+        # API calls if we trigger the task only when we know that we have new emails
+        if user_attr_changed:
+            update_zendesk_identity.delay(user.id, email)
```
LeoMcA, line 219: Though this will work, I'm not a fan of tying zendesk profile updates directly to fxa updates. IMO fxa should trigger a django profile update, and django profile updates should trigger a zendesk update.

kitsune/customercare/zendesk.py
```diff
@@ -74,6 +76,18 @@ def update_user(self, user):
...
         return zendesk_user
 
+    def get_primary_email_identity(self, zendesk_user_id):
+        """Update the primary email of the user."""
+
+        for identity in self.client.users.identities(id=zendesk_user_id):
+            if identity.primary and identity.type == "email":
+                return identity.id
+
+    def update_primary_email(self, zendesk_user_id, identity_id, email):
```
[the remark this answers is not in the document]
LeoMcA: This method also feels very disconnected from the `update_user` method above. Having this functionality, but not including it in the `update_user` method makes me think that method should really be called `update_user_except_for_primary_email`.

Additionally, `update_user` as written *will* update emails, but add them as secondaries. So it's entirely possible - likely even, given the order of when events are fired - that a user with an updated email address will first have their identities set like so, after `update_user` is called:
`[ { primary: true, value: "old@example.com" }, { primary: false, value: "new@example.com" } ]`
and then, after `update_primary_email` is called:
`[ { primary: true, value: "new@example.com" }, { primary: false, value: "new@example.com"} ]`
Zendesk may handle this gracefully and dedupe the emails, but I wouldn't want to count on it.

I think these new methods lend themselves more towards being private methods - with them taking an explicit `zendesk_user_id` and `email` string, rather simply a django user object. `update_user` could then live up to its name, and properly and completely update the user, working out what it needs to call in zendesk to do that.

LeoMcA (commented):
Not even a first pass, but some rough thoughts at the end of my working day.

kitsune/customercare/zendesk.py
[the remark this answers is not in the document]
LeoMcA: Indeed, this is exactly what happens. Updating one's primary email in FxA, then logging in results in:
```
APIException at /fxa/callback/

{"error": "RecordInvalid", "description": "Record validation errors", "details": {"value": [{"description": "Value: lmcardle+1-new-2@mozilla.com is already being used by another user", "error": "DuplicateValue"}]}}
```
Because `update_user` has already added the new email as a secondary.

kitsune/users/auth.py
akatsoulas: We discussed this more and we are going to keep the updates in this method

kitsune/customercare/zendesk.py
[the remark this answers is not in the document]
akatsoulas: Per our discussion we are going to remove email updates from the update_user. Primary email will be updated only through the new method introduced here. TBD about alternate emails

akatsoulas:
@LeoMcA should be good for another pass

kitsune/customercare/zendesk.py
```diff
@@ -47,7 +49,6 @@ def _user_to_zendesk_user(self, user):
         return ZendeskUser(
             id=int(id_str) if id_str else None,
             verified=True,
-            email=user.email,
```
LeoMcA, line 50: This will break the `create_user` method below for existing users, because zendesk matches on email in the `create_or_update` method. We need to do something like:
```python
def _user_to_zendesk_user(self, user, include_email=True):

    ...

    return ZendeskUser(

        ...

        **({ "email": user.email } if include_email else {})
    )

...

def update_user(self, user):

    ...

    zendesk_user = self._user_to_zendesk_user(user, include_email=False)

    ...

```

kitsune/customercare/zendesk.py
```diff
@@ -74,6 +75,19 @@ def update_user(self, user):
         # https://developer.zendesk.com/api-reference/ticketing/users/user_identities/#make-identity-primary
```
LeoMcA, line 75: This block of comment can be removed

kitsune/customercare/zendesk.py
```diff
@@ -74,6 +75,19 @@ def update_user(self, user):
...
         return zendesk_user
 
+    def get_primary_email_identity(self, zendesk_user_id):
+        """Update the primary email of the user."""
```
LeoMcA, line 79: Think this docstring belongs in the method below

kitsune/customercare/zendesk.py
```diff
@@ -74,6 +75,19 @@ def update_user(self, user):
         # https://developer.zendesk.com/api-reference/ticketing/users/user_identities/#make-identity-primary
         return zendesk_user
 
+    def get_primary_email_identity(self, zendesk_user_id):
```
LeoMcA, line 78: nit: I'd make this a private method

kitsune/customercare/zendesk.py
```diff
@@ -74,6 +75,19 @@ def update_user(self, user):
...
+        """Update the primary email of the user."""
+
+        for identity in self.client.users.identities(id=zendesk_user_id):
+            if identity.primary and identity.type == "email":
+                return identity.id
+
+    def update_primary_email(self, zendesk_user_id, email):
```
LeoMcA, line 85: I'd give this the a similar method signature to `update_user` like:
```python
def update_user_primary_email(self, user, email=None):
    zendesk_id = int(user.profile.zendesk_id)
    email = email or user.email
    ...
```

kitsune/customercare/tasks.py
```diff
@@ -1,11 +1,25 @@
...
     user = User.objects.get(pk=user_id)
     if user.profile.zendesk_id:
         zendesk = ZendeskClient()
         zendesk.update_user(user)
+
+
+@task
+def update_zendesk_identity(user_id: int, email: str) -> None:
+    user = User.objects.get(pk=user_id)
+    # email = user.email
```
LeoMcA, line 18: :wave:

kitsune/customercare/tasks.py
```diff
@@ -1,11 +1,25 @@
...
+    zendesk_user_id = user.profile.zendesk_id
+    zendesk = ZendeskClient()
+
+    # fetch identity id
+    identity_id = zendesk.get_primary_email_identity(zendesk_user_id)
+    if identity_id and zendesk_user_id:
+        zendesk.update_primary_email(zendesk_user_id, identity_id, email)
```
LeoMcA, line 25: `update_primary_email` doesn't take a `identity_id` arg any more

LeoMcA (approved):
r+wc

kitsune/customercare/tasks.py
```diff
@@ -1,11 +1,25 @@
...
         zendesk = ZendeskClient()
         zendesk.update_user(user)
+
+
+@task
+def update_zendesk_identity(user_id: int, email: str) -> None:
+    user = User.objects.get(pk=user_id)
+    # email = user.email
+    zendesk_user_id = user.profile.zendesk_id
+    zendesk = ZendeskClient()
```
LeoMcA, line 20: Tests are failing because this is outside the `if user.profile.zendesk_id` conditional

kitsune/customercare/zendesk.py
akatsoulas: This method should be simple. Gets an email and sets it in Zendesk. The control of which email is passed should be handled outside of this method.

PR #4889, opened 2021-09-07
Add subscription info in profiles

akatsoulas:


kitsune/users/jinja2/users/user_subscriptions.html
```diff
@@ -0,0 +1,25 @@
...
+        <img class="card--icon" src="{{ product.image_alternate_url }}" alt="{{ pgettext('DB: products.Product.title', product.title) }}">
+        <div class="card--details">
+          <h3 class="card--title">
+            <a class="title" href="{{ url('products.product', product.slug) }}" data-event-category="link click" data-event-action="product" data-event-label="Firefox">
+              {{ _('<strong>{product}</strong>')|fe(product=pgettext('DB: products.Product.title', product.title)) }}
+            </a>
+          </h3>
+          <a href="{{ settings.FXA_SUBSCRIPTIONS }}">Get Support</a>
+        </div>
+      </div>
```
LeoMcA, line 22: Some thoughts on the cards:
- since the whole card isn't clickable, we should remove the `zoom-on-hover` class - as it implies that the card itself is clickable
- IMO the product title link and the get support link are too close together, easy to click the wrong one
- why does the FXA_SUBSCRIPTIONS link say "Get Support", but instead lead users to somewhere where they can manage their subscription?
- could we instead make clicking on the card take a user to step 2 of the AAQ for that product, and have one overall "manage subscriptions" button which takes users to the FXA_SUBSCRIPTIONS link?

kitsune/users/jinja2/users/user_subscriptions.html
```diff
@@ -0,0 +1,25 @@
...
+    {% for product in products %}
+      <div class="card card--product zoom-on-hover subscriptions">
+        <img class="card--icon" src="{{ product.image_alternate_url }}" alt="{{ pgettext('DB: products.Product.title', product.title) }}">
+        <div class="card--details">
+          <h3 class="card--title">
+            <a class="title" href="{{ url('products.product', product.slug) }}" data-event-category="link click" data-event-action="product" data-event-label="Firefox">
+              {{ _('<strong>{product}</strong>')|fe(product=pgettext('DB: products.Product.title', product.title)) }}
+            </a>
+          </h3>
+          <a href="{{ settings.FXA_SUBSCRIPTIONS }}">Get Support</a>
```
LeoMcA, line 20: Needs to be localisable

kitsune/users/admin.py
```diff
@@ -1,11 +1,22 @@
...
+from kitsune.products.models import Product
 from kitsune.users import monkeypatch
 from kitsune.users.models import AccountEvent, Profile
 
 
 class ProfileAdminForm(forms.ModelForm):
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        # limit the subscriptions choices to the appropriate products
+        associated_proudcts = [product.id for product in self.initial.get("products", [])]
```
LeoMcA, line 14: "associated_proudcts" typo

LeoMcA (approved):
r+wc, left some usability thoughts on the cards themselves

kitsune/users/jinja2/users/user_subscriptions.html
akatsoulas: Regarding the last bullet point let's defer this conversation to the team meeting. For now I will update the string to `Manage Subscriptions` in order to match the functionality that was indicated in the designs. It's an easy change to do later

PR #4892, opened 2021-09-09
Remove 'Contact Us' button

akatsoulas:


kitsune/products/jinja2/products/includes/topic_macros.html
```diff
@@ -63,30 +63,6 @@ <h3 class="card--title">
     </div>
   </div>
 </section>
-
-{% if settings.OIDC_ENABLE and product.id in subscribed_products_ids %}
```
LeoMcA, line 67: We should be able to remove the `subscribed_products_ids` key from the context as well:
https://github.com/mozilla/kitsune/blob/9ea6dbb02be71dad4bce92f8d4c13226fcd3a26d/kitsune/products/views.py#L63

LeoMcA (approved):
r+wc

PR #4898, opened 2021-09-14
add special KB button for opening firefox preferences pages

LeoMcA:
update UITour to add this method

https://github.com/mozilla/sumo-project/issues/950

Enable UITour locally with about:config: https://bedrock.readthedocs.io/en/latest/uitour.html#local-development

Use syntax like the following to add a button: `[[Button:preferences|pane=privacy-locationBar|text=Open Location Bar Privacy Preferences]]`

### DIFF
--- a/kitsune/sumo/static/sumo/js/libs/uitour.js
+++ b/kitsune/sumo/static/sumo/js/libs/uitour.js
@@ -2,272 +2,828 @@
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
+/* eslint valid-jsdoc: ["error", { "requireReturn": false }] */
+
 // create namespace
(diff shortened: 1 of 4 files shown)


kitsune/sumo/static/sumo/js/libs/uitour.js
```diff
@@ -2,272 +2,828 @@
  * License, v. 2.0. If a copy of the MPL was not distributed with this
```
akatsoulas, line 2: Any incompatibilities with what we currently have?

akatsoulas (approved):
r+wc

kitsune/sumo/static/sumo/js/libs/uitour.js
LeoMcA: The only other piece of UITour we're using was `resetFirefox` and that still works fine!

PR #4905, opened 2021-09-21
Zendesk conf

akatsoulas:


k8s/regions/oregon-a/stage.yaml
```diff
@@ -135,3 +135,7 @@ app:
     zendesk_api_token: SECRET
     zendesk_user_email: SECRET
     zendesk_subdomain: SECRET
+    zendesk_category_field_id: 360047206172
+    zendesk_os_field_id: 360018604871
+    zendesk_vpn_field_id: 360047198211
+    zendesk_ticket_form_id: 360000417171
```
LeoMcA, line 141: no newline

k8s/regions/oregon-b/stage.yaml
```diff
@@ -134,3 +134,7 @@ app:
     zendesk_api_token: SECRET
     zendesk_user_email: SECRET
     zendesk_subdomain: SECRET
+    zendesk_category_field_id: 360047206172
+    zendesk_os_field_id: 360018604871
+    zendesk_vpn_field_id: 360047198211
+    zendesk_ticket_form_id: 360000417171
```
LeoMcA, line 140: Only the `ticket_form_id` ought to change between stage/prod, so can we not set the `*_field_id` variables - and pick up the defaults from settings.py - until we definitely need them?

kitsune/customercare/zendesk.py
```diff
@@ -94,11 +87,11 @@ def create_ticket(
         ticket = Ticket(
             subject=subject,
             comment={"body": description},
-            ticket_form_id=TICKET_FORM_ID,
+            ticket_form_id=settings.ZENDESK_TICKET_FORM_ID,
             custom_fields=[
-                {"id": PRODUCT_FIELD_ID, "value": product},
-                {"id": CATEGORY_FIELD_ID, "value": category},
-                {"id": OS_FIELD_ID, "value": os},
+                {"id": settings.ZENDESK_VPN_FIELD_ID, "value": product},
```
LeoMcA, line 92: this should be called `ZENDESK_PRODUCT_FIELD_ID`

kitsune/settings.py
```diff
@@ -1205,3 +1205,8 @@ def filter_exceptions(event, hint):
 ZENDESK_SUBDOMAIN = config("ZENDESK_SUBDOMAIN", default="")
 ZENDESK_API_TOKEN = config("ZENDESK_API_TOKEN", default="")
 ZENDESK_USER_EMAIL = config("ZENDESK_USER_EMAIL", default="")
+# TODO: switch to prod when available
+ZENDESK_VPN_FIELD_ID = config("ZENDESK_VPN_FIELD_ID", default="360047198211", cast=int)
+ZENDESK_OS_FIELD_ID = config("ZENDESK_OS_FIELD_ID", default="360018604871", cast=int)
+ZENDESK_CATEGORY_FIELD_ID = config("ZENDESK_CATEGORY_FIELD_ID", default="360047206172", cast=int)
```
LeoMcA, line 1211: nit: can we swap these so they're in the same order as they appear on the form?

LeoMcA (approved):
r+wc

kitsune/customercare/zendesk.py
akatsoulas: what if we add another product though?

k8s/regions/oregon-b/stage.yaml
akatsoulas: I will go with better explicit that implicit

LeoMcA: With such similar IDs across all the fields (it's hard to glance at them and immediately notice the difference) I think being explicit just makes it more confusing. Could these at least be ordered in the same way as the fields in the form/settings to make comparing them top-to-bottom easier.

kitsune/customercare/zendesk.py
LeoMcA: this is the ID for the product field, not VPN specifically - another product would use the same field, with a different value

PR #4906, opened 2021-09-21
allow pinning articles in the AAQ flow

LeoMcA:
https://github.com/mozilla/sumo-project/issues/912


kitsune/questions/utils.py
```diff
@@ -52,3 +54,35 @@ def get_mobile_product_from_ua(user_agent):
...
+    For pinned articles in WIKI_DEFAULT_LANGUAGE, return a localized version if it exists.
+    For pinned articles in other locales, return them only if `locale` matches.
+    """
+
+    try:
+        pinned_articles = [
+            localized_article
+            for config in product.aaq_configs.all()
+            for article in config.pinned_articles.filter(
+                locale__in=(locale, settings.WIKI_DEFAULT_LANGUAGE)
```
akatsoulas, line 74: Couldn't you squash this double `for` loop with something like the following?
```python
AAQConfig.objects.filter(product=product, pinned_articles__locale=(locales_here)).values_list("pinned_articles")
```
and then add the rest of the logic?

akatsoulas (approved):
r+wc

kitsune/questions/utils.py
LeoMcA: Replaced by just using the first config, as discussed on slack

PR #4923, opened 2021-10-14
move build pipeline to circleci

LeoMcA:
https://github.com/mozilla/sumo-project/issues/910

Opened as a draft, because it's not backwards-compatible with Jenkins (as I'd hoped) because of the changes to `Dockerfile` and `docker-compose.yml`, so we can't merge until we're ready to remove Jenkins (and that requires figuring out the staticfiles -> s3 piece, or an alternative).

Comments welcome, a few steps in the `Dockerfile` might be a bit unusual or convoluted to allow more efficient caching - please point out if you're confused by anything so I know where to add comments.


.circleci/config.yml
```diff
@@ -10,15 +10,19 @@ jobs:
...
             sudo pip install --upgrade pre-commit==2.15.0
             pre-commit run --all-files
+  test-and-build:
+    docker:
+      - image: circleci/python:3.9-buster
+    steps:
+      - checkout
       - setup_remote_docker:
-          version: 17.10.0-ce
+          version: 20.10.7
```
akatsoulas, line 19: That was needed :)
Can we actually bump docker in our current setup - before merging this?

Dockerfile
```diff
@@ -26,83 +16,90 @@ RUN python -m venv /venv
 RUN pip install --upgrade "pip==21.0.1"
 RUN useradd -d /app -M --uid 1000 --shell /usr/sbin/nologin kitsune
 
+RUN apt-get update && apt-get install apt-transport-https && \
```
akatsoulas, line 19: Let's squash these 2 `RUN` commands into one to save ourselves a layer.

Dockerfile
```diff
@@ -26,83 +16,90 @@ RUN python -m venv /venv
...
+#########################
+FROM base AS base-frontend
 
-RUN apt-get update && \
-    apt-get install -y --no-install-recommends  gettext
+COPY ./package*.json /app/
+RUN npm run install-prod && npm run copy:protocol
 
-ENV PATH="/venv/bin:$PATH"
+COPY prepare_django_assets.js .
```
akatsoulas, line 50: Similar here if we can chain a together a few layers it would be great

Dockerfile
```diff
@@ -26,83 +16,90 @@ RUN python -m venv /venv
...
+
+
+##########################
+# Production dependences #
+##########################
+FROM base-frontend AS prod-deps
+
+ARG LOCALE_ENV=master
+ENV LOCALE_ENV=${LOCALE_ENV}
+RUN ./docker/bin/fetch-l10n-files.sh
```
akatsoulas, line 88: same here

LeoMcA:
This should now be feature complete, after implementing s3 push and proper l10n management.

For the latter, I adjusted the approach we used in Jenkins which was:
1. pull mozilla-l10n/sumo-l10n repo
2. lint
3. if lint fails, abort
4. push to mozilla-it/sumo-l10n-prod repo
5. pull from mozilla-it/sumo-l10n-prod repo
6. compile

This approach lacked nuance, and had the effect of losing rather a lot of localisations if just one file failed linting.

The new approach in `l10n-fetch-lint-compile.sh` is something like:
1. pull mozilla-l10n/sumo-l10n repo
2. lint each file individually
3. if lint fails for a file, find what commit the file was last changed in, and checkout the file from that commit
4. lint file again, if it fails goto step 3
5. compile

This way if one locale has a failing file, the rest still keep their most up to date localisations.

Currently all our files pass linting (though msgfmt shows an error for some, which we should fix in the sumo-l10n repo, and then apply that same git backtracking logic to if msgfmt shows any errors as well), but you can test this locally by checking out a known bad commit in the sumo-l10n repo, then running the script:

```bash
cd locale
git checkout 3f359bd6868e3355f46fa81ea00f7edb9455e183
cd ../
./scripts/l10n-fetch-lint-compile.sh
```

You'll see a bunch of `lint returned an error, trying 247e1c5812ef1a749130da96a3e41192912bff48` style messages, and running `git status` on the locale repo will show that the most recent "good" versions of the failing files have been checked out.

You can see a pretend main branch commit here: https://app.circleci.com/pipelines/github/mozilla/kitsune/1561/workflows/85bc7807-9bbd-4864-8f26-159d9e9e60bc/jobs/4283 (with a slightly adjusted config, to avoid uploading static files to the prod s3 bucket).

I think next steps are deploying a docker image from https://hub.docker.com/r/mozilla/kitsune/tags to stage to test the image is good.

docker-compose.ci.yml
```diff
@@ -0,0 +1,19 @@
...
+    build:
+      context: .
+      target: test
+    env_file: .env-test
+    depends_on:
+      - mariadb
+      - elasticsearch7
+      - redis
+
+  prod:
```
akatsoulas, line 13: why is this needed?

docker/bin/entrypoint.sh
```diff
@@ -1,5 +0,0 @@
-#!/bin/bash
-
-source /venv/bin/activate
```
akatsoulas, line 3: isn't this needed any more? Is this replaced by something else?

docker/bin/upload-staticfiles.sh
```diff
@@ -24,11 +23,13 @@ docker/bin/move_hashed_staticfiles.py "${TMP_DIR}" "${TMP_DIR_HASHED}"
 for BUCKET in stage prod; do
     # hashed filenames
     aws s3 sync \
+        --no-progress \
```
akatsoulas, line 26: Doesn't this need `--quiet` and `--only-show-errors` in order to be applied?

kitsune/sumo/jinja2/base.html
```diff
@@ -315,7 +315,7 @@ <h5 class="mzp-c-footer-heading">{{ _('Firefox Accounts') }}</h5>
   </div>
 </footer>
 {% if request.LANGUAGE_CODE %}
-  {% if settings.DEV %}
+  {% if settings.DEV or settings.TEST %}
```
akatsoulas, line 318: you did mention in our chat the reason this is used but could you please go through it again? short memory - sorry :)

akatsoulas (changes requested):
I still need to go through the locales but what's in this review won't be affected by that

.circleci/config.yml
```diff
@@ -28,4 +36,33 @@ jobs:
...
       - run:
           name: Run unit tests
-          command: make test-ci
+          command: ./bin/dc.sh run test ./bin/run-unit-tests.sh
+      - run:
+          name: Run js tests
+          command: ./bin/dc.sh run test ./bin/run-mocha-tests.sh
+      - when:
+          condition:
+            equal: [ main, << pipeline.git.branch >> ]
```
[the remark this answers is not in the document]
LeoMcA: > how about splitting the jobs a bit more (lint, test-js, unittests, build). 

This was my initial approach, however I couldn't immediately figure out how to share the docker context between each job, and so each had to build the image again completely from scratch - which took rather a long time. So I opted for the simpler (faster) approach first, to avoid this PR getting too unwieldily.

But my thought was, once merged, we can explore a bit more of the features Circle has to offer - like "proper" workflows, pretty test result output, and parallel tests.

.circleci/config.yml
```diff
@@ -28,4 +36,33 @@ jobs:
             docker cp ./kitsune/search/dictionaries/synonyms/. project_elasticsearch7_1:/usr/share/elasticsearch/config/synonyms
       - run:
           name: Run unit tests
-          command: make test-ci
+          command: ./bin/dc.sh run test ./bin/run-unit-tests.sh
```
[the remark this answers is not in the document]
LeoMcA: Agreed, see what I said above about exploring more of Circle's features.

Makefile
```diff
@@ -42,44 +37,28 @@ help:
...
 
-.docker-build-pull:
-	${MAKE} pull
-
-.docker-build-full:
-	${MAKE} build-full
-
-build: .docker-build-pull
-	${DC} build node-dev base-dev
+build: .docker-build
```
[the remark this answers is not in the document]
LeoMcA: I think this was the result of me doing a s/.docker-build-pull/.docker-build/ and forgetting to remove this reference. I (think) the rest of the rules make sense, so a `.docker-buld -> make build -> dc build web` happens before running the command.

Dockerfile
```diff
@@ -26,83 +16,84 @@ RUN python -m venv /venv
 RUN pip install --upgrade "pip==21.3.1"
 RUN useradd -d /app -M --uid 1000 --shell /usr/sbin/nologin kitsune
 
-RUN apt-get update && \
+RUN apt-get update && apt-get install apt-transport-https && \
+    curl -sL https://deb.nodesource.com/setup_12.x | bash - && \
```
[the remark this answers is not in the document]
LeoMcA: The old image installed node (this way) - just in a separate step. I moved it earlier to take greater advantage of caching.

The reason for not using a docker node image is - at least currently - we need a python+node environment to build our staticfiles. Once we have webpack out the door, that might change (though I'm not sure exactly where our js locale files would fit in - possibly not a requirement for our javascript tests), and we could/should probably switch to using a node image to build the webpack bundle. I think this will also make splitting up the job into multiple steps in a workflow easier.

Dockerfile
```diff
@@ -26,83 +16,84 @@ RUN python -m venv /venv
...
     gettext build-essential \
     libxml2-dev libxslt1-dev zlib1g-dev git \
     libjpeg-dev libffi-dev libssl-dev libxslt1.1 \
-    libmariadb3 mariadb-client && \
+    libmariadb3 mariadb-client \
+    optipng nodejs zip && \
     rm -rf /var/lib/apt/lists/*
 
 COPY ./requirements/*.txt /app/requirements/
+RUN pip install --no-cache-dir --require-hashes -r requirements/default.txt
```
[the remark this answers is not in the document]
LeoMcA: Since the venv used here is copied into the final prod image, I wanted to avoid installing the dev requirements. The dev image below also isn't ever built in CI, because we have no need for it.

Dockerfile
```diff
@@ -26,83 +16,84 @@ RUN python -m venv /venv
...
-FROM base AS base-dev
-RUN apt-get update && apt-get install apt-transport-https && \
-    curl -sL https://deb.nodesource.com/setup_12.x | bash -
-RUN apt-get update && apt-get install -y --no-install-recommends optipng nodejs zip && \
-    rm -rf /var/lib/apt/lists/*
 
+#########################
+# Frontend dependencies #
+#########################
+FROM base AS base-frontend
```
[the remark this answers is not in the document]
LeoMcA: Ah, right, actually it looks like webpack won't necessary solve our need for a python+node environment because of the need to precompile the nunjucks templates. I'm not sure that necessarily needs to be built with the rest of the javascript though, let me experiment with using a "pure" node image for the rest.

docker-compose.ci.yml
LeoMcA: To allow us to do `docker-compose build prod` and end up with an appropriately tagged image

docker/bin/entrypoint.sh
LeoMcA: I don't think we were using it anywhere!

docker/bin/upload-staticfiles.sh
LeoMcA: I think I may have missed that copying+pasting when being very careful to not accidentally deploy to the prod bucket when testing, I'll recheck.

kitsune/sumo/jinja2/base.html
LeoMcA: Sure: it's to avoid having to minify the django l10n files before running the unit tests in CI - it saves us a few seconds on each run.

Dockerfile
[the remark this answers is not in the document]
LeoMcA: So the `nunjucks_precompile` python script calls the node `nunjucks-precompile` command, which means we do need a node+python environment.

We could, with relative ease, port this script to node, but if we're porting it to anything we might as well port it to a webpack loader (so we can simply `import "./some/nunjucks/template.html"` wherever we need it).

So my strong inclination is to leave things like this for now, then after/as we merge webpack revisit this.

Dockerfile
```diff
@@ -1,16 +1,6 @@
...
-COPY ./kitsune/sumo/static/sumo /app/kitsune/sumo/static/sumo
-RUN npm run production
-
-################################
-# Python dependencies builder
-#
+#######################
+# Common dependencies #
+#######################
 FROM python:3.9-buster AS base
```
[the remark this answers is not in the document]
LeoMcA: It allows building the image on a slower connection, as explained in the commit here:

https://github.com/mozilla/kitsune/commit/91645ed37e337b7c22f59df98d6f180bef3dd32c

I accidentally removed the mechanism to change this arg in the makefile, so I'll add it back.

I wasn't sure that `ARG` values were actually turned into environment variables, but that seems to be the case:

```
ARG PIP_DEFAULT_TIMEOUT=foobar
RUN echo ${PIP_DEFAULT_TIMEOUT}
```
```
Step 4/17 : ARG PIP_DEFAULT_TIMEOUT=foobar
 ---> Running in 8a1ccf651c56
 ---> f2eb3d5c0e63
Step 5/17 : RUN echo ${PIP_DEFAULT_TIMEOUT}
 ---> Running in 64218dd967b9
foobar
```

docker/bin/upload-staticfiles.sh
LeoMcA: Ah, right, `--no-progress` does work alone ("This flag is only applied when the quiet and only-show-errors flags are *not* provided.") - however what it disables is the "dynamic" progress bar which horrifically messes with circle's console output, rather than listing each file which got uploaded.

I think that's still rather a lot of unnecessary output, so I'll change the command to `--only-show-errors`.

.circleci/config.yml
```diff
@@ -28,4 +36,33 @@ jobs:
...
+          command: ./bin/dc.sh run test ./bin/run-mocha-tests.sh
+      - when:
+          condition:
+            equal: [ main, << pipeline.git.branch >> ]
+          steps:
+            - run:
+                name: Build prod image
+                command: DOCKER_BUILDKIT=1 ./bin/dc.sh build --progress=plain prod
+            - run:
+                name: Push prod image
```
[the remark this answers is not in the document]
LeoMcA: I'm not quite sure what you mean here

akatsoulas:
@smarnach could you have a look at this PR? It's mostly done but it would be nice if we could get a second pair of eyes before merging.

.circleci/config.yml
```diff
@@ -10,22 +12,59 @@ jobs:
           command: |
             sudo pip install --upgrade pre-commit==2.15.0
             pre-commit run --all-files
+  test-and-build:
+    docker:
+      - image: circleci/python:3.9-buster
```
smarnach, line 17: Maybe use the `3.9-bullseye` tag, which seems closer to what's used inside the Docker image? Probably doesn't make much of a difference.

smarnach (approved):
The changes here look good from an ops perspective.

It's not quite clear to me how the condition for pushing to Docker Hub works. If I look at the build for the last commit in this PR, it looks like it pushed a new tag to Docker Hub. If I look at the condition, I think it shouldn't have. What's the idea here?

akatsoulas:
> The changes here look good from an ops perspective.
> 
> It's not quite clear to me how the condition for pushing to Docker Hub works. If I look at the build for the last commit in this PR, it looks like it pushed a new tag to Docker Hub. If I look at the condition, I think it shouldn't have. What's the idea here?

The idea here is to push to docker-hub always from the `main` branch but also have the ability to push an image in different branches. In Jenkins there is an image pushed to every branch. In circle-ci we agreed that everything that starts with `prod-` should trigger an image push.

smarnach:
> The idea here is to push to docker-hub always from the `main` branch but also have the ability to push an image in different branches. In Jenkins there is an image pushed to every branch. In circle-ci we agreed that everything that starts with `prod-` should trigger an image push.

That's how I read the condition as well. I still wonder why the last build for this PR pushed an image, since this branch is called `new-build-pipeline`. Shouldn't this only push an image once merged?

akatsoulas:
I believe that condition was added as part of the review process which also was the latest commit that was built from both the `new-build-pipeline` and the `prod-new-build-pipeline` branches. The image hash though in the docker repo seems to be coming from the prod-new-build-pipeline branch. That said, I believe it's worth double checking this by triggering one more build in the current branch and make sure that it won't push to docker hub.

scripts/l10n-fetch-lint-compile.sh
```diff
@@ -0,0 +1,50 @@
...
+
+git clone https://github.com/mozilla-l10n/sumo-l10n.git locale
+cd locale
+
+postatus_file=../kitsune/sumo/static/postatus.txt
+head_hash=$(git log -n1 --format=%H)
+echo "po status for commit ${head_hash}:" > $postatus_file
+echo -e "(if nothing appears below, no errors were detected)\\n" >> $postatus_file
+
+function lintpo() {
```
akatsoulas, line 13: What happens if subsequent commits have errors?

akatsoulas (approved):
There are a few comments that require changes but overall this is looking good. Nice work!
I am more than happy to have another pass if you would like for the new changes. r+wc

docker/bin/entrypoint.sh
LeoMcA: I'm pretty sure that's coming from inside the mariadb image (especially since it's called `docker-entrypoint.sh` not just `entrypoint.sh`)

docker/bin/upload-staticfiles.sh
```diff
@@ -21,18 +20,21 @@ docker rm -f "${CONTAINER_NAME}"
 # separate the hashed files into another directory
 docker/bin/move_hashed_staticfiles.py "${TMP_DIR}" "${TMP_DIR_HASHED}"
 
-for BUCKET in stage prod; do
+# for BUCKET in stage prod; do
```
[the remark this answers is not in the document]
LeoMcA: I think I'd be more comfortable leaving this here until we've tested it on stage (which I'll do after this next commit).

Dockerfile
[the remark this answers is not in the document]
LeoMcA: https://github.com/mozilla/sumo-project/issues/973

scripts/l10n-fetch-lint-compile.sh
LeoMcA: The script will keep backtracking until it finds a file which passes linting - the beauty of doing this on a git repo is that since all the files right now pass linting, we know that there'll always be a file we can (eventually) get back to which passes. That may take rather a while in the worst case, but if we notice there's a locale which is failing for a few builds in a row we can always submit a PR to fix the error.

LeoMcA:
> The image hash though in the docker repo seems to be coming from the prod-new-build-pipeline branch. That said, I believe it's worth double checking this by triggering one more build in the current branch and make sure that it won't push to docker hub.

Yes I think this is a quirk of the circleci/github integration - since I've pushed the same commits to both the new-build-pipeline and prod-new-build-pipeline branches, it's linking to the latter from this commit (I guess because I pushed there first?)

The latest commit is only on this branch, so we should see this stopping after tests: https://circleci.com/gh/mozilla/kitsune/4296

I've deployed https://hub.docker.com/layers/mozilla/kitsune/prod-8b6e26/images/sha256-eec231404190c0018f85e3e85739c2ed76889c52489d4d00ba388c5a01346985?context=explore to stage and everything seems to be working fine!

PR #4934, opened 2021-11-05
Verify access token every hour.

akatsoulas:


kitsune/sumo/middleware.py
```diff
@@ -70,6 +73,48 @@ def process_request(self, request):
...
+            response_token_info = (
+                requests.post(settings.FXA_VERIFY_URL, data={"token": access_token})
+            ).json()
+
+            # if the token is not verified, log the user out
+            if (
+                response_token_info.get("code") == 400
+                and response_token_info.get("message") == "Invalid token"
+            ):
+                logout(request)
```
akatsoulas, line 115: This needs an else clause which is going to reset the counter or retrieve a new access token

akatsoulas:
@LeoMcA this should be good for a rewview

kitsune/sumo/middleware.py
```diff
@@ -70,6 +73,52 @@ def process_request(self, request):
...
+
+            response_token_info = (
+                requests.post(settings.FXA_VERIFY_URL, data={"token": access_token})
+            ).json()
+
+            # if the token is not verified, log the user out
+            if (
+                response_token_info.get("code") == 400
+                and response_token_info.get("message") == "Invalid token"
+            ):
```
akatsoulas, line 114: I guess here we need to also remove the `refresh_token` and double check that any tasks that depend on it first check for its existence

kitsune/settings.py
```diff
@@ -602,12 +604,17 @@ def _username_algo(email):
...
         FXA_LOGOUT_REDIRECT_URL = config("FXA_LOGOUT_REDIRECT_URL", "/")
         FXA_USERNAME_ALGO = config("FXA_USERNAME_ALGO", default=_username_algo)
-        FXA_STORE_ACCESS_TOKEN = config("FXA_STORE_ACCESS_TOKEN", default=False, cast=bool)
+        FXA_STORE_ACCESS_TOKEN = config("FXA_STORE_ACCESS_TOKEN", default=True, cast=bool)
         FXA_STORE_ID_TOKEN = config("FXA_STORE_ID_TOKEN", default=False, cast=bool)
         FXA_SUBSCRIPTIONS = config(
             "FXA_SUBSCRIPTIONS", default="https://accounts.firefox.com/subscriptions"
         )
         FXA_SET_ISSUER = config("FXA_SET_ISSUER", default="https://accounts.firefox.com")
+        FXA_VERIFY_URL = config("FXA_VERIFY_URL", default="")
```
LeoMcA, line 613: Should this have a default value?

kitsune/sumo/middleware.py
```diff
@@ -70,6 +73,55 @@ def process_request(self, request):
...
+        if expiration < now:
+
+            response_token_info = (
+                requests.post(settings.FXA_VERIFY_URL, data={"token": access_token})
+            ).json()
+
+            # if the token is not verified, log the user out
+            if (
+                response_token_info.get("code") == 400
+                and response_token_info.get("message") == "Invalid token"
```
LeoMcA, line 113: We also need to catch the case where a user has no access_token in their session (which will happen once for every session when we deploy this code). The API returns a 400 with "Invalid request parameter", but we probably ought to catch it sooner when `access_token == None`

kitsune/sumo/middleware.py
```diff
@@ -70,6 +73,55 @@ def process_request(self, request):
...
+            ):
+                profile = request.user.profile
+                profile.fxa_refresh_token = ""
+                profile.save()
+                logout(request)
+            else:
+                request.session["oidc_id_token_expiration"] = (
+                    now + settings.FXA_RENEW_ID_TOKEN_EXPIRY_SECONDS
+                )
```
LeoMcA, line 122: We should add a comment here clarifying that, as you explained to me, we don't need to refresh the access_token since the verify endpoint will return fine, even for expired (not invalidated) tokens.

LeoMcA (approved):
r+wc

kitsune/settings.py
akatsoulas: I will add prod to match the rest of the vars.

web-apply:
thanks

PR #4943, opened 2021-11-17
firefox focus beta banners

LeoMcA:
can be shown on articles, and topic pages, with the easy possibility of expanding to more places

deduplicated ff-alert and scam-alert css and merged into sumo-cta-banner class


kitsune/products/models.py
```diff
@@ -188,3 +188,71 @@ class Platform(ModelBase):
 
     def __str__(self):
         return "%s" % self.name
+
+
+class Banner(ModelBase):
```
akatsoulas, line 193: docstring

kitsune/products/models.py
```diff
@@ -188,3 +188,71 @@ class Platform(ModelBase):
...
+        max_length=settings.MAX_FILEPATH_LENGTH,
+        help_text="Must be 72x72px.",
+    )
+    static_image = models.CharField(
+        max_length=255,
+        blank=True,
+        help_text="Path to an image in staticfiles, e.g. 'protocol/img/logos/firefox/logo.svg'. "
+        "Takes precedence over uploaded_image, leave both blank to show no image.",
+    )
+    description = models.TextField()
```
akatsoulas, line 207: Let's make this a CharField. I highly doubt that we will need more than that for a banner. Probably we need to use a `max_length` property too to limit this in something suitable for a banner

kitsune/products/models.py
```diff
@@ -188,3 +188,71 @@ class Platform(ModelBase):
...
+    link = models.CharField(max_length=255, blank=True)
+    button = models.CharField(
```
akatsoulas, line 209: s/button/button_text

kitsune/products/models.py
```diff
@@ -188,3 +188,71 @@ class Platform(ModelBase):
...
+        "Leave blank to associate with all products.",
+    )
+    topics = models.ManyToManyField(
+        Topic,
+        blank=True,
+        help_text="Topics to associate the banner with. Leave blank to associate with all topics.",
+    )
+
+    visible = models.BooleanField()
+    priority = models.IntegerField(
```
akatsoulas, line 233: I would drop this completely. In order to assign a higher priority to one banner over another you still need to locate it and it also adds more logic. Let's just drop it completely

kitsune/products/models.py
```diff
@@ -188,3 +188,71 @@ class Platform(ModelBase):
...
+        blank=True,
+        help_text="Text to show on button, leave blank to show no button.",
+    )
+    modifer_class = models.CharField(
+        max_length=255,
+        blank=True,
+        help_text="Additional classes to add to the banner, space seperated.",
+    )
+
+    products = models.ManyToManyField(
```
akatsoulas, line 220: Given the purpose of this functionality and given that the existing UI space is limited, I doubt that there will be a real world scenario where a 1:N relationship is not adequate. A banner can belong to many products but is there a case for the reverse? Why should a product have multiple banners?

kitsune/products/models.py
```diff
@@ -188,3 +188,71 @@ class Platform(ModelBase):
...
+        Product,
+        blank=True,
+        help_text="Products to associate the banner with. "
+        "Leave blank to associate with all products.",
+    )
+    topics = models.ManyToManyField(
```
akatsoulas, line 226: Similar comment to the product m2m

kitsune/products/models.py
```diff
@@ -188,3 +188,71 @@ class Platform(ModelBase):
...
+    show_to_authenticated = models.BooleanField(
+        default=True,
+        verbose_name="Show to logged in users",
+    )
+    show_to_unauthenticated = models.BooleanField(
+        default=True,
+        verbose_name="Show to logged out users",
+    )
+
+    show_on_topic = models.BooleanField(
```
akatsoulas, line 246: Isn't this redundant? I would assume that if a topic entry exists, it means that it will be shown under the said topic.

kitsune/products/models.py
```diff
@@ -188,3 +188,71 @@ class Platform(ModelBase):
...
+    products = models.ManyToManyField(
+        Product,
+        blank=True,
+        help_text="Products to associate the banner with. "
+        "Leave blank to associate with all products.",
+    )
+    topics = models.ManyToManyField(
+        Topic,
+        blank=True,
+        help_text="Topics to associate the banner with. Leave blank to associate with all topics.",
```
akatsoulas, line 229: Let's be explicit. No topic means that the banner is not displayed.

kitsune/products/models.py
```diff
@@ -188,3 +188,71 @@ class Platform(ModelBase):
...
+        max_length=255,
+        blank=True,
+        help_text="Additional classes to add to the banner, space seperated.",
+    )
+
+    products = models.ManyToManyField(
+        Product,
+        blank=True,
+        help_text="Products to associate the banner with. "
+        "Leave blank to associate with all products.",
```
akatsoulas, line 224: Let's be explicit, no product means the banner is not displayed.

kitsune/products/models.py
```diff
@@ -188,3 +188,71 @@ class Platform(ModelBase):
...
+        default=0, help_text="Higher priority banners will show instead of lower priority ones."
+    )
+
+    show_to_authenticated = models.BooleanField(
+        default=True,
+        verbose_name="Show to logged in users",
+    )
+    show_to_unauthenticated = models.BooleanField(
+        default=True,
+        verbose_name="Show to logged out users",
```
akatsoulas, line 243: s/logged out/non authenticated

kitsune/products/models.py
```diff
@@ -188,3 +188,71 @@ class Platform(ModelBase):
...
+    )
+
+    show_on_topic = models.BooleanField(
+        verbose_name="Show on article listing (topic) pages",
+        help_text="Will only show if the banner is associated "
+        "with the topic and product of the topic page.",
+    )
+    show_on_article = models.BooleanField(
+        verbose_name="Show on articles",
+        help_text="Will only show if the banner is associated "
```
akatsoulas, line 253: the description is a bit confusing. I guess this flag toggles whether it will be displayed in the articles under the specified product and or topics if set?

kitsune/products/models.py
akatsoulas: I would also make this a required field.

kitsune/products/models.py
akatsoulas: Functionality wise I would approach it as follows:

* If a product is set -> Display under that product only.
* If a topic is set -> Display under the product and the set of selected topics
* If the flag is set to display in KB -> Display in all articles under the product if not topic is set, otherwise display to all articles under the topic set

kitsune/products/utils.py
```diff
@@ -0,0 +1,21 @@
+from django.db.models import Q
+
+from kitsune.products.models import Banner
+
+
+def cta_banner(products=[], topics=[], page_type="", request=None):
```
akatsoulas, line 6: I would make `request` the first arg. This is not a view but it helps with consistency

kitsune/products/utils.py
```diff
@@ -0,0 +1,21 @@
...
+    query = Banner.objects
+
+    if products:
```
akatsoulas, line 9: I suspect that the following queries will change after the changes in the schema so I will go through them after the update.

kitsune/settings.py
```diff
@@ -881,6 +884,9 @@ def JINJA_CONFIG():
 # Products
 PRODUCT_IMAGE_PATH = "uploads/products/"
 
+# Banners
+BANNER_IMAGE_PATH = "uploads/banners/"
```
akatsoulas, line 888: Can't we use the product path?

kitsune/products/views.py
```diff
@@ -82,6 +83,8 @@ def document_listing(request, product_slug, topic_slug, subtopic_slug=None):
 
     documents, fallback_documents = documents_for(**doc_kw)
 
+    banner = cta_banner(products=[product], topics=[topic], page_type="topic", request=request)
```
akatsoulas, line 86: nit: I would keep the cta_banner in the context as it's easier to understand what kind of a banner this is

kitsune/sumo/static/sumo/scss/components/_banner.scss
```diff
@@ -0,0 +1,110 @@
+@use '../config' as c;
+@use '../protocol' as p;
+
+// Banners
+//
```
akatsoulas, line 5: how about naming this `_cta_banner.scss` in case we have another in the future?

kitsune/wiki/jinja2/wiki/document.html
```diff
@@ -73,9 +74,8 @@
       </div>
     </header>
 
-    {% if show_cta_banner %}
-      {{ sumo_cta_banner(text=_('Unlock more from the Firefox suite: Sync, Monitor, Lockwise and <a href="https://www.mozilla.org/firefox/accounts/">more.</a>'),
-                         modifier_class="hide-on-mobile") }}
+    {% if banner %}
```
akatsoulas, line 77: similar to the comment I made in the view, let's rename this to cta_banner or something more descriptive

akatsoulas (changes requested):
I had a first pass, mainly targeting the schema of this. I will go through a more thorough review after the changes.

kitsune/products/models.py
LeoMcA: If two banners would be shown in the same place, how do we decide which to show?

kitsune/products/models.py
LeoMcA: No, because you may want to show the banner on KB articles themselves, or just topic pages, or both.

kitsune/products/models.py
LeoMcA: This toggles whether it appears in the article itself (a la the existing FxA banner), the `show_on_article` bool toggles if it appears on a topic page (e.g. this page: https://support.mozilla.org/en-US/products/firefox/privacy-and-security). You may want it to display on either, or both - and this system of logic allows adding other `show_on_` conditionals if we need them (e.g. `show_on_landing`, `show_on_aaq`, etc.)

kitsune/sumo/static/sumo/scss/components/_banner.scss
LeoMcA: I wrote the CSS to be as flexible as possible - it'll function without a CTA button, without a close button, and/or without an image - so it's a generic banner, even though we're using it as a CTA banner in this instance.

kitsune/products/models.py
LeoMcA: This is prone to be quite variable in length, given HTML is supported - I think a TextField is still more appropriate.

akatsoulas: Do you foresee anything that will require more than 65535 characters? I would also suggest to add a smaller max_length in the charifeild. We can start with 1024 chars or something. Charfield will help also with the validation if we ever need it

kitsune/products/models.py
akatsoulas: Let's start with something simpler and if we ever need a second banner we can revisit

kitsune/settings.py
LeoMcA: I think it ought to be different, like the topic image path is.

akatsoulas (changes requested):
While I was reviewing this I was thinking that we should probably follow the same paradigm with the AAQ and introduce a ProductConfig. Every product can have a single config entry and we can change those configurations easily through the admin interface. Taking this path means that a CTA banner is part of that config. For simplicity, we can start with a single banner per config

akatsoulas:
Discussing this PR more through slack, we decided to remove all of the DB related parts. The current schema is limiting us from having a product with multiple configurations without investing a big amount of time which is well beyond the scope of this PR.

LeoMcA:
Updated to remove all the db elements, the previous code is available here if we ever want to reference it: https://github.com/LeoMcA/kitsune/tree/db-backed-cta-banner

Here's how the banners look using the preliminary text from Abby:

Screenshot 2021-11-24 at 15-56-35 Firefox Focus for Android Firefox Focus Help - 127 0 0 1
Screenshot 2021-11-24 at 15-56-08 Firefox Focus for iOS Firefox Focus Help - 127 0 0 1

We'll need to add the proper link to the wiki article in the Android banner once it's added to the KB.

mimi8321 (commented):
****

kitsune/products/jinja2/products/documents.html
```diff
@@ -70,6 +70,18 @@ <h1 class="topic-title sumo-page-heading">
         </div>
       </header>
 
+      {% if topic.slug == "Focus-ios" %}
+        {{ sumo_banner(
+          id="focus-ios-beta-banner",
```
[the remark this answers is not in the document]
LeoMcA: Manually - our quest for an effective jinja formatter goes on!

PR #4944, opened 2021-11-17
Question queries

akatsoulas:


kitsune/questions/views.py
```diff
@@ -187,9 +187,15 @@ def question_list(request, product_slug):
             question_qs = question_qs.done()
 
     question_qs = question_qs.select_related("creator", "last_answer", "last_answer__creator")
-    question_qs = question_qs.prefetch_related("topic", "topic__product")
+    # Exclude questions over 90 days old without an answer or
+    # older than 5 years or
+    # created by deactivated users
+    today = date.today()
+    question_qs = question_qs.exclude(
+        created__lt=today - timedelta(days=90), num_answers=0, creator__is_active=False
```
LeoMcA, line 195: The `creator__is_active=False` exclude needs to be put in a separate `exclude()`, otherwise we're only excluding questions over 90 days old without an answer AND created by deactivated users.

And rather than doing `.exclude(creator__is_active=False)` I think it'd be preferable to avoid the double negative with `.filter(creator__is_active=True)`.

kitsune/questions/views.py
```diff
@@ -187,9 +187,15 @@ def question_list(request, product_slug):
...
+    ).filter(updated__gt=today - timedelta(days=365 * 5))
```
LeoMcA, line 196: I'd be inclined to leave this at 1 year, for now, or a lesser increase like 2 years - we can always bump it up further in future if performance is really good.

LeoMcA (approved):
r+wc

I went a little bit down the rabbit hole while reviewing this and have a few more optimisations to this query through greater use of select/prefetch_related, and some changes to make sure the template actually ends up using those select/prefetches, as well as - ironically - removing some redis caching which speeds things up. Do you want to wait for those, or shall I submit them as a separate PR?

akatsoulas:
Please feel free to open a new PR with the optimizations. I am merging this so that we can release it asap

PR #4964, opened 2021-12-16
cache webpack_static helper to reduce json loading/parsing

LeoMcA:
https://github.com/mozilla/sumo-project/issues/874

### DIFF
--- a/kitsune/sumo/utils.py
+++ b/kitsune/sumo/utils.py
@@ -3,6 +3,7 @@ import re
 import sys
 from contextlib import contextmanager
 from datetime import datetime
+from functools import lru_cache
 
 from django.conf import settings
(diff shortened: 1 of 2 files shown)


kitsune/sumo/utils.py
```diff
@@ -353,6 +354,7 @@ def check_for_spam_content(data):
     return is_toll_free or is_nanp_number or has_links
 
 
+@lru_cache(maxsize=0 if settings.DEV else 128)
```
akatsoulas, line 357: shall we move this logic to settings?

```python
WEBPACK_LRU_CACHE=128

if DEV:
    WEBPACK_LRU_CACHE=0
```

akatsoulas (approved):
r+wc

PR #4965, opened 2021-12-16
Webpack imports

LeoMcA:
Highly recommend reviewing commit-by-commit, I deliberately made things atomic.

Also highly recommend ignoring whitespace changes (append `?w=1` to your github diff view url, or select the option in the dropdown) as there's a number of big indentation changes when removing the IIFEs to allow exporting of functions.

On the issue of IIFEss I had multiple changes of heart over the course of this PR, so some files have it totally removed to allow exporting functions, whereas in others I just move the functions I need to export outside to ensure a minimal git diff. We have absolutely no need for them anymore, so I ought to go back and totally remove them in the files where I just moved functions around a little. However - would you prefer one big commit at the start of the PR removing all the IIFEs (at least in the files being changed here), or to remove them as and when I need to in order to export functions (which is what I've done so far)?

Tests are failing, I've got a commit fixing that which I need to clean up, which I'll append once I'm done. It might also require editing a few of these commits in an interactive rebase, but it should be minor changes.


akatsoulas (approved):
r+ with CircleCI happy. Really nice work!

PR #4971, opened 2021-12-22
Update K8s README and EKS related details

rvaralda:


k8s/README.md
```diff
@@ -48,54 +48,33 @@ The production branch maintains an image that can be released in production with
     -   ./regions/oregon-b/stage-secrets.yaml
     -   ./regions/oregon-b/prod-secrets.yaml
 
-#### Configuration for deploying into a Kops cluster
-
--   Create symbolic links to `kubectl` and `kubeconfig` to use in each region. Note that `kubectl` should match the version of the cluster, or have a version skew of 1 minor version.
```
akatsoulas, line 53: The mention of `kubectl` matching the version of the cluster or having a skew of 1 minor version should probably stay.

k8s/README.md
```diff
@@ -48,54 +48,33 @@ The production branch maintains an image that can be released in production with
     -   ./regions/oregon-b/stage-secrets.yaml
     -   ./regions/oregon-b/prod-secrets.yaml
```
akatsoulas, line 49: The kops cluster had a specific way of handling secrets. Will this be preserved in `eks`?

akatsoulas: Probably this section needs updating too

akatsoulas (changes requested):

k8s/README.md
akatsoulas: Not necessarily in this section. Probably somewhere in the beginning of the file would make sense.

k8s/commander.sh
```diff
@@ -84,7 +102,7 @@ function compare-client-server-versions {
 	fi
 }
 
-if ![ -f "$PYENV_FILE" ]; then
```
akatsoulas, line 87: why this change?

akatsoulas (changes requested):
Can you add a comment about the related to EKS changes that you are making?

akatsoulas:
This PR is now in the `commander-eks` branch. Any updates should target that branch.

PR #4989, opened 2022-01-10
Refresh access_token every 12 hours.

akatsoulas:


kitsune/settings.py
```diff
@@ -620,7 +620,7 @@ def _username_algo(email):
         )
         # Defaults to an hour
```
LeoMcA, line 621: "Defaults to 12 hours"

kitsune/users/auth.py
```diff
@@ -51,6 +51,30 @@ def get_token(self, payload):
         self.refresh_token = token_info.get("refresh_token")
         return token_info
 
+    @classmethod
+    def refresh_access_token(cls, refresh_token):
```
LeoMcA, line 55: probably want a `ttl` argument here, currently we just request a 5-minute access token (so it expires shortly after the profile update task completes), but in other scenarios we may want a longer-lived token

kitsune/users/auth.py
```diff
@@ -51,6 +51,30 @@ def get_token(self, payload):
...
+        """Gets a new access_token by using a refresh_token.
+
+        returns: the actual token or None
```
LeoMcA, line 58: should this be "returns: contents of the token_info endpoint, or an empty dict"?

kitsune/users/tasks.py
```diff
@@ -66,22 +68,14 @@ def process_event_profile_change(event_id):
     event = AccountEvent.objects.get(id=event_id)
     refresh_token = event.profile.fxa_refresh_token
 
-    if not refresh_token:
+    fxa = FXAAuthBackend()
+    token_info = fxa.refresh_access_token(refresh_token)
```
LeoMcA, line 72: since refresh_access_token is a classmethod, would `token_info = FXAAuthBackend.refresh_access_token(refresh_token)` be better?

kitsune/sumo/middleware.py
```diff
@@ -103,29 +92,21 @@ def process_request(self, request):
         expiration = request.session.get("oidc_id_token_expiration", 0)
         now = time.time()
         access_token = request.session.get("oidc_access_token")
```
LeoMcA, line 94: We don't actually do anything with the access_token, so I wonder if there's any point in storing this in the session?
Similarly the `oidc_id_token_expiration` value is something we're setting ourselves.

So I wonder, could we just use something like a `session["fxa_session_expiry"]` value, and bump that value by `FXA_RENEW_SESSION_EXPIRY_SECONDS` when using the refresh_token works (but then immediately discarding the access_token).

Though, if we wanted to (now or in the future) hit the verify endpoint with the access_token, before attempting a refresh when that fails, there's value in keeping the access_token in the session. In that case - we probably don't want to send a `ttl` argument in our refresh request, to allow FxA to dictate how often we hit the more expensive refresh route.

kitsune/sumo/middleware.py
```diff
@@ -103,29 +92,21 @@ def process_request(self, request):
...
-                profile.save()
-                logout(request)
-            else:
-                # we don't need to refresh the token,
-                # an expired token is valid if the FxA session is not
-                # revoked from the FxA settings
+            token_info = FXAAuthBackend.refresh_access_token(profile.fxa_refresh_token)
+            new_access_token = token_info.get("access_token")
+            if new_access_token:
+                request.session["oidc_access_token"] = new_access_token
```
LeoMcA, line 102: For sure, storing a 5-minute access token in the session - which then expires 11 and a bit hours before we hit the refresh route again - is a bit weird, I can see us forgetting this and being confused why calling some other fxa route with `session["oidc_access_token"]` fails in the future.

LeoMcA (commented):
Just some comments from a visual inspection, will test locally now:

LeoMcA (approved):
:rocket:  r+wc

kitsune/users/tasks.py
akatsoulas: That was my initial approach but since we do need to initialize the class because of the rest of the methods that we are using, I thought to use the same object everywhere to avoid confusion with different calling methods

kitsune/sumo/middleware.py
akatsoulas: I am updating the PR to exclude `ttl` from here. It will be an optional param used by the `process_event_profile_change` task

PR #5018, opened 2022-02-02
whitenoise + webpack

LeoMcA:
- configure whitenoise to recognize webpack files
- do gzip and brotli compression in webpack

### DIFF
--- a/webpack.prod.js
+++ b/webpack.prod.js
@@ -1,5 +1,7 @@
 const { mergeWithCustomize, unique } = require("webpack-merge");
 const MiniCssExtractPlugin = require("mini-css-extract-plugin");
+const CompressionPlugin = require("compression-webpack-plugin");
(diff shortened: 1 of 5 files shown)


webpack.prod.js
```diff
@@ -15,6 +17,27 @@ module.exports = mergeWithCustomize({
     new MiniCssExtractPlugin({
       filename: "[name].[contenthash].css",
     }),
+    new CompressionPlugin({
```
akatsoulas, line 20: do we need both gzip and brotli?

akatsoulas (approved):
nice work! r+wc

webpack.prod.js
akatsoulas: let's keep them I guess for a while and we can remove one of them in the future. Merge away!

PR #5033, opened 2022-02-18
Add ALB template service configuration; fix service deletion command

rvaralda:


akatsoulas (approved):
I don't see anything wrong here but maybe @smarnach could also take a look?

PR #5036, opened 2022-02-21
Revert EKS changes in deployments scripts.

akatsoulas:


akatsoulas:
@LeoMcA @rvaralda This is the revert of all the changes we did in commander script for k8s. I squashed all commits to this PR in order to revert easily when we are ready for EKS again. Fell free to review and merge when you need it for the migration.

rvaralda (approved):

LeoMcA:
I made use of this earlier in the migration by just checking out this branch, I think we can continue to do that if/when we need to, so I won't merge to avoid locking ourselves out of deploying to the EKS clusters still in use for frankfurt and staging.

akatsoulas:
@smarnach last commit updates the memory limit to 1.5 Gi

PR #5063, opened 2022-03-23
Use a context manager for DB connections.

akatsoulas:


kitsune/dashboards/models.py
```diff
@@ -54,11 +54,11 @@ def reload_period_from_analytics(cls, period, verbose=False):
...
 
             # Instead, we use raw SQL!
-            cursor = connection.cursor()
-            cursor.execute(
-                "DELETE FROM `dashboards_wikidocumentvisits`" "    WHERE `period` = %s", [period]
-            )
-
+            with connection.cursor() as cursor:
+                cursor.execute(
+                    "DELETE FROM `dashboards_wikidocumentvisits`" "    WHERE `period` = %s",
```
LeoMcA, line 59: we can remove the unnecessary whitespace: `s/" "   //`

LeoMcA (approved):
r+wc: there seem to be a couple more places we use db cursors:
https://github.com/mozilla/kitsune/blob/213e92345e19c3d4c987aca62c35c2a35f493838/kitsune/dashboards/readouts.py#L159
https://github.com/mozilla/kitsune/blob/268a9c5db4fb75a225645ad215d0a4614ac2c7a6/kitsune/kpi/api.py#L412

akatsoulas:
We should probably need to just delete those since we are not using a multidb setup here. I will file a ticket about it

PR #5082, opened 2022-03-31
hide subscription products from question forums lists

LeoMcA:
https://github.com/mozilla/sumo/issues/1008

### DIFF
--- a/kitsune/questions/views.py
+++ b/kitsune/questions/views.py
@@ -104,7 +104,7 @@ def product_list(request):
     return render(
         request,
         "questions/product_list.html",
-        {"products": Product.objects.filter(questions_locales__locale=request.LANGUAGE_CODE)},
+        {"products": Product.objects.with_question_forums(request)},
     )
 
 
@@ -410,7 +410,7 @@ def question_details(
 
     extra_kwargs.update(ans_)
 
-    products = Product.objects.filter(visible=True)
+    products = Product.objects.with_question_forums(request)
     topics = topics_for(product=question.product)
 
     related_documents = question.related_documents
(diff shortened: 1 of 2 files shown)


LeoMcA:
Blocked for now, from Seburo:
> Can you hold fire on actioning that?  We need to put something else in place to direct questions to the right place before we close them

kitsune/questions/views.py
```diff
@@ -410,7 +410,7 @@ def question_details(
 
     extra_kwargs.update(ans_)
 
-    products = Product.objects.filter(visible=True)
+    products = Product.objects.with_question_forums(request)
```
akatsoulas, line 413: Won't this change the current behavior?

LeoMcA: Yes, this is needed to fix https://github.com/mozilla/sumo/issues/1008: we don't want to add products without a support forum to the edit product dropdown

akatsoulas: My comment was related to visible. Don't we need to filter on both the `with_question_forums` and `visible=True`?

akatsoulas (approved):
r+ with regards to my last comment

kitsune/questions/views.py
LeoMcA: Since the end-user-facing logic which dictates whether a question can be posted to a product forum doesn't take into account that product's `visible` state (instead requiring its existence in the question `config.py` file, and an appropriate QuestionLocale row), I don't think so, no. Otherwise we could get into the state where end-users can still post into a product forum, but contributors can't move questions into it, nor see it listed on the support forums page.

PR #5085, opened 2022-04-04
Bump FRA mem limit to 2GB.

akatsoulas:


akatsoulas:
@smarnach this is only for FRA EKS. I will add a commit on the PR related to Kops

smarnach:
I think you can increase the request amount as well.

akatsoulas:
I updated the request to 512MB

PR #5089, opened 2022-04-06
stop parsing errors from crashing the whole page

LeoMcA:
https://bugzilla.mozilla.org/show_bug.cgi?id=1762414

### DIFF
--- a/kitsune/sumo/parser.py
+++ b/kitsune/sumo/parser.py
@@ -5,6 +5,7 @@ from django.conf import settings
 from django.template.loader import render_to_string
 from django.utils.translation import gettext_lazy as _lazy, ugettext as _
 
+from sentry_sdk import capture_exception
 from wikimarkup.parser import Parser, ALLOWED_TAGS
 
 from kitsune.gallery.models import Image, Video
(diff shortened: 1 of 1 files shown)


kitsune/sumo/parser.py
```diff
@@ -244,16 +245,22 @@ def parse(
...
+                    text,
+                    show_toc=show_toc,
+                    tags=tags or ALLOWED_TAGS,
+                    attributes=attributes or ALLOWED_ATTRIBUTES,
+                    styles=styles or ALLOWED_STYLES,
+                    nofollow=nofollow,
+                    strip_comments=True,
+                    **kwargs,
+                )
+            except Exception as e:
```
akatsoulas, line 259: let's make this a `TypeError`

akatsoulas (approved):
r+wc

kitsune/sumo/parser.py
LeoMcA: I wanted to catch all errors, not just the specific one in the bug above, to guard against any future parsing bugs we may encounter.

akatsoulas: let's catch the one we know about instead of silencing everything.

PR #5090, opened 2022-04-07
Route traffic to Service.

akatsoulas:


smarnach:
As tested yesterday, both approaches should work, but I believe routing directly to the pods is preferable, so let's close this one.

PR #5101, opened 2022-04-21
Remove unused svc

akatsoulas:
Since we are using `target-type: ip` we are routing traffic directly to pods making the NodePort svc redundant.


smarnach:
I don't think the whole service resource is redundant. It's referenced as the target of the ingress. We can probably remove the port specification from the service, though.

PR #5137, opened 2022-05-20
Import django-tidings files

akatsoulas:
* Update imports
* Use only python3


kitsune/tidings/models.py
```diff
@@ -0,0 +1,152 @@
+from django.conf import settings
+from django.contrib.auth.models import AnonymousUser
+from django.contrib.contenttypes.fields import GenericForeignKey, GenericRelation
+from django.contrib.contenttypes.models import ContentType
+from django.contrib.sites.models import Site
+from django.db import connections, models, router
+
+from .utils import import_from_setting, reverse
```
escattone, line 8: Nit: Perhaps for consistency with all of the other kitsune.tidings imports within this sub-package, we should make this an absolute import?
```python
from kitsune.tidings.utils import import_from_setting, reverse
```

kitsune/questions/tasks.py
```diff
@@ -2,14 +2,14 @@
...
-# NOTE: This import is just so _fire_task gets registered with celery.
-import tidings.events  # noqa
 from celery import task
 from django.conf import settings
 from django.contrib.auth.models import User
 from django.db import connection, transaction
 from sentry_sdk import capture_exception
 
+# NOTE: This import is just so _fire_task gets registered with celery.
+import kitsune.tidings.events  # noqa
```
escattone, line 12: I was wondering if it might be better to register the tidings Celery tasks in the app config? What if we change the tidings app config to something like:

```python
class TidingsConfig(AppConfig):
    name = "kitsune.tidings"
    default_auto_field = "django.db.models.AutoField"

    def ready(self):
        # Register the tidings Celery tasks.
        from kitsune.tidings import events, tasks  # noqa
```

and then move `kitsune.tidings` such that it's the first of the kitsune apps? Something we can consider later if at all, just thinking out loud.

escattone:
@akatsoulas I had the two thoughts above (which you can do with as you wish), and one more thing. I think we should add the unit tests from https://github.com/mozilla/django-tidings as well. Overall, this looks excellent to me.

PR #5150, opened 2022-06-10
disallow association of images if not owner

escattone:
Blocks the association of an image with objects which are not owned by the requester.

### DIFF
--- a/kitsune/upload/views.py
+++ b/kitsune/upload/views.py
@@ -42,6 +42,10 @@ def up_image_async(request, model_name, object_pk):
         message = _("Object does not exist.")
         return HttpResponseNotFound(json.dumps({"status": "error", "message": message}))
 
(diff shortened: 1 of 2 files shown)


kitsune/upload/views.py
```diff
@@ -42,6 +42,10 @@ def up_image_async(request, model_name, object_pk):
         message = _("Object does not exist.")
         return HttpResponseNotFound(json.dumps({"status": "error", "message": message}))
 
+    if request.user != getattr(obj, "creator", obj):
```
akatsoulas, line 45: Let's also whitelist superusers here.

akatsoulas (approved):
r+wc

kitsune/upload/views.py
escattone: Great catch, will do!

PR #5152, opened 2022-06-13
clear user inbox and outbox when anonymizing

escattone:
Fixes https://github.com/mozilla/sumo/issues/674

### DIFF
--- a/kitsune/users/tests/__init__.py
+++ b/kitsune/users/tests/__init__.py
@@ -2,6 +2,7 @@ import factory
 from django.contrib.auth.models import Group, Permission, User
 from django.contrib.contenttypes.models import ContentType
 
+from kitsune.messages.utils import send_message
 from kitsune.sumo.tests import FuzzyUnicode, LocalizingClient, TestCase
 from kitsune.tidings.models import Watch
 from kitsune.users.models import CONTRIBUTOR_GROUP, AccountEvent, Profile, Setting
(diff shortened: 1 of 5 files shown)


kitsune/users/utils.py
```diff
@@ -98,9 +99,10 @@ def anonymize_user(user):
     profile.fxa_uid = "{user_id}-{uid}".format(user_id=user.id, uid=str(uid))
     profile.save()
 
-    # Deactivate the user and change key information
+    # Change key information, clear the user's inbox/outbox, and deactivate the user.
     user.username = f"user{uid.int}"
     user.email = f"{uid.int}@example.com"
+    clear_inbox_and_outbox(user)
```
akatsoulas, line 105: I feel that adding a method that will perform two `delete` queries adds unnecessary complexity without any benefits. Let's just trigger the `delete`s here

kitsune/users/tests/test_views.py
```diff
@@ -442,3 +448,37 @@ def test_id_token(self, key):
...
+            username="ringo", email="ringo@beatles.com", groups=[GroupFactory()]
+        )
+        self.client.login(username=self.user.username, password="testpass")
+        super(UserCloseAccountTests, self).setUp()
+
+    def test_close_account(self):
+        """Test the closing of a user's account."""
+        # Populate inboxes and outboxes with messages between the user and other users.
+        conv = ConversationFactory(primary_user=self.user, number_of_other_users=2)
+        assert self.user.is_active
```
akatsoulas, line 465: Let's use the assert methods from unittest.

akatsoulas: This applies to all new/edited tests that are not using the the assert methods from the `unittest` framework.

kitsune/users/tests/__init__.py
```diff
@@ -104,3 +105,30 @@ class Meta:
...
+    def __init__(self, primary_user=None, number_of_other_users=2):
+        # Set or create the primary user and create some other users.
+        self.user = primary_user or UserFactory()
+        self.other_users = UserFactory.create_batch(number_of_other_users)
+        # Populate the inboxes and outboxes of the users.
+        for sender in self.other_users:
+            send_message([self.user], "foo", sender=sender)
+        send_message(self.other_users, "bar", sender=self.user)
+        # Confirm our expectations.
+        assert self.user.outbox.count() == 1
```
akatsoulas, line 124: Assertions should not be part of the factory. Factories should prepare the objects that we need for the testing and we can perform all assertions in the test itself

kitsune/users/tests/__init__.py
```diff
@@ -104,3 +105,30 @@ class Meta:
...
+        assert self.user.inbox.count() == len(self.other_users)
+        assert self.inboxes_and_outboxes_of_others_are_populated()
+
+    def inbox_and_outbox_of_primary_user_are_empty(self):
```
akatsoulas, line 128: I feel that we should test this in any test that we need to and not in the factory. It's not DRY but for the tests shouldn't be a problem. Let's keep each test as atomic as possible

kitsune/users/tests/__init__.py
```diff
@@ -104,3 +105,30 @@ class Meta:
...
+        return (self.user.inbox.count() == 0) and (self.user.outbox.count() == 0)
+
+    def inboxes_and_outboxes_of_others_are_populated(self):
```
akatsoulas, line 131: Similar as above

akatsoulas (changes requested):

escattone:
Thanks for the review @akatsoulas! I believe I've addressed everything, and this is ready for your final review.

akatsoulas:
Let's hold off merging until we release tomorrow

akatsoulas:
feel free to merge

PR #5153, opened 2022-06-15
fix showfor on revision pages

escattone:
fix mozilla/sumo#1043

The "show_for" selector on revision pages currently doesn't work because the revision pages include `kitsune/sumo/static/sumo/js/wiki.js` but that code doesn't call `new ShowFor()` for revision pages.

This PR does two things:
- It includes `kitsune/sumo/static/sumo/js/document.js` instead of `kitsune/sumo/static/sumo/js/wiki.js` for revision pages, because `kitsune/sumo/static/sumo/js/document.js` calls `new ShowFor()` by default, and because it seems to be better suited to the revision page in the sense that it doesn't include superfluous JS code.
- It also uses the `include_showfor=True` keyword argument of the `document_tools` macro to generate the `show_for` selector instead of using the `show_for` macro with a header. It makes for a better looking page that's more consistent with the document pages.

### DIFF
--- a/kitsune/wiki/jinja2/wiki/revision.html
+++ b/kitsune/wiki/jinja2/wiki/revision.html
@@ -1,11 +1,11 @@
 {% extends "wiki/base.html" %}
 {% from "wiki/includes/sidebar_modules.html" import document_tools with context %}
-{% from "wiki/includes/document_macros.html" import show_for with context %}
 {# L10n: {n} is the revision number, {t} is the title of the document. {c} is the category. #}
 {% set title = _('Revision {n} | {t} | {c}')|f(n=revision.id, t=document.title, c=document.get_category_display()) %}
 {% set crumbs = [(document.get_absolute_url(), document.title),
                  (None, _('Revision {num}')|f(num=revision.id))] %}
 {% set classes = 'document' %}
+{% set scripts = ('document',) %}
 {% set meta = [('robots', 'noindex, nofollow')] %}
 
 {% block content %}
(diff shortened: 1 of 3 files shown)


kitsune/wiki/jinja2/wiki/revision.html
```diff
@@ -91,11 +91,5 @@ <h1 class="sumo-page-heading">{{ document.title }}</h1>
 {% endblock %}
 
 {% block side_top %}
-  {{ document_tools(document, document.parent, user, '', settings) }}
-{% endblock %}
-
-{% block side_bottom %}
```
akatsoulas, line 97: To be honest I would prefer to keep `show_for`instead of hiding the same call (with a different arg) a level deep. I would even go a step further and say that we should remove the `show_for` option from the doucument_tools and just use it directly. 

I don't object to passing a header of False but at the same time these pages are used heavily by contributors. I do have a concern that we will confuse users by removing just `Article is for:` without adding much to the aesthetics or the UX of the page.

kitsune/wiki/jinja2/wiki/revision.html
```diff
@@ -1,11 +1,11 @@
 {% extends "wiki/base.html" %}
 {% from "wiki/includes/sidebar_modules.html" import document_tools with context %}
-{% from "wiki/includes/document_macros.html" import show_for with context %}
 {# L10n: {n} is the revision number, {t} is the title of the document. {c} is the category. #}
 {% set title = _('Revision {n} | {t} | {c}')|f(n=revision.id, t=document.title, c=document.get_category_display()) %}
 {% set crumbs = [(document.get_absolute_url(), document.title),
                  (None, _('Revision {num}')|f(num=revision.id))] %}
 {% set classes = 'document' %}
+{% set scripts = ('document',) %}
```
akatsoulas, line 8: Does it makes sense to add a revision.js file which will import showFor and initialize it? Then we can use just that file

akatsoulas (changes requested):

kitsune/wiki/jinja2/wiki/revision.html
escattone: @akatsoulas Yes, I think that's better. I'll do that.

akatsoulas (approved):
r+ squash and merge

PR #5160, opened 2022-06-17
fix ratelimit and is_ratelimited

escattone:
mozilla/sumo#1052

This PR changes the `ratelimit` and `is_ratelimited` SUMO utils to respect the `name` and `method` arguments. This will introduce the following changes from the way things currently work:
- only POST requests will count against rate limits, unless of course another method or methods are explicitly specified using the `method` argument (as far as I can tell, this will not change anything, since all rate-limited endpoints currently only accept POST requests anyway)
- `forum-post` and `aaq-day` will be counted separately (so a limit of 5 POST requests for each separately)
- `kbforum-post-day` and `primate-message-day` will be counted separately (so a limit of 50 POST requests for each separately)
- `question-vote`, `answer-vote`, and `document-vote` will be counted separately (so a limit of 10 POST requests for each separately)

I suspect those changes will make SUMO work as originally intended, but if one or more of those aren't acceptable, we can of course change the names to consolidate groups as desired. So for example, if we'd like to limit the number of any kind of vote to 10, we could change each of the names `question-vote`, `answer-vote`, and `document-vote` to just `vote`.

Some final points:
- Instead of importing and using `ratelimit.utils.is_ratelimited`, I now import and use `ratelimit.core.is_ratelimited`. As of `django-ratelimit` version `3`, `ratelimit.core` is the preferred public interface, so we should use that going forward. Note that unlike `ratelimit.utils.is_ratelimited`, `ratelimit.core.is_ratelimited` does **not** set `request.limited`, so that's now handled completely within `kitsune.sumo.utils.is_ratelimited`.
- I removed the `skip_if` argument from the `ratelimit` and `is_ratelimited` functions since it wasn't used anywhere.
- When recording a rate limit event, [we're currently using`request.META.get("HTTP_X_CLUSTER_CLIENT_IP", request.META["REMOTE_ADDR"])`](https://github.com/mozilla/kitsune/blob/db6aefb19b65eb27dfb400293c97c0f1de5072f5/kitsune/sumo/utils.py#L278) to get the offending IP address. We should use the same IP address that `ratelimit.core.is_ratelimited` uses, so I've changed that to use `request.META["REMOTE_ADDR"]`.

### DIFF
--- a/kitsune/sumo/tests/test_utils.py
+++ b/kitsune/sumo/tests/test_utils.py
@@ -140,29 +140,38 @@ class ChunkedTests(TestCase):
 
 
 class IsRatelimitedTest(TestCase):
-    def test_ratelimited(self):
+    def test_ratelimited_grouping_by_name_and_rate(self):
+        """Ensure that both the name and the rate differentiate the counting group."""
         request = Mock()
         request.user = UserFactory()
         request.limited = False
         request.method = "POST"
 
-        # Ensure that both the name and the rate differentiate the counting group.
         self.assertEqual(is_ratelimited(request, "test1", "1/min"), False)
         self.assertEqual(is_ratelimited(request, "test2", "1/min"), False)
         self.assertEqual(is_ratelimited(request, "test1", "1/d"), False)
         self.assertEqual(is_ratelimited(request, "test1", "1/min"), True)
+
         request.limited = False
         self.assertEqual(is_ratelimited(request, "test2", "1/min"), True)
+
         request.limited = False
         self.assertEqual(is_ratelimited(request, "test1", "1/d"), True)
 
-        # Ensure the method(s) is(are) respected.
+    def test_ratelimited_restriction_by_http_method(self):
+        """Ensure that the HTTP method(s) is(are) respected."""
+        request = Mock()
+        request.user = UserFactory()
         request.limited = False
+
+        request.method = "POST"
         self.assertEqual(is_ratelimited(request, "test3", "1/min", "GET"), False)
         self.assertEqual(is_ratelimited(request, "test3", "1/min", "GET"), False)
+
         request.method = "GET"
         self.assertEqual(is_ratelimited(request, "test3", "1/min", "GET"), False)
         self.assertEqual(is_ratelimited(request, "test3", "1/min", "GET"), True)
+
         request.limited = False
         request.method = "GET"
         self.assertEqual(is_ratelimited(request, "test3", "1/min", ("PUT", "POST")), False)
@@ -171,10 +180,16 @@ class IsRatelimitedTest(TestCase):
         request.method = "POST"
         self.assertEqual(is_ratelimited(request, "test3", "1/min", ("PUT", "POST")), True)
 
-        # If the request is already limited, then it should remain limited,
-        # but still increment when called.
-        request.method = "POST"
+    def test_ratelimited_when_already_limited(self):
+        """
+        Ensure that if the request is already limited, then it should remain
+        limited, but the counting group should still be incremented when called.
+        """
+        request = Mock()
+        request.user = UserFactory()
         request.limited = True
+        request.method = "POST"
+
         num_records_before = Record.objects.count()
         # This call should return True since the request was already limited,
         # but it should still increment the "test4 1/min" group.
@@ -182,6 +197,7 @@ class IsRatelimitedTest(TestCase):
         # It should not have logged a record of the ratelimit since it
         # occurred prior to this call.
         self.assertEqual(Record.objects.count(), num_records_before)
+
         request.limited = False
         # Let's confirm that the previous call really did increment.
         self.assertEqual(is_ratelimited(request, "test4", "1/min"), True)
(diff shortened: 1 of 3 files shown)


kitsune/sumo/utils.py
```diff
@@ -254,32 +254,34 @@ def _wr(self, s):
...
-    Reimplement ``ratelimit.helpers.is_ratelimited``, with sumo-specific details:
+    Wraps ``ratelimit.core.is_ratelimited`` with sumo-specific details:
 
     * Always check for the bypass rate limit permission.
     * Log times when users are rate limited.
     * Always uses ``user_or_ip`` for the rate limit key.
+    * Adds the Boolean attribute "limited" to the request.
     """
-    if skip_if(request) or request.user.has_perm("sumo.bypass_ratelimit"):
+    if request.user.has_perm("sumo.bypass_ratelimit"):
```
akatsoulas, line 266: You could refactor the next block to:

```python
if request.user.has_perm("sumo.bypass_ratelimit"):
    return False
if (limited := is_ratelimited_core(...):
``` 
This will remove the `else` and an `if` statement, simplifying the code a bit

kitsune/sumo/utils.py
```diff
@@ -254,32 +254,34 @@ def _wr(self, s):
...
-        # old/deleted custom user_or_ip method
-        rl_is_ratelimited(
-            request, increment=True, group="sumo.utils.is_ratelimited", rate=rate, key="user_or_ip"
+        limited = is_ratelimited_core(
+            request, group=name, key="user_or_ip", rate=rate, method=method, increment=True
         )
-        if request.limited:
+        if limited:
+            # We only record a ratelimit event for this counter.
             if hasattr(request, "user") and request.user.is_authenticated:
```
akatsoulas, line 274: is the `hasattr` needed?

kitsune/sumo/utils.py
```diff
@@ -254,32 +254,34 @@ def _wr(self, s):
...
                 key = 'user "{}"'.format(request.user.username)
             else:
-                ip = request.META.get("HTTP_X_CLUSTER_CLIENT_IP", request.META["REMOTE_ADDR"])
-                key = "anonymous user ({})".format(ip)
+                key = "anonymous user ({})".format(request.META["REMOTE_ADDR"])
```
akatsoulas, line 277: we could use an f string here
`f"anonymous user {request.META['REMOTE_ADDR']}"`

kitsune/sumo/utils.py
```diff
@@ -254,32 +254,34 @@ def _wr(self, s):
...
             Record.objects.info(
                 "sumo.ratelimit", "{key} hit the rate limit for {name}", key=key, name=name
             )
+
+        # If the request was already limited, keep it that way.
+        request.limited = getattr(request, "limited", False) or limited
```
akatsoulas, line 283: isn't `limited` always present in the request?

kitsune/sumo/utils.py
escattone: No, good catch, I'll remove that.

kitsune/sumo/utils.py
escattone: No, `limited` would only be present on the incoming `request` if this same function (`is_ratelimited()`) had already been called earlier, which is the case, for example, when stacking `@ratelimit()` decorators (see https://github.com/mozilla/kitsune/blob/3b5f420f58630dff3b4203e353bcbc6929fd0acb/kitsune/questions/views.py#L696).

kitsune/sumo/utils.py
escattone: Nice! I haven't used the "walrus" operator much yet.

kitsune/sumo/utils.py
```diff
@@ -265,22 +265,22 @@ def is_ratelimited(request, name, rate, method="POST"):
...
-        limited = is_ratelimited_core(
-            request, group=name, key="user_or_ip", rate=rate, method=method, increment=True
+        return False
+
+    if limited := is_ratelimited_core(
+        request, group=name, key="user_or_ip", rate=rate, method=method, increment=True
+    ):
+        # We only record a ratelimit event for this counter.
+        if request.user.is_authenticated:
+            key = f'user "{request.user.username}"'
```
akatsoulas, line 275: super nit: I would prefer double quotes externally and single internally just b/c we have all strings with double quotes because of black.

kitsune/sumo/tests/test_utils.py
```diff
@@ -141,16 +141,52 @@ def test_chunked(self):
...
-        request.user = u
+        request.user = UserFactory()
         request.limited = False
         request.method = "POST"
 
-        # One call to the rate limit won't trigger it.
-        self.assertEqual(is_ratelimited(request, "test-ratelimited", "1/min"), False)
-        # But two will
-        self.assertEqual(is_ratelimited(request, "test-ratelimited", "1/min"), True)
+        # Ensure that both the name and the rate differentiate the counting group.
```
akatsoulas, line 149: Let's break different cases in different tests.

akatsoulas (approved):
r+wc. Please squash the 2 ocmmits into a single one before merging

PR #5166, opened 2022-06-24
fix feeds and search of forums

escattone:
mozilla/sumo#1058

Fixes issue with the feeds and search endpoints for forums.


kitsune/forums/feeds.py
```diff
@@ -12,7 +13,10 @@ class ThreadsFeed(Feed):
     feed_type = Atom1Feed
 
     def get_object(self, request, forum_slug):
-        return get_object_or_404(Forum, slug=forum_slug)
+        forum = get_object_or_404(Forum, slug=forum_slug)
+        if not forum.allows_viewing_by(request.user):
```
akatsoulas, line 17: I am wondering whether it's a better idea to add a method in the `Forum` class similar to `authorized_forums_for_user` for consistency. In any case this is a nit

kitsune/forums/tests/test_feeds.py
```diff
@@ -1,17 +1,27 @@
...
+)
 from kitsune.sumo.tests import get
+from kitsune.users.tests import UserFactory
+
 
 YESTERDAY = datetime.now() - timedelta(days=1)
 
 
-class ForumTestFeedSorting(ForumTestCase):
+class ForumTestFeeds(ForumTestCase):
```
akatsoulas, line 22: Shall we also test that emails are not sent?

akatsoulas (approved):
r+wc

kitsune/forums/feeds.py
escattone: Hm, I don't see how that would work here, but maybe I'm misunderstanding something? It looks like the `authorized_forums_for_user` class method is not used anywhere.

akatsoulas: It could be used but I am starting to second guess the efficiency of that method. Pls ignore

PR #5168, opened 2022-06-27
upgrade to celery 5.2.7

escattone:
mozilla/sumo#1010

This PR upgrades Celery to `5.2.7`. I did the following:
- `poetry add celery@5.2.7`
- replaced `celery worker` option `--maxtasksperchild` with `--max-tasks-per-child` since `--maxtasksperchild` is no longer supported in Celery version 5
- replaced `from celery import task` with `from celery import shared_task`
- removed the `ignore_result=True` keyword argument from some `@shared_task` calls, since we set `CELERY_TASK_IGNORE_RESULT` to True so all task results are ignored by default


akatsoulas (approved):
nice! how about the `CELERY_ACCEPT_CONTENT` setting? Can we remove pickle?

escattone:
> how about the CELERY_ACCEPT_CONTENT setting? Can we remove pickle?
No, not yet, but I have another branch started to remove `pickle`, and it's fairly easy. I'll have that ready later today.

PR #5169, opened 2022-06-28
remove need for pickle serializer for celery tasks

escattone:
The only remaining challenge is to find a nice way to avoid `pickle` when `kitsune.tidings.events` are fired.


kitsune/tidings/tasks.py
```diff
@@ -1,13 +1,14 @@
 from celery import shared_task
+from django.contrib.auth import get_user_model
 
 from kitsune.tidings.models import Watch
 
 
 @shared_task
-def claim_watches(user):
```
escattone, line 7: We could also just delete this file `kitsune/tidings/tasks.py`, since it's not used anywhere as far as I can tell.

kitsune/tidings/utils.py
```diff
@@ -94,3 +99,15 @@ def import_from_setting(setting_name, fallback):
...
 reverse = import_from_setting("TIDINGS_REVERSE", django_reverse)  # no QA
+
+
+def get_class(module_name, class_name):
+    """
+    Convenience function for extracting a class from the given module name using
+    the given class name. Raises a DisallowedModule exception if the module name
+    is outside of kitsune.
+    """
+    if not module_name.startswith("kitsune."):
```
escattone, line 110: This is really important for security. I wonder if we need to lock this down even more? I don't think so, but would like to hear your thoughts. For example, I was thinking that we could lock it down even more by checking that the `module_name` is one of the installed apps, or something like that.

kitsune/sumo/management/commands/enqueue_lag_monitor_task.py
```diff
@@ -9,4 +9,4 @@ class Command(BaseCommand):
     help = "Fire a task that measures the queue lag."
 
     def handle(self, **options):
-        measure_queue_lag.delay(datetime.now())
+        measure_queue_lag.delay(datetime.now().isoformat())
```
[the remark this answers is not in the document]
escattone: My understanding is that the JSON serializer that Celery uses doesn't handle `datetime` objects by default, and since the `datetime.now()` function returns a `datetime` object (which would require `pickle` serialization or a custom JSON encoder/decoder), I instead converted the datetime object to a string in `ISO` format.

akatsoulas:
Let's extract the `generate_thumbnail`, `compress_image` and `measure_queue_lag` in a separate PR

kitsune/tidings/events.py
```diff
@@ -108,8 +104,30 @@ class Event(object):
...
+                instance_cls_module_name=instance.__class__.__module__,
+                instance_cls_name=instance.__class__.__name__,
+                instance_id=instance.id,
+            )
+        if exclude:
+            if not isinstance(exclude, Sequence):
+                exclude = [exclude]
+            kwargs.update(exclude_user_ids=[user.id for user in exclude])
+
+        fire.delay(cls.__module__, cls.__name__, **kwargs)
```
akatsoulas, line 126: This is very confusing. The fact that the Event class has a method `fire` but we also have a `fire` task

kitsune/tidings/tasks.py
```diff
@@ -1,13 +1,55 @@
...
+
+
+@shared_task
+def fire(
+    event_cls_module_name,
+    event_cls_name,
+    instance_cls_module_name=None,
+    instance_cls_name=None,
+    instance_id=None,
+    exclude_user_ids=None,
```
akatsoulas, line 27: The calling signature and the actual function signature are not the same. There's an implicit unpacking happening here but the `kwargs` pass to this function from the `Event` class could be anything. Passing `kwargs` is probably cleaner and you can do the unpacking within the function

kitsune/tidings/tasks.py
```diff
@@ -1,13 +1,55 @@
...
+    Celery task that is JSON-serializer friendly, and that fires the event specified by
+    the "event_cls_*" arguments while excluding the users specified by "exclude_user_ids"
+    if provided. The event will be constructed with the database model instance specified
+    by the "instance_*" arguments if provided.
+    """
+    # Get the event class.
+    event_cls = get_class(event_cls_module_name, event_cls_name)
+
+    # Construct the event, either with or without an instance.
+    if instance_cls_module_name and instance_cls_name and instance_id is not None:
```
akatsoulas, line 39: better be consistent on how we are checking here.

akatsoulas: nit you could use `all()`

kitsune/tidings/tasks.py
```diff
@@ -1,13 +1,55 @@
...
+        # Get the instance class, and then the instance itself from the database.
+        instance_cls = get_class(instance_cls_module_name, instance_cls_name)
+        instance = instance_cls.objects.get(id=instance_id)
+        event = event_cls(instance)
```
akatsoulas, line 43: Is this trying to mimic the `EventSubclass(instance_obj).fire(*args)` type of calling?

kitsune/tidings/tasks.py
```diff
@@ -1,13 +1,55 @@
...
+    else:
+        event = event_cls()
+
+    # Get the excluded users, if any.
+    if exclude_user_ids is None:
+        exclude = None
+    else:
+        if not isinstance(exclude_user_ids, Sequence):
+            exclude_user_ids = [exclude_user_ids]
+        exclude = list(get_user_model().objects.filter(id__in=exclude_user_ids).all())
```
akatsoulas, line 53: Why are you forcing evaluation here?

kitsune/tidings/tasks.py
```diff
@@ -1,13 +1,55 @@
...
+    Attach any anonymous watches having a user's email to that user.
     Call this from your user registration process if you like.
-
     """
+    user = get_user_model().objects.get(id=user_id)
     Watch.objects.filter(email=user.email).update(email=None, user=user)
+
+
+@shared_task
+def fire(
```
akatsoulas, line 21: I have mixed feelings about this approach. We have this task which is called by a `classmethod` that gets an instance sometimes and the task constructs the object. Wouldn't be simpler to just move the task in each subclass of Event? The main event could have a synchronous `run`, `send_emails` or any other suitable name that each `fire` method from the subclass would run asynchronously

akatsoulas (changes requested):

kitsune/tidings/events.py
```diff
@@ -108,8 +104,30 @@ class Event(object):
     #: ``set(['color', 'flavor'])``
     filters = set()
 
-    def fire(self, exclude=None, delay=True):
-        """Notify everyone watching the event.
+    @classmethod
+    def fire_async(cls, instance=None, exclude=None):
```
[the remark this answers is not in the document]
escattone: We still need some method that synchronously performs the work of the `fire` method, because the Celery task needs to call that, but we could call that something else instead, like `send_emails`, to make it less confusing?

kitsune/tidings/events.py
```diff
@@ -108,8 +104,30 @@ class Event(object):
...
+        if instance:
+            kwargs.update(
+                instance_cls_module_name=instance.__class__.__module__,
+                instance_cls_name=instance.__class__.__name__,
+                instance_id=instance.id,
+            )
+        if exclude:
+            if not isinstance(exclude, Sequence):
+                exclude = [exclude]
+            kwargs.update(exclude_user_ids=[user.id for user in exclude])
```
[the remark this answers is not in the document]
escattone: The `exclude` keyword argument can be either a single user object (that's why we check for a non-`Sequence`) or an iterable of user objects, but we need the list comprehension because we have to convert those user objects into JSON-friendly information that the Celery task accepts, so we convert them into a list of integers (ids).

kitsune/tidings/events.py
escattone: Sorry, I think that's a bad choice of mine to name the Celery task `fire` as well (too many meanings for `fire`). We can change the name of the Celery task if we stay with this approach.

kitsune/tidings/tasks.py
escattone: There's a lot to discuss here, so probably better to discuss in a video meeting, but I have three thoughts.
- If we move the task to each subclass of `Event`, those tasks would no longer be auto-discoverable by https://github.com/mozilla/kitsune/blob/2b40d5841ceeb6bd2f8108ebfd3decbec97a168f/kitsune/celery.py#L10.
- The `fire_async` `classmethod` serves the purpose of translating the incoming arguments that reference Python objects, like `instance` and `exclude`, into integers and strings and then calling the Celery task with those JSON-friendly arguments.
- If you don't like this general approach of having a single Celery task in tidings, there is a [version of this PR (an earlier commit)](https://github.com/mozilla/kitsune/pull/5169/commits/11f050149a0d22dee56f48a136742a44c4022426) that creates separate Celery tasks for each Event sub-class.

kitsune/tidings/tasks.py
escattone: No, it doesn't work as a queryset.

kitsune/tidings/utils.py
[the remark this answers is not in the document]
escattone: After discussing with you earlier today, you convinced me that I should get rid of this.

kitsune/tidings/events.py
```diff
@@ -121,29 +134,37 @@ def fire(self, exclude=None, delay=True):
...
+            )
+        return result
+
+    def get_constructor_instance(self):
+        """
+        Event subclasses constructed from a database model instance should override this
+        method and provide the database model instance used to construct themselves.
+        """
+        return None
+
```
escattone, line 167: @tasos Based on our discussion earlier today, you might prefer that we get rid of `get_constructor_instance()` and make `serialize()` an abstract method that must be implemented in each subclass, so for example, each subclass would do something like this:

```python
def serialize(self):
    """
    Serialize this event into a JSON-friendly dictionary.
    """
    return {
        "event": {
            "module": "kitsune.wiki.events"
            "class": "EditDocumentEvent"
        },
        "instance": {
            "module": "kitsune.wiki.models",
            "class": "Revision",
            "id": self.revision.id
        }
    }
```

and we can still do that if you think that's easier to understand -- in fact I already have that option implemented on my laptop-- but just wanted to see what you thought of this approach first.

escattone:
@akatsoulas This is ready for you to take another look. I kept the same `fire` interface, but changed the Celery task name to `send_emails` and have the event subclasses implement the `get_constructor_instance()` method (but see https://github.com/mozilla/kitsune/pull/5169#discussion_r921637891 above).

kitsune/tidings/events.py
akatsoulas: I think this is a bit clearer on intentions :+1:

kitsune/tidings/events.py
```diff
@@ -109,7 +105,24 @@ class Event(object):
...
 
     def fire(self, exclude=None, delay=True):
-        """Notify everyone watching the event.
+        """
+        Notify everyone watching the event, either synchronously or asynchronously.
+        """
+        if delay:
+            event_info = self.serialize()
+            if exclude:
+                if not isinstance(exclude, Sequence):
```
akatsoulas, line 114: I do feel that we are adding unnecessary  complexity here. Let's change the calling of this function to always pass a list of ids (optionally) and we can only do:

```python
if not exclude:
    exclude = []
```

In the case where we are dealing with querysets let's use the `values_list` . By doing this we can omit all this check since we will be always passing a list. We can also add that  in the docstring of the fuction

akatsoulas (commented):
I believe we are very close. I mostly had a nit and with your suggestion this should be good to go

akatsoulas (approved):
r+wc Squash and merge

PR #5170, opened 2022-06-28
strengthen get_next_url

escattone:
mozilla/sumo#1063

Thanks to @bkochendorfer for his suggestions and help with this!

### DIFF
--- a/kitsune/sumo/utils.py
+++ b/kitsune/sumo/utils.py
@@ -128,6 +128,10 @@ def get_next_url(request):
     else:
         url = request.META.get("HTTP_REFERER")
 
+    # Remove any newline or carriage-return characters.
+    if url:
+        url = url.replace("\n", "").replace("\r", "")
+
     if not settings.DEBUG and not is_safe_url(
         url, allowed_hosts={Site.objects.get_current().domain}
     ):
--- a/kitsune/sumo/utils.py
+++ b/kitsune/sumo/utils.py
@@ -4,15 +4,16 @@ import sys
 from contextlib import contextmanager
 from datetime import datetime
 from functools import lru_cache
+from urllib.parse import urlparse
 
 from django.conf import settings
 from django.contrib.sites.models import Site
-from django.core.exceptions import DisallowedRedirect
 from django.db import models
 from django.db.models.signals import pre_delete
-from django.http import HttpResponseRedirect
 from django.templatetags.static import static
 from django.utils import translation
+from django.utils.encoding import iri_to_uri
+
 from django.utils.http import url_has_allowed_host_and_scheme, urlencode
 from ratelimit.core import is_ratelimited as is_ratelimited_core
 from timeout_decorator import timeout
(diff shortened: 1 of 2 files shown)


kitsune/sumo/utils.py
```diff
@@ -128,6 +128,10 @@ def get_next_url(request):
     else:
         url = request.META.get("HTTP_REFERER")
 
+    # Remove any newline or carriage-return characters.
+    if url:
+        url = url.replace("\n", "").replace("\r", "")
```
smarnach, line 133: The `\t` character also needs to be removed, according to https://url.spec.whatwg.org/#url-parsing. However, I'd tend to simply reject URLs containing these characters instead of trying to sanitize them.

kitsune/sumo/utils.py
```diff
@@ -128,6 +128,10 @@ def get_next_url(request):
...
+
     if not settings.DEBUG and not is_safe_url(
```
smarnach, line 135: `is_safe_url()` has been renamed to `url_has_allowed_host_and_scheme()`, and the old name is deprecated, so we could change this while we are here.

escattone:
@smarnach I made some adjustments based on your feedback as well as your conversation with Django security. Thanks so much for your help!

bkochendorfer (approved):
Good tests and looks good to me

escattone:
Thanks for your review @bkochendorfer!

smarnach (approved):
Thanks, this looks like a reasonable approach, after what we learned from the Django folks.

kitsune/sumo/utils.py
```diff
@@ -128,7 +130,16 @@ def get_next_url(request):
     else:
         url = request.META.get("HTTP_REFERER")
 
-    if not settings.DEBUG and not is_safe_url(
+    if not url:
+        return None
+
+    # Ensure that the URL is allowed as a redirect.
+    try:
+        HttpResponseRedirect(url)
```
akatsoulas, line 138: I am not sure that I completely understand this solution. Can you maybe elaborate on why this is working?

akatsoulas, line 138: I am not sure that I completely understand this solution. Can you maybe elaborate on why this is working?

akatsoulas: I am trying to follow the code. This ends up being something like `abc:abc` which the urlparse does not recognize as a valid scheme (abc) when initializes the HttResponseRedirect

akatsoulas: I had to dive into the code to start understanding why this is a solution to the problem. Personally I find this a bit cryptic on why it's happening. The short version is:

- request object from HttpRequest reconstruct the query_string by using the `parse_qsl` from urllib.parse
- HttpResposneRedirect constrcuctor is using the HttpResponseRedirectBase where:
  - adds the location header by using the iri_to_uri
  - uses the urlparse from urllib on the passed url from the request which returns a schme of `javascript` in our example
  
Personally I feel it would be easier to understand this code if we had something similar to:
```python
parsed_url = urlparse(url)
if parsed_url.scheme and parsed_url.scheme not in ['http', 'https']:
    return None

if not settings.DEBUG and not url_has_allowed_host_and_scheme(
        url, allowed_hosts={Site.objects.get_current().domain}
    ):
        return None
return iri_to_uri(url)

escattone: @akatsoulas First of all, thanks for digging into this in more detail! Here are my thoughts:

akatsoulas (approved):

PR #5181, opened 2022-07-08
restrict access to documents that have no approved content

escattone:
mozilla/sumo#566

This PR addresses the problem where the title and slug, which could be "spammy", of freshly-created documents (English and non-English) are publicly available before their content is approved.

This PR changes our code such that these freshly created documents -- that have no approved content yet (`current_revision__isnull=True`) -- can only be "seen" by their creator, or superusers, or users with special permission. The users with special permission include locale leaders, locale reviewers, and users with `wiki.review_revision` and/or `wiki.delete_document` permission. For all other users, these documents and translations effectively do not exist.

Although the essence of this PR is fairly simple, it affects almost every URL in our `wiki` and `kbforums` apps, and consequently their tests.

- When fixing existing tests, the primary fix was to login as a user with permission to view documents without any approved content, but sometimes it was clearer to simply add an approved revision to the document created for the test.
- For the `helpful_vote` view, I decided to simply limit that endpoint to only allow votes on approved revisions.
- The `json_view` view already filters on `current_revision__isnull=False`, so there was no need to modify it.

### DIFF
--- a/kitsune/wiki/managers.py
+++ b/kitsune/wiki/managers.py
@@ -4,58 +4,66 @@ from django.db.models import Exists, OuterRef, Q
 from kitsune.wiki.permissions import can_delete_documents_or_review_revisions
 
 
-class DocumentManager(models.Manager):
+class VisibilityManager(models.Manager):
+    """Abstract base class for the Document and Revision Managers."""
+
+    # For managers of models related to documents, provide the name of the model attribute
+    # that provides the related document. For example, for the manager of revisions, this
+    # should be "document".
+    document_relation = None
+
+    def get_creator_condition(self, user):
+        """
+        Return a conditional (e.g., a Q or Exists object) that is only true
+        when the given user is the creator of this document or revision.
+        """
+        raise NotImplementedError
+
     def visible(self, user, **kwargs):
         """
         Documents are effectively invisible when they have no approved content,
         and the given user is not a superuser, nor allowed to delete documents or
         review revisions, nor a creator of one of the (yet unapproved) revisions.
         """
-        locale = kwargs.get("locale")
+        prefix = f"{self.document_relation}__" if self.document_relation else ""
+
+        locale = kwargs.get(f"{prefix}locale")
+
         qs = self.filter(**kwargs)
 
         if not user.is_authenticated:
             # Anonymous users only see documents with approved content.
-            return qs.filter(current_revision__isnull=False)
+            return qs.filter(**{f"{prefix}current_revision__isnull": False})
 
         if not (
             user.is_superuser or can_delete_documents_or_review_revisions(user, locale=locale)
         ):
             # Authenticated users without permission to see documents that
             # have no approved content, can only see those they have created.
-            from kitsune.wiki.models import Revision
-
-            rev_created_by_user = Revision.objects.filter(document=OuterRef("pk"), creator=user)
-            return qs.filter(Q(current_revision__isnull=False) | Exists(rev_created_by_user))
-
-        return qs
-
-    def get_visible(self, user, **kwargs):
-        return self.visible(user, **kwargs).get()
-
-
-class RevisionManager(models.Manager):
-    def visible(self, user, **kwargs):
-        """
-        Revisions are effectively invisible when their document has no approved content,
-        and the given user is not a superuser, nor allowed to delete documents or review
-        revisions, nor the creator.
-        """
-        locale = kwargs.get("document__locale")
-        qs = self.filter(**kwargs)
-
-        if not user.is_authenticated:
-            # Anonymous users only see revisions of documents with approved content.
-            return qs.filter(document__current_revision__isnull=False)
-
-        if not (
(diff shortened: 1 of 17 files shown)


escattone:
@akatsoulas I think I should add more tests (but it's a bit overwhelming given the number of endpoints), but I think it's at least ready for an initial review.

escattone:
@akatsoulas This is ready for review.

akatsoulas (commented):
I was examining the existing code-base along with the changes in the PR and I found this like which (I assume) seems to set a hook for our case 
https://github.com/mozilla/kitsune/blob/main/kitsune/wiki/views.py#L126

I am wondering whether we could just augment the logic there and return a 404 if the requirements that you are setting in the custom manager about the perms and login state of the user are not met. Thoughts?

akatsoulas:
Regardless though if we go with that solution, the mentioned hook is not used anywhere (I will double check that) so we certainly need to do some cleaning up if we are not going to use it.

kitsune/wiki/models.py
```diff
@@ -58,6 +61,34 @@ class _NotDocumentView(Exception):
...
+                rev_created_by_user = Revision.objects.filter(
+                    document=OuterRef("pk"), creator=user
+                )
+                qs = qs.filter(Q(current_revision__isnull=False) | Exists(rev_created_by_user))
+        else:
+            # Anonymous users only see documents with approved content.
+            qs = qs.filter(current_revision__isnull=False)
+        return qs
+
+    def get_visible(self, user, **kwargs):
```
akatsoulas, line 88: shouldn't we have more required args here to ensure a successful get operation?

kitsune/wiki/views.py
```diff
@@ -106,6 +106,14 @@ def _doc_page_cache_view(request, document_slug, *args, **kwargs):
     return _doc_page_cache_view
 
 
+def get_visible_document_or_404(user, **kwargs):
```
akatsoulas, line 109: we should probably move those helper methods to wiki/utils.py

kitsune/wiki/models.py
escattone: No, I don't think so. In this regard, `get_visible()` is  no different than Django's `get()` , which also does not ensure a successful operation in that either a `Model.DoesNotExist` or `Model.MultipleObjectsReturned` could be returned.

escattone:
> I was examining the existing code-base along with the changes in the PR and I found this like which (I assume) seems to set a hook for our case https://github.com/mozilla/kitsune/blob/main/kitsune/wiki/views.py#L126
> 
> I am wondering whether we could just augment the logic there and return a 404 if the requirements that you are setting in the custom manager about the perms and login state of the user are not met. Thoughts?

That's the path I initially went down as well, but changed my mind for two reasons:

escattone:
> Regardless though if we go with that solution, the mentioned hook is not used anywhere (I will double check that) so we certainly need to do some cleaning up if we are not going to use it.

It's actually used indirectly, because a `fallback_reason` of `no_content` ensures that we fall into the default `else` clause [here](https://github.com/mozilla/kitsune/blob/7ee97662c62058ce8fa004dd9ad8b1d7c7ba3adc/kitsune/wiki/jinja2/wiki/includes/document_macros.html#L106).

kitsune/wiki/tests/test_views.py
```diff
@@ -23,12 +23,327 @@
...
     TemplateDocumentFactory,
+    TestCaseBase,
+    TranslatedRevisionFactory,
     new_document_data,
 )
 from kitsune.wiki.views import _document_lock_check, _document_lock_clear, _document_lock_steal
 
 
-class RedirectTests(TestCase):
+class VisibilityTestsBase(TestCaseBase):
```
akatsoulas, line 33: With the tests let's be very repetitive and verbose so it can be very obvious what the testis about. I know that it will be add bolierplate code but this will make each test much more atomic and self contained

akatsoulas (changes requested):

kitsune/wiki/tests/__init__.py
```diff
@@ -25,6 +25,27 @@ class TestCaseBase(TestCase):
 
     client_class = LocalizingClient
 
+    def login_with_permission(self, permission_codename):
```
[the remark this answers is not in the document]
escattone: I wasn't sure I liked this either, so I decided to just get rid of it. :)

kitsune/wiki/managers.py
```diff
@@ -0,0 +1,61 @@
+from django.db import models
+from django.db.models import Exists, OuterRef, Q
+
+from kitsune.wiki.permissions import can_delete_documents_or_review_revisions
+
+
+class DocumentManager(models.Manager):
```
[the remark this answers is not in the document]
escattone: I like that idea a lot. See what you think about what I did there.

akatsoulas (approved):
This looks good! Let's move it to staging for some testing! Let's squash all commits into one. I am adding a `do not merge` label until we get the release out of the way

akatsoulas:
@escattone squash and merge away! Nice work!

PR #5190, opened 2022-07-28
add readiness probe

escattone:


escattone:
Closing this since the SRE team will be adding a readiness probe in our Helm chart that we'll be using in the near future.

PR #5192, opened 2022-07-29
[BUG 1680203] Layout issue for localized user profiles

smithellis:
Certain languages produce lengthy words in the user profile page.
I altered --details-label to allow for 170px width, which appears
to accomodate the longest word I could locate.

A better answer long term would be to have flex items here that
fit themselves, but that is possibly included in project level work in
the near future.


escattone (approved):
@smithellis Congrats on your first SUMO PR! 🎉

PR #5193, opened 2022-07-29
centralize locale fallback and add to what-links-here

escattone:
mozilla/sumo#477

This PR arose out of reviewing #4805, and is an alternative approach. That PR is well-written, and I like that it centralized the handling of locale-related fallbacks (e.g., used for redirecting URL's like`/de/<english-slug/` to `/de/<slug-of-german-translation/`), but I felt that its combination of middleware and a view decorator to mark exceptions made the wiki views more difficult to understand. I also felt that the locale-related fallback didn't make sense for most of the wiki endpoints.

This PR also centralizes the fallback code, but takes a more narrow approach in that it only changes the behavior of the `wiki.what_links_here` endpoint.

### DIFF
--- a/kitsune/wiki/utils.py
+++ b/kitsune/wiki/utils.py
@@ -1,4 +1,3 @@
-from enum import Enum, auto
 import random
 
 import requests
@@ -12,11 +11,6 @@ from django.utils.http import urlencode
 from kitsune.dashboards import LAST_7_DAYS
 from kitsune.dashboards.models import WikiDocumentVisits
 from kitsune.wiki.models import Document, Revision
-
-
-class FallbackCodes(Enum):
-    NO_TRANSLATION = auto()
-    TRANSLATION_AT_DIFFERENT_SLUG = auto()
 
 
 class BitlyException(Exception):
@@ -185,26 +179,43 @@ def get_featured_articles(product=None, locale=settings.WIKI_DEFAULT_LANGUAGE):
 
 
 def get_visible_document_or_404(user, **kwargs):
-    return get_object_or_404(Document.objects.visible(user, **kwargs))
+    """
+    Get the document specified by the keyword arguments and visible to the given user, or 404.
+    """
+
+    look_for_translation_via_parent = kwargs.pop("look_for_translation_via_parent", False)
+    return_parent_if_no_translation = kwargs.pop("return_parent_if_no_translation", False)
+
+    try:
+        return Document.objects.get_visible(user, **kwargs)
+    except Document.DoesNotExist:
+        pass
+
+    if (
+        not look_for_translation_via_parent
+        or not (locale := kwargs.get("locale"))
+        or (locale == settings.WIKI_DEFAULT_LANGUAGE)
+    ):
+        # We either don't want to try to find the translation via its parent, or it doesn't
+        # make sense, because we're not making a locale-specific request or the locale we're
+        # requesting is already the default locale.
+        raise Http404
+
+    # We couldn't find a visible translation in the requested non-default locale, so let's
+    # see if we can find a visible translation via its parent.
+    kwargs.update(locale=settings.WIKI_DEFAULT_LANGUAGE)
+    parent = get_object_or_404(Document.objects.visible(user, **kwargs))
+
+    # If there's a visible translation of the parent for the requested locale, return it.
+    if translation := parent.translated_to(locale, visible_for_user=user):
+        return translation
+
+    # Otherwise, we're left with the parent.
+    if return_parent_if_no_translation:
+        return parent
+
+    raise Http404
 
 
 def get_visible_revision_or_404(user, **kwargs):
     return get_object_or_404(Revision.objects.visible(user, **kwargs))
-
-
-def get_visible_document_or_fallback_or_404(user, locale, slug):
-    """Return a document with its fallback code, or 404."""
-    try:
-        return (Document.objects.get_visible(user, locale=locale, slug=slug), None)
-    except Document.DoesNotExist:
-        if locale == settings.WIKI_DEFAULT_LANGUAGE:
-            # Don't repeat the query if it's going to be the same one we just tried.
-            raise Http404
-        # No visible document exists in the requested locale so let's try the default locale.
-        parent = get_visible_document_or_404(
-            user, locale=settings.WIKI_DEFAULT_LANGUAGE, slug=slug
-        )
-        # If there's a visible translation to the requested locale, redirect to it.
-        if translation := parent.translated_to(locale, visible_for_user=user):
-            return (translation, FallbackCodes.TRANSLATION_AT_DIFFERENT_SLUG)
-        return (parent, FallbackCodes.NO_TRANSLATION)
--- a/kitsune/wiki/utils.py
+++ b/kitsune/wiki/utils.py
@@ -1,15 +1,22 @@
+from enum import Enum, auto
 import random
 
 import requests
 from django.conf import settings
 from django.contrib.auth.models import User
 from django.db.models import Prefetch, Q
+from django.http import Http404
 from django.shortcuts import get_object_or_404
 from django.utils.http import urlencode
 
 from kitsune.dashboards import LAST_7_DAYS
 from kitsune.dashboards.models import WikiDocumentVisits
 from kitsune.wiki.models import Document, Revision
+
+
+class FallbackCodes(Enum):
+    NO_TRANSLATION = auto()
+    TRANSLATION_AT_DIFFERENT_SLUG = auto()
 
 
 class BitlyException(Exception):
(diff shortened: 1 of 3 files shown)


kitsune/wiki/utils.py
```diff
@@ -183,3 +190,21 @@ def get_visible_document_or_404(user, **kwargs):
...
 def get_visible_revision_or_404(user, **kwargs):
     return get_object_or_404(Revision.objects.visible(user, **kwargs))
+
+
+def get_visible_document_or_fallback_or_404(user, locale, slug):
+    """Return a document with its fallback code, or 404."""
+    try:
+        return (Document.objects.get_visible(user, locale=locale, slug=slug), None)
+    except Document.DoesNotExist:
+        if locale == settings.WIKI_DEFAULT_LANGUAGE:
```
akatsoulas, line 200: I would prefer to move the logic outside of the `except` clause

kitsune/wiki/utils.py
```diff
@@ -183,3 +190,21 @@ def get_visible_document_or_404(user, **kwargs):
 
 def get_visible_revision_or_404(user, **kwargs):
     return get_object_or_404(Revision.objects.visible(user, **kwargs))
+
+
+def get_visible_document_or_fallback_or_404(user, locale, slug):
+    """Return a document with its fallback code, or 404."""
```
akatsoulas, line 196: Are we returning always the fallback code?

kitsune/wiki/utils.py
```diff
@@ -183,3 +190,21 @@ def get_visible_document_or_404(user, **kwargs):
...
+    try:
+        return (Document.objects.get_visible(user, locale=locale, slug=slug), None)
+    except Document.DoesNotExist:
+        if locale == settings.WIKI_DEFAULT_LANGUAGE:
+            # Don't repeat the query if it's going to be the same one we just tried.
+            raise Http404
+        # No visible document exists in the requested locale so let's try the default locale.
+        parent = get_visible_document_or_404(
```
akatsoulas, line 204: Basically  we have multiple levels of nesting with functions calling just other functions. Maybe it's better to change the naming of this function altogether and maybe re-approach what each one is doing. Basically this call is doing exactly the same thing with the `try` statement apart the 404. We could simplify this code and all the functions related to visible documents

kitsune/wiki/views.py
```diff
@@ -67,7 +67,12 @@
     send_contributor_notification,
     send_reviewed_notification,
 )
-from kitsune.wiki.utils import get_visible_document_or_404, get_visible_revision_or_404
+from kitsune.wiki.utils import (
+    FallbackCodes,
+    get_visible_document_or_404,
+    get_visible_document_or_fallback_or_404,
```
akatsoulas, line 73: maybe this visualization here helps to better understand what I am trying to say in my comment above.

kitsune/wiki/views.py
```diff
@@ -118,35 +123,20 @@ def document(request, document_slug, document=None):
...
-            url = urlparams(url, query_dict=request.GET)
-            return HttpResponseRedirect(url)
-        elif doc.current_revision:
-            # There is no translation, so we'll fall back to the approved parent,
+
+    doc, fallback_code = get_visible_document_or_fallback_or_404(
+        request.user, request.LANGUAGE_CODE, document_slug
+    )
+
+    if fallback_code is FallbackCodes.TRANSLATION_AT_DIFFERENT_SLUG:
```
akatsoulas, line 131: shouldn't be `==`?

kitsune/wiki/views.py
```diff
@@ -118,35 +123,20 @@ def document(request, document_slug, document=None):
...
+        url = doc.get_absolute_url()
+        url = urlparams(url, query_dict=request.GET)
+        return HttpResponseRedirect(url)
+
+    if fallback_code is FallbackCodes.NO_TRANSLATION:
```
akatsoulas, line 136: same as above

kitsune/wiki/views.py
```diff
@@ -606,24 +591,17 @@ def preview_revision(request):
...
-            url = reverse("wiki.document_revisions", args=[translation.slug], locale=locale)
-            return HttpResponseRedirect(url)
-        else:
-            raise Http404
+
+    doc, fallback_code = get_visible_document_or_fallback_or_404(
+        request.user, locale, document_slug
+    )
+
+    if fallback_code is FallbackCodes.TRANSLATION_AT_DIFFERENT_SLUG:
```
akatsoulas, line 599:
```suggestion
    if fallback_code == FallbackCodes.TRANSLATION_AT_DIFFERENT_SLUG:
```

kitsune/wiki/views.py
```diff
@@ -606,24 +591,17 @@ def preview_revision(request):
...
+        url = reverse("wiki.document_revisions", args=[doc.slug], locale=locale)
+        return HttpResponseRedirect(url)
+
+    if fallback_code is FallbackCodes.NO_TRANSLATION:
```
akatsoulas, line 603:
```suggestion
    if fallback_code == FallbackCodes.NO_TRANSLATION:
```

kitsune/wiki/views.py
```diff
@@ -1634,7 +1612,17 @@ def recent_revisions(request):
 def what_links_here(request, document_slug):
     """List all documents that link to a document."""
     locale = request.GET.get("locale", request.LANGUAGE_CODE)
-    doc = get_visible_document_or_404(request.user, locale=locale, slug=document_slug)
+
+    doc, fallback_code = get_visible_document_or_fallback_or_404(
+        request.user, locale, document_slug
+    )
+
+    if fallback_code is FallbackCodes.TRANSLATION_AT_DIFFERENT_SLUG:
```
akatsoulas, line 1620:
```suggestion
    if fallback_code == FallbackCodes.TRANSLATION_AT_DIFFERENT_SLUG:
```

kitsune/wiki/views.py
```diff
@@ -1634,7 +1612,17 @@ def recent_revisions(request):
...
+        url = reverse("wiki.what_links_here", args=[doc.slug], locale=locale)
+        return HttpResponseRedirect(url)
+
+    if fallback_code is FallbackCodes.NO_TRANSLATION:
```
akatsoulas, line 1624:
```suggestion
    if fallback_code == FallbackCodes.NO_TRANSLATION:
```

kitsune/wiki/views.py
akatsoulas, line 1620: This pattern appears elsewhere too. I wonder if it makes sense to have a helper function here which will accept a translation as an arg and redirect or raise the appropriate HTTP errors

akatsoulas (changes requested):
I didn't spend much time on this iteration on the logic but rather on the code. I will have another pass focusing on how this is solved on my second pass.

kitsune/wiki/utils.py
escattone: @akatsoulas Good point. I got rid of `get_visible_document_or_fallback_or_404()` and the fallback codes, and refactored `get_visible_document_or_404()` to handle the additional cases. Let's see how you feel about that.

kitsune/wiki/utils.py
```diff
@@ -178,7 +179,42 @@ def get_featured_articles(product=None, locale=settings.WIKI_DEFAULT_LANGUAGE):
 
 
 def get_visible_document_or_404(user, **kwargs):
-    return get_object_or_404(Document.objects.visible(user, **kwargs))
+    """
```
akatsoulas, line 182: If you prefer to have `kwargs` here instead of default values, let's define in the docstring the standard/default args this function awaits

akatsoulas (approved):
r+wc :+1:

akatsoulas:
Please remember to squash the changes into a single commit

kitsune/wiki/utils.py
```diff
@@ -178,7 +179,42 @@ def get_featured_articles(product=None, locale=settings.WIKI_DEFAULT_LANGUAGE):
...
+    return_parent_if_no_translation = kwargs.pop("return_parent_if_no_translation", False)
+
+    try:
+        return Document.objects.get_visible(user, **kwargs)
+    except Document.DoesNotExist:
+        pass
+
+    if (
+        not look_for_translation_via_parent
+        or not (locale := kwargs.get("locale"))
```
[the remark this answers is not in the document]
escattone: I think that would break the logic, since I'm checking that either `locale` is not provided or it's set to the value of `settings.WIKI_DEFAULT_LANGUAGE`. If I collapsed the lines, the case when `locale` is not provided would never trigger.

kitsune/wiki/utils.py
escattone: I was on the fence about what to do, so if you have a preference, we can do that. On further reflection, I think default values for `look_for_translation_via_parent` and `return_parent_if_no_translation` are probably clearer.

kitsune/wiki/utils.py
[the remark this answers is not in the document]
akatsoulas: If it's not provided, it would default to an empty string which will not be equal to the default locale. In case it is provided, it will check against the default which is the second part. This is very nit. I am just curious about it to figure out what I am missing

kitsune/wiki/utils.py
akatsoulas: Agreed about the default values

kitsune/wiki/utils.py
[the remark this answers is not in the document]
escattone: > If it's not provided, it would default to an empty string which will not be equal to the default locale.

That wouldn't match the current logic, since it would be `False` and not trigger the clause and raise an Http404 exception. So to match the logic, it would have to instead be:

```python
kwargs.get("locale", settings.WIKI_DEFAULT_LANGUAGE) == settings.WIKI_DEFAULT_LANGUAGE
```

so that's why I kept it as I had it.

PR #5195, opened 2022-08-02
add pre-rendering of svelte templates to run-bootstrap.sh

escattone:


bin/run-bootstrap.sh
```diff
@@ -4,5 +4,7 @@ set -ex
 
 # install Node dependencies
 npm run development && npm run webpack:build
+# pre-render the svelte-based templates
+python manage.py compilejsi18n && npm run webpack:build:pre-render && python manage.py collectstatic --noinput
```
akatsoulas, line 8: Why do we need `compliejsi18n` and `collectstatic` here?

akatsoulas: even if we do need them, let's add the compilejsi18n first, then group together all the npm commands and leave the collectstatic to run after npm

akatsoulas (approved):
r+wc

bin/run-bootstrap.sh
escattone: @akatsoulas Without running `compilejsi18n` prior, the `npm run webpack:build:pre-render` will fail, but perhaps we can remove the `collectstatic` step? I'll look into that.

PR #5197, opened 2022-08-04
disable short URL generation cron on stage

escattone:
mozilla/sumo#1080


escattone:
@akatsoulas Adding you as a reviewer just as an FYI. I decided to resolve mozilla/sumo#1080 this way -- instead of removing the Bitly secret for stage -- because it just disables the cron task without also preventing us from testing Bitly from stage altogether.

PR #5199, opened 2022-08-10
remove celery-flower from local docker config

escattone:
Remove `celery-flower` from local Docker configuration. It currently fails, and I don't see it as useful enough to fix.


smithellis:
Looks good @escattone

smithellis (commented):
I see I didn't put my review in the right place.  This looks fine to me - and since it is now merged, doubly so.

PR #5214, opened 2022-09-06
Fix broken markup

akatsoulas:


smithellis (changes requested):
Should the ln 560 select tag read : `<select></select>` ?  This change makes it `<select><select/>` .Otherwise  I approve and thank you!

smithellis (approved):
r+wc

PR #5223, opened 2022-09-12
add pylint-django to our dev dependencies

escattone:
We [load the `pylint_django` plugin](https://github.com/mozilla/kitsune/blob/98b2452c8c433acf720552c8eb5f05263a4ac182/.vscode/settings.json#L14) when running `pylint` within the `vscode` container, but I suspect it's been silently failing, because we don't actually install the `pylint-django` package. When moving to the latest version of `pylint`, the load failure is reported.

Assuming we want to continue using `pylint`, this PR fixes the loading of the `pylint_django` plugin.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -108,6 +108,7 @@ recommonmark = "^0.7.1"
 sphinx-rtd-theme = "^1.0.0"
 sphinxcontrib-httpdomain = "^1.8.0"
 tabulate = "^0.8.9"
+pylint-django = "^2.5.3"
 
 
 [build-system]
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -11,7 +11,8 @@
     "python.linting.pylintArgs": [
         "--disable=C0111",
         "--enable=W0614",
-        "--load-plugins=pylint_django"
+        "--load-plugins=pylint_django",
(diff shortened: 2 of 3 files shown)


pyproject.toml
```diff
@@ -110,6 +110,9 @@ sphinxcontrib-httpdomain = "^1.8.0"
 tabulate = "^0.8.9"
 
 
+[tool.poetry.group.dev.dependencies]
```
akatsoulas, line 113: We are using the `dev-dependencies` notation which will be slowly deprecated in favor of `dev.dependencies`. I am in favor of the new notation but we need to migrate the existing one otherwise it's failing with the following error
```
The Poetry configuration is invalid:
#0 0.589     - Additional properties are not allowed ('group' was unexpected)
```

From the poetry documentation:
```
Poetry will slowly transition away from the dev-dependencies notation which will soon be deprecated, so it’s advised to migrate your existing development dependencies to the new group notation.
```

akatsoulas (approved):
r+wc assuming a successful build

pyproject.toml
escattone: Sorry, I used the wrong version of `poetry` and it used the new format. I'll fix all of this.

PR #5230, opened 2022-09-22
Bump mem request to 1Gi for prod, 512Mb for stage

akatsoulas:


escattone (approved):
r+ Based on https://kubernetes.io/docs/concepts/configuration/manage-resources-containers/#how-pods-with-resource-limits-are-run, it seems that this change could help spread-out the pods when scheduling.

PR #5247, opened 2022-10-10
jquery regressions - .proxy and this.off corrections

smithellis:
This allows the code the run and populate the editor toolbar.

### DIFF
--- a/kitsune/sumo/static/sumo/js/libs/jquery.autoresize.js
+++ b/kitsune/sumo/static/sumo/js/libs/jquery.autoresize.js
@@ -88,12 +88,12 @@
 
 		bind: function() {
 
-			var check = $.proxy(function(){
+			var check = $.bind(function(){
 				this.check();
 				return true;
 			}, this);
 
-			this.off();
+			this.off && this.off();
 
 			this.el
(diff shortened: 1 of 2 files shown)


kitsune/sumo/static/sumo/js/libs/jquery.autoresize.js
```diff
@@ -88,12 +88,12 @@
 
 		bind: function() {
 
-			var check = $.proxy(function(){
+			var check = $.bind(function(){
 				this.check();
 				return true;
 			}, this);
```
escattone, line 94: It seems that `Function.prototype.bind()` should be called like this instead?
```suggestion
			var check = (function(){
				this.check();
				return true;
			}).bind(this);
```

escattone (changes requested):
Other than my one suggested change, this looks good and fixes https://github.com/mozilla/sumo/issues/1107.

kitsune/sumo/static/sumo/js/libs/jquery.autoresize.js
smithellis: I believe you are correct.

akatsoulas (approved):

PR #5274, opened 2022-10-24
adjust pipeline to handle svelte pre-render variations

escattone:
This PR adjusts the Circle-CI pipeline and the k8s deployment script to build and deploy separate Docker images into the stage and prod environments. With the dependency of the `svelte` pre-rendering on the value of `SUMO_URL`, some of the pre-rendered JS, CSS, and HTML files can vary depending on the value of `SUMO_URL`, which will vary depending on the target environment. That means we can no longer deploy or upload static assets from a single Docker image into the stage and prod (and dev) environments. We must build, deploy, and upload static assets from an image specifically built for each target environment.

### DIFF
--- a/.circleci/build.html
+++ b/.circleci/build.html
@@ -18,6 +18,7 @@
   <body>
     build completed!<br />
     hash: <span class="hash"></span><br />
+    image: mozilla/kitsune:stage-<span class="hash"></span>
     image: mozilla/kitsune:prod-<span class="hash"></span>
     <script>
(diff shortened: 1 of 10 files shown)


escattone:
I'm closing this PR in favor of https://github.com/mozilla/kitsune/pull/5275 for now. It's may be worth saving this work however. In the future, we may need to specify settings that vary between target environments, and since we pre-render our `svelte` components -- which I think we have to do since our `svelte` components are currently embedded within Jinja2 templates -- we may want to do something like what this PR does.

PR #5275, opened 2022-10-25
use relative urls to avoid SUMO_URL consequences

escattone:
By using relative URLs, we guarantee that we use the same origin that the page was loaded from and we avoid:
- building separate Docker images for each target environment (e.g., stage, prod) -- at least for now, since we don't yet have any other settings that vary between the target environments
- the need to provide the extra code and configuration required to set `SUMO_URL` for each target environment, including local development -- for example, like parts of https://github.com/mozilla/kitsune/pull/5274

### DIFF
--- a/svelte/contribute/Contribute.svelte
+++ b/svelte/contribute/Contribute.svelte
@@ -5,11 +5,7 @@
     import Linkable from "./Linkable.svelte";
     import { gettext } from "../lib/utils";
     import { createClient, setContextClient } from "@urql/svelte";
-    import {
-        SUMO_URL,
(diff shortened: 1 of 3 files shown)


escattone:
FYI, the code in this PR was deployed to [stage](https://support.allizom.org/en-US/contribute) at around 3pm PST.

akatsoulas (approved):
That's perfect! :100: :rocket:

escattone:
Everything from this PR was included in https://github.com/mozilla/kitsune/pull/5278 except for the typo fix, and since https://github.com/mozilla/kitsune/pull/5278 was merged, we can close this one.

PR #5279, opened 2022-10-26
replace django-authority with django-guardian

escattone:
https://github.com/mozilla/sumo/issues/1123

This PR uses `django-guardian` to replace `django-authority`.

# Reviewer's Notes
- Since `django-guardian` builds on Django's standard interface for implementing object-level permissions, the check for an object-level permission is the usual `user.has_perm()` but with the object added as an additional argument, for example `user.has_perm("forums.move_forum_thread", forum)`.
- Object-level permissions in SUMO are currently used only for forums (instances of `Forum`).

- ##### Restricted Forums
  - The standard way to use permissions within an application is to restrict access to a class of objects (or a specific object in the case of object-level permissions), and only users or groups with the necessary permission can gain access.
  - With forums this is true for editing, deleting, locking, and moving threads, making threads sticky, as well as editing and deleting thread posts. Currently, every forum will check if a user has one of those permissions for that forum in order for the user to perform that permission's respective action.
  - It is not true when restricting access to view or post-in forums. Most forums are free from these two restrictions, so any user can view or post without permission, but a few forums are restricted in viewing or posting or both.
  - Currently, using `django-authority`, a restricted forum is identified by having either a `forums_forum.view_in_forum` or `forums_forum.post_in_forum` permission or both defined on itself. The `forums_forum.view_in_forum` permission restricts viewing, and the `forums_forum.post_in_forum` permission restricts posting.
  - So, in other words, every time a user wants to view or post in a forum, the forum first has to be checked to see if it's restricted in viewing or posting, in order to determine whether or not the user should be checked for the proper permission. This means a database query for every check, which is wasteful given that nearly every forum is unrestricted.
  - With this PR, this is no longer the case. Each forum has a `restrict_viewing` attribute that defaults to `False` but can be set to `True` for forums that restrict viewing, and similarly, a `restrict_posting` attribute. Only if the attribute is set to `True` will the user be checked for the proper permission. This PR contains a standard database migration to create the two new columns and their indexes on the forum table, and also a data migration to ensure these attributes will be set properly for every forum based on the existing `django-authority` permissions.
- Many of the forum views were using the `permission_required_or_403` or `has_perm_or_owns_or_403` decorator. Not only were they un-intuitive, but they duplicated the query for the forum object. They have been removed, and replaced with more readable and efficient code within the view function itself. Also, these decorators sometimes returned a 403 when the user wasn't even permitted to view the forum, "leaking" information that a restricted forum existed. With this PR, a 404 will be returned in those cases.
- `django-authority` remains as a package as well as an installed app. It and its database tables will be removed in a later PR, but only after all databases (local, stage, prod) have been successfully migrated with this PR.
- Note that the `kitsune.access.utils.has_perm` function, which is used in all of the forum views and templates (as well as the `Forum` model), checks both the standard "global" permission as well as the object-level permission.
- **I think the global permissions will be useful for giving groups like `Forum Moderators` power over all current and future forums without having to manually add all of the permissions to the group for each new forum. This PR doesn't do that, but it does enable us to do it if we wish.**
- I think the `has_perm_or_owns` function obfuscated more than it helped. I found the code more readable to just replace it with its underlying two checks: whether or not the user was the author/creator, and a `has_perm()` call.


smithellis (approved):
This all makes sense and seems to work as designed.

PR #5281, opened 2022-10-27
Incresed width and added overflow to handle long usernames

smithellis:
Improves username length issue, however this page is a good candidate for some further CSS improvements...should resolve 1091.


escattone (approved):
r+

PR #5282, opened 2022-10-28
Replace es7 and ES7 references with es and ES

smithellis:


akatsoulas (approved):
This looks good! Let's make sure that we have an issue open for our SREs to update the env vars in GCP too

PR #5290, opened 2022-11-04
remove obsolete messages from po files

escattone:
https://github.com/mozilla/sumo/issues/1135

This is a companion PR for https://github.com/mozilla-l10n/sumo-l10n/pull/107.

This PR changes the `./manage.py merge` command so that it excludes obsolete messages from the updated PO files.

[Obsolete messages](https://www.gnu.org/software/gettext/manual/html_node/Obsolete-Entries.html) are no-longer used messages in PO files which are commented-out with `#~` and only exist for the convenience of being able to resurrect their translation if needed again sometime in the future. We should **not** be including obsolete messages in the PO files, since we have all of the message history in `git`, but also because they can and do trigger fatal duplicate message errors during compilation, since the `msgfmt` command treats them the same as active messages.

This PR also disables fuzzy matching when updating (merging) the PO files (via `--no-fuzzy-matching`). We shouldn't be using fuzzy matching. The translators in [our Pontoon community](https://pontoon.mozilla.org/projects/sumo/) will do a much better job than any fuzzy guesses at translations that the merge command attempts when adding new strings.

### DIFF
--- a/kitsune/sumo/management/commands/merge.py
+++ b/kitsune/sumo/management/commands/merge.py
@@ -32,7 +32,11 @@ class Command(BaseCommand):
                 continue
 
             if Path(f"locale/{locale}").is_dir():
-                sub_cmd = "update"
+                command = [
+                    "pybabel",
+                    "update",
+                    "--ignore-obsolete",
+                ]
             elif not is_supported_for_init(locale):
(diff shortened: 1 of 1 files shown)


escattone:
Closing in favor of https://github.com/mozilla/kitsune/pull/5292

PR #5299, opened 2022-11-15
Bump mozilla-django-oidc to 3

smithellis:


smithellis:
#5300 was opened to fix the flake8 issue that is causing build fails.

akatsoulas (approved):

PR #5300, opened 2022-11-15
flake8 move to github in pre-commit

smithellis:
Builds are breaking now in CI because flake8 moved from gitlab to github ; updating out pre-commit to handle the shift.


smithellis:
Hm.  Betting my CI will fail because of the pre-commit issue this PR is fixing.  Inception!

PR #5305, opened 2022-11-18
Remove django-adminplus

smithellis:
django-adminplus lacks support, so we are removing it.

Remove adminplus refs in monkeypatch
Remove kadmin and adminplus from settings


smithellis:
Should I just remove that entire directory, @akatsoulas ?

akatsoulas (approved):
This looks good!

PR #5317, opened 2022-12-02
add GCS as a media backend

escattone:
This PR is work by @ahoneiser. I'm just creating this PR for him for review purposes.

### DIFF
--- a/kitsune/settings.py
+++ b/kitsune/settings.py
@@ -784,7 +784,7 @@ DEFAULT_FILE_STORAGE = "kitsune.upload.storage.RenameFileStorage"
 
 # GCP storage settings
 GS_BUCKET_NAME = config("GS_BUCKET_NAME", default="")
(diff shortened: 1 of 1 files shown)


kitsune/upload/storage.py
```diff
@@ -6,10 +6,19 @@
 from django.conf import settings
 from django.core.files.storage import FileSystemStorage
 
+from storages.backends.gcloud import GoogleCloudStorage
 from storages.backends.s3boto3 import S3Boto3Storage
 
+DjangoStorage = None
```
escattone, line 12: Nit. Since the `else` clause takes care of the default case, we could drop this if you'd like.

escattone (commented):
r+

kitsune/upload/storage.py
ahoneiser: thanks @escattone! removed that line :+1: also, removed `GS_LOCATION`, it's not required for the GCS backend to work :slightly_smiling_face:

PR #5325, opened 2022-12-07
Fix migrations after Django4

akatsoulas:


akatsoulas:
Merging this to fix the migrations in the `wagtail-cms` branch - tested it locally

PR #5320, opened 2022-12-07
SEO adjustments for KB forums and user-related links

escattone:
mozilla/sumo#1131

This PR does the following:
- adds `rel="nofollow"` to all user links (NOTE: I didn't consider user links as "content" so I didn't use `ugc`)
- disallows crawling of the KB forums, which are user-generated discussions of KB articles, via `robots.txt`
- adds `rel="ugc nofollow"` to all KB forum links
- adds tests
- adds a doc for notes and policies related to SEO

PR #5331, opened 2022-12-10
add rel='nofollow' to user links

escattone:
mozilla/sumo#1153

This PR adds `rel="nofollow"` to all user links.

### DIFF
--- a/kitsune/sumo/jinja2/layout/breadcrumbs.html
+++ b/kitsune/sumo/jinja2/layout/breadcrumbs.html
@@ -3,7 +3,12 @@
   <ol id="breadcrumbs" class="breadcrumbs--list">
   {% for target, label in breadcrumbs %}
     {% if target %}
-      <li><a href="{{ target }}">{{ label|truncate(length=30) }}</a></li>
+      {% if target is string %}
+        <li><a href="{{ target }}">{{ label|truncate(length=30) }}</a></li>
+      {% else %}
+        {% set target, extra = target %}
+        <li><a {{ extra|safe }} href="{{ target }}">{{ label|truncate(length=30) }}</a></li>
+      {% endif %}
     {% else %}
       <li>{{ label|truncate(length=40) }}</li>
     {% endif %}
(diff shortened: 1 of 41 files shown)


kitsune/sumo/jinja2/layout/breadcrumbs.html
```diff
@@ -3,7 +3,12 @@
   <ol id="breadcrumbs" class="breadcrumbs--list">
   {% for target, label in breadcrumbs %}
     {% if target %}
-      <li><a href="{{ target }}">{{ label|truncate(length=30) }}</a></li>
+      {% if target is string %}
```
akatsoulas, line 6: Why are you introducing this change here? Even if it is a valid change, shall we include it in a new PR since it does not seem related to the the `nofollow` directive.

akatsoulas (commented):
Overall this is looking good but I do have a couple of comments:

* This PR seems very similar to #5320. They do share some common functionality and more or less there are about the same thing. I believe we should tackle everything in a single PR and move the work from here to #5320 and clean that PR up
* Why did you remove `ugc`?
* This PR applies the `nofollow` directive to user related links. Though there is some overlap with the use of the same directive in base pages. For example in the kbforums in PR 5320 you add the `nofollow` which is also applied in the user links of `discussions.html`. I do understand that you are focusing here on the user links but I am also wondering if there is a way to have some guidelines/rules on how we apply this. Eg first on the base.html of each app if applicable, second to specific pages if we do not want it on the base.html etc. Some kind of policy
* We need to document that policy in the documentation

kitsune/sumo/jinja2/layout/breadcrumbs.html
escattone: This is needed to support the new breadcrumbs format, which allows extra attributes (in this case `rel="nofollow"`) to be attached to breadcrumbs specified by sub-classes, for example, `kitsune/users/jinja2/users/edit_profile.html` below.

PR #5336, opened 2022-12-14
Remove leading / from messages urls

smithellis:
Drop the / slash form messages urls


escattone (commented):
If you remove the leading `/`'s from here, just remember to add a trailing `/` to the root route if it doesn't already have one, in this case [`r"^messages/"`](https://github.com/mozilla/kitsune/blob/afe3f1ae30d2ee5ddb5ab668be70da2b381f0cd2/kitsune/urls.py#L30). If you don't, things like `reverse("messages.outbox")` will return `'/messagessent'` instead of `'/messages/sent'`.

akatsoulas (approved):

PR #5348, opened 2022-12-19
Improve error message for duplicate e-mail

yecril71pl:
Use the term <INS >e-mail address</INS > for clarity!
- *e-mail* means:
  1. a form for a letter adapted for being exchanged electronically,
  2. the mechanism used for exchanging the former.
- *e-mail address* means a part of an *e-mail* containing information about the sender or a recipient.


akatsoulas (approved):
Thanks @yecril71pl !

PR #5360, opened 2023-01-03
Remove leading / from questions urls

smithellis:


escattone (approved):
You did the right thing on the test! Looks great. r+

PR #5361, opened 2023-01-05
Simplify circleci config

akatsoulas:


escattone (commented):
With this change we're no longer building the `prod` target of the Dockerfile, so we're no longer running (and therefore testing) `npm run webpack:build:pre-render`. It seems like it would be nice to somehow continue to run `npm run webpack:build:pre-render` as a test to ensure that it runs without failure. We can merge this as-is for now, but I think we should consider finding a nice way to add that back into the CI test flow.

escattone:
FYI, I submitted #5362 to address my concerns.

escattone (approved):
r+

PR #5375, opened 2023-01-21
fix update_user method to work without a request

escattone:
mozilla/sumo#1177

It seems that our processing of FxA user profile events has been failing on every event. This PR resolves that.

### DIFF
--- a/kitsune/users/auth.py
+++ b/kitsune/users/auth.py
@@ -79,9 +79,11 @@ class FXAAuthBackend(OIDCAuthenticationBackend):
 
     def update_contributor_status(self, profile):
         """Register user as contributor."""
-        if contribution_area := self.request.session.get("contributor"):
+        # The request attribute might not be set.
+        request = getattr(self, "request", None)
+        if request and (contribution_area := request.session.get("contributor")):
             add_to_contributors(profile.user, profile.locale, contribution_area)
-            del self.request.session["contributor"]
+            del request.session["contributor"]
 
     def create_user(self, claims):
         """Override create user method to mark the profile as migrated."""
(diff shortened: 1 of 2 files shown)


kitsune/users/auth.py
```diff
@@ -195,30 +197,36 @@ def update_user(self, user, claims):
...
         # There is a change in the email in Firefox Accounts. Let's update user's email
         # unless we have a superuser
         if user.email != email and not user.is_staff:
             if User.objects.exclude(id=user.id).filter(email=email).exists():
-                msg = _(
-                    "The e-mail address used with this Firefox Account is already "
-                    "linked in another profile."
-                )
-                messages.error(self.request, msg)
+                if request:
```
akatsoulas, line 224:
```suggestion

```

kitsune/users/auth.py
```diff
@@ -195,30 +197,36 @@ def update_user(self, user, claims):
...
+            if request:
+                # This is the first time an existing user is using FxA. Redirect to profile edit
+                # in case the user wants to update any settings.
+                request.session["oidc_login_next"] = reverse("users.edit_my_profile")
+                messages.info(request, "fxa_notification_updated")
 
         # There is a change in the email in Firefox Accounts. Let's update user's email
         # unless we have a superuser
         if user.email != email and not user.is_staff:
             if User.objects.exclude(id=user.id).filter(email=email).exists():
```
akatsoulas, line 223:
```suggestion
            if request and User.objects.exclude(id=user.id).filter(email=email).exists():
```

kitsune/users/auth.py
```diff
@@ -195,30 +197,36 @@ def update_user(self, user, claims):
         # Check if the user has active subscriptions
         subscriptions = claims.get("subscriptions", [])
 
+        # The request attribute might not be set.
+        request = getattr(self, "request", None)
+
         if not profile.is_fxa_migrated:
```
akatsoulas, line 203: You can move the check for the existence of the request object here. This error is triggered from events that are coming from FxA thus it makes sense to assume that the profiles are migrated already. This way you can avoid the 2 checks under this if statement

```python
if (request := getattr(self, "request", None)) and not profile.is_fxa_migrated:
```

akatsoulas: Another way to go about this would be to check for the absence of the request, do the relevant tasks and then return.

```python
if not request:
   we_are_doing_all_the_common_tasks_for_both_the_login_flow_and_the_celery_task()
   return
we_are_now_only_in_the_login_flow_and_we_have_the_request_object_steps

akatsoulas (approved):
r+wc nice catch!

kitsune/users/auth.py
escattone: I decided against making this change because we still want to check for an existing user having that email even if request is `None`.

PR #5362, opened 2023-01-05
strengthen test target in dockerfile

escattone:
mozilla/sumo#1167

PR #5381, opened 2023-01-27
Add l10n process to bootstrap script

smithellis:
Unit tests now depend on a functional locale folder (see [this pull request](https://github.com/mozilla/kitsune/pull/5362 )). Rather than modify documentation for local development, we can add the step to `run-bootstrap.sh` so the process can be part of the make automation.


escattone (approved):
r+

PR #5389, opened 2023-02-06
remove patch no longer needed for django-activity-stream

escattone:
mozilla/sumo#1143

This PR removes a patch that is no longer needed now that [version `1.4.2` of `django-activity-stream`](https://github.com/justquick/django-activity-stream/blob/890d13fd23d5c165a9f28841bd8df9b92f6ad662/actstream/registry.py#L55) has been released and incorporated into our codebase.

### DIFF
--- a/kitsune/questions/apps.py
+++ b/kitsune/questions/apps.py
@@ -1,16 +1,6 @@
 from django.apps import AppConfig
 
 
-def is_installed(model_class):
-    """
-    This is a fix for django-activity-stream 1.4.0 that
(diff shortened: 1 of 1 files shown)


escattone:
@akatsoulas I was tempted to just merge this without review since it's so simple and I've confirmed that version `1.4.2` of `django-activity-stream` contains the exact same code as our temporary patch, but decided against that. I think you could safely review and merge this before the release tomorrow if you'd like.

akatsoulas (approved):

PR #5392, opened 2023-02-10
remove React dependency within JS tests

escattone:
mozilla/sumo#1185

This PR removes the dependency on React within the JS tests. React was used only within our tests, and only for the convenience of using JSX to create HTML for each test. Whatever purpose it served originally, no longer exists, and it has become a unnecessary complication that's best removed. With this PR, it is replaced with simple JQuery (`$("body").empty().html(...)`).

Some notes:
- This PR only changes the way the HTML is created for the tests. Nothing else has been changed, except for the case of the `showAndHide` tests (see note below).
- The `showAndHide` tests are the only exception. I had to add `showFor.initShowFuncs();` to its `beforeEach()` set-up function. To understand why, you first have to understand a misleading, subtle issue in the React-based `kitsune/sumo/static/sumo/js/tests/showfortests.js`. In the React-based version, the same DOM elements are re-used for every test case, because the HTML created for each test case is **never** unmounted. This means that any DOM changes (like changing a data attribute of an element) persist from test to test, and suddenly the order of the tests can make a difference. The only reason why the `showAndHide` tests worked in the React-based version is because the `initShowFuncs` tests were run just prior, and the DOM changes they introduced with `showFor.initShowFuncs();` persisted into the run of the `showAndHide` tests, which depended on them. In my new version of `kitsune/sumo/static/sumo/js/tests/showfortests.js`, the DOM is created fresh for each test (because of the `$('body').empty()`), so `showFor.initShowFuncs();` has to be explicitly called before the `showAndHide` tests (which is what happens in real usage anyway, because `initShowFuncs()` is called in the `showFor` constructor, which is bypassed in the tests).
- One of the primary tasks, when creating this PR, was translating React's JSX into plain, vanilla HTML. So, for example, attributes like `className` and `defaultValue` become `class` and `value`, and `<input ... />` becomes `<input ... >`.
- Removing `react` from our dev dependencies also meant that we could remove `@babel/plugin-transform-react-jsx` as well.

### DIFF
--- a/kitsune/sumo/static/sumo/js/tests/ajaxpreviewtests.js
+++ b/kitsune/sumo/static/sumo/js/tests/ajaxpreviewtests.js
@@ -1,4 +1,3 @@
-import React from 'react';
 import {expect} from 'chai';
 import sinon from 'sinon';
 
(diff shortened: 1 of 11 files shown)


akatsoulas (approved):
This is amazing! Great work1

PR #5394, opened 2023-02-13
Announcement check

akatsoulas:


kitsune/announcements/templatetags/jinja_helpers.py
```diff
@@ -4,5 +4,8 @@
 
 
 @library.global_function
-def get_announcements():
+def get_announcements(request):
+    user = request.user if request.user.is_authenticated else None
+    if user:
+        return Announcement.get_for_groups(user.groups.values_list("id", flat=True))
```
escattone, line 10: The group ids passed here to `Announcement.get_for_groups` are actually a `QuerySet` rather than a `list`, which works fine of course, but do we need to add `QuerySet | list[int]` to the type hint for the `group_ids` argument of `Announcement.get_for_groups`?

escattone (approved):
r+ wc

PR #5397, opened 2023-02-15
add mypy type-checking to pre-commit

escattone:
mozilla/sumo#1120

This PR adds type checking via `mypy` to our `pre-commit` flow, which is also run as part of the linting step of the CI flow.

## Notes and Comments
- The default option passed to the `mypy` `pre-commit` hook is `--ignore-missing-imports`.
- I limited type-checking to all of the Python files under the`kitsune` folder (excluding the Django database migration files). Do we also want to run this on our Python scripts under the `scripts` folder? It didn't seem worth it, but maybe I'm wrong about that.
- The bulk of the changes in this PR are simply to get `mypy` passing on all of our Python files. There's a reason for that. When running `mypy` in `pre-commit`, it'll be called with any Python files that have changed, but it also checks each of their imported files as well. This can be turned-off (for example, one way is to use `--follow-imports=silent`) but `mypy` strongly recommends using the default setting. Since so many extra files get pulled into the type-checking, in the end it only makes sense to start by getting `mypy` to run on all of our Python files.
- The `additional_dependencies` (type-hint stubs) I added in the `.pre-commit-config.yaml` file for the `mypy` hook were the ones recommended by `mypy` itself when it was run on our Python files. I tried adding `django-stubs` as well, but it's not yet approved for `mypy` version `1.0.0` or greater. I think that would be good to add in the future when possible.
- Since we're not yet enabling the `--check-untyped-defs` option of `mypy`, some type errors won't be detected as you might expect. For example, although we provide type hints for the [Announcement.get_for_groups method](https://github.com/mozilla/kitsune/blob/f26ad9a75f27717f5a3facba9130a0ce14c9f2bc/kitsune/announcements/models.py#L67), if we were to make a mistake of using `Announcement.get_for_groups(23)` instead of [what we have currently](https://github.com/mozilla/kitsune/blob/f26ad9a75f27717f5a3facba9130a0ce14c9f2bc/kitsune/announcements/templatetags/jinja_helpers.py#L10), that type error won't be detected because it won't be checked. It's not checked because the `get_announcements()` function definition doesn't have any type hints. If we want to turn on `--check-untyped-defs`, we'd have to fix another 143 errors.
- I'm not convinced this work is worth the effort. I'm not a fan of typing. I think the extra time and effort is much better spent on improving tests. I'm willing to give it a try though.
- I added an ADR to the docs for type checking. Even though I'm not a fan, I still tried to explain the benefits and why we're using it as best I could.

### DIFF
--- a/kitsune/announcements/admin.py
+++ b/kitsune/announcements/admin.py
@@ -19,7 +19,7 @@ class AnnouncementAdmin(admin.ModelAdmin):
             return "Yes (always)"
         return ""
 
-    is_visible.short_description = "Visible?"
+    is_visible.short_description = "Visible?"  # type: ignore
 
     def save_model(self, request, obj, form, change):
         if not obj.pk:
--- a/kitsune/dashboards/readouts.py
+++ b/kitsune/dashboards/readouts.py
@@ -486,7 +486,7 @@ class Readout(object):
     column3_label = _lazy("Visits in last 30 days")
     column4_label = _lazy("Status")
     modes = [(MOST_VIEWED, _lazy("Most Viewed")), (MOST_RECENT, _lazy("Most Recent"))]
-    default_mode = MOST_VIEWED
+    default_mode: int | None = MOST_VIEWED
 
     def __init__(self, request, locale=None, mode=None, product=None):
         """Take request so the template can use contextual macros that need it.
@@ -1216,13 +1216,13 @@ class CannedResponsesReadout(Readout):
 
 # L10n Dashboard tables that have their own whole-page views:
 L10N_READOUTS = OrderedDict(
-    (t.slug, t)
+    (t.slug, t)  # type: ignore
     for t in [MostVisitedTranslationsReadout, TemplateTranslationsReadout, UnreviewedReadout]
 )
 
 # Contributors ones:
 CONTRIBUTOR_READOUTS = OrderedDict(
-    (t.slug, t)
+    (t.slug, t)  # type: ignore
     for t in [
         MostVisitedDefaultLanguageReadout,
         TemplateReadout,
@@ -1240,7 +1240,7 @@ READOUTS = L10N_READOUTS.copy()
 READOUTS.update(CONTRIBUTOR_READOUTS)
 
 GROUP_L10N_READOUTS = OrderedDict(
-    (t.slug, t) for t in [MostVisitedTranslationsReadout, UnreviewedReadout]
+    (t.slug, t) for t in [MostVisitedTranslationsReadout, UnreviewedReadout]  # type: ignore
 )
 # English group locale is the same as l10n dashboard.
 GROUP_CONTRIBUTOR_READOUTS = CONTRIBUTOR_READOUTS
--- a/kitsune/search/base.py
+++ b/kitsune/search/base.py
@@ -2,7 +2,7 @@ from abc import ABC, abstractmethod
 from dataclasses import dataclass
 from dataclasses import field as dfield
 from datetime import datetime
-from typing import Union, overload
+from typing import Union, Self, overload
 
 from django.conf import settings
 from django.core.paginator import EmptyPage, PageNotAnInteger
@@ -29,6 +29,11 @@ from kitsune.search.parser.tokens import TermToken
 
 class SumoDocument(DSLDocument):
     """Base class with common methods for all the different documents."""
+
+    # Controls if a document should be indexed or updated in ES.
+    #   True: An update action will be performed in ES.
+    #   False: An index action will be performed in ES.
+    update_document = False
 
     indexed_on = field.Date()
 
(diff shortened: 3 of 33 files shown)


kitsune/announcements/admin.py
```diff
@@ -19,7 +19,7 @@ def is_visible(self, obj):
             return "Yes (always)"
         return ""
 
-    is_visible.short_description = "Visible?"
+    is_visible.short_description = "Visible?"  # type: ignore
```
escattone, line 22: This is required due to an annoying aspect (a flaw, really) of `mypy` where it flags as an error the setting of an attribute on any object if that attribute doesn't already exist.

kitsune/dashboards/readouts.py
```diff
@@ -1216,13 +1216,13 @@ def _format_row(self, row):
 
 # L10n Dashboard tables that have their own whole-page views:
 L10N_READOUTS = OrderedDict(
-    (t.slug, t)
+    (t.slug, t)  # type: ignore
```
escattone, line 1219: Another strange thing in `mypy`. It flags `t.slug` as an attribute error thinking that the `slug` attribute doesn't exist, even though it's defined on all of classes provided.

kitsune/wiki/models.py
```diff
@@ -990,7 +990,7 @@ def previous(self):
             return None
 
 
-class DraftRevision(ModelBase, AbstractRevision):
+class DraftRevision(ModelBase, AbstractRevision):  # type: ignore
```
escattone, line 993: This is required due to another strange `mypy` behavior. It complains that the `ModelBase` and `AbstractRevision` classes have incompatible `Meta` class attributes, when in fact they're both exactly the same. It's probably due to Django's special handling of `Meta` class attributes (it sets `Meta.abstract` to `False` for all sub-classes).

kitsune/search/search.py
```diff
@@ -323,7 +323,7 @@ class CompoundSearch(SumoSearch):
     _children: list[SumoSearch] = dfield(default_factory=list, init=False)
     _parse_query: bool = True
 
-    @property
+    @property  # type: ignore
```
escattone, line 326: Even though this code works perfectly, `mypy` complains that `Dataclass attribute may only be overridden by another attribute`, so that's why the `# type: ignore` is here.

kitsune/search/documents.py
```diff
@@ -38,12 +40,6 @@ class WikiDocument(SumoDocument):
     class Index:
         pass
 
-    @classmethod
-    @property
-    def update_document(cls):
```
akatsoulas, line 43: This change will make the `index_object` to fail. If I recall correctly it operates on class types and not on instantiated objects.

kitsune/search/base.py
```diff
@@ -117,16 +122,6 @@ def alias_points_at(cls, alias):
 
         return aliased_indices[0] if aliased_indices else None
 
-    @classmethod
-    @property
-    def update_document(cls):
```
akatsoulas, line 122: This change will make the index_object to fail. If I recall correctly it operates on class types and not on instantiated objects.

kitsune/search/documents.py
akatsoulas: Why are we making these changes?

kitsune/settings.py
```diff
@@ -983,26 +983,26 @@ def show_toolbar_callback(*args):
 
 # CORS Setup
 CORS_ALLOW_ALL_ORIGINS = True
-CORS_URLS_REGEX = [
```
akatsoulas, line 986: Is this change needed for mypy?

kitsune/sumo/utils.py
```diff
@@ -21,9 +22,12 @@
 from kitsune.lib.tlds import VALID_TLDS
 from kitsune.sumo import paginator
 
+
 POTENTIAL_LINK_REGEX = re.compile(r"[^\s/]+\.([^\s/.]{2,})")
 POTENTIAL_IP_REGEX = re.compile(r"(?:[0-9]{1,3}\.){3}[0-9]{1,3}")
 
+User: models.Model = get_user_model()
```
akatsoulas, line 29: very interesting!

kitsune/users/widgets.py
```diff
@@ -20,6 +20,10 @@ class MonthYearWidget(Widget):
     month_field = "%s_month"
     year_field = "%s_year"
 
+    @classmethod
+    def id_for_label(cls, id_):
+        return "%s_month" % id_
```
akatsoulas, line 25: f strings please!

akatsoulas (changes requested):
Really nice work! Addressing a couple of comments and this should be good to go

akatsoulas (approved):

kitsune/search/documents.py
escattone: NOTE: @akatsoulas and I resolved this during a video discussion. My addition of `update_document` as a class-level boolean attribute does the same thing in a simpler way.

kitsune/settings.py
escattone: Yes, this was because `mypy` inferred that it was a list initially, and then complained when we tried to turn it into an instance of `re.Pattern`

PR #5407, opened 2023-02-22
Use m2m for Announcement groups

akatsoulas:


kitsune/announcements/models.py
```diff
@@ -85,8 +82,8 @@ def _visible_query(cls, **query_kwargs):
 
 
 def connector(sender, instance, created, **kw):
-    # Only email new announcements in a group. We don't want to email everyone.
-    if created and instance.group and instance.send_email:
+    # Only email new announcements in specified groups. We don't want to email everyone.
+    if created and instance.groups and instance.send_email:
```
escattone, line 86: Since `instance.groups` is always true -- I think it's an instance of `QuerySet` -- this should be `instance.groups.count()` instead. However, I've only just realized there's a fundamental problem now that we're using `groups`, a many-to-many relationship, rather than `group`. The `post_save` signal will only be fired when the `Announcement` is saved, but that's always **before** any groups have been added, so no emails will ever be sent.

It seems that we need to know when the Admin has finished adding all of the groups to the `Announcement`, and only then we can send the emails, but I'm not sure of the best way to do that. In any case, I don't think we can use the `post_save` signal with this `connector` any longer. I suppose we have to create a new signal that we fire once the admin has finished adding all of the groups?

escattone (changes requested):

akatsoulas:
@escattone pr updated

kitsune/announcements/tasks.py
```diff
@@ -30,7 +30,10 @@ def send_group_email(announcement_id):
 
     @safe_translation
     def _make_mail(locale, user):
-        subject = _("New announcement for {group}").format(group=group.name)
+        if groups.count() == 1:
+            subject = _(f"New announcement for {groups[0].name}")
+        else:
+            subject = _(f"New announcement for groups [{', '.join([g.name for g in groups])}])")
```
escattone, line 36: Nit. Remove unmatched `)` at the end of the string.
```suggestion
            subject = _(f"New announcement for groups [{', '.join([g.name for g in groups])}]")
```

escattone (approved):
Nice! r+ wc

PR #5410, opened 2023-03-07
restrict html in messages to what bleach allows

escattone:
This is a significant change from what we allow currently in messages, but it's much safer.


akatsoulas (approved):
r+ with tests passing

PR #5417, opened 2023-03-21
Added the base of our Selenium UI framework:

emilghittasv:
- Added the core classes (Base Page & Test Utilities).
- Added the messages classes for Homepage and Contribute page.
- Added the pages classes for (Contribute & Homepage pages and Footer & Top Navbar sections).
- Added the conftest.py file.
- Added a simple test file as a "POC" which contains two tests (test_homepage.py).
- Integrated the logging mechanism with the HTML reporting.
- Added the pytest.ini file and defined our markers.
- Added the following dependencies to poetry: pytest selenium webdriver-manager pytest-html pytest-xdist
- Added .pytest_cache/, geckodriver.log and chromedriver.log to be ignored by git.

**Note**

In order to test it you need to:

1.  Navigate inside the selenium_tests directory `cd selenium_tests`
2. Run the following command: `pytest -m smokeTest -n 2 --html=reports/report.html --capture=tee-sys`


akatsoulas:
Closing in favor of #5437. It's the same PR rebased against `main` and it has one additional commit which syncs the poetry.lock and removes poetry from the dependencies

PR #5430, opened 2023-04-03
support SVG for product images

escattone:
mozilla/sumo#312

# Notes
- This PR adds a new `kitsune.sumo.fields.ImagePlusField` class that adds SVG support on top of Django's `models.ImageField`, and uses that class for the image fields of the `Product` and `Topic` classes within `kitsune.products.models`. Images for those fields are added/changed via the admin console.
- I originally wrote the PR using the `cairosvg` package to clean the incoming SVG image (and overwrite the incoming image with the cleaned image), but that pulled me into the "rabbit hole" of understanding and handling the various cases of incoming file objects that the related form field (Django's `django.forms.ImageField`) handles.
- I finished that idea of using the `cairosvg` package, and it's still available in this PR as the first commit. However, since the `cairosvg.svg2svg` function modifies the incoming SVG image, and since the intention is to only allow SVG images in trusted situations, in the end I regretted the time spent on trying to "clean" the incoming SVG image and decided to just accept the incoming SVG image unmodified and unchecked (only the file extension is checked). That decision is reflected in the final state of this PR.
- One possibility for redeeming the `cairosvg` work, is that we could use `cairosvg.svg2png` (or even `cairosvg.svg2svg`) to add support in the future for SVG images in the gallery.


akatsoulas:
Since you already did the work, is there a reason not to accept the 1st commit in order to have something more generic that we can use throughout the platform?

escattone:
> Since you already did the work, is there a reason not to accept the 1st commit in order to have something more generic that we can use throughout the platform?

My only concern is that `cairosvg.svg2svg` significantly alters the SVG image XML, even if it's safe, so in trusted situations like the product and topic images, there's no way to attach SVG images unaltered. However, it doesn't alter the actual look of the image, except, as far as I understand, in some cases where `cairosvg` doesn't yet support some things in SVG.

So, there's really no hard reason for not accepting the first commit. We could go with that one and see how we like it, and like you said, it allows us to use the `kitsune.sumo.fields.ImagePlusField` more generally, in untrusted situations.

escattone:
@akatsoulas I updated the PR to go with the first commit. However, if after further reflection you'd prefer to not clean/modify the incoming SVG image, I can easily remove the `cairosvg` stuff.

akatsoulas:
I am marking this PR as blocked until we fix the migrations in stage and prod.

escattone:
I removed the `Blocked` tag since the Django DB migrations have been fixed on stage and prod.

PR #5438, opened 2023-04-11
include templates in kb_overview_rows

escattone:
mozilla/sumo#1232

# Notes
It looks like templates have been excluded from the following pages since [October 15, 2014](https://github.com/mozilla/kitsune/commit/a5715c27d3a9f12b779981761f4f6f565b644459):

- `/<locale>/contributors`
- `/<locale>/contributors/kb-overview`
- `/<locale>/contributors/overview-rows`

We allow filtering by category for the first two endpoints, but the `Templates` category never worked because they were always excluded.


akatsoulas (approved):
Nice catch! Let's hold on merging this until today's release is out

PR #5447, opened 2023-04-26
Fix document edit page - Issue 1234

smithellis:
- Created separate page for document meta editing
- Modified sidebar navigation to support separate pages
- Added view for new page
- Modified .js to be appropriate per-page
- Only staff can edit document meta
- Slug no longer prepopulates if changed
- Modify tests to reflect new process


kitsune/wiki/jinja2/wiki/edit.html
```diff
@@ -19,83 +20,59 @@
...
-          <details {% if disclose_description %} open="open"{% endif %}>
-            <summary class="sumo-page-subheading">{{ _('Edit Description') }}</summary>
-            {{ errorlist(document_form) }}
-            <form action="" method="post" data-json-url="{{ url('wiki.json') }}" data-document-id="{{ document.id }}">
-              {% csrf_token %}
-                {% for field in document_form.visible_fields() if
-                   (field.name != 'is_localizable' or not document.translations.exists()) %}
-                  <div class="field {% if field.name == 'needs_change_comment' %}has-large-textarea comment{% endif %}">
-                    {{ field|label_with_help }}{{ field }}
+            {% if request.user.is_staff %}
```
akatsoulas, line 31: why are we introducing this change?

kitsune/wiki/views.py
```diff
@@ -435,7 +435,89 @@ def steal_lock(request, document_slug, revision_id=None):
...
+        return translate(request, doc.parent.slug, revision_id)
+    if revision_id:
+        rev = get_object_or_404(Revision, pk=revision_id, document=doc)
+    else:
+        rev = doc.current_revision or doc.revisions.order_by("-created", "-id")[0]
+
+    rev_form = None
+
+    if request.method == "GET":
+        if doc.allows(user, "create_revision"):
```
akatsoulas, line 466: You could reduce the level of nesting by using something like:

```python
if doc.allows(user, "create_revision"):
    rev_form = RevisionForm(request.POST or None, instance=rev, ...)

    # request.GET part goes here
    ....
    
    if rev_form.is_valid():  # request.POST goes here
        ....
```
of course this will require a restructure to the whole view but most probably it's going to improve readability and make it more compact.

If the above is not possible you could remove the check for the GET and just check for post

kitsune/wiki/jinja2/wiki/edit.html
smithellis: When I spoke to Abby about this effort, she asked that only staff be allowed to edit the Document Meta (title, slug, etc).

kitsune/wiki/views.py
smithellis: hm. I'll look at this!  Thanks.

kitsune/wiki/views.py
```diff
@@ -435,7 +435,89 @@ def steal_lock(request, document_slug, revision_id=None):
...
+            rev_form = RevisionForm(request.POST)
+            rev_form.instance.document = doc  # for rev_form.clean()
+            if rev_form.is_valid():
+                _save_rev_and_notify(rev_form, user, doc, base_rev=rev)
+                if "notify-future-changes" in request.POST:
+                    EditDocumentEvent.notify(request.user, doc)
+                return HttpResponseRedirect(
+                    reverse("wiki.document_revisions", args=[document_slug])
+                )
+        else:
```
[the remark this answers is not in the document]
smithellis: Whoops!

akatsoulas:
@smithellis is this ready for another pass?

kitsune/wiki/jinja2/wiki/edit.html
smithellis: I wasn't enforcing in the view - I'll look at that today.

kitsune/wiki/views.py
```diff
@@ -455,7 +455,89 @@ def steal_lock(request, document_slug, revision_id=None):
...
+            raise PermissionDenied
```
escattone, line 519: What do you think about the following as a way to reduce nesting? I'm also assuming that the `doc.allows(user, "create_revision")` check is done as recommended above.

```suggestion
    if request.method == "GET":
        rev_form = RevisionForm(instance=rev, initial={"based_on": rev.id, "comment": ""})
        show_revision_warning = _show_revision_warning(doc, rev)
        locked, locked_by = _document_lock(doc.id, user.username)
        return render(
            request,
            "wiki/edit.html",
            {
                "revision_form": rev_form,
                "document": doc,
                "show_revision_warning": show_revision_warning,
                "locked": locked,
                "locked_by": locked_by,
            },
        )

    _document_lock_clear(doc.id, user.username)
    rev_form = RevisionForm(request.POST)
    rev_form.instance.document = doc  # for rev_form.clean()
    if rev_form.is_valid():
        _save_rev_and_notify(rev_form, user, doc, base_rev=rev)
        if "notify-future-changes" in request.POST:
            EditDocumentEvent.notify(request.user, doc)
        return HttpResponseRedirect(
            reverse("wiki.document_revisions", args=[document_slug])
        )
```

kitsune/wiki/views.py
```diff
@@ -455,7 +455,89 @@ def steal_lock(request, document_slug, revision_id=None):
...
+            "show_revision_warning": show_revision_warning,
+            "locked": locked,
+            "locked_by": locked_by,
+        },
+    )
+
+
+@require_http_methods(["GET", "POST"])
+@login_required
+def edit_document_doc(request, document_slug, revision_id=None):
```
escattone, line 539: What do you think of `edit_document_metadata` instead?

kitsune/wiki/urls.py
```diff
@@ -11,6 +11,7 @@
     re_path(r"^/revision/(?P<revision_id>\d+)$", views.revision, name="wiki.revision"),
     re_path(r"^/history$", views.document_revisions, name="wiki.document_revisions"),
     re_path(r"^/edit$", views.edit_document, name="wiki.edit_document"),
+    re_path(r"^/edit/document$", views.edit_document_doc, name="wiki.edit_document_doc"),
```
escattone, line 14: What do you think of the following instead?
```suggestion
    re_path(r"^/edit/metadata$", views.edit_document_metadata, name="wiki.edit_document_metadata"),
```

kitsune/wiki/jinja2/wiki/edit.html
```diff
@@ -22,83 +23,59 @@
...
+              <p><a href="{{ url('wiki.edit_document_doc', document.slug) }}" class="sumo-button primary-button">{{ _('Edit Article Description Instead?') }}</a></p>
```
escattone, line 35: I think we should remove the `Instead?`. What do you think?
```suggestion
              <p><a href="{{ url('wiki.edit_document_doc', document.slug) }}" class="sumo-button primary-button">{{ _('Edit Article Description') }}</a></p>
```

kitsune/wiki/views.py
escattone: It was difficult to know where to put an alternative for this view, so I'll add my suggestion here:
```python
@require_http_methods(["GET", "POST"])
@login_required
def edit_document_metadata(request, document_slug, revision_id=None):
    """Edit document metadata."""
    user = request.user

    doc = get_visible_document_or_404(
        user,
        locale=request.LANGUAGE_CODE,
        slug=document_slug,
        look_for_translation_via_parent=True,
        return_parent_if_no_translation=True,
    )

    if doc.locale != request.LANGUAGE_CODE:
        # We've fallen back to the parent, since no visible translation existed.
        url = reverse("wiki.translate", locale=request.LANGUAGE_CODE, args=[document_slug])

    # If this document has a parent, then the edit is handled by the
    # translate view. Pass it on.
    if doc.parent:
        return translate(request, doc.parent.slug, revision_id)

    if not doc.allows(user, "edit") and user.is_staff:
        raise PermissionDenied

    can_edit_needs_change = doc.allows(user, "edit_needs_change")
    can_archive = doc.allows(user, "archive")

    if revision_id:
        rev = get_object_or_404(Revision, pk=revision_id, document=doc)
    else:
        rev = doc.current_revision or doc.revisions.order_by("-created", "-id")[0]

    if request.method == "GET":
        doc_form = DocumentForm(
            initial=_document_form_initial(doc),
            can_archive=can_archive,
            can_edit_needs_change=can_edit_needs_change,
        )

    else:  # POST
        _document_lock_clear(doc.id, user.username)

        post_data = request.POST.copy()
        post_data.update({"locale": request.LANGUAGE_CODE})

        topics = []
        for t in post_data.getlist("topics"):
            topics.append(int(t))
        post_data.setlist("topics", topics)

        doc_form = DocumentForm(
            post_data,
            instance=doc,
            can_archive=can_archive,
            can_edit_needs_change=can_edit_needs_change,
        )
        if doc_form.is_valid():
            # Get the possibly new slug for the imminent redirection:
            try:
                doc = doc_form.save(None)
            except (TitleCollision, SlugCollision) as e:
                # TODO: .add_error() when we upgrade to Django 1.7
                errors = doc_form._errors.setdefault("title", ErrorList())
                message = "The {type} you selected is already in use."
                message = message.format(
                    type="title" if isinstance(e, TitleCollision) else "slug"
                )
                errors.append(_(message))
            else:
                # Do we need to rebuild the KB?
                _maybe_schedule_rebuild(doc_form)

                return HttpResponseRedirect(
                    urlparams(reverse("wiki.document", args=[doc.slug]))
                )

    show_revision_warning = _show_revision_warning(doc, rev)
    locked, locked_by = _document_lock(doc.id, user.username)

    return render(
        request,
        "wiki/edit_metadata.html",
        {
            "document_form": doc_form,
            "document": doc,
            "show_revision_warning": show_revision_warning,
            "locked": locked,
            "locked_by": locked_by,
        },
    )
```

escattone (commented):
@smithellis I thought I'd add my comments so far, and I'll keep reviewing more tomorrow. I agree with you that the current experience is poor when you're shown the same view after saving your metadata edits. Now that you've split this into two kinds of edits, I think we should redirect to the article itself after saving metadata edits. What do you think?

kitsune/wiki/views.py
```diff
@@ -455,7 +455,89 @@ def steal_lock(request, document_slug, revision_id=None):
...
+    # If this document has a parent, then the edit is handled by the
+    # translate view. Pass it on.
+    if doc.parent:
+        return translate(request, doc.parent.slug, revision_id)
+    if revision_id:
+        rev = get_object_or_404(Revision, pk=revision_id, document=doc)
+    else:
+        rev = doc.current_revision or doc.revisions.order_by("-created", "-id")[0]
+
+    rev_form = None
```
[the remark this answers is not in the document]
smithellis: Whoops.

kitsune/wiki/views.py
smithellis: Nice - and agreed.

kitsune/wiki/views.py
smithellis: I'm going to go ahead and include the change.

kitsune/wiki/jinja2/wiki/edit.html
smithellis: Sure thing - keep it simple!

kitsune/wiki/jinja2/wiki/edit.html
```diff
@@ -22,83 +23,59 @@
...
-            <form action="" method="post" data-json-url="{{ url('wiki.json') }}" data-document-id="{{ document.id }}">
-              {% csrf_token %}
-                {% for field in document_form.visible_fields() if
-                   (field.name != 'is_localizable' or not document.translations.exists()) %}
-                  <div class="field {% if field.name == 'needs_change_comment' %}has-large-textarea comment{% endif %}">
-                    {{ field|label_with_help }}{{ field }}
+            {% if request.user.is_staff %}
+              <p><a href="{{ url('wiki.edit_document_metadata', document.slug) }}" class="sumo-button primary-button">{{ _('Edit Article Description') }}</a></p>
+            {% endif %}
+            {% if revision_form %}
```
escattone, line 37: With your changes in this PR, `revision_form` is always defined when using `wiki/edit.html`, so I think we could get rid of this `{% if revision_form %}{% endif %}` wrapper.

kitsune/wiki/jinja2/wiki/edit_metadata.html
```diff
@@ -0,0 +1,62 @@
...
+    <article id="edit-document" {% if locked %}class="locked"{% endif %} data-slug="{{ document.slug }}">
+      <h1 class="sumo-page-heading">{{ _('<em>Editing Description For:</em><br>{title}')|fe(title=document.title) }}</h1>
+      {% if document.needs_change %}
+        <p>
+          {% trans comment=document.needs_change_comment %}
+            <label>Needs Change:</label> {{comment}}
+          {% endtrans %}
+        </p>
+      {% endif %}
+      {% if document_form %}
```
escattone, line 31: With your changes in this PR, `document_form` is always defined when using `wiki/edit_metadata.html`, so I think we could get rid of this `{% if document_form %}{% endif %}` wrapper.

kitsune/wiki/views.py
```diff
@@ -471,106 +471,144 @@ def edit_document(request, document_slug, revision_id=None):
...
+        # We've fallen back to the parent, since no visible translation existed.
+        url = reverse("wiki.translate", locale=request.LANGUAGE_CODE, args=[document_slug])
+        return HttpResponseRedirect(url)
+
+    # If this document has a parent, then the edit is handled by the
+    # translate view. Pass it on.
+    if doc.parent:
+        return translate(request, doc.parent.slug, revision_id)
+
+    if not doc.allows(user, "edit") and user.is_staff:
```
escattone, line 553: The logic won't be correct unless we wrap everything after the `not` in parentheses.
```suggestion
    if not (doc.allows(user, "edit") and user.is_staff):
```

kitsune/sumo/static/sumo/js/wiki.js
```diff
@@ -25,7 +24,6 @@ import ShowFor from "sumo/js/showfor";
 
     $('select.enable-if-js').prop("disabled", false);
 
-    initPrepopulatedSlugs();
```
escattone, line 28: 🎉 I'm really happy to see this go! 😄

kitsune/wiki/views.py
```diff
@@ -471,106 +471,144 @@ def edit_document(request, document_slug, revision_id=None):
...
-                        EditDocumentEvent.notify(request.user, doc)
+                # Do we need to rebuild the KB?
+                _maybe_schedule_rebuild(doc_form)
 
-                    return HttpResponseRedirect(
-                        reverse("wiki.document_revisions", args=[document_slug])
-                    )
-            else:
-                raise PermissionDenied
+                return HttpResponseRedirect(urlparams(reverse("wiki.document", args=[doc.slug])))
```
escattone, line 602: If the user changes the slug, the current document (with the original slug) becomes a redirect to the document at the new slug, so in that case, this code would redirect the user to original slug first, which would, in turn, then redirect to the new slug. I suspect this leaves users confused about what happened to the document at the original slug, so maybe we should add `?redirect="no"` so that the user is shown the original document. That way, users see what happened to the document they just edited, which has been turned into a redirect with a link to the new document slug in the body, and also clearly see that they can simply click that link if they'd like to go to the new document slug. I don't know. Maybe it's best to ask the content team what they'd prefer?
```suggestion
                return HttpResponseRedirect(urlparams(reverse("wiki.document", args=[doc.slug]), redirect="no"))
```

escattone: By the way, if we decide not to do this, I just wanted to note that the `urlparams` call is not needed and can be removed.
```suggestion
                return HttpResponseRedirect(reverse("wiki.document", args=[doc.slug]))
```

escattone (changes requested):
This is such a nice improvement, thanks @smithellis! I've requested some minor changes, but this is really close as far as I'm concerned.

kitsune/wiki/jinja2/wiki/edit_metadata.html
```diff
@@ -0,0 +1,62 @@
...
+                    {% endfor %}
+                  {% if document.translations.exists() %}
+                    {{ document_form.is_localizable.as_hidden()|safe }}
+                  {% endif %}
+                  <input type="hidden" name="form" value="doc" />
+                  <div class="submit sumo-button-wrap align-end">
+                    <button class="sumo-button primary-button" type="submit">{{ _('Save description') }}</button>
+                  </div>
+                </form>
+            </div>
```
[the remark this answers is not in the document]
smithellis: I either fixed this or made it worse.

kitsune/wiki/views.py
smithellis: I hadn't thought about that at all! Good catch.

smithellis:
I adopted your suggestions, @escattone . Thanks for the great feedback.

kitsune/wiki/views.py
```diff
@@ -471,106 +471,146 @@ def edit_document(request, document_slug, revision_id=None):
...
+                # Do we need to rebuild the KB?
+                _maybe_schedule_rebuild(doc_form)
 
-                    return HttpResponseRedirect(
-                        reverse("wiki.document_revisions", args=[document_slug])
-                    )
-            else:
-                raise PermissionDenied
+                return HttpResponseRedirect(
+                    urlparams(reverse("wiki.document", args=[doc.slug]), redirect="no")
```
escattone, line 603: I'm sorry @smithellis. I'm not only doubting my suggestion to use `redirect="no"`, but realized that it doesn't even work unless we save the original slug before we save the `doc_form` on line 591, and use that original slug here instead of `doc.slug`. If you prefer to keep `redirect="no"`, you'd have to do those things, but my new suggestion is to just remove it, which means you can remove `urlparams` as well:
```suggestion
                    reverse("wiki.document", args=[doc.slug])
```

escattone (changes requested):
This looks great except I discovered one more thing in my local testing.

kitsune/wiki/views.py
smithellis: No sweat - and I get a demerit for only running unit tests and not checking that link manually.

escattone (approved):

kitsune/wiki/views.py
```diff
@@ -471,106 +471,144 @@ def edit_document(request, document_slug, revision_id=None):
...
+
     if revision_id:
         rev = get_object_or_404(Revision, pk=revision_id, document=doc)
     else:
         rev = doc.current_revision or doc.revisions.order_by("-created", "-id")[0]
 
-    disclose_description = bool(request.GET.get("opendescription"))
-    doc_form = rev_form = None
-    if doc.allows(user, "create_revision"):
+    if request.method == "GET":
```
akatsoulas, line 487: Why is this needed?

akatsoulas: Usually we are checking against a post request that has some kind of a payload. GET requests should be the default and we just create the context and render the page. In my previous comment there was a suggestion to initialize the form for both if possible. 

For example you could do something like 
`rev_form = RevisionForm(request.POST or None, ....)

kitsune/wiki/views.py
```diff
@@ -471,106 +471,144 @@ def edit_document(request, document_slug, revision_id=None):
...
+    if rev_form.is_valid():
+        _save_rev_and_notify(rev_form, user, doc, base_rev=rev)
+        if "notify-future-changes" in request.POST:
+            EditDocumentEvent.notify(request.user, doc)
+        return HttpResponseRedirect(reverse("wiki.document_revisions", args=[document_slug]))
+
+    show_revision_warning = _show_revision_warning(doc, rev)
+    locked, locked_by = _document_lock(doc.id, user.username)
+
+    return render(
```
akatsoulas, line 516: This is also part of the GET request. Given that you return on the above check and redirect on POST, when does the program flow reach this point?

kitsune/wiki/views.py
```diff
@@ -471,106 +471,144 @@ def edit_document(request, document_slug, revision_id=None):
...
+        # We've fallen back to the parent, since no visible translation existed.
+        url = reverse("wiki.translate", locale=request.LANGUAGE_CODE, args=[document_slug])
+        return HttpResponseRedirect(url)
+
+    # If this document has a parent, then the edit is handled by the
+    # translate view. Pass it on.
+    if doc.parent:
+        return translate(request, doc.parent.slug, revision_id)
+
+    if not (doc.allows(user, "edit") and user.is_staff):
```
akatsoulas, line 553: Shouldn't this be:
```python
if not doc.allows(user, "edit") or not user.is_staff:
   do_stuff
```

akatsoulas: Also this check should be one of the first things we check, closer to the top. If this evaluates to a permission denied, there is no need to do the checks currently above this line.

kitsune/wiki/views.py
```diff
@@ -471,106 +471,144 @@ def edit_document(request, document_slug, revision_id=None):
...
+
+    can_edit_needs_change = doc.allows(user, "edit_needs_change")
+    can_archive = doc.allows(user, "archive")
+
+    if revision_id:
+        rev = get_object_or_404(Revision, pk=revision_id, document=doc)
+    else:
+        rev = doc.current_revision or doc.revisions.order_by("-created", "-id")[0]
+
+    if request.method == "GET":
```
akatsoulas, line 564: Again, let's check for POST. This will reduce the nesting in the code because you won't need the `else` clause and will improve readability

akatsoulas (changes requested):

kitsune/wiki/views.py
escattone: Just wanted to note that `not doc.allows(user, "edit") or not user.is_staff` is equivalent to `not (doc.allows(user, "edit") and user.is_staff)`, so we could use whatever you guys think is most readable.

kitsune/wiki/views.py
smithellis: Ok, after much toying with this - you can't avoid an explicit check for GET or POST, as there doesn't seem to be a way to designate a single rev_form definition without creating a problem.  Doing something like: `rev_form=RevisionForm(request.POST or None, instance=rev, initial={"based_on": rev.id, "comment": ""})` is problematic, since the initial values being defined will overwrite the POSTED data and break tests. You could assign the rev_form content based on a check for POST or GET, but that makes the code more difficult to understand AND still requires you to implicitly check for the request TYPE. Latest commit shows what I think is the best solution.

kitsune/wiki/views.py
smithellis: I changed this to @akatsoulas recommendation - I think it makes reading easier.

kitsune/wiki/views.py
smithellis: I changed this to check for POST, but an `else` is still required as there is a GET specific bit of code, then a bit of code that is GET specific, then a bit of code that is for GET as well as fallback for a bad POST/failed is_valid().

kitsune/wiki/views.py
smithellis: Fixed. See latest commit.

kitsune/wiki/views.py
```diff
@@ -483,94 +483,123 @@ def edit_document(request, document_slug, revision_id=None):
...
+        },
+    )
 
-        _document_lock_clear(doc.id, user.username)
 
-        if which_form == "doc":
-            if doc.allows(user, "edit"):
-                post_data = request.POST.copy()
-                post_data.update({"locale": request.LANGUAGE_CODE})
+@require_http_methods(["GET", "POST"])
```
akatsoulas, line 520: Is this needed?

kitsune/wiki/views.py
```diff
@@ -483,94 +483,123 @@ def edit_document(request, document_slug, revision_id=None):
...
-                        errors = doc_form._errors.setdefault("title", ErrorList())
-                        message = "The {type} you selected is already in use."
-                        message = message.format(
-                            type="title" if isinstance(e, TitleCollision) else "slug"
-                        )
-                        errors.append(_(message))
-                    else:
-                        # Do we need to rebuild the KB?
-                        _maybe_schedule_rebuild(doc_form)
+    if not doc.allows(user, "edit") or not user.is_staff:
```
akatsoulas, line 534: Similar comment with the template, how about superusers and Admins?

kitsune/wiki/views.py
```diff
@@ -483,94 +483,123 @@ def edit_document(request, document_slug, revision_id=None):
...
-                        EditDocumentEvent.notify(request.user, doc)
+    if doc.locale != request.LANGUAGE_CODE:
+        # We've fallen back to the parent, since no visible translation existed.
+        url = reverse("wiki.translate", locale=request.LANGUAGE_CODE, args=[document_slug])
+        return HttpResponseRedirect(url)
 
-                    return HttpResponseRedirect(
-                        reverse("wiki.document_revisions", args=[document_slug])
-                    )
+    can_edit_needs_change = doc.allows(user, "edit_needs_change")
```
akatsoulas, line 542: There's a big copy/paste block that's shared between the two views. Is it possible to make this a bit more DRY?

kitsune/wiki/views.py
```diff
@@ -483,94 +483,123 @@ def edit_document(request, document_slug, revision_id=None):
...
+                message = "The {type} you selected is already in use."
+                message = message.format(type="title" if isinstance(e, TitleCollision) else "slug")
+                errors.append(_(message))
             else:
-                raise PermissionDenied
+                # Do we need to rebuild the KB?
+                _maybe_schedule_rebuild(doc_form)
+                return HttpResponseRedirect(reverse("wiki.document", args=[doc.slug]))
+    else:  # GET
+        # We can't drop this ELSE because we need to account for bad POSTs
```
akatsoulas, line 586: If the form is not valid, won't the form's clean method raise the errors before reaching this point?

akatsoulas (changes requested):

akatsoulas (changes requested):
Let me know when it's updated to have another pass

akatsoulas:
See also the comments made in PR #5567

kitsune/wiki/jinja2/wiki/includes/sidebar_modules.html
```diff
@@ -24,9 +24,14 @@ <h3 class="sidebar-subheading large-only">{{ _('Editing Tools') }}</h3>
...
           {% endif %}
           {% if document and not document.is_archived and (document.allows(user, 'create_revision') or document.allows(user, 'edit')) and fallback_reason != "fallback_locale" %}
-            <li {{ active|class_selected('edit') }}>
-              <a href="{{ url('wiki.edit_document', document.slug) }}">{{ _('Edit Article') }}</a>
+          <li {{ active|class_selected('edit') }}>
+            <a href="{{ url('wiki.edit_document', document.slug) }}">{{ _('Edit Article') }}</a>
+          </li>
+            {% if request.user.is_staff %}
+            <li {{ active|class_selected('edit_doc') }}>
+              <a href="{{ url('wiki.edit_document_metadata', document.slug) }}">{{ _('Edit Article Description') }}</a>
```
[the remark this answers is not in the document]
smithellis: Yeah, I wobbled on that wondering if people would understand "Metadata" vs. "Description". I guess the user editing articles would understand the idea of Metadata, so...I'll change it!

kitsune/wiki/jinja2/wiki/edit.html
```diff
@@ -22,83 +23,54 @@
...
-                {% for field in document_form.visible_fields() if
-                   (field.name != 'is_localizable' or not document.translations.exists()) %}
-                  <div class="field {% if field.name == 'needs_change_comment' %}has-large-textarea comment{% endif %}">
-                    {{ field|label_with_help }}{{ field }}
-                  </div>
-                {% endfor %}
-              {% if document.translations.exists() %}
-                {{ document_form.is_localizable.as_hidden()|safe }}
+      {% if request.user.is_staff %}
+        <p><a href="{{ url('wiki.edit_document_metadata', document.slug) }}" class="sumo-button primary-button">{{ _('Edit Article Description') }}</a></p>
```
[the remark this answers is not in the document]
smithellis: Same as above.

kitsune/wiki/jinja2/wiki/edit_metadata.html
```diff
@@ -0,0 +1,60 @@
...
+{% from "includes/common_macros.html" import content_editor with context %}
+{% from "wiki/includes/document_macros.html" import edit_messages, submit_revision %}
+{% from "wiki/includes/document_macros.html" import document_lock_warning with context %}
+{% set title = _('Edit Article Metadata | {document}')|f(document=document.title) %}
+{# TODO: Change KB url to landing page when we have one #}
+{% set crumbs = [(document.get_absolute_url(), document.title),
+                 (None, _('Edit Article Metadata'))] %}
+
+{% set scripts = ('wiki', 'wiki.diff', 'wiki.editor') %}
+{% set classes = 'edit_doc' %}
```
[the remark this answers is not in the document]
smithellis: Sure thing.

kitsune/wiki/jinja2/wiki/includes/sidebar_modules.html
```diff
@@ -24,9 +24,14 @@ <h3 class="sidebar-subheading large-only">{{ _('Editing Tools') }}</h3>
             </li>
           {% endif %}
           {% if document and not document.is_archived and (document.allows(user, 'create_revision') or document.allows(user, 'edit')) and fallback_reason != "fallback_locale" %}
-            <li {{ active|class_selected('edit') }}>
-              <a href="{{ url('wiki.edit_document', document.slug) }}">{{ _('Edit Article') }}</a>
+          <li {{ active|class_selected('edit') }}>
+            <a href="{{ url('wiki.edit_document', document.slug) }}">{{ _('Edit Article') }}</a>
+          </li>
+            {% if request.user.is_staff %}
```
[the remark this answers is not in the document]
smithellis: If you are a `superuser` you are automatically given the `staff` role. Since someone in the `administrator` group could not be staff, in order to do what was asked, I'm only looking for staff.

kitsune/wiki/views.py
```diff
@@ -466,14 +466,14 @@ def edit_document(request, document_slug, revision_id=None):
...
 
+    if not doc.allows(user, "create_revision"):
+        raise PermissionDenied
+
     if doc.locale != request.LANGUAGE_CODE:
         # We've fallen back to the parent, since no visible translation existed.
         url = reverse("wiki.translate", locale=request.LANGUAGE_CODE, args=[document_slug])
         return HttpResponseRedirect(url)
 
-    can_edit_needs_change = doc.allows(user, "edit_needs_change")
```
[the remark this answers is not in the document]
smithellis: It's only used in the metadata edit, not the document.

kitsune/wiki/views.py
smithellis: If `is_valid()` fails we want to render the same form submission, with the errors. So if `is_valid()` fails during a `POST`, execution continues to the end. We won't create a new instance of DocumentForm (which we do for a GET) but instead use the same `doc_form`. So you get the errors with your form data vs. new form data.

kitsune/wiki/jinja2/wiki/edit.html
```diff
@@ -22,83 +23,54 @@
...
-              {% csrf_token %}
-                {% for field in document_form.visible_fields() if
-                   (field.name != 'is_localizable' or not document.translations.exists()) %}
-                  <div class="field {% if field.name == 'needs_change_comment' %}has-large-textarea comment{% endif %}">
-                    {{ field|label_with_help }}{{ field }}
-                  </div>
-                {% endfor %}
-              {% if document.translations.exists() %}
-                {{ document_form.is_localizable.as_hidden()|safe }}
+      {% if request.user.is_staff %}
```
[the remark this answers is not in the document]
smithellis: I thought that if you were used to it behaving the old way, you'd be trying to figure out what happened to the Metadata piece you used to see - this gives you another visual cue about the change. It does look out of place though.  I'll take it off.

kitsune/wiki/views.py
smithellis: Can do.

smithellis:
Just to clarify the main points of the latest notes/edits:

I broke out several recurring features into separate functions:
`_get_user_and_doc` - gets the `request.user` and the `appropriate doc`
`_check_edit_perms` - checks for `doc.allows` perms based on document type (`doc` or `meta`)
`_get_rev` - returns the proper revision 

I couldn't think of a meaningful way to break out the locale check that returns a `HttpResponseRedirect` that made anything easier to read. I could have broken it out and done some weird walrus stuff like:
```
if (ret := _check_locale(request, document_slug, doc) is not None:
    return ret
```
But that seems a little less readable to me than just checking and returning the `wiki.translate` page from the same spot....

I was able to move things around so in both `edit_document` and `edit_metadata` we are only checking for POST, but I wasn't able to use the suggested fix of assigning `rev_form` on a single line with `rev_form = RevisionForm(request.POST or None, instance=rev, initial={"based_on": rev.id, "comment": ""})` . This fix breaks testing, because it alters functionality.  For a `POST` request, the instance and initial values are not discarded when the form is re-instantiated with `POST` data. Therefore, when `is_valid()` is called, the `POST` data is validated against the model instance's constraints and the `initial` data. 

I've also integrated the the changes from the now defunct https://github.com/mozilla/kitsune/pull/5567 PR.

kitsune/wiki/views.py
```diff
@@ -452,125 +452,154 @@ def steal_lock(request, document_slug, revision_id=None):
...
+        doc_form = DocumentForm(
+            post_data,
+            instance=doc,
+            can_archive=can_archive,
+            can_edit_needs_change=can_edit_needs_change,
+        )
+        if doc_form.is_valid():
+            # Get the possibly new slug for the imminent redirection:
+            try:
+                doc = doc_form.save(None)
```
akatsoulas, line 581: Why don't we pass the logic from here to the form save method?

akatsoulas: There's also validation logic in the form that may be a better fit for this

kitsune/wiki/views.py
```diff
@@ -452,125 +452,154 @@ def steal_lock(request, document_slug, revision_id=None):
...
+        rev = get_object_or_404(Revision, pk=revision_id, document=doc)
+    else:
+        rev = doc.current_revision or doc.revisions.order_by("-created", "-id")[0]
+
+    return rev
+
+
+@require_http_methods(["GET", "POST"])
+@login_required
+def edit_document(request, document_slug, revision_id=None):
```
akatsoulas, line 487: `edit_document` and `edit_document_metadata` share a lot of code. We could use something like 

```python
def base_edit_checks(request, document_slug, revision_id=None, doctype="doc"):

    user = request.user
    doc = get_visible_document_or_404(
        user,
        locale=request.LANGUAGE_CODE,
        slug=document_slug,
        look_for_translation_via_parent=True,
        return_parent_if_no_translation=True,
    )

    if doctype == "doc":
        if not doc.allows(user, "create_revision"):
            raise PermissionDenied
    if doctype == "meta":
        if not doc.allows(user, "edit") or not user.is_staff:
            raise PermissionDenied

    if doc.locale != request.LANGUAGE_CODE:
        # We've fallen back to the parent, since no visible translation existed.
        url = reverse("wiki.translate", locale=request.LANGUAGE_CODE, args=[document_slug])

    # If this document has a parent, then the edit is handled by the
    # translate view. Pass it on.
    if doc.parent:
        return translate(request, doc.parent.slug, revision_id)

    if revision_id:
        rev = get_object_or_404(Revision, pk=revision_id, document=doc)
    else:
        rev = doc.current_revision or doc.revisions.order_by("-created", "-id")[0]

    return user, doc, rev
```
and remove all the 3 helper functions used that start with the underscore. The `edit_document_metadata` view needs to add a forth param like `user, doc = base_edit_checks(request, document_slug, revision_id=None, doctype="meta")`

akatsoulas: The above will remove a lot of duplicate code and each view can focus on its respective function. With that said, I believe it's worth adding an issue to rethink how the forms are structured. I believe we can remove a lot of boilerplate code. That said, it's working fine and it's not a priority

akatsoulas (commented):
With the last couple of comments, this should be good to go

kitsune/wiki/views.py
smithellis: The locale piece that returns an `HttpResponseRedirect` and the doc.parent check that returns `translate` both have to be returned again in the main view function. So, for instance, when I tried to break the locale piece into it's own function, I still had to have this in the view:
```
if (ret := _check_locale(request, document_slug, doc) is not None:
    return ret
```

Which seemed less readable than just leaving the code in place.  Thoughts?

smithellis: I'll implement - thanks.

kitsune/wiki/views.py
smithellis: I moved the try/except/else out only to replace it with a try/except/else. I'm not 100% sure this is what you intended the outcome to be.  I looked at moving some of the other validation into the view, but it seems like there is a series of events that has to happen before each item in the save process happens. You need a real `doc` to get a `doc.pk`, you need the `save_m2m()` to get the products relationship....not sure how to make it better off hand?  Let me know your thoughts.

akatsoulas: I was thinking to replace everything here with just `doc = doc_form.save()` or doc = doc_form.save(commit=False)` and do all the checks in the form. The errors can live in the clean method and the object manipulation in the save. Let's remove these changes and just merge it

smithellis: Will do!

akatsoulas (approved):
r+wc. Let's remove the changes in the save method and leave things as they were. Please do not forget to squash before merging. Nice work!

Alahiki:
Save m2m


-- 
Ifeanyi Winner

Alahiki:
Merged #5447


-- 
Ifeanyi Winner

Alahiki:
-- 
Ifeanyi Winner

PR #5455, opened 2023-05-02
switching devices -- add wiki "UI" hook for device-migration wizard

escattone:
https://github.com/mozilla/sumo/issues/1247

This PR adds a new Wiki "hook" called `UI`, which can be used to pull predefined UI components into any KB article. The predefined UI components are not restricted in any way, so for example normally-disallowed HTML elements will not be escaped. This PR also adds a placeholder file (`kitsune/wiki/jinja2/wikiparser/hook_device_migration_wizard.html`) for a predefined UI component called `device_migration_wizard`, which can be included into any KB article using the already familiar Wiki syntax of:

```wiki
[[UI:device_migration_wizard]]
```

## Testing
I've tested this code locally, and it works great. I've also added unit tests.

### DIFF
--- a/kitsune/sumo/parser.py
+++ b/kitsune/sumo/parser.py
@@ -306,9 +306,7 @@ class WikiParser(Parser):
         """
 
         for name in self.ui_components:
-            html = html.replace(
-                UI_COMPONENT_PLACEHOLDER % name, render_to_string(f"wikiparser/hook_{name}.html")
-            )
+            html = html.replace(UI_COMPONENT_PLACEHOLDER % name, generate_ui_component_embed(name))
 
         return html
 
@@ -445,6 +443,11 @@ def generate_video(v, params=[]):
 def generate_youtube_embed(video_id):
     """Takes a youtube video id and returns the embed markup."""
     return render_to_string("wikiparser/hook_youtube_embed.html", {"video_id": video_id})
+
+
+def generate_ui_component_embed(name):
+    """Takes a UI component name and returns the embed markup."""
+    return render_to_string(f"wikiparser/hook_{name}.html")
 
 
 def _get_video_url(video_file):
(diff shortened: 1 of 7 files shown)


kitsune/sumo/parser.py
```diff
@@ -44,6 +44,8 @@
 }
 VIDEO_PARAMS = ["height", "width", "modal", "title", "placeholder"]
 YOUTUBE_PLACEHOLDER = "YOUTUBE_EMBED_PLACEHOLDER_%s"
+UI_COMPONENT_PLACEHOLDER = "UI_COMPONENT_EMBED_PLACEHOLDER_%s"
+ALLOWED_UI_COMPONENTS = frozenset(["device_migration_wizard"])
```
akatsoulas, line 48: why are you using a frozenset here? 
Also the idea of using the iterable is to allow multiple UI components in the future?

kitsune/sumo/parser.py
```diff
@@ -211,8 +213,10 @@ def __init__(self, base_url=None):
         self.registerInternalLinkHook("Video", self._hook_video)
         self.registerInternalLinkHook("V", self._hook_video)
         self.registerInternalLinkHook("Button", self._hook_button)
+        self.registerInternalLinkHook("UI", self._hook_ui_component)
 
         self.youtube_videos = []
+        self.ui_components = set()
```
akatsoulas, line 219: Shall we also use a `set` for youtube_videos too (or the other way around)?

kitsune/sumo/parser.py
```diff
@@ -224,6 +228,7 @@ def parse(
         locale=settings.WIKI_DEFAULT_LANGUAGE,
         nofollow=False,
         youtube_embeds=True,
+        ui_component_embeds=True,
```
akatsoulas, line 231: Let's add this arg in the docstring too

akatsoulas (approved):
r+wc

kitsune/sumo/parser.py
escattone: > why are you using a frozenset here?
Just to indicate that it's immutable, that additional UI component names can't be added dynamically.
> Also the idea of using the iterable is to allow multiple UI components in the future?
Yes, just to remain open to that possibility.

PR #5457, opened 2023-05-04
switching devices -- add stage/prod FxA metrics-flow endpoints to connect-src CSP

escattone:
mozilla/sumo#1253


kitsune/settings.py
```diff
@@ -1217,6 +1217,8 @@ def filter_exceptions(event, hint):
     "'self'",
     "https://*.google-analytics.com",
     "https://location.services.mozilla.com",
+    "https://accounts.firefox.com/metrics-flow",
```
akatsoulas, line 1220: Shall we whitelist  the whole domain?

akatsoulas (approved):
r+wc

PR #5459, opened 2023-05-04
Use the is_switching_devices tag in document.html for script injection.

mikeconley:
This also adds some default values for the FIREFOX_SWITCHING_DEVICES_ARTICLES and FIREFOX_SWITCHING_DEVICES_TOPIC settings.


mikeconley:
I wasn't sure if there was a less static-y way of setting those `FIREFOX_SWITCHING_DEVICES_ARTICLES` and `FIREFOX_SWITCHING_DEVICES_TOPIC` values, so this is what I went with. Definitely open to alternatives in case there's a right-er way.

escattone (approved):
r+

PR #5460, opened 2023-05-04
switching devices -- fix slug check to account for translations

escattone:
Fix the `is_switching_devices` check to account for translations.


escattone:
Closing in favor of https://github.com/mozilla/kitsune/pull/5462

PR #5462, opened 2023-05-04
load device-migration-wizard JS in all places needed

escattone:
This PR does the following:
- a bit of renaming that I hope helps separate what the JS and SCSS files are for (the device migration wizard -- the HTML file is already named in this way [hook_device_migration_wizard.html](https://github.com/mozilla/kitsune/blob/switching-devices/kitsune/wiki/jinja2/wikiparser/hook_device_migration_wizard.html)) from where the JS bundle is needed (when displaying the content of the `switching-devices` KB article)
    - the `document.switching-devices.js` JS file → `device-migration-wizard.js`
    - the `document.switching-devices` JS bundle name → `switching-devices`
    - the `_switching-devices.scss` SCSS file → `_device-migration-wizard.scss`
- loads the `switching-devices` JS bundle in the additional places it's needed
- provides a function (which can also be used within Jinja templates) to check for a `switching-devices` document that accounts for translations (so this PR replaces my PR mozilla/kitsune#5460)


kitsune/wiki/templatetags/jinja_helpers.py
```diff
@@ -16,3 +17,10 @@ def diff_table(content_from, content_to):
 @library.global_function
 def generate_video(v):
     return Markup(parser.generate_video(v))
+
+
+@library.global_function
+def is_switching_devices_document(doc):
+    return (
+        doc.parent.slug if doc.parent else doc.slug
```
akatsoulas, line 25: shall we make this a property on the model?

akatsoulas (approved):
r+[wc]?

mikeconley (approved):
This works for me! Thanks!

kitsune/wiki/templatetags/jinja_helpers.py
escattone: Yes, that's better, thanks @akatsoulas.

PR #5463, opened 2023-05-05
Make SwitchingDevicesWizardManager request FxA metrics immediately after construction. (#1261)

mikeconley:
This also:

* exposes the settings Django object to UI components
* adds a new FXA_ROOT settings value

### DIFF
--- a/kitsune/sumo/static/sumo/js/switching-devices-wizard-manager.js
+++ b/kitsune/sumo/static/sumo/js/switching-devices-wizard-manager.js
@@ -8,9 +8,115 @@
  */
 export default class SwitchingDevicesWizardManager {
   #formWizard = null;
-  #state = {};
+  #fxaRoot = null;
+  #state = {
+    // These UTM parameters are fallback values in the event that we
+    // don't get any passed through to us from the entrypoint to this
+    // SUMO page. These are what will ultimately get passed through the
+    // FxA sign-in flow.
+    utm_source: "support.mozilla.org",
+    utm_campaign: "migration",
+    utm_medium: "mozilla-websites",
+
+    entrypoint: "fx-new-device-sync",
+    entrypoint_experiment: null,
+    entrypoint_variation: null,
+    flow_id: null,
+    flow_begin_time: null,
+    context: null,
+  };
 
   constructor(formWizard) {
     this.#formWizard = formWizard;
+
+    if (!this.#formWizard.hasAttribute("fxa-root")) {
+      throw new Error(
+        "Expected fxa-root on element passed to " +
+          "SwitchingDevicesWizardManager constructor."
+      );
+    }
+    this.#fxaRoot = this.#formWizard.getAttribute("fxa-root");
+
+    // If the page was loaded with any UTM or entrypoint parameters, we'll want
+    // to pass those along, so now we check for those and add them to our state
+    // if need be. Since this occurs during construction, we don't need to call
+    // #updateState().
+    let searchParams = new URLSearchParams(window.location.search);
+
+    for (let paramName of [
+      "utm_source",
+      "utm_campaign",
+      "utm_medium",
+      "entrypoint",
+      "entrypoint_experiment",
+      "entrypoint_variation",
+    ]) {
+      if (searchParams.has(paramName)) {
+        this.#state[paramName] = searchParams.get(paramName);
+      }
+    }
+
+    // We need to get some query parameters from the FxA server before we
+    // show the user any kind of form to create or sign-in to an account.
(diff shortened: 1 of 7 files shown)


mikeconley:
Switching reviewers since I think it's after @akatsoulas 's EOD.

kitsune/sumo/static/sumo/js/switching-devices-wizard-manager.js
```diff
@@ -8,9 +8,113 @@
...
+
+  /**
+   * Updates the internal state object by replacing it with a copy of newState.
+   *
+   * @param {object}
+   *   The new state object.
+   */
+  #updateState(newState) {
+    this.#state = Object.assign({}, newState);
+  }
```
hannajones, line 85:
```suggestion
  #updateState(newState) {
    this.#state = Object.assign(#this.state, newState);
  }
```

Might be nice to be able to pass in an objecting representing the diff, rather than having to always clone `this.#state` before modifying it

kitsune/sumo/static/sumo/js/switching-devices-wizard-manager.js
```diff
@@ -8,9 +8,113 @@
...
+    ]) {
+      if (searchParams.has(paramName)) {
+        this.#state.fxa[paramName] = searchParams.get(paramName);
+      }
+    }
+
+    // We need to get some query parameters from the FxA server before we
+    // show the user any kind of form to create or sign-in to an account.
+    // See https://mozilla.github.io/ecosystem-platform/relying-parties/reference/metrics-for-relying-parties#relying-party-hosted-email-form.
+    this.#requestMetricsParams();
```
hannajones, line 64: Curious if we need to add error handling in `#requestMetricsParams` - like If we hit some kind of error here do we want/need to prevent the user from even trying to move on to the next step and sign in to FxA?

hannajones (approved):
LGTM, thanks!

One small question about error handling, but that might be something we need to think about more generally/outside of this PR.

kitsune/sumo/static/sumo/js/switching-devices-wizard-manager.js
```diff
@@ -8,9 +8,113 @@
...
+      "entrypoint",
+      "entrypoint_variation",
+      "entrypoint_experiment",
+    ]) {
+      if (this.#state.fxa[paramName]) {
+        params.set(paramName, this.#state.fxa[paramName]);
+      }
+    }
+    let response = await window.fetch(
+      `${this.#fxaRoot}/metrics-flow?${params}`
```
escattone, line 108: @mikeconley Will `/metrics-flow` be the only path used with `this.#fxaRoot` within `SwitchingDevicesWizardManager`? I'm just wondering if I'll perhaps need to broaden our current [CSP `connect-src`](https://github.com/mozilla/kitsune/blob/d2a822f229a4426d73e42535be39cb816b75b965/kitsune/settings.py#L1221) setting?

escattone (approved):
r+

kitsune/sumo/static/sumo/js/switching-devices-wizard-manager.js
mikeconley: I think we probably want to err on the side of still letting the user go down the happy path. Like, the bad news is that we won't be able to measure the success rate of user's going through this flow in the event that something falls over... but I suspect that measurement problems shouldn't cause us to block the user from signing in.

kitsune/sumo/static/sumo/js/switching-devices-wizard-manager.js
mikeconley: Via an XHR? Yes, I believe so. I believe we'll end up submitting a `<form>` similar to [this one](https://github.com/mozilla/activity-stream/blob/06aeeb331e9dd497e4d115d0e6cba51b9b25b36c/content-src/asrouter/templates/StartupOverlay/StartupOverlay.jsx#L124C1-L144) to get the user into the sign-in flow, but I don't _believe_ that requires a connect-src CSP change. We may, however, need to add the FxA server to the form-action list: https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Security-Policy/form-action

mikeconley: I've filed a ticket to get the FxA server added to the form-action CSP list: https://github.com/mozilla/sumo/issues/1264

escattone: @mikeconley Good catch, thanks!

PR #5465, opened 2023-05-08
Form wizard setup

hannajones:
This patch sets up the basic slot mechanism of the `<form-wizard>` and adds a few simple tests to ensure the active slot updates when the steps change.

### DIFF
--- a/kitsune/sumo/static/sumo/js/form-wizard.js
+++ b/kitsune/sumo/static/sumo/js/form-wizard.js
@@ -1,0 +1,65 @@
+/**
+ * A custom element for displaying multi-step forms.
+ *
+ * Uses a named slot and "#activeStep" state to determine which step of the form
+ * should be shown. All child elements where the "name" attribute doesn't match
+ * the "#activeStep" will not be assigned to the named slot.
+ *
+ */
+export class FormWizard extends HTMLElement {
+  #steps = null;
+  #activeStep = null;
+
+  constructor() {
+    super();
+    let shadow = this.attachShadow({ mode: "open" });
+    let activeSlot = document.createElement("slot");
+    activeSlot.setAttribute("name", "active");
+    shadow.appendChild(activeSlot);
+  }
+
+  connectedCallback() {
+    if (this.activeStep) {
+      // Make sure the active step is shown.
+      this.#setActiveStepAttributes();
+    } else {
+      // If there's no active step, default to the first step.
+      this.activeStep = this.firstElementChild?.getAttribute("name");
+    }
+  }
+
+  get activeStep() {
+    return this.#activeStep;
+  }
+
+  set activeStep(name) {
+    this.#activeStep = name;
+    this.#setActiveStepAttributes();
+  }
+
+  get steps() {
+    return this.#steps;
+  }
+
+  set steps(nextSteps) {
+    this.#steps = nextSteps;
+    this.activeStep = nextSteps.find((step) => step.status === "active")?.name;
+  }
+
+  /**
+   * Set the slot attribute on all children to ensure only
+   * the active step is shown.
+   */
+  #setActiveStepAttributes() {
+    let { activeStep } = this;
+    for (let step of this.children) {
+      let name = step.getAttribute("name");
+      if (name === activeStep) {
+        step.slot = "active";
+      } else {
+        step.slot = "";
+      }
+    }
+  }
+}
+customElements.define("form-wizard", FormWizard);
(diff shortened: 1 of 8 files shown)


webpack/entrypoints.js
```diff
@@ -81,6 +81,7 @@ const entrypoints = {
   ],
   "wiki.editor": [
     "sumo/js/wiki.editor.js",
+    "sumo/js/form-wizard.js",
```
hannajones, line 84: @escattone I added this since we need the element to be able to accurately preview the wizard when editing, but let me know if there's a different way we want to handle this.

kitsune/sumo/static/sumo/js/tests/form-wizard-tests.js
```diff
@@ -0,0 +1,55 @@
...
+
+  it("should show a different step when the active step changes", () => {
+    let nextSteps = [
+      {
+        name: "first",
+        status: "done",
+      },
+      {
+        name: "second",
+        status: "active",
```
hannajones, line 44: @mikeconley curious what you think about possible statuses. It seems like we might actually have five possible states/statuses for form steps (though still tbd if we'll actually allow clicking back and forth between steps in the switching devices wizard):

- active
- completed, can be navigated back to
- completed, can't be navigated back to
- not completed, can be navigated to
- not completed, can't be navigated to

webpack/entrypoints.js
escattone: @hannajones Hm, I thought I had covered those additional places where the `switching-device.js` bundle should be loaded with the merge of https://github.com/mozilla/kitsune/pull/5462, so that you wouldn't need to do this, but I see that I missed the pages where the content can be previewed. Sorry about that.

I think we'd like to keep the JS for this only loaded for the pages that use your wizard (at least for now), so would you mind removing this and putting your `form-wizard.js` code within `device-migration-wizard.js`, and then I'll create a new PR that ensures that the `switching-device.js` bundle is loaded for all of the pages that allow the preview of content while editing? Does that make sense?

escattone (changes requested):
@hannajones This looks great, except see my comment below.

kitsune/sumo/static/sumo/js/form-wizard.js
```diff
@@ -0,0 +1,65 @@
...
+ * the "#activeStep" will not be assigned to the named slot.
+ *
+ */
+export class FormWizard extends HTMLElement {
+  #steps = null;
+  #activeStep = null;
+
+  constructor() {
+    super();
+    let shadow = this.attachShadow({ mode: "open" });
```
mikeconley, line 14: I suppose the open shadow root makes this easier for the automated test. I wonder - @escattone, do we have a way of knowing at runtime whether or not we're in the test environment? If so, we maybe can gain greater encapsulation if we have a closed shadow root here in the non-testing case.

kitsune/sumo/static/sumo/js/tests/form-wizard-tests.js
mikeconley: Per our last meeting with UX, it sounds like we're not going to support going back through the wizard for now. So I think the states are:

- active
- completed
- not yet unavailable

webpack/entrypoints.js
hannajones: That makes total sense, I reverted my change to `entrypoints.js`. Thanks for the clarification!

kitsune/sumo/static/sumo/js/tests/form-wizard-tests.js
hannajones: Actually disregard this - probably moot since we're not going to be navigating back and forth between steps. I updated the `<form-wizard>` to only rely on `activeStep` for now in order to start with the simplest possible implementation.

kitsune/sumo/static/sumo/js/form-wizard.js
escattone: @mikeconley @hannajones Hm, I'm not sure what's best. Maybe we could use https://www.npmjs.com/package/detect-mocha or create something similar of our own?

escattone (approved):

webpack/entrypoints.js
escattone: @hannajones I created https://github.com/mozilla/kitsune/pull/5466 to cover the preview cases. Thanks again for catching those!

kitsune/sumo/static/sumo/js/form-wizard.js
mikeconley: I wouldn't block this patch on that, but if you're willing to file a follow-up ticket, it might be worth pursuing just to maximize encapsulation and prevent unnecessary information leakage.

mikeconley (approved):

PR #5467, opened 2023-05-08
switching devices -- add stage & prod FxA servers to form-action CSP

escattone:
mozilla/sumo#1264


mikeconley (approved):
Looks good, thanks!

PR #5473, opened 2023-05-10
Create initial step computation mechanism and UA qualification mechanism for the SwitchingDevicesWizardManager. (#1268)

mikeconley:
* Adds UA string checking to make sure that the wizard is only able to be used if the client is a Firefox Desktop browser.
* Adds a series of step definitions within SwitchingDevicesWizardManager with specific exit conditions and a method for turning the internal state of the SwitchingDevicesWizardManager into a payload that can be used to configure the selected step.

### DIFF
--- a/kitsune/sumo/static/sumo/js/switching-devices-wizard-manager.js
+++ b/kitsune/sumo/static/sumo/js/switching-devices-wizard-manager.js
@@ -1,3 +1,5 @@
+import BrowserDetect from "./browserdetect";
+
 /**
  * This class is responsible for managing the state for a wizard that
  * guides a user through setting up a Firefox Account for the purposes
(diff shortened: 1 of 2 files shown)


escattone (approved):
r+

escattone:
@mikeconley While reviewing this, I saw the following CORS error on the metrics-flow call. Is that something that the FxA team is already planning on addressing?

image

mikeconley:
> Oops, never mind, I see you've already created an [issue](https://github.com/mozilla/sumo/issues/1265) for this!

Yep! This should be live in production on the 17th, but should be on FxA staging sometime later today.

kitsune/sumo/static/sumo/js/switching-devices-wizard-manager.js
```diff
@@ -18,15 +26,105 @@ export default class SwitchingDevicesWizardManager {
...
+   *   for configuring the displayed step.
+   */
+
+  // An array of SwitchingDevicesWizardSteps that are evaluated during each
+  // state change to determine which step we should put the user in with the
+  // <form-wizard>. This is intentionally public so that the steps can be
+  // unit tested individually.
+  steps = Object.freeze([
+    {
+      name: "sign-into-fxa",
```
hannajones, line 72: Just a note (maybe to self) - we'll need to make sure the `name` attributes in `hook_device_migration_wizard.html` match what we've got here in order to get things working. Doesn't necessarily have to be done here though.

hannajones (approved):
Looks great, thanks! One small note, but I think that's something I can handle in the PR where I add the `setState` method.

kitsune/sumo/static/sumo/js/switching-devices-wizard-manager.js
mikeconley: Thanks! Yeah, I put these in as placeholders for now. If we decide there are better names, updating these and the tests is fairly straight-forward.

PR #5474, opened 2023-05-10
Wizard step indicator

hannajones:
This PR adds a super basic/un-styled implementation of the step indicator piece of the `<form-wizard>`. I was originally going to create this as a separate component, but since it's relatively simple and more or less needs the same data as the rest of the wizard it felt like keeping the two components in sync might be more trouble than it was worth.

We aren't yet initializing `step` data for the wizard, but for testing purposes something like this can be added to `device-migration-wizard.js` in order to make the step indicator visible:

```js
let wizard = document.querySelector("#switching-devices-wizard");
let steps = [
  {
    name: "sign-into-fxa",
    status: "done",
    label: "Create an account",
  },
  {
    name: "configure-sync",
    status: "active",
    label: "Sync your data",
  },
  {
    name: "setup-new-device",
    status: "unavailable",
    label: "Set up your new device",
  },
];
wizard.steps = steps;
```

### DIFF
--- a/kitsune/sumo/static/sumo/js/form-wizard.js
+++ b/kitsune/sumo/static/sumo/js/form-wizard.js
@@ -1,14 +1,28 @@
 /**
  * A custom element for displaying multi-step forms.
  *
- * Uses a named slot and "#activeStep" state to determine which step of the form
- * should be shown. All child elements where the "name" attribute doesn't match
- * the "#activeStep" will not be assigned to the named slot.
+ * Uses a named slot and `#activeStep` state to determine which step of the form
+ * should be shown. All child elements where the `name` attribute doesn't match
+ * the `#activeStep` will not be assigned to the named slot.
+ *
+ * Can be initialized with step data via the `steps` setter.
+ *
+ * wizard.steps = [
+ *  { name: "first", status: "done", label: "foo" },
+ *  { name: "second", status: "active", label: "bar" },
+ *  { name: "third", status: "unavailable", label: "baz" },
+ * ];
+ *
+ * The active step can be changed by calling `setStep` and passing in a step
+ * name along with new data representing that step.
+ *
+ * wizard.setStep("third", { name: "third", status: "active", label: "baz" });
  *
  */
 export class FormWizard extends HTMLElement {
   #activeStep = null;
   #progressIndicator = null;
+  #stepIndicator = null;
   #steps = [];
 
   static get markup() {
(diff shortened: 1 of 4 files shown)


mikeconley (approved):
Looks good to me! Solid inline documentation too.

kitsune/sumo/static/sumo/js/form-wizard.js
```diff
@@ -69,9 +75,84 @@ export class FormWizard extends HTMLElement {
...
+   * @param {string} name - Name of the active step.
+   * @param {FormWizardStep} data - Data about the active step.
+   */
+  setStep(name, data) {
+    let nextSteps = this.#steps.map((step) => {
+      if (step.status === "active") {
+        return { ...step, status: "done" };
+      }
+      return step.name === name ? { ...step, ...data } : step;
+    });
```
mikeconley, line 123: I just realized we're not setting the active step here based on `name`. Is that something we planned for a follow-up?

hannajones: Do you mean doing something like this?

```js
{ ...step, ...data, status: "active" }
```

I could definitely do that here/as part of this patch. I wasn't totally sure if the status was going to be part of the `data` passed to `setStep`.

mikeconley: Ohhh, I see. I think it'd be simpler on the manager end of things if `setStep` also did the work of making the step associated with `name` active.

kitsune/sumo/static/sumo/js/form-wizard.js
```diff
@@ -69,9 +75,100 @@ export class FormWizard extends HTMLElement {
...
+  setStep(name, data) {
+    let currentStatus = this.#steps.find((step) => step.name === name)?.status;
+    let isUnavailable = currentStatus === "unavailable";
+    let nextSteps = this.#steps.map((step) => {
+      if (step.status === "active" && isUnavailable) {
+        return { ...step, status: "done" };
+      } else if (step.name === name && isUnavailable) {
+        return { ...step, ...data, status: "active" };
+      } else if (step.name === name) {
+        return { ...step, ...data };
```
mikeconley, line 131: I suppose when the form-step stuff gets built out, we won't need to attach `data` here? My understanding based on our conversation was that the data / payload would get delegated down to the associated `<form-step>` to do its setup.

mikeconley (approved):
Thanks!

kitsune/sumo/static/sumo/js/form-wizard.js
hannajones: Yeah I think that'll probably be the way to go. For now this made it possible to test and be sure `setStep` was doing what it needed to do. Once the `<form-step>` piece is in place I'll make the updates and change the tests to validate that `setState` gets called with `data` (or something along those lines).

PR #5475, opened 2023-05-11
First take on tabbed menu

akatsoulas:


kitsune/wiki/jinja2/wiki/includes/document_macros.html
```diff
@@ -411,4 +411,70 @@ <h3 class="card--title">
...
+  <section class="topic-list" id="kb-tab-all">
+    {% for document in all_docs %}
+      <article class="topic-article">
+        <div class="topic-article--icon-wrap">
+          <img class="topic-article--icon" src="{{ webpack_static('protocol/img/icons/reader-mode.svg') }}" alt="" />
+        </div>
+        <div class="topic-article--text">
+          <h2 class="sumo-card-heading">
+            <a href="{{ document['url'] }}" data-event-category="link click" data-event-action="document"
+              data-event-label="{{ document['document_title'] }}">
```
escattone, line 449: What do you think about using the following instead?
- `data-event-category="device-migration-tabbed-panel"`
- `data-event-action="link-click"`
- `data-event-label="{{document.title}}"`

kitsune/wiki/jinja2/wiki/includes/document_macros.html
```diff
@@ -411,4 +411,70 @@ <h3 class="card--title">
...
+        <a class="tabs--link is-active" href="#kb-subtopic-all" data-event-category="link click">
+          <span>{{ _('View All') }}</span>
+        </a>
+      </li>
+      {% for subtopic in subtopics %}
+        <li class="tabs--item">
+          <a class="tabs--link" id="kb-subtopic-{{ subtopic.title }}" 
+             href="#kb-subtopic-{{ subtopic.title }}" 
+             data-event-category="link click" 
+             data-event-action="subtopic sidebar" data-event-label="{{subtopic.title}}">
```
escattone, line 429: What do you think about using the following instead?
- `data-event-category="device-migration-tabbed-panel"`
- `data-event-action="tab-click"`
- `data-event-label="{{subtopic.title}}"`

kitsune/wiki/jinja2/wiki/includes/document_macros.html
```diff
@@ -411,4 +411,70 @@ <h3 class="card--title">
...
+    <section class="topic-list" id="tab-{{ subtopic.title }}">
+      {% for document in documents %}
+        <article class="topic-article">
+          <div class="topic-article--icon-wrap">
+            <img class="topic-article--icon" src="{{ webpack_static('protocol/img/icons/reader-mode.svg') }}" alt="" />
+          </div>
+          <div class="topic-article--text">
+            <h2 class="sumo-card-heading">
+              <a href="{{ document['url'] }}" data-event-category="link click" data-event-action="document"
+                data-event-label="{{ document['document_title'] }}">
```
escattone, line 470: Same here. What do you think about using the following instead?
- `data-event-category="device-migration-tabbed-panel"`
- `data-event-action="link-click"`
- `data-event-label="{{document.title}}"`

escattone (commented):
I know this work will probably change significantly, but I had the following comments so far.

kitsune/sumo/static/sumo/js/device-migration-wizard.js
```diff
@@ -3,6 +3,32 @@ import trackEvent from "sumo/js/analytics";
...
+  tabItems.forEach((item) => {
+    item.addEventListener("click", (ev) => {
+      toggleTabContent(ev);
+    });
+  });
+}
+
+function toggleTabContent(ev) {
+  let tabContent = document.getElementsByClassName("topic-list");
+  [].forEach.call(tabContent, (el) => {
```
escattone, line 23: Would this work instead?
```suggestion
  tabContent.forEach((el) => {
```

kitsune/sumo/static/sumo/js/device-migration-wizard.js
```diff
@@ -3,6 +3,32 @@ import trackEvent from "sumo/js/analytics";
...
 
-$(document).ready(function() {
+document.addEventListener("DOMContentLoaded", function () {
   new SwitchingDevicesWizardManager(document.querySelector("#switching-devices-wizard"))
+  kbTabsInit();
 });
+
+function kbTabsInit() {
+  document.getElementById("kb-tab-all").style.display = "block";
+  let tabItems = document.querySelectorAll('[data-event-category="link click"]');
```
escattone, line 13: Shouldn't this be just the tabs? As it's written, I think it includes all of the article links as well.

escattone:
```suggestion
  let tabItems = document.querySelectorAll('.tabs--link');
```

kitsune/sumo/static/sumo/js/device-migration-wizard.js
```diff
@@ -3,6 +3,37 @@ import trackEvent from "sumo/js/analytics";
...
+  [].forEach.call(tabContent, (el) => {
+    el.style.display = "none";
+  });
+
+  let tabItems = document.getElementsByClassName("tabs--link");
+  [].forEach.call(tabItems, (item) => {
+    item.classList.remove("is-active");
+  });
+
+  ev.target.parentElement.classList.add("is-active");
```
mikeconley, line 37: I think if I had any critique, it'd be here. This is the SUMO team's codebase, and so I defer to your styles, preferences and idioms - but what we tend to do on the Firefox side of things for stuff like this is put most of the hiding and showing rules in our CSS, and then use classes / attributes from script. So what I've highlighted here, we might have as:

```css
.topic-list:not(.is-active) {
  display: none;
}
```

```js
// Remove the active state from the previously selected tab / tab content
let oldSelectedTabContent = document.querySelector(".topic-list.is-active");
let oldSelectedTab = document.querySelector(".tabs--link.is-active");

oldSelectedTabContent.classList.remove("is-active");
oldSelectedTab.classList.remove("is-active");

// Put the newly selected tab / tab content into the active state.
let selectedTab = ev.target.parentElement;
let selectedTabContent = document.querySelector(`#tab-${ev.target.parentElement.dataset.eventLabel}`);

selectedTab.classList.add("is-active");
selectedTabContent.classList.add("is-active");
```

(I haven't tested this code, but hopefully it conveys the general idea)

I think the reason why we tend to avoid setting styles inline with script is because it keeps CSS as the domain where we declare the style characteristics of things, which helps separate concerns a little bit. (Inline CSS from script is also more difficult to override from stylesheets without hacks like `!important`.)

This is a minor critique and, I think, ultimately a matter of preference and convention. I wouldn't block on my suggestion.

smithellis: I think that what @mikeconley says here makes sense; separating concerns is never a bad move.

kitsune/wiki/jinja2/wiki/includes/document_macros.html
```diff
@@ -411,4 +411,71 @@ <h3 class="card--title">
...
+
+{% macro kb_subtopic_documents(documents, tab_content_id, is_fallback=False) %}
+  {% for document in documents %}
+      <article class="topic-article">
+        <div class="topic-article--icon-wrap">
+          <img class="topic-article--icon" src="{{ webpack_static('protocol/img/icons/reader-mode.svg') }}" alt="" />
+        </div>
+        <div class="topic-article--text">
+          <h2 class="sumo-card-heading">
+            <a {% if is_fallback %} href="{{ document['url']|replace('/'+ settings.WIKI_DEFAULT_LANGUAGE +'/', '/' + request.LANGUAGE_CODE + '/') }}" {% else %} href="{{ document['url'] }}" {% endif %}
```
smithellis, line 424: I like this - more elegant than what I threw down.

smithellis (approved):
Everything functions and all tests pass. Code looks good overall.

mikeconley (approved):
Yeah, to be clear, I approve the change. :) The things I mentioned are either fodder for follow-up or can be dropped if they're not in keeping with Kitsune conventions.

escattone (approved):
r+wc

kitsune/sumo/static/sumo/js/device-migration-wizard.js
escattone: Ah, I think `getElementById(id)` handles spaces in the `id` just fine, but if I'm not mistaken I think `querySelector(cssSelector)` would not if the `cssSelector` contained spaces.

escattone: So I guess my concern boils down to this:
- Using the `subtopic.title` within HTML element `id`s may introduce spaces into the `id`s
- Spaces in `id`s make them difficult/problematic to query -- you'd have to know to escape them if querying by CSS selector, or use `getElementById` instead

akatsoulas: Thanks @mikeconley that's great feedback. I restructured the HTML a bit to make room for the snippet above. It's much cleaner now.

PR #5476, opened 2023-05-11
Use UITour in SwitchingDevicesWizardManager to get FxA signed-in status (#1276)

mikeconley:
This is based on the current changes in #5474, and I'll wait for that to land before rebasing and requesting review for this.


escattone (approved):
LGTM! r+

mikeconley:
> I'm definitely getting a flash of the first step on page load before I get moved to the second step. Not a huge deal for now but it might be worth re-visiting a loading state down the line.

Agreed. We might want to alert Tati to this now that we've got something that's somewhat less theoretical.

PR #5477, opened 2023-05-11
Have SwitchingDevicesWizardManager attempt to query for the SUMO user email address. (#1277)

mikeconley:
For users that are already signed into SUMO, we can pre-populate the
email address field for signing into / creating a Firefox Account.
This patch adds that capability.


mikeconley:
This is based on mikeconley:uitour-usage. I'll publish this for review once that other one gets reviewed and landed.

escattone (approved):
@mikeconley Just reviewed the SUMO graphQL request for the current user's email, which I tested locally, and it looks good!

hannajones (approved):
LGTM! Seeing my fake user's email in step 1 with the latest changes from `main` applied 🥳

PR #5479, opened 2023-05-12
Add BaseFormStep and sketch in the migration wizard form steps

hannajones:
This PR adds a `BaseFormStep` element that is intended to be used in tandem with `FormWizard` to create and update custom form steps.

I'll file follow up tickets for filling in the specific markup/logic we need for the three form steps. For now I've just implemented something super basic so we can see that they're working.


mikeconley (approved):
This works really well. Great job, @hannajones!

PR #5480, opened 2023-05-12
Add form wizard specific styles

hannajones:
I'm not totally set on this approach, but I wanted to see if it would be possible to load an external stylesheet into the `<form-wizard>` for styling. It seems doable in dev with minimal webpack related tweaks. My big hesitation here is I think we'll ultimately need some way to get the built file name, since in production the filenames follow the pattern `[name].[contenthash].css`, and we would need to know what that hashed filename is. There's something similar already happening for images via the [asset-json-plugin](https://github.com/mozilla/kitsune/blob/main/webpack/asset-json-plugin.js), which provides a mapping of image paths to built filenames. Maybe we could re-purpose that for css files?

I didn't go down the rabbit hole of figuring that out yet. I wanted to run this by people and see if there's maybe a simpler approach - like maybe we don't worry about the cache busting hash for `form-wizard.css`? Or maybe we just go with embedded `<style>` tags for the wizard + steps, at least for now. Open to thoughts/feedback/suggestions.

The good news is the styles I wrote will work either way, and the wizard now looks like this locally:

Screenshot 2023-05-12 at 3 50 03 PM


kitsune/sumo/static/sumo/js/form-wizard.js
```diff
@@ -46,10 +44,17 @@ export class FormWizard extends HTMLElement {
     super();
     let shadow = this.attachShadow({ mode: "open" });
 
+    let stylesheet = document.createElement("link");
+    stylesheet.setAttribute("rel", "stylesheet");
+    stylesheet.setAttribute(
+      "href",
+      this.getAttribute("static-url") + "form-wizard-styles.css"
```
hannajones, line 51: This is the part I think will be problematic in production, since the filename will be `form-wizard-styles.[some-hash].css`

kitsune/sumo/static/sumo/scss/form-wizard.scss
```diff
@@ -0,0 +1,45 @@
...
+  width: 100%;
+  height: 8px;
+  appearance: none;
+  border: none;
+  border-bottom-left-radius: 8px;
+  border-bottom-right-radius: 8px;
+}
+
+progress::-moz-progress-bar,
+progress::-webkit-progress-value {
```
mikeconley, line 43: Kinda unfortunate that we still have to do this sort of thing. :/

kitsune/sumo/static/sumo/scss/form-wizard.scss
```diff
@@ -0,0 +1,45 @@
...
+  line-height: 1;
+  color: var(--color-heading);
+}
+
+ul {
+  margin-inline-end: 48px;
+}
+
+section {
+  display: flex;
```
mikeconley, line 30: It looks like this needs to be rebased on top of your BaseFormStep stuff - but we probably want to have those BaseFormStep's flex too, if they're not already doing that. Maybe with a class or something?

kitsune/sumo/static/sumo/scss/form-wizard.scss
```diff
@@ -0,0 +1,45 @@
...
+}
+
+progress {
+  width: 100%;
+  height: 8px;
```
mikeconley, line 35: If we change this to `max-height: 8px;`, then we can use `flex: 1` to flex to the total width. I think we generally prefer using flex rules rather than percentages when we can use them.

kitsune/sumo/static/sumo/scss/form-wizard.scss
```diff
@@ -0,0 +1,45 @@
...
+
+ul {
+  margin-inline-end: 48px;
+}
+
+section {
+  display: flex;
+}
+
+progress {
```
mikeconley, line 33: We should reach out to Tati and make sure this wasn't meant to animate. The progress bar in about:welcome, for example, is constructed using two <div>'s and some clever CSS transition properties. I don't think we need to worry about that for this version, but it might be worth asking about as follow-up work.

mikeconley (approved):
I think this needs a rebase so I can see it more in context with the other changes that have gone into main. I have some small suggestions, but otherwise the CSS looks good to me. Thanks!

escattone (commented):
After thinking about this, my vote would be to use internal styles for the shadow DOM of the form-wizard instead of an external file, but I'll happily defer to anyone with a better argument than mine, which is simply that it seems easiest and all other options don't seem worth the effort. 😄

kitsune/sumo/static/sumo/scss/form-wizard.scss
hannajones: Good call! Left a comment on the Figma mocks to start the discussion there: https://www.figma.com/file/qe9yX8x51jjbt6DJYgayg2?node-id=3362:38070#444445016

kitsune/sumo/static/sumo/scss/form-wizard.scss
hannajones: For easier tracking (since this all got moved to a different file) - I've changed the rule to:

```css
section,
::slotted([slot="active"])  {
    display: flex;
    flex: 1;
}
```

Let me know if you had something else in mind!

escattone (approved):
Thanks @hannajones!

mikeconley:
Yep, let's go with this. Thanks!

PR #5486, opened 2023-05-15
Implement configure step for the switching-devices <form-wizard>.

mikeconley:


kitsune/sumo/static/sumo/js/device-migration-wizard.js
```diff
@@ -123,21 +125,133 @@ class ConfigureStep extends BaseFormStep {
     return `
       <template>
         <div>
-          <p>This is where the sync step will go.</p>
-          <p id="sync-status"></p>
+          <p id="header">
+            <img class="icon" src="${infoImageURL}" aria-hidden="true"></img>
+            <span>${gettext("You are now logged in to Firefox Accounts")}</span>
+            <a href="#">${gettext("Forgot password?")}</a>
```
escattone, line 131: @mikeconley Do you think we should report clicks on this to GA? If so, I guess we could use:
- `data-event-category="device-migration-wizard"`
- `data-event-action="click"`
- `data-event-label="forgot-password"`

mikeconley: Good idea. I'll add that.

mikeconley:
Based on a conversation with @hannajones, I'm going to rejig this a little bit to use an external style sheet, based on the work in #5489.

escattone (approved):
r+

mikeconley:
This pull request also includes a patch to move the sign-in-step's CSS into a separate external stylesheet.

kitsune/sumo/static/sumo/js/device-migration-wizard.js
```diff
@@ -122,21 +118,78 @@ class ConfigureStep extends BaseFormStep {
...
+  }
+
   render(prevState, state) {
     if (this.state.syncEnabled !== prevState.syncEnabled) {
       let statusEl = this.shadowRoot.getElementById("sync-status");
       statusEl.textContent = this.state.syncEnabled
-        ? "Syncing: ON"
-        : "Syncing: OFF";
+        ? gettext("ON")
+        : gettext("OFF");
```
hannajones, line 164: Do we want to do this via styles / `text-transform: uppercase` instead? Seems slightly safer than relying on translations being all caps.

kitsune/sumo/static/sumo/scss/form-wizard-configuration-step.styles.scss
```diff
@@ -0,0 +1,55 @@
...
+  font-weight: 700;
+  color: var(--color-white);
+  border-radius: 34px;
+  padding: 2px 16px;
+  margin-inline-start: 12px;
+
+  /**
+   * TODO: Should this be in the color swatches somewhere?
+   */
+  background-color: #E22850;
```
hannajones, line 32: Seems like that's a Protocol color but the Protocol colors aren't available 🤔

I wonder if we should go with --color-warning: https://github.com/mozilla/kitsune/blob/main/kitsune/sumo/static/sumo/scss/config/_project-theme.scss#L12

Otherwise we can look into bringing in the Protocol colors for the form steps.

hannajones (approved):
This is great, thanks for moving the CSS! One question that's more for my own learning/clarity.

kitsune/sumo/static/sumo/js/device-migration-wizard.js
escattone: @hannajones Good catch!

mikeconley: Oh, that's a great idea. Thanks, will do before landing!

kitsune/sumo/static/sumo/scss/form-wizard-configuration-step.styles.scss
mikeconley: `--color-error` is probably closer to the red that Tati was looking for. I'll go with that, and file a follow-up bug for trying to bring in some more colours from Protocol.

kitsune/sumo/static/sumo/js/device-migration-wizard.js
```diff
@@ -123,21 +125,133 @@ class ConfigureStep extends BaseFormStep {
...
+        align-self: end;
+      }
+    `;
+    return style;
+  }
+
+  constructor() {
+    super();
+    let buttons = this.shadowRoot.querySelector("#buttons");
+    buttons.addEventListener("click", this);
```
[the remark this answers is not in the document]
hannajones: Gotcha that makes sense - so the cleanup is mostly important if we're adding listeners to any elements outside of the web component itself. Thanks for clarifying!

PR #5488, opened 2023-05-16
Fix tab js hitbox - 1290

smithellis:
This fixes the issue where the wrong selector was being called on tab clicks.  I also added the toggling feature directly to the event listener; which is a little more consolidated, but I'm open to feedback on whether it makes things less readable. 

Also also, I thought about pulling the `kbTabsInit();` call out of the` eventListener` code for `SwitchingDevicesWizardManager `and just add an additional `DOMcontentLoaded` listener just for tabs, but couldn't decide if that mattered - on the one hand, moving it out would get our tab code all the way out of the device switching code, but on the other hand it would be a little redundant.

Thoughts appreciated, but just plain ole' approving is also accepted.


escattone (approved):
@smithellis Nicely done! I like the way you solved this, getting rid of the need to deal with the event.

PR #5489, opened 2023-05-16
Use external CSS files from .styles.scss source

hannajones:
This PR provides a potential solution for using external stylesheets in our wizard web components. The webpack config changes basically:

* takes `.scss` source code and translates it to CSS using `sass-loader`
* specifies `type: "asset/resource"` which will emit a separate file and export the URL of that generated file by default
* specifies that the naming structure for the file that gets emitted to ensure it's a `.css` file

Once we've generated the CSS file we can import the `.scss` source in our web component to get the URL of the built file, then create a link pointing to that URL.

This whole process doesn't seem to work if we run things through `MiniCssExtractPlugin.loader`, so I've given the file a `.styles.scss` extension so that we can specify different Webpack rules for CSS we want to load into web components. (This was inspired by [this article](https://dev.to/m4thieulavoie/how-i-managed-to-use-scss-inside-web-components-3lk9) - I'm open to other naming conventions.)

I'm not 100% sure this is going to work in production environments, but it's working well for me locally, so seems like a starting point at least.

--

As a little experiment to check that we could load Protocol styles, I copied the contents of `_protocol.scss` into `form-wizard.styles.scss` and added an import for the protocol button component. This was the result:

Screenshot 2023-05-16 at 12 48 09 PM

### DIFF
--- a/kitsune/sumo/static/sumo/js/form-wizard.js
+++ b/kitsune/sumo/static/sumo/js/form-wizard.js
@@ -1,3 +1,5 @@
+import wizardStylesURL from "../scss/form-wizard.styles.scss";
+
 /**
  * A custom element for displaying multi-step forms. Designed to be used with
  * child elements that inherit from `BaseFormStep`, or which implement a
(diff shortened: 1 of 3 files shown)


mikeconley (approved):
This works great for me! Let's hope it works on staging!

PR #5490, opened 2023-05-16
Have alternative links from switching devices manager pass redirect_immediately=true.

mikeconley:
This was also being hardcoded into the <form>, but for consistency sake, is now read and set from the state object sent from the SwitchingDevicesWizardManager.


hannajones (approved):
Thanks!

PR #5491, opened 2023-05-17
Add common form step styles, bring protocol elements into steps

hannajones:
This PR provides a way to pull Protocol styles/other common styles into all of our form steps via `BaseFormStep`. It also adds some styling/polish to the existing form steps. Here's how things are looking now:

Screenshot 2023-05-17 at 11 02 13 AM

Screenshot 2023-05-17 at 11 02 00 AM

### DIFF
--- a/kitsune/sumo/static/sumo/scss/form-wizard-configuration-step.styles.scss
+++ b/kitsune/sumo/static/sumo/scss/form-wizard-configuration-step.styles.scss
@@ -1,3 +1,8 @@
+.configure-step-wrapper {
+  display: flex;
+  flex-direction: column;
+}
+
 #header,
 #sync-status-container {
   display: flex;
@@ -7,6 +12,7 @@
 #header {
   padding-block-end: 12px;
   border-bottom: 1px solid var(--color-light-gray-03);
+  font-size: 0.8125rem;
 }
 
 .icon {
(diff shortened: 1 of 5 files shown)


hannajones:
@mikeconley I just realized protocol doesn't seem to provide the `:invalid` form styles from the mockups 🤔 Might take a stab at that here, in which case more changes incoming

kitsune/sumo/static/sumo/scss/form-wizard-configuration-step.styles.scss
```diff
@@ -54,4 +63,16 @@
 
 #next {
   align-self: end;
+  margin-block-end: 16px;
+  margin-block-start: 8px;
```
mikeconley, line 67:
```suggestion
  margin-block: 8px 16px;
```

mikeconley (approved):
Brilliant as usual - just one small suggestion before landing. Thanks @hannajones!

hannajones:
Small update - I filed a separate ticket for the custom validation message, since it's a decent chunk of work and should probably be reviewed separately: https://github.com/mozilla/sumo/issues/1305

hannajones:
> > @mikeconley I just realized protocol doesn't seem to provide the `:invalid` form styles from the mockups 🤔 Might take a stab at that here, in which case more changes incoming
> 
> Ah, just saw this part of the pull request. Okay, I'll re-review once you add these.

I'm actually going to tackle those changes in a separate patch - there's a decent chunk of JS work involved so it feels a little distinct from these other style changes

PR #5492, opened 2023-05-17
Configure step improvements

mikeconley:
This pull request fixes mozilla/sumo#1303, and also fixes the issue where the "Forgot password" link wasn't having its href attribute set. This also splits out the configure step into its own module, to make it easier to test in isolation.


kitsune/sumo/static/sumo/js/tests/form-wizard-configure-step-tests.js
```diff
@@ -0,0 +1,82 @@
...
+    target.addEventListener(eventName, resolve, { once: true });
+  })
+}
+
+describe("configure-step custom element", () => {
+  let step;
+  let slot;
+
+  before(() => {
+    customElements.define("configure-step", ConfigureStep);
```
hannajones, line 15: do we want to just do the `customElements.define` once in `form-wizard-configure-step.js` instead of both here and in `device-migration-wizard.js`? I think we could then this in `device-migration-wizard.js`:

```js
import "sumo/js/form-wizard-configure-step";
```

and everything would just work™️ 

curious if there's another reason for delaying the `customElements.define` call

hannajones (commented):
This is great! I did hit one case that we might want/need to handle still. I took these steps:

* from step 2 with sync enabled, clicked "Next" to get taken to step 3
* opened about:preferences > sync in a different tab and disabled syncing
* went back to the switching devices article, and I was still on step 3, but maybe we should actually kick users back to step 2 with the "Next" button disabled?

If I understand correctly there's not much value in them using the link in step 3 if sync is turned off. Also, if I refresh the page after disabling sync I'm taken back to step 2.

kitsune/sumo/static/sumo/js/tests/form-wizard-configure-step-tests.js
mikeconley: Ah, yes, I can do that. I just recall us doing things in Firefox-land where we try to avoid side-effects from imports: https://searchfox.org/mozilla-central/rev/45b20d00311bc2170dfbdae8a91bc4dca29717d8/browser/components/storybook/stories/migration-wizard.stories.mjs#7

but that's all just old reflex. What you're suggesting makes sense.

mikeconley:
> This is great! I did hit one case that we might want/need to handle still. I took these steps:
> 
>     * from step 2 with sync enabled, clicked "Next" to get taken to step 3
> 
>     * opened about:preferences > sync in a different tab and disabled syncing
> 
>     * went back to the switching devices article, and I was still on step 3, but maybe we should actually kick users back to step 2 with the "Next" button disabled?
> 

> 
> If I understand correctly there's not much value in them using the link in step 3 if sync is turned off. 

Ah yes, I'll fix that here, thanks.

> Also, if I refresh the page after disabling sync I'm taken back to step 2.

That's fine - if the user is signed in but not syncing, I think we want to show them Step 2 so that they can enable sync.

hannajones (approved):
Thanks!

PR #5494, opened 2023-05-17
use document for label on tabbed-menu link clicks

escattone:
I was reviewing our GA calls, and noticed that for `link-click` actions within the `device-migration-tabbed-menu` category we're reporting the topic's slug as the label. These are clicks on document links within any of the tabs, so it seems better to report the document rather than the topic, so either the document's title or slug.

### DIFF
--- a/kitsune/wiki/jinja2/wiki/includes/document_macros.html
+++ b/kitsune/wiki/jinja2/wiki/includes/document_macros.html
@@ -427,9 +427,9 @@
           <div class="topic-article--text">
             <h2 class="sumo-card-heading">
               <a {% if is_fallback %} href="{{ document['url']|replace('/'+ settings.WIKI_DEFAULT_LANGUAGE +'/', '/' + request.LANGUAGE_CODE + '/') }}" {% else %} href="{{ document['url'] }}" {% endif %}
-                data-event-category="device-migration-tabbed-menu" 
+                data-event-category="device-migration-tabbed-menu"
                 data-event-action="link-click"
-                data-event-label="{{ active_topic_slug }}">
+                data-event-label="{{ document['document_title'] }}">
                 {{ document['document_title'] }}
               </a>
             </h2>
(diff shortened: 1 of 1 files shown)


kitsune/wiki/jinja2/wiki/includes/document_macros.html
```diff
@@ -444,18 +444,18 @@ <h2 class="sumo-card-heading">
...
-        data-event-action="tab-click" 
+        data-event-action="tab-click"
         data-event-label="view-all">
         <span>{{ _('View All') }}</span>
       </button>
       </li>
       {% for subtopic in subtopics %}
         <li class="tabs--item">
-          <button class="tabs--link" id="kb-subtopic-{{ subtopic.title }}" 
+          <button class="tabs--link" id="kb-subtopic-{{ subtopic.title }}"
```
akatsoulas, line 456: In the initial implementation of the tabbed menu, we were using the `subtopic.title` but we swapped that for the `slug` b/c of the [concern](https://github.com/mozilla/kitsune/pull/5475#discussion_r1192606494) that introducing spaces can be prone to errors. I agree that title is better especially for GA, I just want to make sure that we don't foresee any problems with the spaces.

kitsune/wiki/jinja2/wiki/includes/document_macros.html
```diff
@@ -427,9 +427,9 @@ <h2 class="sumo-page-subheading">{{ _('In English') }}</h2>
           <div class="topic-article--text">
             <h2 class="sumo-card-heading">
               <a {% if is_fallback %} href="{{ document['url']|replace('/'+ settings.WIKI_DEFAULT_LANGUAGE +'/', '/' + request.LANGUAGE_CODE + '/') }}" {% else %} href="{{ document['url'] }}" {% endif %}
-                data-event-category="device-migration-tabbed-menu" 
+                data-event-category="device-migration-tabbed-menu"
                 data-event-action="link-click"
-                data-event-label="{{ active_topic_slug }}">
+                data-event-label="{{ document['document_title'] }}">
```
akatsoulas, line 432: In the initial implementation of the tabbed menu, we were using the `subtopic.title` but we swapped that for the `slug` b/c of the [concern](https://github.com/mozilla/kitsune/pull/5475#discussion_r1192606494) that introducing spaces can be prone to errors. I agree that title is better especially for GA, I just want to make sure that we don't foresee any problems with the spaces.

kitsune/wiki/jinja2/wiki/includes/document_macros.html
```diff
@@ -444,18 +444,18 @@ <h2 class="sumo-card-heading">
...
       </button>
       </li>
       {% for subtopic in subtopics %}
         <li class="tabs--item">
-          <button class="tabs--link" id="kb-subtopic-{{ subtopic.title }}" 
+          <button class="tabs--link" id="kb-subtopic-{{ subtopic.title }}"
              data-event-category="device-migration-tabbed-menu"
-             data-event-action="tab-click" 
+             data-event-action="tab-click"
              data-event-label="{{subtopic.slug}}">
```
akatsoulas, line 459: Depending on the answer in the previous comment, we also have a `slug` here

kitsune/wiki/jinja2/wiki/includes/document_macros.html
escattone: @akatsoulas I had the same concern, thinking that perhaps we should use the document slug instead. It looks like we've used the document title for other links in the past for GA, but I still wonder about it.

PR #5496, opened 2023-05-18
Use slug instead of title in id

akatsoulas:
If we introduce a subtopic with spaces, the JS part will break.


escattone (approved):
Good catch, thanks!

PR #5499, opened 2023-05-18
Add disqualified state for the form-wizard. (#1274)

mikeconley:


kitsune/sumo/static/sumo/js/form-wizard.js
```diff
@@ -150,8 +160,15 @@ export class FormWizard extends HTMLElement {
   }
 
   // To be implemented as part of mozilla/sumo #1274.
```
hannajones, line 162:
```suggestion
```

Done!

kitsune/sumo/static/sumo/js/form-wizard.js
```diff
@@ -34,14 +35,23 @@ export class FormWizard extends HTMLElement {
...
+            <h2>${gettext("Wizard header")}</h2>
+            <section>
+              <ul id="step-indicator"></ul>
+              <slot name="active"></slot>
+            </section>
+          </div>
+          <progress max="100" value="0"></progress>
+          <div class="form-wizard-disqualified">
+            <div class="disqualification">
+              <img class="warning-icon" src="${warningImageURL}"/>
```
hannajones, line 49:
```suggestion
              <img class="warning-icon" src="${warningImageURL}" aria-hidden="true" />
```

Let's keep this icon thing going

kitsune/sumo/static/sumo/js/form-wizard.js
```diff
@@ -34,14 +35,23 @@ export class FormWizard extends HTMLElement {
...
+              <span class="disqualification-header"></span>
```
hannajones, line 50: Screenshot 2023-05-18 at 5 41 11 PM


Splitting hairs here a bit, but the icon isn't quite vertically centered with the text. Maybe worth wrapping them in a flex container?

hannajones (approved):
Thanks! Couple of small suggestions, but this looks great! I even tested it in Chrome and we're dong what we need to do there 😎

mikeconley:
Hey @hannajones! Thanks for the review. I think I pushed a slight alteration just as you were publishing it! I've changed things slightly, where now the strings for the disqualification messages are defined in the markup.

I also added the UITour-is-broken disqualification message even though it isn't being used just yet - mainly to optimize for getting our strings into the tree. I'll wire it up to UITour failure tomorrow.

hannajones (approved):
LGTM, thanks!

kitsune/sumo/static/sumo/js/form-wizard.js
```diff
@@ -34,18 +35,31 @@ export class FormWizard extends HTMLElement {
...
+          </div>
+          <progress max="100" value="0"></progress>
+          <div class="form-wizard-disqualified">
+            <div class="disqualification" reason="need-fx-desktop">
+              <span class="disqualification-header"><img class="warning-icon" src="${warningImageURL}" aria-hidden="true"/>${gettext("Use Firefox to continue")}</span>
+              <div class="disqualification-message">${gettext("To use the setup assistant or access the settings for backing up your Firefox data, please visit this page using Firefox on a desktop device.")}</div>
+            </div>
+
+            <div class="disqualification" reason="uitour-broken">
+              <div class="disqualification-message">${gettext("The setup assistant is currently unavailable for your version of Firefox. However, you can still perform a manual backup of your data by following the steps outlined in")} <a href="/kb/profile-manager-create-remote-switch-firefox-profiles">${gettext("this article.")}</a></div>
```
flodolo, line 54: Please never do this.
https://mozilla-l10n.github.io/documentation/localization/dev_best_practices.html#avoid-concatenations-use-placeholders-instead

These two strings landed as separate strings with zero context, I assume concatenation only because `this article.` stars with lowercase and ends with a period.

This is effectively impossible to localize if a language requires "this article" in the middle of the other sentence.

mikeconley: Thanks, @flodolo. What is the correct way to localize this in the JavaScript layer of this Django application?

escattone: @flodolo 👍 That makes sense. I'm at fault for recommending this approach.

hannajones: hey @flodolo let me know if this approach using `interpolate` helps address the issue: https://github.com/mozilla/kitsune/pull/5517

if not I'm open to other suggestions!

PR #5500, opened 2023-05-18
Implement the setup device form step

hannajones:
This PR adds the third and final setup device form step. There are a handful of design things I think will need to be followed up on, specifically:

* there's no feedback when the user clicks on the "copy link" button
* the contrast between the tooltip and the white background probably isn't accessible

Those still need design input, so this seems alright as a first pass.

Screenshot 2023-05-18 at 3 22 03 PM

### DIFF
--- a/kitsune/sumo/static/sumo/js/form-wizard-setup-device-step.js
+++ b/kitsune/sumo/static/sumo/js/form-wizard-setup-device-step.js
@@ -7,9 +7,10 @@ export class SetupDeviceStep extends BaseFormStep {
     return `
       <template>
         <div>
-          <h3>${gettext("Congratulatory headline they’ve done all they need to do on THIS device")}</h3>
+          <h3>${gettext("You’re all set on this device!")}</h3>
           <ul>
-            <li>${gettext("Let user know this is the point in the flow they will need to access their new device - if they don’t have it yet, that’s okay but.. (give next step - come back here when __ , do ___ until")}</li>
+            <li>${gettext("To continue with the download and sync process on your new device, click the copy-paste button and save the link in a place that you will remember, such as a note-taking app or send it to your email. When you're on your new device, simply paste the download link into the default browser of your new device to download Firefox and we'll be there to assist you with the next steps.")}</li>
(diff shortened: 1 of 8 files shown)


kitsune/sumo/static/sumo/js/form-wizard-setup-device-step.js
```diff
@@ -1,19 +1,63 @@
...
+    let copyButton = this.shadowRoot.getElementById("copy-button");
+    copyButton.removeEventListener("click", this.copyLink);
+  }
+
+  copyLink() {
+    let downloadLink = this.shadowRoot.getElementById("download-link");
+    navigator.clipboard.writeText(downloadLink.href);
+  }
+
   render() {
```
hannajones, line 59: @mikeconley this is kind of an unfortunate side effect of having the `render` method in `BaseFormStep` throw an error...do you think it would be better if we made it a noop there so that implementing `render` in the subclasses is optional?

kitsune/sumo/static/sumo/scss/form-step.styles.scss
```diff
@@ -19,41 +19,60 @@
...
-.error-tooltip:not(.hidden) + input:invalid,
-input:not(:focus):not(:placeholder-shown):invalid {
-  border: 2px solid p.$color-red-60;
-}
-
-.error-tooltip:not(.hidden) + input:focus:invalid {
-  box-shadow: 0 0 0 2px rgba(226, 40, 80, 0.5);
+.visible {
+  visibility: visible;
 }
```
mikeconley, line 24: This should probably go after the .tooltip rule, because it's being overridden by the `visibility: hidden;` rule on line 38, even with the visible class.

mikeconley (approved):
Hi, @hannajones! This looks great! Just a few things to address before landing:

1. The setup device step strings have settled, so let's get those in now. See [the doc here](https://docs.google.com/document/d/1bVcQz5tDDfbPf9t6zXfgjmHfkNqXfWW_xBqw4WV5TTw/edit) for reference.
2. The tooltip for the sign in step does not appear anymore due to a CSS priority bug (see below).

With those addressed, I think you're good to land!

PR #5501, opened 2023-05-18
Land first batch of finalized copy for the switching devices wizard

mikeconley:


kitsune/sumo/static/sumo/js/form-wizard-sign-in-step.js
```diff
@@ -57,7 +57,7 @@ export class SignInStep extends BaseFormStep {
 
           <p class="warning">
             <img class="key-icon" src="${keyImageURL}" aria-hidden="true"></img>
-            ${gettext("Provide any warnings about unexpected password situations (like noted in mock)")}
+            ${gettext("Make sure to create a recovery key in case you need to reset your password. For security purposes, Mozilla isn't able to retain your synced data when you reset your password without one. If you must reset your password without a recovery key, only do so on the machine that has the data you want re-synced.")}
```
mikeconley, line 60: Hey @escattone - so one of the things that I think UX / Content design wants to do is to link to another document from here. Specifically, to this page:

https://support.mozilla.org/en-US/kb/reset-your-firefox-account-password-recovery-keys

and to have the term "recovery key" link to it. How do we do that that is localizable?

kitsune/sumo/static/sumo/scss/form-wizard-sign-in-step.styles.scss
```diff
@@ -1,11 +1,17 @@
+h3 {
+  font-size: 1rem;
+  line-height: 1.5;
```
hannajones, line 3: This looks very similar to `h3` styles in the setup device step (I was also using a 26px margin there, but changed it based on how things looked in the copy doc). Maybe we move this to `form-step-styles.scss`?

hannajones (approved):
🚢

kitsune/sumo/static/sumo/js/form-wizard-sign-in-step.js
mikeconley: Sorted this out with @escattone on Slack.

kitsune/sumo/static/sumo/scss/form-wizard-sign-in-step.styles.scss
mikeconley: Good idea!

PR #5503, opened 2023-05-19
Style the form wizard step indicator

hannajones:
This PR adds styling to the form wizard step indicator, as well as some extra space to the bottom of the third form step to avoid jarring height changes between the steps.

Screenshot 2023-05-19 at 10 51 12 AM


hannajones:
@mikeconley hadn't finished my coffee when I pushed this earlier - just realized this is missing some disabled styling + bold text for the active step 😓 working on that now!

mikeconley (approved):
My local FxA setup is kinda borked (working with the FxA crew on that), so I had to do some console trickery to view each of the states - so I wasn't able to "properly" test this end to end, but from inspection this looks great!

PR #5506, opened 2023-05-19
Change copy button styles

hannajones:
The copy button styles in the setup device step changed pretty significantly in order to provide feedback to the user when they click on the button. This PR applies those changes, and reverts some of the work I did with the tooltip since it's no longer being re-used.

Screenshot 2023-05-19 at 12 15 40 PM
Screenshot 2023-05-19 at 12 15 37 PM


mikeconley (approved):
Wonderful, thank you!

PR #5507, opened 2023-05-19
Copy updates

hannajones:
Took another look at the doc and only found one word change and an extra whitespace.

### DIFF
--- a/kitsune/sumo/static/sumo/js/form-wizard-sign-in-step.js
+++ b/kitsune/sumo/static/sumo/js/form-wizard-sign-in-step.js
@@ -57,7 +57,7 @@ export class SignInStep extends BaseFormStep {
 
           <p class="warning">
             <img class="key-icon" src="${keyImageURL}" aria-hidden="true"></img>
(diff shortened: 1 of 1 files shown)


kitsune/sumo/static/sumo/js/form-wizard-sign-in-step.js
```diff
@@ -20,7 +20,7 @@ export class SignInStep extends BaseFormStep {
           </p>
 
           <p class="for-sign-in">
-            ${gettext("A Firefox account allows you to securely sync your data when signing in from other devices and gives access to other fantastic Mozilla services.")}
+            ${gettext("A Firefox account allows you to securely sync your data when logging in from other devices and gives access to other fantastic Mozilla services.")}
```
mikeconley, line 23: Hey @hannajones - it turns out that we _did_ want this to be "signing", but the document was never updated to reflect that. I've corrected the document, and then I think we can revert this change.

mikeconley (approved):

kitsune/sumo/static/sumo/js/form-wizard-sign-in-step.js
hannajones: thanks for clarifying! In that case I'll just commit the whitespace removal I guess 😂

PR #5508, opened 2023-05-19
Wait for 1.5s before assuming UITour is broken and cannot be communicated with. (#1299)

mikeconley:
This can be tested manually by going into about:config and setting `browser.uitour.enabled` to `false`.


mikeconley:
Hey @escattone - just adding you as a reviewer to make sure that I'm doing the right thing with those GA `trackEvent` calls.

escattone (approved):
The `trackEvent` calls look perfect, thanks @mikeconley!

hannajones (approved):
Works well, thanks!

PR #5510, opened 2023-05-19
Progress bar enhancements

hannajones:
Found kind of a weird/nifty way of animating progress elements via transforming width/animating padding here: https://codepen.io/snookca/pen/KRdRzZ

Actually seemed nicer than replacing the progress element with a div or something.

Also added an aria-label to the progress element - @escattone is it alright to do string interpolation for `gettext`/translation purposes? Let me know if there's something else I should be doing.

### DIFF
--- a/kitsune/sumo/static/sumo/js/form-wizard.js
+++ b/kitsune/sumo/static/sumo/js/form-wizard.js
@@ -225,6 +225,11 @@ export class FormWizard extends HTMLElement {
       let progress =
         Math.ceil((activeStepIndex / (this.#steps.length - 1)) * 100) || 10;
       this.#progressIndicator.value = progress;
+      this.#progressIndicator.setAttribute(
+        "aria-label",
+        gettext(`Step ${activeStepIndex + 1} of 3`)
+      );
+      this.#progressIndicator.style.setProperty("--progress", progress + "%");
     }
   }
 
@@ -317,7 +322,7 @@ export class BaseFormStep extends HTMLElement {
    * Method that gets run whenever the element's state changes. Can be used to
    * specify how the DOM should update in response to state changes. `prevState`
    * and `nextState` are provided for making comparisons.
-   * 
+   *
    * This must be implemented in the subclass if the form step needs to respond
    * to state updates.
    *
--- a/kitsune/sumo/static/sumo/scss/form-wizard.styles.scss
+++ b/kitsune/sumo/static/sumo/scss/form-wizard.styles.scss
@@ -94,6 +94,15 @@ progress::-webkit-progress-value {
   background-color: var(--color-progress);
 }
 
+progress::-moz-progress-bar {
+  transition: padding-bottom 0.6s;
+  padding-left: 8px;
+  height: 0;
+  padding-bottom: var(--progress);
+  transform-origin: 0 0;
+  transform: rotate(-90deg) translateX(-8px);
+}
+
 p {
   margin: 0;
   color: var(--color-dark-gray-10);


kitsune/sumo/static/sumo/scss/form-wizard.styles.scss
```diff
@@ -94,6 +94,15 @@ progress::-webkit-progress-value {
   background-color: var(--color-progress);
 }
 
+progress::-moz-progress-bar {
+  transition: padding-bottom 0.6s;
```
mikeconley, line 98: Generally speaking, I think we try to avoid this kind of transition because it's not something that runs on the compositor (aka, it's not GPU accelerated). That means if, during the animation, some event on the main thread causes us to blow out the 16ms frame budget, then the animation will stutter.

The CSS transitions that we _can_ use on the GPU are the matrix-y ones like `translate` and `scale`. Those ones we can run smoothly in almost all scenarios, even if the main thread janks a little.

This is, I believe, how the about:welcome progress bar works. If you open about:welcome and examine it with the Inspector, there's a container element and the indicator itself, and the indicator is translated horizontally based on a `--progress-bar-progress` CSS variable, using this calculation:

```
transition: var(--progress-bar-transition);
translate: calc(var(--progress-bar-progress, 0%) - 100%);
```

where `--progress-bar-transition` is `none` in the prefers-reduced-motion case, and `--progress-bar-transition: 0.6s translate;` otherwise.

I suspect we might want to go this route, mainly because I think Tati expected the progress bar to behave the same as about:welcome's. FWIW, this sort of thing might make a good reusable component.

mikeconley (changes requested):

kitsune/sumo/static/sumo/js/form-wizard.js
```diff
@@ -225,6 +225,11 @@ export class FormWizard extends HTMLElement {
       let progress =
         Math.ceil((activeStepIndex / (this.#steps.length - 1)) * 100) || 10;
       this.#progressIndicator.value = progress;
+      this.#progressIndicator.setAttribute(
+        "aria-label",
+        gettext(`Step ${activeStepIndex + 1} of 3`)
```
escattone, line 230: I was again surprised by the limitations of even the GNU `xgettext` extractor when working with JavaScript files. I'm sorry to say that it can't yet handle this case. What you'll have to do instead, which is what we do in other places in our JavaScript code, for example [here](https://github.com/mozilla/kitsune/blob/d9b73f19471180f3fbb606c585321d9e7a1c4d75/kitsune/sumo/static/sumo/js/wiki.js#L604), is use a function called `interpolate` that we define [here](https://github.com/mozilla/kitsune/blob/d9b73f19471180f3fbb606c585321d9e7a1c4d75/kitsune/sumo/static/sumo/js/i18n.js#L21):

```suggestion
        interpolate(gettext("Step %s of 3"), [activeStepIndex + 1])
```

I've tested that locally, and it both works as expected, and is extracted properly into our [JavaScript POT file](https://github.com/mozilla-l10n/sumo-l10n/blob/master/templates/LC_MESSAGES/djangojs.pot) that we share with Pontoon.

escattone (changes requested):

kitsune/sumo/static/sumo/scss/form-wizard.styles.scss
hannajones: Thanks! That's super interesting and good to know. I was kind of just fixated on avoiding the two `<div>` approach in favor of using the built in `<progress>` element, but that's maybe not worth the hassle. The `about:welcome` progress bar also has some nice `aria-` attrs we should probably add here too. May keep tinkering for a bit.

This would be a great candidate for a reusable component - there's at least one other progress bar I can think of off the top of my head in Firefox view.

mikeconley:
I think the issue here is that there's a race between our JavaScript that's setting the progress variable, and our style/layout system that's calculating the initial position of the progress indicator.

What we ideally want is for the progress indicator to _start_ translated at -100%. However, if the `progress` style variable is set _before_ style/layout has had a chance to run, the starting position will actually be at 0% (so that it fills the container), before transitioning to -50%, which causes it go from right-to-left.

What we need to do is start the progress indicator at -100% without having animated there, and only after it has reached that point (and we've done a style flush), tell it to translate to the progress variable. That way, our style engine knows to tell our compositor "translate from -100% to -50%" or whathaveyou, rather than "translate from your initial position (0%) to -50%".

This is a workaround that seems to work for me - wdyt?:

```
diff --git a/kitsune/sumo/static/sumo/js/form-wizard.js b/kitsune/sumo/static/sumo/js/form-wizard.js
index bee3f7ef1..6afa68291 100644
--- a/kitsune/sumo/static/sumo/js/form-wizard.js
+++ b/kitsune/sumo/static/sumo/js/form-wizard.js
@@ -228,6 +228,12 @@ export class FormWizard extends HTMLElement {
         Math.ceil((activeStepIndex / (this.#steps.length - 1)) * 100) || 10;
       let indicator = this.#progressBar.querySelector(".indicator");
       indicator.style.setProperty("--progress", progress + "%");
+      requestAnimationFrame(() => {
+        requestAnimationFrame(() => {
+          indicator.toggleAttribute("ready", true);
+        });
+      });
+
       this.#progressBar.setAttribute("aria-valuenow", activeStepCount);
       this.#progressBar.setAttribute(
         "aria-label",
diff --git a/kitsune/sumo/static/sumo/scss/form-wizard.styles.scss b/kitsune/sumo/static/sumo/scss/form-wizard.styles.scss
index 367f44832..30cd767cd 100644
--- a/kitsune/sumo/static/sumo/scss/form-wizard.styles.scss
+++ b/kitsune/sumo/static/sumo/scss/form-wizard.styles.scss
@@ -90,9 +90,13 @@ section {
   border-bottom-right-radius: 8px;
 }

-.indicator {
+.indicator[ready] {
   transition: 0.6s translate;
   translate: calc(var(--progress, 0%) - 100%);
+}
+
+.indicator {
+  translate: -100%;
   background-color: var(--color-progress);
   flex: 1;
 }
```

kitsune/sumo/static/sumo/js/form-wizard.js
```diff
@@ -222,9 +223,23 @@ export class FormWizard extends HTMLElement {
...
+      requestAnimationFrame(() => {
+        requestAnimationFrame(() => {
+          indicator.toggleAttribute("ready", true);
+        });
+      });
+      this.#progressBar.setAttribute("aria-valuenow", activeStepCount);
+      this.#progressBar.setAttribute(
+        "aria-label",
+        interpolate(gettext("Step %s of 3"), [activeStepCount])
+      );
```
mikeconley, line 242: Seems like we're hard-coding a total of `3` both here and in the markup. Since the form-wizard is meant to be a bit more general, I think we can probably write that value in based on `this.#steps.length - 1`.

mikeconley (approved):
Thanks, @hannajones! Just one thing to address before landing.

kitsune/sumo/static/sumo/js/form-wizard.js
hannajones: good catch, thank you!

PR #5517, opened 2023-05-24
use interpolate to make error message translatable

hannajones:
Attempting to address the points raised in [this comment](https://github.com/mozilla/kitsune/pull/5499/files#r1204155619).

### DIFF
--- a/kitsune/sumo/static/sumo/js/form-wizard.js
+++ b/kitsune/sumo/static/sumo/js/form-wizard.js
@@ -51,7 +51,12 @@ export class FormWizard extends HTMLElement {
             </div>
 
             <div class="disqualification" reason="uitour-broken">
-              <div class="disqualification-message">${gettext("The setup assistant is currently unavailable for your version of Firefox. However, you can still perform a manual backup of your data by following the steps outlined in")} <a href="/kb/profile-manager-create-remote-switch-firefox-profiles">${gettext("this article.")}</a></div>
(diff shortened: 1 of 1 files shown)


muffinresearch:
As another possible solution, here's an [example of a way to do this that we've used on AMO in the past that helps keep the context within the same string](https://github.com/mozilla/addons-frontend/blob/df28945d5c4521bd044a5dc5a43426e547856503/src/amo/pages/Block/index.js#L165).  This uses placeholders to define the start of the link and the end of the link to be what's wrapped thus making it possible to have the full string context for localizers and without exposing the link to be inadvertently translated. To do this it can be a good idea to have linting or a test that guards against a localizer accidentally reversing the order.

As an additional aside, I also wonder if we should include more content in the link text so that if someone skips directly to the link (e.g. using assistive tech) it has context. As it stands  "this article" out of context doesn't.

hannajones:
Thanks all! @flodolo that approach of keeping the whole string intact and just interpolating the URL that we don't want translated seems to work with Kitsune's localization tools, so I've switched to what you suggested.

mikeconley (approved):
Thanks for fixing this up!

escattone:
@flodolo @muffinresearch Thanks for your comments! They'll certainly inform my approach to this kind of thing in the future. I really liked the example you noted @muffinresearch since I think it satisfies both my desire to avoid embedding HTML in the translatable string without sacrificing context.

PR #5518, opened 2023-05-24
Make it possible to update step labels via wizard manager

hannajones:
This is a fix for [the issue](https://github.com/mozilla/sumo/issues/1316) identified by @emilghittasv where the step indicator label doesn't change based on sign in state.

The fix ended up being a little more involved since we basically need to be able to pass data about the steps themselves when state changes. Previously we only passed that kind of step data to the wizard to initialize the steps, so it wasn't intended to be dynamic.

### DIFF
--- a/kitsune/sumo/static/sumo/js/form-wizard.js
+++ b/kitsune/sumo/static/sumo/js/form-wizard.js
@@ -148,23 +148,27 @@ export class FormWizard extends HTMLElement {
    * currently "unavailable".
    *
    * @param {string} name - Name of the step to be updated.
-   * @param {object} data - Data to update the step with.
+   * @param {object} payload - Data to pass to the active form step.
+   * @param {object} metadata - Data about the step for form wizard updates.
    */
-  setStep(name, data) {
+  setStep(name, payload, metadata) {
     // Find the step getting updated and set its state.
     let formStep = this.getRootNode().querySelector(`[name="${name}"]`);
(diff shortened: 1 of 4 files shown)


mikeconley (changes requested):
This works, but I think it misses a case - when I was testing this signed into SUMO, it did properly update the Step 1 label. After I signed into FxA though and was redirected back to the page, Step 1 was labelled "Create an account" despite being signed in:

image

So I think we might have to take into account - that we might want the Step 1 label updated even if it's not the currently active step.

hannajones:
@mikeconley poking around at this and I found a different bug that's a bit more severe - when I'm not logged in to SUMO the second step always shows as "unavailable." I think it has something to do with the order in which we receive updates and the fact that we don't update the step `status` in `switching-device-wizard-maanger.js`, so when we pass metadata to `<form-wizard>.setStep` we're always overwriting the `status` to `unavailable` 😞

Might go back to the drawing board on this one and see if there's a more straightforward way to just change the labels.

Screenshot 2023-05-25 at 2 35 35 PM

hannajones:
@mikeconley got this working, but it definitely feels pretty clunky 😕 Going to keep looking to see if I can find something simpler, definitely open to suggestions here

mikeconley (approved):
This works brilliantly. Much cleaner. Thanks so much, @hannajones!

PR #5523, opened 2023-05-29
Restyle switching devices form wizard disqualification warnings on mobile. (#1304)

mikeconley:
This styles the message to be more like a traditional SUMO warning, and hides much of the underlying wizard, which takes up too much space on mobile device screen widths.


hannajones (approved):
LGTM, thanks!

PR #5527, opened 2023-05-30
loadnow for docs

smithellis:
* Force images to load without lazyload for revisions, in order to restore anchor functionality
* Clean up spacing


escattone (changes requested):
This doesn't work for me, and I don't think it works in general because most of our images don't include the `lazy` class, and `$("img.lazy")` only selects images with that class.

I'm definitely not an expert, but the only thing I could find that seems like it might work would be something like https://css-tricks.com/aspect-ratio-boxes/, but I don't see a way to do that without making some significant changes.

I'm not sure what the best next step for this might be.

smithellis:
For whatever reason, I thought the lazyload jquery plugin was adding the lazy class in wiki docs. I've looked again, and it appears I was 100% incorrect about that.

smithellis:
Ok, the wikiparser *does* add the lazy class to images in wiki documents.

smithellis:
So the lazy class is added by wikiparser ( see `hook_image.html` and `hook_image_lazy.html` ) - confirmed.  However, on retesting, I see that I can still get the problem (not sure how I missed it yesterday...). So I added a chunk of code to `document.js` that waits for all page content to load before going to the right hash. So now I'm 
1) Forcing immediate image load ( `.loadnow()` ) 
2) Going to the right place after everything loads

escattone:
> So the lazy class is added by wikiparser ( see `hook_image.html` and `hook_image_lazy.html` ) - confirmed.

Sorry @smithellis! 🙏 I see now that I missed all of the behind-the-scenes work of the WikiParser's image handling!

akatsoulas (commented):
Can we target just the page which has this specific problem at least until we have a better solution for all of the KB? I am concerned about the performance implications this will have if we applied to the whole of the KB

smithellis:
I'm sure I can figure that out, @akatsoulas  - but I think it's a safe bet that this issue exists on other pages. I'll target the page for this one, then work on something more graceful for the whole KB.

escattone (commented):
This works, and seems to me the quickest way to resolve mozilla/sumo#1213. Thanks @smithellis, and sorry again for my earlier misunderstanding!

It seems to me that with this approach we're eagerly loading all of our images, waiting until everything is loaded, and then re-applying the hash. So if we decide we'd like to stick with this approach permanently, we should consider modifying [`hook_image.html`](https://github.com/mozilla/kitsune/blob/28f3d8bb1bbb64148f13de8a261bc2574506ae9a/kitsune/sumo/jinja2/wikiparser/hook_image.html) to remove the lazy handling altogether, and then we could skip the `$("img.lazy").loadnow();` call. If we did that, we'd also have to re-parse all of our documents with something like `doc.html = doc.current_revision.content_parsed` to rebuild all of the image tags without the lazy approach.

Alternatively, in the future it might also be worth exploring the [aspect ratio approach](https://css-tricks.com/aspect-ratio-boxes/), which would allow us to lazily load all of our images while still reserving the right amount of height (depending on the device).

smithellis:
Per @akatsoulas I've targeted just the relay-integration page. I'll look into a broader fix, but this should solve for this singular issue.

akatsoulas (approved):

akatsoulas:
Merging per [comment](https://github.com/mozilla/kitsune/pull/5527#pullrequestreview-1457863143)

PR #5529, opened 2023-05-31
Make sure GA instrumentation works for shadow DOM nodes as well.

mikeconley:


kitsune/sumo/static/sumo/js/form-wizard-sign-in-step.js
```diff
@@ -1,3 +1,4 @@
+import trackEvent from "sumo/js/analytics";
```
hannajones, line 1:
```suggestion
```

Looks like this isn't needed

hannajones (approved):
One very small piece of cleanup, but otherwise looks great!

I am curious if we want/need to replicate the 250ms delays that are implemented in `analytics.js` to ensure that the `trackEvent` happens before form submission or navigation, but I'll defer to @escattone on that.

hannajones:
>  I added `alert()` calls both pre and post the `trackEvent()` call, and when I clicked on the `Forgot password?` link, I did see the `alert()` call post `trackEvent()` before leaving the page, so I think we're OK?

Sounds good - I did something similar with `console.log` and was seeing those come through as well 😆

PR #5531, opened 2023-06-01
Prevent keyboard interaction when the wizard is disabled

hannajones:
Emil identified an issue where it was still possible to navigate into the form steps even when the wizard was "disqualified." Digging into this a bit it seemed like a good opportunity to take the [relatively new](https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/inert#browser_compatibility) `inert` attribute for a spin, since it prevents focus events on an element and all of its descendants.

If this feels heavy handed/we have any concerns about this approach I'm happy to try something else.


mikeconley (approved):
This is a wonderfully clean solution. Great stuff!

PR #5532, opened 2023-06-01
Prevent tooltip from hiding input label

hannajones:
Full disclosure I was testing this out by editing the CSS sources in the browser on my Windows machine since I don't have the kitsune repo set up locally there.


mikeconley (approved):
Works wonderfully - tested in macOS and Windows, both with and without high contrast. Thanks!

PR #5533, opened 2023-06-02
Selenium Tests - Added contribute page tests with soft asserts 

emilghittasv:
- Added pytest-check to poetry for soft assertions.
- Added a couple of tests for the Contribute pages section.
- Fixed flake8 issues which are passing locally now.


emilghittasv:
I'm going to close this PR in favor of an upcoming work that is going to happen (New tests & Selenium framework updates, enabling the capability of signing into SUMO using FxA accounts (for our tests), adding GitHub secrets for FxA accounts and plugging the framework in GitHub Actions workflow.

Some of the changes in this PR will be present in a follow-up.

PR #5544, opened 2023-06-09
Fix issues in ui.js

smithellis:
* Undo https://github.com/mozilla/kitsune/pull/2604
* Code doesn't appear to be needed and is causing issues with scrolling


escattone (approved):
Nice catch @smithellis! This worked great for me when I tested it locally.

## Before
image

## After
image

PR #5550, opened 2023-06-15
Exclude archived documents from featured

akatsoulas:


escattone (commented):
It seems that we'd have to exclude archived documents from the documents that we get from the original `visits` query as well? So something like:
```python
    visits = (
        WikiDocumentVisits.objects.filter(period=LAST_7_DAYS)
        .exclude(document__products__slug__in=settings.EXCLUDE_PRODUCT_SLUGS_FEATURED_ARTICLES)
        .exclude(document__is_archived=False)
        .order_by("-visits")
        .select_related("document")
    )
```

PR #5557, opened 2023-06-23
use Django ORM for weekly votes

escattone:
mozilla/sumo#1341

This PR is another step in preparation for the migration to Postgres, and replaces another chunk of raw SQL with the Django ORM.

## Notes
- The `fix_weekly_votes` management command is redundant, so I deleted it. The `update_weekly_votes` command accomplishes the same thing but more efficiently.


kitsune/questions/tasks.py
```diff
@@ -30,30 +30,25 @@ def update_question_votes(question_id):
...
+    log.info("Calculating past week votes for %s questions." % len(question_ids))
+
+    past_week = (datetime.now() - timedelta(days=7)).replace(
+        hour=0, minute=0, second=0, microsecond=0
+    )
+
+    Question.objects.filter(id__in=question_ids).update(
+        num_votes_past_week=Subquery(
+            QuestionVote.objects.filter(question_id=OuterRef("id"), created__gte=past_week)
+            .order_by()
```
akatsoulas, line 46: do we need the `order_by`?

akatsoulas (approved):
r+wc

kitsune/questions/tasks.py
escattone: I think so, at least to be safe, but I'm not absolutely certain. Here's my reference for using it: https://docs.djangoproject.com/en/4.2/topics/db/aggregation/#interaction-with-order-by. My thinking was that I need to clear any default ordering on the `QuestionVote`, which could mess-up the annotations (for example, the grouping by `question_id`). `QuestionVote` doesn't have a default ordering, but if one was added in the future, it could cause problems in a way that would be difficult to understand.

PR #5562, opened 2023-06-26
Issue 1335 - Content scrolling

smithellis:
Avoid the chance of a race condition
* Decouple url test from the scroll to hash function
* Remove lazyload actions from the window.load section

I was never able to replicate the issues here: https://github.com/mozilla/sumo/issues/1335 but I was able to determine that it didn't make sense to handle the lazyload decision AFTER the page fully loads - that seemed like a mistake.


escattone (approved):
Love it, thanks @smithellis! I think this will resolve https://github.com/mozilla/sumo/issues/1335

PR #5563, opened 2023-06-28
1163 - Fix n+1 query

smithellis:
* Added prefetch_related to view

### DIFF
--- a/kitsune/wiki/views.py
+++ b/kitsune/wiki/views.py
@@ -1659,7 +1659,7 @@ def what_links_here(request, document_slug):
         return HttpResponseRedirect(url)
 
     links = {}
-    links_to = doc.links_to().prefetch_related("linked_from")
+    links_to = doc.links_to().select_related("linked_from")
     for link_to in links_to:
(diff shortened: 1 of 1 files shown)


kitsune/wiki/views.py
```diff
@@ -1659,7 +1659,8 @@ def what_links_here(request, document_slug):
         return HttpResponseRedirect(url)
 
     links = {}
-    for link_to in doc.links_to():
+    links_to = doc.links_to().prefetch_related("linked_from")
```
escattone, line 1662: I think `select_related()` works here and would be a bit more efficient, since it would all be done in the same query?
```suggestion
    links_to = doc.links_to().select_related("linked_from")
```

escattone (changes requested):

kitsune/wiki/views.py
smithellis: Ah - you are right!

escattone (approved):
Thanks @smithellis!

elmatrwy (approved):
Thankyou

PR #5575, opened 2023-07-12
Do not index questions without answers

akatsoulas:


kitsune/search/documents.py
```diff
@@ -150,8 +150,13 @@ class Index:
...
         # Add a discard field in the document if the following conditions are met
-        # Question document is spam
-        if instance.is_spam:
+        # Question document is spam or question doesn't have any answers
+        if all(
+            [
+                isinstance(instance, Question),
+                any([instance.is_spam, instance.num_answers == 0]),
+            ]
+        ):
```
escattone, line 159: The tests are failing because the `any()` will be called when creating the list for the `all()`, so `instance.num_answers` will be executed before the `isinstance(instance, Question)` check, causing this to fail when the `instance` is an `Answer`. 
```suggestion
        if isinstance(instance, Question) and (instance.is_spam or (instance.num_answers == 0)):
```

escattone (changes requested):

kitsune/search/documents.py
akatsoulas: The reason I went for this check was mainly because the `any` had more items in the list. The `all` is probably too much here but this is the correct behavior. Similar to `(condition) and (condition or condition)`

akatsoulas:
@escattone pr updated

kitsune/search/documents.py
escattone: My suggested change works fine too -- I tested it -- since the `isinstance(instance, Question)` check is done first and if it fails, the remainder of the checks are skipped, but your update is equivalent.

escattone: What I mean is that my suggested change implements the same `(condition) and (condition or condition)` that your update does, but of course you're using the `any` to implement the `or`.

PR #5577, opened 2023-07-13
disable graphql introspection

escattone:
mozilla/sumo#1362

This PR also adds tests for the `graphql` endpoint.

### DIFF
--- a/kitsune/settings.py
+++ b/kitsune/settings.py
@@ -1250,9 +1250,11 @@ TRUSTED_GROUPS = [
 ]
 
 # GraphQL configuration
-GRAPHENE = {
+GRAPHENE: dict[str, str | tuple[str, ...]] = {
     "SCHEMA": "kitsune.schema.schema",
 }
+if not DEV:
+    GRAPHENE["MIDDLEWARE"] = ("kitsune.graphql.middleware.DisableIntrospectionMiddleware",)
 
 # Contributor Groups
 LEGACY_CONTRIBUTOR_GROUPS = [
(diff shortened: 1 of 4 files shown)


kitsune/settings.py
```diff
@@ -1252,6 +1252,7 @@ def filter_exceptions(event, hint):
 # GraphQL configuration
 GRAPHENE = {
     "SCHEMA": "kitsune.schema.schema",
+    "MIDDLEWARE": ("kitsune.graphql.middleware.DisableIntrospectionMiddleware",),
```
akatsoulas, line 1255: Maybe we can add the middleware only when DEV is False?

akatsoulas: There's also this in the graphene documention
https://docs.graphene-python.org/en/latest/execution/queryvalidation/#disable-introspection

escattone: I saw that too, but couldn't make sense of how to use the validator, because the `validate` call appears to be something that has to be embedded in the query processing flow.

escattone: @akatsoulas For ease of testing, I decided to always add `DisableIntrospectionMiddleware`, and dynamically check for `settings.DEV` in the middleware itself. If you'd still prefer that `DisableIntrospectionMiddleware` be added conditionally, I'll do that and remove the `DEV=True` part of the tests.

escattone: @akatsoulas After digging into the source code of the `graphene` and `graphene-django` packages, I'm convinced that the middleware approach, rather than the validation rule, is by far the simplest way to achieve the same goal of disabling introspection.

akatsoulas (approved):

PR #5581, opened 2023-07-16
optipng won't compile on mac silicon

smithellis:
- create a Make option 'init-mac' to account for it
- allow run-boostrap.sh to process the new --optipng-fix flag
- update doc

Open to better ideas (yes, move to linux is valid)
- Determining if you are on Apple silicon from run-bootstrap.sh directly appeared to be problematic since it is running in a container
- Doing it prior in the make call seemed to make sense
  - change nothing for exisiting processes but add a simpler way forward for Mac folk

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -47,6 +47,9 @@ run: .docker-build
 init: .docker-build
 	${DC} run web bin/run-bootstrap.sh
 
+init-mac: .docker-build
+	${DC} run web bin/run-bootstrap.sh --optipng-fix
+
 shell: .docker-build
 	${DC} run web bash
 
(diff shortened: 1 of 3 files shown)


escattone (commented):
Tried it locally, and it worked great! Will think about it more and finalize my review soon.

akatsoulas:
@smithellis / @escattone do we still need this or shall we close the PR?

smithellis:
I'm pretty sure optipng compiles properly now. Let's close and reopen if necessary.

escattone:
Reopening since this is still an issue.

escattone (approved):

escattone:
Since this is a small feature addition that doesn't affect anyone unless they're a Mac M1/M2 user that needs it, I'm going to merge it.

PR #5582, opened 2023-07-18
Ensure translations inherit Topics

smithellis:
This ensures that when translations are made, they inherit parent topics.

NOTE: There appears to be code that specifically removes Products from the form save - is this desired behavior?

```
        if parent_doc:
            # Products are not set on translations.
            doc.products.remove(*[p for p in doc.products.all()])
```

### DIFF
--- a/kitsune/wiki/forms.py
+++ b/kitsune/wiki/forms.py
@@ -229,6 +229,8 @@ class DocumentForm(forms.ModelForm):
         if parent_doc:
             # Products are not set on translations.
             doc.products.remove(*[p for p in doc.products.all()])
+            # A child always inherits parent topics.
+            doc.topics.add(*[t for t in parent_doc.topics.all()])
 
         return doc
 


akatsoulas (approved):

kitsune/wiki/forms.py
```diff
@@ -229,6 +229,8 @@ def save(self, parent_doc, **kwargs):
         if parent_doc:
             # Products are not set on translations.
             doc.products.remove(*[p for p in doc.products.all()])
+            # A child always inherits parent topics.
+            doc.topics.add(*[t for t in parent_doc.topics.all()])
```
akatsoulas, line 233: Why are you adding this in the DocumentForm?

akatsoulas (commented):
I was too fast on the last review. I am curious as to why we are adding the topics in the Document Form here

kitsune/wiki/forms.py
smithellis: Honestly I put it there because the removal of `Products` removal code was there, which seemed to be analogous. I can try to add it to the view if you prefer, but I would probably need to modify the model's `save()` and some other bits (and probably some tests) so I can get an `.id` assigned to the `doc_form` so I can then assign it values from this relationship.  
I can look at that if you want, just let me know.

akatsoulas:
@smithellis is this ready for review?

smithellis:
@akatsoulas This was ready unless you were wanting me to move the code out of DocumentForm. If it's good there, I can close out.

akatsoulas:
I had another pass, doing some local testing. This looks good to me. Let's wait to merge it after the release today

PR #5587, opened 2023-07-26
switching-devices -- add "details" wiki syntax pair

escattone:
mozilla/sumo#1383

Please suggest a new pair of names if you'd prefer something else! 😄


mikeconley (approved):
This looks good to me! Thanks!

akatsoulas (approved):

PR #5590, opened 2023-07-31
Fix regression 

smithellis:
* There is one case where you *want* the slug to prepopulate, on a net new kb article. Added back the prepop code.


escattone (approved):
Works great, thanks @smithellis!

elmatrwy (approved):
Thankyou

PR #5592, opened 2023-08-01
Update the switching devices form-wizard to bring it more in line with the spec

mikeconley:
This fixes mozilla/sumo#1378, mozilla/sumo#1377, mozilla/sumo#1384.

It's broken up into several separate commits. I'm not sure if that makes it less or more annoying to review in GitHub. I hope the former!

### DIFF
--- a/kitsune/sumo/static/sumo/js/form-wizard-sign-in-step.js
+++ b/kitsune/sumo/static/sumo/js/form-wizard-sign-in-step.js
@@ -14,12 +14,8 @@ export class SignInStep extends BaseFormStep {
           <h3 class="for-sign-up">${gettext("Create an account")}</h3>
           <h3 class="for-sign-in">${gettext("Sign in to your account")}</h3>
 
-          <p class="for-sign-up">
-            ${gettext("With an account, your data remains encrypted and protected, while also giving you access to other Mozilla products and services like Firefox Relay and Pocket.")}
-          </p>
-
-          <p class="for-sign-in">
-            ${gettext("A Firefox account allows you to securely sync your data when signing in from other devices and gives access to other fantastic Mozilla services.")}
+          <p>
+            ${gettext("You’ll be able to sign into this account on another device to sync your data.")}
           </p>
 
           <form method="get" novalidate>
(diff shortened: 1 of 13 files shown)


mikeconley:
Also note that there's still work in mozilla/sumo#1403 to update the syncing indicator in links, as I realize those are currently not working with this patch stack.

kitsune/sumo/static/sumo/js/form-wizard-sign-in-step.js
```diff
@@ -38,14 +33,14 @@ export class SignInStep extends BaseFormStep {
...
-              <input id="email" name="email" type="email" required="true" placeholder="${gettext("Enter your email address")}"/>
+              <input id="email" name="email" type="email" required="true" placeholder="${gettext("user@example.com")}"/>
             </div>
 
-            <button class="for-sign-up mzp-c-button mzp-t-product" type="submit" data-event-category="device-migration-wizard" data-event-action="click" data-event-label="signup-button">${gettext("Sign up")}</button>
-            <button class="for-sign-in mzp-c-button mzp-t-product" type="submit" data-event-category="device-migration-wizard" data-event-action="click" data-event-label="sign-in-button">${gettext("Sign in")}</button>
+            <button class="for-sign-up mzp-c-button mzp-t-product" type="submit" data-event-category="device-migration-wizard" data-event-action="click" data-event-label="signup-button">${gettext("Continue")}</button>
+            <button class="for-sign-in mzp-c-button mzp-t-product" type="submit" data-event-category="device-migration-wizard" data-event-action="click" data-event-label="sign-in-button">${gettext("Continue")}</button>
```
hannajones, line 43: Since we're using the same label now I wonder if we could simplify/clean things up a bit by just changing the `data-label` via JS.

hannajones (approved):
Everything looks good to me, thanks! One potential bit of cleanup, but I don't think it'll make too much of a difference

kitsune/sumo/static/sumo/js/form-wizard-sign-in-step.js
mikeconley: Great suggestion, will do, thanks!

PR #5593, opened 2023-08-01
Order of init issue

smithellis:
mozilla/sumo#1402


escattone (approved):
Verified locally that it fixes https://github.com/mozilla/sumo/issues/1402 🎉

PR #5594, opened 2023-08-01
Permission check needed parens

smithellis:


kitsune/wiki/views.py
```diff
@@ -478,7 +478,7 @@ def edit_init_and_perms(request, document_slug, revision_id=None, doctype="doc")
         if not doc.allows(user, "create_revision"):
             raise PermissionDenied
     if doctype == "meta":
-        if not doc.allows(user, "edit") or not user.is_staff:
+        if (not doc.allows(user, "edit") or not user.is_staff):
```
escattone, line 481: The logic, as it is, is definitely wrong, but I think you meant the following instead:
```suggestion
        if not (doc.allows(user, "edit") or user.is_staff):
```
or you could do it this way too:
```suggestion
        if not doc.allows(user, "edit") and not user.is_staff):
```

escattone, line 481: This won't make any difference, but I'm not sure what the logic should be here. Should only staff members who also have edit permission be able to do this? If so, then the logic here is correct. Or should staff members **or** users with edit permission be allowed? If that's the case, then this needs to change to:
```python
        if not (doc.allows(user, "edit") or user.is_staff):
```

escattone: I suspect the existing logic is correct, that only staff with edit perms should be allowed. If that's the case, this PR can be closed, because https://github.com/mozilla/sumo/issues/1401 does not apply. It only tested based on whether the user was staff or non-staff, but it also has to take into account edit perms.

smithellis: Closing - the proposed fix changes nothing, but it turns out this works as intended.

PR #5600, opened 2023-08-02
Issue #1403 - Update sync your data step to more closely align with the specification.

mikeconley:
Perhaps most importantly, this patch also updates the syncing indicator and switches the configuration / turn on syncing buttons into links embedded in the instruction text.


hannajones (approved):
Works for me, thanks!

PR #5601, opened 2023-08-03
Add fox doodle that updates based on the form step

hannajones:
Still working on a potential implementation for the feature flag to enable A/B testing, but figured I would get this up so we at least have the doodle landed (assuming we want to add it if A/B testing doesn't end up being feasible for some reason).

The styling is a little bit funky since the doodle belongs to the `form-wizard` instead of each individual step. It could also probably be better optimized for smaller window sizes, but the whole wizard gets a bit weird if you shrink the window enough so I didn't push too hard there.


kitsune/sumo/static/sumo/js/form-wizard.js
```diff
@@ -2,6 +2,24 @@ import trackEvent from "sumo/js/analytics";
...
+import setupDoodleURL from "sumo/img/fox-doodle-setup.png";
+
+const STEP_TO_DOODLE_MAP = {
+  "sign-into-fxa": {
+    src: signInDoodleURL,
+    alt: gettext("Fox tail sticking out from behind a stack of boxes."),
+  },
+  "configure-sync": {
+    src: configureDoodleURL,
+    alt: gettext("Fox sticking it's head our from behind a stack of boxes."),
```
mikeconley, line 16: Typo: "our" -> "out"

mikeconley (approved):
This looks great! Just one teeny tiny nit. Great work, @hannajones!

emilghittasv:
Testing is done for this ticket (performed on both Windows 10 and macOS 11).

* I can confirm that the fox doodle is successfully displayed inside the wizard if the `dwf_show_device_migration_doodle` cookie is set to `True`. If the `dwf_show_device_migration_doodle` is set to `False` the fox doodle is no longer displayed.

* I can confirm that the fox doodle transitions to the expected state/image when advancing from one wizard step to another & after navigating back by: refreshing the page while on step 3 (goes back to step 2), disconnecting from FxA (goes back to step 1) or by disabling/enabling the FxA sync.

* I can confirm that sometimes after clearing the browser cookies the switching-devices page is loaded with `dwf_show_device_migration_doodle` set on `False` meaning that the rule of having two separate pool of users is successfully applied.

* I can confirm that the fox doodle is not displayed if the `dwf_show_device_migration_doodle` cookie is set to `False` even if the user logs in to SUMO with a normal user account or connects to FxA. 

* Signing in with a superuser account into SUMO will make the fox doodle to appear regardless of the `dwf_show_device_migration_doodle` cookie value.

* I can confirm that the fox doodle image is successfully displayed while using High Contrast themes & the `alt` value is successfully read when using NVDA screen reader.

While verifying this ticket I've noticed that the fox doodle is overlapping several wizard elements when the browser window width is reduced: https://github.com/mozilla/sumo/issues/1418 

@hannajones, @mikeconley Ryan confirmed that the GA events for the fox doodle are working


Moving this ticket inside the "Release" column.

PR #5604, opened 2023-08-04
allow articles to have restricted visibility

escattone:
mozilla/sumo#1387

# Notes

- This feature touches so many things (e.g., the edit/create document UI, the WIki API, the dashboards, creating and editing documents, a management command, and so on), so this is a large PR. The most complex part, by far, was the dashboard-related work, the work to convert the raw SQL queries in `readouts.py` to Django ORM queries so we could take advantage of the `unrestricted` method and its thorough testing.
- I'm not a UI/UX designer, so of course I'm happy to change any of the UI! 😄 
- I could have spent another few days on just adding more tests, but I decided to instead invest heavily in strategic testing, mainly testing the `VisibilityManager.visible` method, which includes the new `unrestricted` method, in `kitsune/wiki/tests/test_models.py`. I also improved some other tests, like some tests in `kitsune/wiki/tests/test_facets.py` and `kitsune/dashboards/tests/test_readouts.py`.
- As part of the review process, I suspect we'll want to push this work to the `dev` environment, not only for thorough QA testing, but also for the content team to try it out and be able to provide feedback.
- I decided to use the `unrestricted` method instead of the `visible` method within `readouts.py`, only because it's a tiny bit more permissive than the `visible` method, which I suspected would be more useful to the dashboard audience, but it would be trivial to switch it to `visible`.
- This PR doesn't attempt to address the [`related_documents` issue](https://github.com/mozilla/sumo/issues/302). That issue can be addressed in a different PR whenever we're ready/willing, and that work would have to take into account document visibility. 
- I decided that [get_featured_articles()](https://github.com/mozilla/kitsune/blob/e7301f33f5adcf40f12619ec0aee689d890332e7/kitsune/wiki/utils.py#L106) should only return articles that are unrestricted for everyone.

# Screen Shots
image

image

image

image

image


smithellis (approved):
I've looked at what I can and I believe this is in order - QA can help us double check this big PR!

kitsune/wiki/jinja2/wiki/new_document.html
```diff
@@ -20,7 +20,25 @@ <h1 class="sumo-page-heading">{{ _('Create a New Knowledge Base Article') }}</h1
...
           {% csrf_token %}
           <ul>
-            {% for field in document_form.visible_fields() if (field.name != 'is_localizable' or settings.WIKI_DEFAULT_LANGUAGE == request.LANGUAGE_CODE) %}
+            {% if request.user.is_staff %}
+            <li class="mzp-c-emphasis-box field has-large-textarea">
+              <fieldset>
+                  <legend>{{ _("Restrict Visibility") }}</legend>
+                  <p>{{ _("If one or both are selected, the document will be visible only to those users.") }}</p>
+                  <span class="field has-large-textarea checkbox {% if document_form.restrict_to_staff.errors %}has-error{% endif %}">
+                    {{ document_form.restrict_to_staff|safe }}
```
akatsoulas, line 29: Why are we marking this as `|safe`?

kitsune/wiki/jinja2/wiki/new_document.html
```diff
@@ -20,7 +20,25 @@ <h1 class="sumo-page-heading">{{ _('Create a New Knowledge Base Article') }}</h1
...
+                    {{ document_form.restrict_to_staff|label_with_help }}
+                    {{ document_form.restrict_to_staff.errors }}
+                  </span>
+                  <span class="field has-large-textarea {% if document_form.restrict_to_group.errors %}has-error{% endif %}">
+                    {{ document_form.restrict_to_group|label_with_help }}
+                    {{ document_form.restrict_to_group|safe }}
```
akatsoulas, line 35: Similar comment as above

kitsune/wiki/management/commands/send_weekly_ready_for_review_digest.py
```diff
@@ -47,19 +47,21 @@ def _send_mail(locale, user, context):
         products = Product.objects.all()
 
         for loc in locales:
-            docs = revs.filter(document__locale=loc).values_list("document", flat=True).distinct()
-            docs = Document.objects.filter(id__in=docs)
+            doc_ids = (
```
akatsoulas, line 50: Obviously this query is not introduced now, but we definitely have `N+1` queries here. I am wondering if there's away to reshape this part to avoid that problem

kitsune/wiki/management/commands/send_weekly_ready_for_review_digest.py
```diff
@@ -47,19 +47,21 @@ def _send_mail(locale, user, context):
...
                 reviewers = Locale.objects.get(locale=loc).reviewers.all()
-                users = list(set(chain(leaders, reviewers)))
+                users = set(user for user in chain(leaders, reviewers) if user.is_active)
             except ObjectDoesNotExist:
                 # Locale does not exist, so skip to the next locale
                 continue
 
             for user in users:
                 docs_list = []
+                docs = Document.objects.unrestricted(user, id__in=doc_ids)
```
akatsoulas, line 64: marking this here also for compliteness

kitsune/wiki/managers.py
```diff
@@ -12,26 +12,77 @@ class VisibilityManager(models.Manager):
...
         """
         raise NotImplementedError
 
-    def visible(self, user, **kwargs):
+    def unrestricted(self, user=None, **kwargs):
+        """
+        Documents are unrestricted for a given user if the user is a superuser, or
+        if they're not restricted to staff or a specific group, or when the user is
+        staff and they're restricted by staff, or when the user is a member of the
+        specific group to which they're restricted. A translation (i.e., a document
```
akatsoulas, line 32: Regardless of the restriction staff members should always have visibility to restricted documents. I haven't looked at that part of the code through my review yet but by this wording I get the impression that a staff member will not have access to the document if it's restricted to a group by a non staff member.

kitsune/wiki/managers.py
```diff
@@ -12,26 +12,77 @@ class VisibilityManager(models.Manager):
...
+        def unrestricted_condition(parent_prefix=""):
+            """
+            Create the unrestricted condition for either the document or
+            the document's parent.
+            """
+            restrict_to_staff = f"{prefix}{parent_prefix}restrict_to_staff"
+            restrict_to_group = f"{prefix}{parent_prefix}restrict_to_group"
+
+            parent_condition = Q(**{f"{prefix}parent__isnull": not bool(parent_prefix)})
+            not_restricted = Q(**{restrict_to_staff: False}) & Q(
```
akatsoulas, line 49: Why aren't you using `filter/exclude(restrict_to_staff=False, restrict_to_group__isnull=True)?

kitsune/wiki/models.py
```diff
@@ -71,6 +71,16 @@ class Document(NotificationsMixin, ModelBase, BigVocabTaggableMixin, DocumentPer
     # Is this document localizable or not?
     is_localizable = models.BooleanField(default=True, db_index=True)
 
+    restrict_to_staff = models.BooleanField(default=False, db_index=True)
```
akatsoulas, line 74: Copied from the forms.py - Probably this is better suited here

`I would also argue here that if we use a m2m field we can also drop the restrict_to_staff field which is basically a convenience.`

akatsoulas:
Let's squash those commits into one before merging since we can't have any of those as a standalone.

kitsune/search/documents.py
```diff
@@ -93,11 +93,11 @@ def get_model(cls):
     @classmethod
     def get_queryset(cls):
         return (
-            # do not include any documents without an approved revision
-            wiki_models.Document.objects.exclude(current_revision__isnull=True)
-            # all documents will need their current revision:
+            # Only include documents that are visible to everyone.
+            wiki_models.Document.objects.visible()
```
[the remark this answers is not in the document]
escattone: Yes. The `visible()` method, when called without a user (or the user is None), excludes articles without an approved revision.

kitsune/wiki/forms.py
```diff
@@ -139,6 +140,26 @@ def __init__(self, *args, **kwargs):
...
+        initial=False,
+        label=_lazy("Staff"),
+        required=False,
+        help_text=_lazy(
+            "The document will be restricted to staff members "
+            "(and also group members if a group is selected)."
+        ),
+    )
+
+    restrict_to_group = forms.ModelChoiceField(
```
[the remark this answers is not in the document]
escattone: Ok, I'll remove `restrict_to_staff` and change `restrict_to_group` to `restrict_to_groups`.

kitsune/wiki/managers.py
escattone: Yes, you're right on. I changed this in both the `unrestricted` method as well as the `visible` method.

kitsune/wiki/managers.py
escattone: It's because this inner function has to return a Q object, because ultimately, in the outer method, the Q object it returns is used within the "or" condition `unrestricted_condition() | unrestricted_condition("parent__")`.

kitsune/wiki/jinja2/wiki/new_document.html
escattone: Good catch!

kitsune/wiki/jinja2/wiki/new_document.html
escattone: Good catch!

kitsune/wiki/management/commands/send_weekly_ready_for_review_digest.py
escattone: After further investigation, I don't think we can eliminate the `N+1` queries here, because the locales determine the users, and each user could possibly see a different set of visible docs.

kitsune/wiki/management/commands/send_weekly_ready_for_review_digest.py
escattone: After further investigation, I don't think we can eliminate the `N+1` queries here, because each user could possibly see a different set of visible docs.

akatsoulas (approved):
r+wc. Let's squash those commits and move it to QA.

kitsune/wiki/jinja2/wiki/document.html
```diff
@@ -76,14 +76,22 @@
       </div>
     </header>
 
+    {% if document.is_restricted %}
+      <div class="warning-box mzp-c-notification-bar mzp-t-warning">
+        {{ _('This document is restricted to members of the group(s):') }}<br />
+        {% for group_name in document.restrict_to_groups.order_by('name').values_list('name', flat=True) %}
```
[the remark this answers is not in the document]
escattone: Hm, good point. I'll just say the document is restricted for now.

kitsune/sumo/static/sumo/js/wiki.editor.js
```diff
@@ -1 +1,8 @@
-// TODO: extract editor-related functionality from wiki.js in a later commit
+import "@selectize/selectize";
+
+(function($) {
+  // Improve the interface for restricting articles.
+  $("select[id='id_restrict_to_groups']").selectize({
+    placeholder: "No group(s) selected ",
```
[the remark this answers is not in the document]
escattone: That's a really good suggestion. I'll change it.

kelimuttu:
Has this been pushed to prod, @escattone ?

emilghittasv:
@kelimuttu not yet. This is live in `stage` and under `QA`. The plan is to have it released this Wednesday (if nothing critical comes up and we don't have to push our release to next week).

kelimuttu:
Thanks for replying @emilghittasv We'll need to communicate this to the community so it won't be a surprise to them, even though this feature isn't for them. 

Can you elaborate which group this will be restricted to?

emilghittasv:
@kelimuttu 

* Only staff memebers will be able to set article visibility restrictions (The newly added field/functionality will not be available/displayed for non-staff members).
* The Restrict Visibility field/functionality will be availabe (for staff members) in two places: 1. When creating a new kb article (in the /new form). 2. When editing the article metadata (via the "Edit Article Metadata" option from the "Editing Tools" section).
* From the restricted visibility field you will be able to select one or more groups. The article will then be available/viewable by staff level accounts & members belonging to those selected groups only.

A few things worth mentioning:
* The article for which a visibility restriction was applied should not be available/displayed throughout the entire SUMO (forums, dashboards,etc) for non-whitelisted group members. (only whitelisted group members & staff members will be able to see those). 
* The article for which a visibility restriction was applied should not be returned in search results (even for whitelisted group members & staff members)
* Non whitelisted group members will get a 404 page when trying to navigate directly to the restricted visibility article link (if they do have that link at hand somehow) 
* Non whitelisted group members (and non staff accounts) should not receive e-mail notifications for restricted visibility articles.

This was a big change and multiple areas of SUMO were touched. @kelimuttu I think that we should keep an eye on the community and see if reports are coming in for some weird broken links (users ending up seeing a 404 page when navigating to a restricted kb visibility article)

kelimuttu:
That's very helpful, @emilghittasv . Thank you so much for sharing more context about this change. 

Just one clarification question. When you say that it's only for staff members, you're referring to the `is_staff` field from the user database, right? Asking because we don't have that data exposed in the UI currently, and not all admins can change that data from the admin panel (myself included). This may create a potential bottleneck in the future if we need to share a restricted article with staff members from another team, for example. We may need a super-admin intervention to change their `is_staff` status, in that case, unless we have a setting for that expose in the UI. 

It'll be a different story if, for example, we use a group system, dedicated only to staff members. Because we can easily set up and modify a group from the SUMO interface.

emilghittasv:
Sorry for missing this! Yes, only users with `is_staff` checked.

^ cc @escattone for visibility over Kiki's concern/use case.

escattone:
Thanks for helping out with the detailed responses @emilghittasv! Only users with `is_staff` can restrict an article and add/remove allowed groups, but all users within those allowed groups will be able to see the article itself. So in regard to @kelimuttu's concern, as long as the users we need to share an article with are within one of the article's allowed groups, they'll be able to see it. Does that answer your concern @kelimuttu?

kelimuttu:
@escattone I think so. But we may need your help nevertheless to create a new `Staff` group to make it easier for us to share the link with the partner team. I can't seem to create a new group from https://support.mozilla.org/groups/ either.

kelimuttu:
Minor feedback after trying this feature on the staging site: 

From the description provided, it seems that groups are additional audience on top of the default target (those with `is_staff` status). However, from trying this feature, it appears that adding group(s) is required to enable this functionality. This, ofc, doesn't hamper this feature to properly functioning. However, this eliminates the case where we only want to share the article with those who have `is_staff` status, since `is_staff` status isn't a group. Is this intentional?

escattone:
@kelimuttu I just created a `Staff` group, and assigned all current staff members to that group, so if you want to restrict an article to staff only, you can add the `Staff` group.

kelimuttu:
@escattone can you share the group link? I'm afraid I can't find it from https://support.mozilla.org/groups/

escattone:
@kelimuttu The `Staff` group isn't a publicly managed/viewable group (it doesn't have a `GroupProfile`), so it's one of those groups that doesn't have a public link to a management page.

kelimuttu:
@escattone The idea is to have a publicly managed group so we can easily allow someone from a partner team to review the restricted articles.

Alternatively, if we don't want to expose the group publicly, can we automate everybody with LDAP to be added to the Staff group?

PR #5608, opened 2023-08-08
Hide/show fox doodle based on a feature flag

hannajones:
So I'm still not 100% sure this will work for A/B testing, but it seems to work to hide/show the doodle locally. This patch:

* reintroduces waffleJS
* uses waffleJS to check the value/presence of a `show_device_migration_doodle` feature flag
* if the flag is on, shows the fox doodle and sends an event to GA

I set up the feature flag through my local admin dashboard with the following configuration:

Screenshot 2023-08-08 at 2 57 54 PM

If this seems like I'm on the right track we'll probably need someone on data science to weigh in on what percentage of users we want to show the doodle to, as well as see if they have opinions on how the GA event should be formatted/tagged.

### DIFF
--- a/kitsune/sumo/jinja2/base.html
+++ b/kitsune/sumo/jinja2/base.html
@@ -333,5 +333,7 @@
 {% if settings.STAGE %}
   <script src="https://pontoon.mozilla.org/pontoon.js"></script>
 {% endif %}
+
+  <script src="{{ url("wafflejs") }}"></script>
 </body>
 </html>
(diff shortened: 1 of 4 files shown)


mikeconley (approved):
Thanks, @hannajones!

This makes sense to me, but I very much think you should take my review with a grain of salt - @akatsoulas or @escattone should make the final call if this should land as-is.

escattone (approved):
This looks good to me.
- I thought we might only load `wafflejs` for the switching-devices document, but that file is so small it seems fine to always load it. ✅ 
- Using `wafflejs` allows us to avoid the document caching. ✅

akatsoulas (approved):
@hannajones there's a failing test. Other than that it should be good to go

PR #5609, opened 2023-08-10
flip the switch to postgres

escattone:
mozilla/sumo#1395
mozilla/sumo#1396
mozilla/sumo#1397
mozilla/sumo#1437
mozilla/sumo#1449
mozilla/sumo#1452
mozilla/sumo#1455
mozilla/sumo#1464
mozilla/sumo#1485
mozilla/sumo#1492

This PR, when merged, effectively flips the switch to use Postgres instead of MySQL, but depends on a properly configured Postgres database in its target Kubernetes environments, as well as the prior merging of a complementary PR in the `webservices-infra` repository that includes the proper `DATABASE_URL` setting. **Do not merge until we're ready to flip the switch**.

# Notes
- **This PR introduces a notable change to usernames**. Prior to this PR, the usernames `joe` and `Joe` would clash, only one or the other was allowed. This is because MySQL used a case-insensitive collation. Once  the username `joe` was used, no other case-differing version of it, like `Joe` or `JOE` was allowed. With this PR, because we're using a default, case-sensitive collation for Postgres, the usernames `joe` and `Joe` can co-exist as different usernames. Note that this change matches the default behavior of Django's default user model.
- The raw SQL queries within the files `kitsune/dashboards/readouts.py` and `cache_most_unhelpful_kb_articles.py`, just didn't seem worth spending a lot of time converting to Django ORM queries, at least at this point in time. We can revisit that later if we wish.
- One change in `kitsune/dashboards/readouts.py` that requires some explanation is the line `MAX(dashboards_wikidocumentvisits.visits)`. MySQL was so forgiving in this case, that it allowed the non-aggregated `dashboards_wikidocumentvisits.visits` even though that column has multiple rows in each group (groups are organized by document ID), and so should have required an aggregation. The value of that column for each of the rows in a group is always the same, so that's probably why MySQL was forgiving. Postgres is more strict, so it required an aggregation. Since the value of the column for all of the rows in each group are the same, we could use `MAX` or `MIN` or perhaps other aggregation functions, but I chose `MAX`.
- Another change that requires explanation in `kitsune/dashboards/readouts.py` is the commit [account for postgres ordering of num_visits](https://github.com/mozilla/kitsune/pull/5609/commits/ce1bea231ba8b3823210573c4d55428dac7b02ee). Postgres, by default, considers `NULL` values the largest value, while MySQL considers them the lowest.


akatsoulas (approved):
:rocket: :guitar:

PR #5610, opened 2023-08-11
Double UITour timeout for the switching-devices wizard to 3 seconds.

mikeconley:
Fixes mozilla/sumo#1419.


hannajones (approved):
Thanks!

PR #5613, opened 2023-08-14
remove is_staff metadata restriction

smithellis:
Leave all formatting changes, but remove any restrictions to metadata editing that require someone to be staff.


escattone (approved):
r+wc

@akatsoulas @emilghittasv Should we still hide the `Edit Article Metadata` link in the sidebar if the user doesn't have edit permission? I think we should, but maybe that's a separate small PR?

akatsoulas:
@escattone that's a good point. I will merge this to give @emilghittasv to test the functionality and I will have a look at hiding the link. I believe it's also tracked in another issue

akatsoulas:
I was wrong about the other issue. I opened a new one https://github.com/mozilla/sumo/issues/1424

emilghittasv:
@akatsoulas I think that was covered in https://github.com/mozilla/sumo/issues/1406 ? hmmm

PR #5615, opened 2023-08-16
Catch redirected Pocket articles

smithellis:
* Match urls with /kb/pocket/articleid-article-slug
* View tries to find kb document with article-slug - if exists it goes there
* If it doesn't exist, the view fails back to /products/pocket
* If it has to fail, it posts a message to the /products/pocket page

### DIFF
--- a/kitsune/wiki/tests/test_views.py
+++ b/kitsune/wiki/tests/test_views.py
@@ -2320,3 +2320,24 @@ class FallbackSystem(TestCaseBase):
         trans_content = "This article is translated into fr"
         assert en_content not in doc_content
         assert trans_content in doc_content
+
+
+class PocketArticleTests(TestCaseBase):
+    """Pocket article redirect tests."""
+
+    def setUp(self):
+        ProductFactory()
+
+    def test_pocket_redirect_to_kb_for_document_that_exists(self):
+        """Are we redirecting to our article page?"""
+        doc = DocumentFactory(title="an article title", slug="article-title")
+        RevisionFactory(document=doc, is_approved=True)
+        pocket_style_url = f"/kb/pocket/1111-{doc.slug}"
+        response = self.client.get(pocket_style_url, follow=True)
+        self.assertEqual(response.redirect_chain[-1][0], "/en-US/kb/article-title")
+
+    def test_pocket_redirect_when_kb_article_doesnt_exist(self):
+        """No match found, we should be sent to /products/pocket"""
+        pocket_style_url = "/kb/pocket/1111-wont-match"
+        response = self.client.get(pocket_style_url, follow=True)
+        self.assertEqual(response.redirect_chain[-1][0], "/products/pocket")
--- a/kitsune/wiki/urls.py
+++ b/kitsune/wiki/urls.py
@@ -161,3 +161,14 @@ urlpatterns += [
         r"^/discuss/watch_locale$", kbforums_views.watch_locale, name="wiki.discuss.watch_locale"
     ),
 ]
+
+urlpatterns += [
+    # Redirect for pocket articles
+    # This assumes pocket redirects take the form of:
+    # /pocket/<article_id>-<document_slug>
+    re_path(
+        r"^/pocket/(?:(?P<article_id>\d+)-)?(?P<document_slug>[\w-]+)(?P<extra_path>/[\w/-]*)?/?$",
+        views.pocket_article,
+        name="wiki.pocket_article",
+    ),
+]
--- a/kitsune/wiki/views.py
+++ b/kitsune/wiki/views.py
@@ -17,7 +17,7 @@ from django.db.models import Aggregate, TextField  # TODO: Delete once we move t
 from django.db.models.functions import TruncDate
 from django.forms.utils import ErrorList
 from django.http import Http404, HttpResponse, HttpResponseBadRequest, HttpResponseRedirect
-from django.shortcuts import get_object_or_404, render
+from django.shortcuts import get_object_or_404, redirect, render
 from django.template.loader import render_to_string
 from django.utils.translation import gettext as _
 from django.utils.translation import gettext_lazy as _lazy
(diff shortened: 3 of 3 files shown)


kitsune/wiki/urls.py
```diff
@@ -161,3 +161,14 @@
         r"^/discuss/watch_locale$", kbforums_views.watch_locale, name="wiki.discuss.watch_locale"
     ),
 ]
+
+urlpatterns += [
+    # Redirect for pocket articles
+    # This assumes pocket redirects take the form of:
+    # /kb/pocket/article/<article_id>-<document_slug>
+    re_path(
```
akatsoulas, line 169: Pocket will have the same URL structure for the KB similar to the other products. For example: 
Pocket article `https://help.getpocket.com/article/1142-firefox-new-tab-recommendations` will redirect to `https://support.mozilla.org/en-US/kb/irefox-new-tab-recommendations`

Basically we are keeping everything that follows after the `ID-` and we are considering the slug. If we don't have such an article we are redirecting to the product landing page. I suspect that will be `https://support.mozilla.org/en-US/products/pocket`

smithellis: We need an additional piece in the URL for incoming pocket redirects (aka the /pocket/) to give us the ability to know these were incoming pocket articles vs. any other type of article request.  So if we set up the redirect from pocket to land at just `/kb/article-slug` we are going to have to know in some other way where the request came from so we don't just fail **all** bad kb article requests to `/products/pocket` . I think having the `/pocket/` piece in there (doesn't have to be called that, or in that order, etc.) is easier than trying to catch the redirect sources in Django. 
I guess alternately we can get a redirect to `/kb/my-article-from-pocket?pocket=true' or something like that. The slug or data passed doesn't really matter to me, but I don't think there is another way to know for sure that the visitor tried to go to `help.getpocket.com` first.
Hope I understood what you were after here - let me know!

kitsune/wiki/views.py
```diff
@@ -1768,3 +1769,18 @@ def parse_accept_lang_header(lang_string):
     # Changed here to get the locale name only
     result = [k for k, v in result]
     return result
+
+
+def pocket_article(request, article_id=None, document_slug=None):
+    """Pocket articles migrated to SUMO are redirected to the new URL"""
+    try:
+        # If we migrated the document, let's get the doc
+        doc = Document.objects.get(locale=request.LANGUAGE_CODE, slug=document_slug)
```
akatsoulas, line 1778: Does this fallback to the default locale if there's not a translation?

kitsune/wiki/urls.py
```diff
@@ -161,3 +161,14 @@
         r"^/discuss/watch_locale$", kbforums_views.watch_locale, name="wiki.discuss.watch_locale"
     ),
 ]
+
+urlpatterns += [
```
akatsoulas, line 165: So, I believe this should go under products and we should account for more cases.

* A user visits `https://help.getpocket.com/`. We should get a redirect with no params landing users to the pocket product page.
* A user visits `https://help.getpocket.com/category/857-category`. We should get the `category` and `857-category` as params. During processing, we are going to ignore those params and redirect to the product landing page.
* A user visits anything in the `help.getpocket.com/` domain that is not followed by `article`, we should redirect to the product landing page.
* If and only if there is url of the form `help.getpocket.com/article/number-slug` then we are extracting the slug part and we are redirecting to the KB page if exists. If not we are redirect to the product landing page.

akatsoulas (changes requested):
I added a couple of comments. Let me know if you have more questions

kitsune/wiki/urls.py
akatsoulas: By going under products, I mean that we should aim at `https://support.mozilla.org/en-US/products/pocket`. That url needs to accept optional parameters that we need to unpack and do the processing.

Alternatively we can ask our SREs to redirect everything to `https://support.mozilla.org/en-US/products/pocket` and `help.getpocket.com/article/*` to a view similar with the one you created. I would choose the path of least resistance. If SREs can help us with two redirects, we can optimize what we have here

smithellis: I assume that refractr will take a root url request and/or anything not an article and send it directly to '/products/pocket', and then my code handles articles and fails back gracefully to product. I think this is the path of least resistance, right? So your use cases - `/category/857-category` is going to just redirect right from refractr to `/products/pocket`, as will 'help.getpocket.com`.  If that's the case, I don't think we should move the view under products because it really is just a `kb` view.

kitsune/wiki/tests/test_views.py
```diff
@@ -2320,3 +2320,30 @@ def test_custom_locale_mapping_for_header_locale(self):
...
+    def setUp(self):
+        ProductFactory()
+
+    def test_pocket_redirect_to_kb_for_document_that_exists(self):
+        """Are we redirecting to our article page?"""
+        doc = DocumentFactory(title="an article title", slug="article-title")
+        RevisionFactory(document=doc, is_approved=True)
+        pocket_style_url = f"/kb/pocket/1111-{doc.slug}"
+        response = self.client.get(pocket_style_url, follow=True)
+        self.assertContains(response, doc.title)
```
akatsoulas, line 2337: Can we maybe have an exact match here on the url with `assertEqual`?

kitsune/wiki/tests/test_views.py
```diff
@@ -2320,3 +2320,30 @@ def test_custom_locale_mapping_for_header_locale(self):
...
+
+    def test_pocket_redirect_when_kb_article_doesnt_exist(self):
+        """No match found, we should be sent to /products/pocket"""
+        pocket_style_url = "/kb/pocket/1111-wont-match"
+        response = self.client.get(pocket_style_url, follow=True)
+        path_found = False
+        for resp in response.redirect_chain:
```
akatsoulas, line 2344: Let's just follow the redirects and check whether the final one is `products/pocket`?

kitsune/wiki/tests/test_views.py
smithellis: Sure thing.

kitsune/wiki/tests/test_views.py
smithellis: Ah, I have over-complicated this.  Changing.

kitsune/wiki/urls.py
smithellis: @akatsoulas Did I sway youabout leaving it under wiki, or do you still prefer I move the view to products ?

kitsune/wiki/views.py
smithellis: Dropped the `locale=` bit - this should now be agnostic to locale since we are literally looking for a specific slug.  Really just being sure it exists, then letting wiki.document handle things.

akatsoulas:
@smithellis is this ready for a final pass?

akatsoulas (approved):

PR #5617, opened 2023-08-18
require a reason when reporting abuse

escattone:
mozilla/sumo#1435

## Notes
- MySQL allows the [`reason`](https://github.com/mozilla/kitsune/blob/6efbb5c1ac7ea2a1153cc9a827a2258fb29d7e91/kitsune/flagit/models.py#L43) field to be empty (`null`) when reporting abuse, because the column in the database is actually out-of-sync with the Django model, which requires a `reason` to be specified. Postgres is fully synchronized with the Django model, so it doesn't allow the `reason` to be `null`.
- With this PR, a `reason` is required, and also, if `Other` is selected, more detail must be provided as well.
- I also added another `reason` option to the form -- `Abusive content` -- because it was specified [here](https://github.com/mozilla/kitsune/blob/6efbb5c1ac7ea2a1153cc9a827a2258fb29d7e91/kitsune/flagit/models.py#L25), but never added to the form.

## Screenshots
The `Abusive content` option has been added:
image

When clicking the submit button without selecting a reason:
image

When clicking the submit button and not specifying more detail after selecting `Other`:
image

### DIFF
--- a/kitsune/sumo/static/sumo/js/reportabuse.js
+++ b/kitsune/sumo/static/sumo/js/reportabuse.js
@@ -6,10 +6,24 @@
   'use strict';
 
   $(function() {
-    $('#report-abuse [type="submit"]').on('click', function(ev) {
+    $('#report-abuse input[type=radio][name=reason]').on('change', function(ev) {
+      $(this).closest('form').siblings('.message').text('');
+    });
+
+    $('#report-abuse [type=submit]').on('click', function(ev) {
       ev.preventDefault();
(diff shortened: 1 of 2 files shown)


akatsoulas:
After discussing this with @escattone in person we decided to add a default value in the model that equals to an empty string and open a new issue to replace our current form with a modelform going forward.

escattone:
@akatsoulas and I discussed this in person, and we decided to close this PR in favor of #5618.

PR #5618, opened 2023-08-25
Add default value to reason when reporting abuse

akatsoulas:
mozilla/sumo#1435


escattone:
mozilla/sumo#1435

## Notes
- MySQL allows the [`reason`](https://github.com/mozilla/kitsune/blob/6efbb5c1ac7ea2a1153cc9a827a2258fb29d7e91/kitsune/flagit/models.py#L43) field to be empty (`null`) when reporting abuse, because the column in the database is actually out-of-sync with the Django model, which requires a `reason` to be specified. Postgres is fully synchronized with the Django model, so it doesn't allow the `reason` to be `null`.
- With this PR, `spam` becomes the default `reason`, and, correspondingly,  the `Spam or other unrelated content` radio button in the form is selected by default.
- Also, another option for `reason` has been added to the form -- `Abusive content` -- because it was specified [here](https://github.com/mozilla/kitsune/blob/6efbb5c1ac7ea2a1153cc9a827a2258fb29d7e91/kitsune/flagit/models.py#L25), but never added to the form.

## Screenshots
The `Abusive content` option has been added and the `Spam or other unrelated content` radio button is selected by default:
image

escattone (approved):

markoenix (commented):
Ok

PR #5621, opened 2023-08-29
1404 - Search focus issue with instant search

smithellis:
* replaced .on('keyup') with an .on('input')
* removed some deprecated .focus() events


escattone (approved):
Nice! Thanks @smithellis!

PR #5622, opened 2023-08-29
532 - Product name being paramed vs. slug

smithellis:
* Change to use slug


escattone (approved):
Works like a charm, thanks @smithellis!

escattone:
This should fix https://github.com/mozilla/sumo/issues/1444 as well as https://github.com/mozilla/sumo/issues/532! 🎉

PR #5624, opened 2023-08-30
fix set-up and loading of GA gtag function

escattone:
The SUMO [`gtm-snippet.js` code](https://github.com/mozilla/kitsune/blob/main/kitsune/sumo/static/sumo/js/gtm-snippet.js) defines a `window.gtag` function and, if do-not-track is not enabled, also kicks-off an `async` load of https://www.googletagmanager.com/gtag/js. As [recommended by Google](https://developers.google.com/analytics/devguides/collection/gtagjs), the `window.gtag` function must be defined inline -- or, alternatively, the JS code that does that must be fetched in a blocking mode -- immediately after the `head` tag. This is to ensure that `window.gtag` is defined prior to the execution of any other JS code that's loaded.

Currently, we're not doing that. We're fetching the `gtm-snippet.js` code in `defer` mode, so there's no guarantee that `window.gtag` is defined prior to GA calls. In fact, that's what we're seeing. For example, we're seeing that `window.gtag` is often not defined prior to the `trackEvent()` call [here](https://github.com/mozilla/kitsune/blob/02ec5146204adbf149ed425dc5c9553b01897a2e/kitsune/sumo/static/sumo/js/form-wizard.js#L114).

This PR resolves that by synchronously loading (i.e., blocking until complete) the `gtm-snippet.js` code immediately after the `head` tag. This is exactly what `www.mozilla.org` does as well.

image

## Before
image

## After
image

### DIFF
--- a/kitsune/sumo/jinja2/base.html
+++ b/kitsune/sumo/jinja2/base.html
@@ -6,6 +6,8 @@
 <!DOCTYPE html>
 <html class="no-js" lang="{{ request.LANGUAGE_CODE }}"{% if DIR %} dir="{{ DIR }}"{% endif %}{% if settings.GTM_CONTAINER_ID %} data-gtm-container-id="{{ settings.GTM_CONTAINER_ID }}"{% endif %}>
 <head>
+  {% include "includes/google-analytics.html" %}
+
   {% block head_top %}{% endblock %}
   <meta http-equiv="Content-Type" content="text/html; charset=utf-8"/>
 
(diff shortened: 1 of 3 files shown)


escattone:
I'm going to merge this, just so we can test it on stage. It should still be reviewed, and if necessary, we can revert it.

escattone:
I can confirm that with this code active in stage, the doodle GA events are showing-up in stage GA .
image

PR #5657, opened 2023-09-26
Update images to bookworm

akatsoulas:


escattone (approved):

escattone:
@akatsoulas I've looked into this, but I don't understand why `npm` is not found. It should be included in the `Node.js` install. I suspected it might be a Docker image caching issue, but after clearing the cache, the same error occurred. Any ideas?

escattone:
@akatsoulas I took the liberty of adding a commit to your branch that resolves the `npm` issue, but I'm going to wait to have you look at this and merge when you're happy with it.

PR #5662, opened 2023-10-04
Use `rest_framework.status` instead of hard coded one

heka1024:
Instead of using hard coded integer, use `rest_framework.status` to make more readable code.


escattone (approved):
Thank you @heka1024!

PR #5667, opened 2023-10-06
fix PlusToSpaceMiddleware now that we use Django i18n

escattone:
mozilla/sumo#1496

Alternatively, maybe we can get rid of the `PlusToSpaceMiddleware`?


akatsoulas:
I am pretty sure there will be somewhere a `+` sign in the url especially in places where we apply filters in the get requests.

akatsoulas (approved):

PR #5669, opened 2023-10-09
use Django's parse_accept_lang_header and ensure proper case in locale checks

escattone:
mozilla/sumo#1499

### DIFF
--- a/kitsune/wiki/views.py
+++ b/kitsune/wiki/views.py
@@ -1,6 +1,5 @@
 import json
 import logging
-import re
 import time
 from datetime import datetime, time as datetime_time
 from functools import wraps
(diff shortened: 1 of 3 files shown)


kitsune/wiki/views.py
```diff
@@ -1676,16 +1677,15 @@ def get_fallback_locale(doc, request):
     """Get best fallback local based on locale mapping"""
 
     # Get locales that the current article is translated into.
-    translated_locales = doc.translations.values_list("locale", flat=True).exclude(
-        current_revision=None
+    translated_locales = set(
```
akatsoulas, line 1680: Why are you using `set` here?

akatsoulas (approved):
r+(wc)? :guitar:

kitsune/wiki/views.py
escattone: It's not a big speed improvement, but it does make the `x in translated_locales` checks more efficient (O(1) rather than O(n) where n is the number of translated locales).

PR #5670, opened 2023-10-10
Work to set up a loginless form for MA

smithellis:
- Alter form to allow for logged in and loginless
- Alter view to allow non-logged-in to get to aaq3
- Alter texts in question_editing_frame.html, form labels, etc.
- Hide unusued form fields
- Utilize settings.LOGINLESS_EXCEPTIONS in case we grow this

### DIFF
--- a/kitsune/customercare/forms.py
+++ b/kitsune/customercare/forms.py
@@ -1,17 +1,50 @@
 from django import forms
+from django.conf import settings
 from django.utils.translation import gettext_lazy as _lazy
 
-from kitsune.customercare.zendesk import CATEGORY_CHOICES, OS_CHOICES, ZendeskClient
+from kitsune.customercare.zendesk import ZendeskClient
 
 PRODUCTS_WITH_OS = ["firefox-private-network-vpn"]
+
+# See docs/zendesk.md for details about getting the valid choice values for each field:
+CATEGORY_CHOICES = [
+    (None, _lazy("Select a reason for contacting")),
+    ("payment", _lazy("Payments & Billing")),
+    ("accounts", _lazy("Accounts & Login")),
+    ("technical", _lazy("Technical")),
+    ("feedback", _lazy("Provide Feedback/Request Features")),
+    ("not_listed", _lazy("Not listed")),
+]
+
+CATEGORY_CHOICES_LOGINLESS = [
+    (None, _lazy("Select a reason for contacting")),
+    ("fxa-reset-password", _lazy("I forgot my password")),
+    ("fxa-emailverify-lockout", _lazy("I can't recover my account using email")),
+    ("fxa-remove3rdprtylogin", _lazy("I'm having issues signing in with my Google or Apple ID")),
+    ("fxa-2fa-lockout", _lazy("My security code isn't working or is lost")),
+    ("other-account-issue", _lazy("I have another sign in issue")),
+]
+
+OS_CHOICES = [
+    (None, _lazy("Select platform")),
+    ("win10", _lazy("Windows")),
+    ("mac", _lazy("Mac OS")),
+    ("linux", _lazy("Linux")),
+    ("android", _lazy("Android")),
+    ("ios", _lazy("iOS")),
+    ("other", _lazy("Other")),
+]
 
 
 class ZendeskForm(forms.Form):
     """Form for submitting a ticket to Zendesk."""
 
+    required_css_class = "required"
+
     product = forms.CharField(disabled=True, widget=forms.HiddenInput)
+    email = forms.EmailField(label=_lazy("Contact Email"), required=True, widget=forms.HiddenInput)
     category = forms.ChoiceField(
-        label=_lazy("What do you need help with?"), choices=CATEGORY_CHOICES
+        label=_lazy("What do you need help with?"), required=True, choices=CATEGORY_CHOICES
     )
     os = forms.ChoiceField(
         label=_lazy("What operating system does your device use?"),
@@ -19,12 +52,21 @@ class ZendeskForm(forms.Form):
         required=False,
     )
     subject = forms.CharField(label=_lazy("Subject"), required=False)
-    description = forms.CharField(label=_lazy("Your message"), widget=forms.Textarea())
+    description = forms.CharField(
+        label=_lazy("Tell us more"), widget=forms.Textarea(), required=False
+    )
     country = forms.CharField(widget=forms.HiddenInput)
 
-    def __init__(self, *args, product, **kwargs):
+    def __init__(self, *args, product, user=None, **kwargs):
         kwargs.update({"initial": {"product": product.slug}})
         super().__init__(*args, **kwargs)
+        if product.slug in settings.LOGIN_EXCEPTIONS and not user.is_authenticated:
+            self.fields["email"].widget = forms.EmailInput()
+            self.fields["category"].choices = CATEGORY_CHOICES_LOGINLESS
+        else:
+            self.fields["email"].initial = user.email
+        self.label_suffix = ""
+
         if product.slug not in PRODUCTS_WITH_OS:
             del self.fields["os"]
 
--- a/kitsune/customercare/zendesk.py
+++ b/kitsune/customercare/zendesk.py
@@ -1,29 +1,8 @@
 from django.conf import settings
-from django.utils.translation import gettext_lazy as _lazy
 from zenpy import Zenpy
 from zenpy.lib.api_objects import Identity as ZendeskIdentity
 from zenpy.lib.api_objects import Ticket
 from zenpy.lib.api_objects import User as ZendeskUser
-
-# See docs/zendesk.md for details about getting the valid choice values for each field:
-CATEGORY_CHOICES = [
-    (None, _lazy("Select a topic")),
-    ("payment", _lazy("Payments & Billing")),
-    ("accounts", _lazy("Accounts & Login")),
-    ("technical", _lazy("Technical")),
-    ("feedback", _lazy("Provide Feedback/Request Features")),
-    ("not_listed", _lazy("Not listed")),
-]
-
-OS_CHOICES = [
-    (None, _lazy("Select platform")),
-    ("win10", _lazy("Windows")),
-    ("mac", _lazy("Mac OS")),
-    ("linux", _lazy("Linux")),
-    ("android", _lazy("Android")),
-    ("ios", _lazy("iOS")),
-    ("other", _lazy("Other")),
-]
 
 
 class ZendeskClient(object):
@@ -38,27 +17,42 @@ class ZendeskClient(object):
         }
         self.client = Zenpy(**creds)
 
-    def _user_to_zendesk_user(self, user, include_email=True):
-        fxa_uid = user.profile.fxa_uid
-        id_str = user.profile.zendesk_id
+    def _user_to_zendesk_user(self, user, ticket_fields=None):
+        if not user.is_authenticated:
+            name = "Anonymous User"
+            locale = "en-US"
+            id = None
+            external_id = None
+            user_fields = None
+        else:
+            fxa_uid = user.profile.fxa_uid
+            id_str = user.profile.zendesk_id
+            id = int(id_str) if id_str else None
+            name = user.profile.display_name
+            locale = user.profile.locale
+            user_fields = {"user_id": fxa_uid}
+            external_id = fxa_uid
         return ZendeskUser(
-            id=int(id_str) if id_str else None,
+            id=id,
             verified=True,
-            email=user.email if include_email else "",
-            name=user.profile.display_name,
-            locale=user.profile.locale,
-            user_fields={"user_id": fxa_uid},
-            external_id=fxa_uid,
+            email=ticket_fields.get("email") or user.email,
+            name=name,
+            locale=locale,
+            user_fields=user_fields,
+            external_id=external_id,
         )
 
-    def create_user(self, user):
+    def create_user(self, user, ticket_fields=None):
         """Given a Django user, create a user in Zendesk."""
-        zendesk_user = self._user_to_zendesk_user(user)
+        zendesk_user = self._user_to_zendesk_user(user, ticket_fields=ticket_fields)
         # call create_or_update to avoid duplicating users FxA previously created
         zendesk_user = self.client.users.create_or_update(zendesk_user)
 
-        user.profile.zendesk_id = str(zendesk_user.id)
-        user.profile.save(update_fields=["zendesk_id"])
+        # We can't save anything to AnonymousUser Profile
+        # as it has none
+        if user.is_authenticated:
+            user.profile.zendesk_id = str(zendesk_user.id)
+            user.profile.save(update_fields=["zendesk_id"])
 
         return zendesk_user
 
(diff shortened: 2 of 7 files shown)


kitsune/settings.py
```diff
@@ -1159,6 +1159,9 @@ def filter_exceptions(event, hint):
 ZENDESK_OS_FIELD_ID = config("ZENDESK_OS_FIELD_ID", default="360018604871", cast=int)
 ZENDESK_COUNTRY_FIELD_ID = config("ZENDESK_COUNTRY_FIELD_ID", default="360026463511", cast=int)
 
+# Products that allow un-authenticated users to submit support requests
+LOGINLESS_PRODUCTS = ["mozilla-account"]
```
akatsoulas, line 1163: nit: I would go for something like `LOGIN_EXCEPTIONS`

kitsune/questions/views.py
```diff
@@ -617,9 +631,14 @@ def aaq_step2(request, product_key):
     return aaq(request, product_key=product_key, step=2)
 
 
-@login_required
 def aaq_step3(request, product_key, category_key=None):
     """Step 3: Show full question form."""
+
+    # Since removing the @login_required decorator for MA form
```
akatsoulas, line 637: We should supply a `next` url here so that users can continue from where they left the process. I believe the `login_required` decorator uses the `redirect_to_login` method from `django.contrib.auth.views` which accepts the `next` url and the login url as params

kitsune/questions/views.py
```diff
@@ -617,9 +631,14 @@ def aaq_step2(request, product_key):
...
+    # need to catch unauthenticated, non-MA users here """
+    if product_key not in settings.LOGINLESS_PRODUCTS and not request.user.is_authenticated:
```
akatsoulas, line 639: Maybe instead of doing this check (which is a repeated block of code), you could supply an optional parameter in the function signature (eg `is_loginless=False` or `needs_login=True`) and modify that flag from the calling function. According to the flag state you may or may not redirect to a login wall.

kitsune/questions/views.py
```diff
@@ -533,8 +534,20 @@ def aaq(request, product_key=None, category_key=None, step=1):
 
             return HttpResponseRedirect(path)
 
-        if has_subscriptions:
-            zendesk_form = ZendeskForm(data=request.POST or None, product=product)
+        # If product allows loginless support
+        #  and not logged in, use CATEGORY_CHOICES_LOGINLESS
+        if product_key in settings.LOGINLESS_PRODUCTS and not request.user.is_authenticated:
```
akatsoulas, line 539: Since this check refers to the `Zendesk` form, maybe it should come after the next check where we try to figure out if we need support through zendesk or the forums

akatsoulas: The check for the existence of the `product_key` in the list happens in more than one places. Shall we replace it with something like
`render_loginless_categories = product_key in settings.LOGINLESS_PRODUCTS`

kitsune/questions/jinja2/questions/includes/question_editing_frame.html
```diff
@@ -90,12 +90,20 @@ <h1 class="sumo-page-heading">{% block headline %}{% endblock %}</h1>
               {% endif %}
 
               <div class="info card shade-bg highlight mb">
+              {% if current_product.key == 'mozilla-account' %}
+                {% trans %}
+                  Can't sign in to your account and need help? 
+                  You've found the right place.  Complete the form below
```
akatsoulas, line 96: Let's remove the extra space

kitsune/questions/jinja2/questions/includes/question_editing_frame.html
```diff
@@ -90,12 +90,20 @@ <h1 class="sumo-page-heading">{% block headline %}{% endblock %}</h1>
               {% endif %}
 
               <div class="info card shade-bg highlight mb">
+              {% if current_product.key == 'mozilla-account' %}
```
akatsoulas, line 93: shall we pass a flag from the view instead? The more we hardcode the value in multiple places, the harder will be to change it when needed. Ideally we should define that once in settings

kitsune/questions/config.py
```diff
@@ -639,6 +639,17 @@
                 ),
             },
         ),
+        (
+            "mozilla-account",
```
akatsoulas, line 643: This is already part of another PR. We should keep one out of the two.

kitsune/customercare/zendesk.py
```diff
@@ -38,22 +47,38 @@ def __init__(self, **kwargs):
         }
         self.client = Zenpy(**creds)
 
-    def _user_to_zendesk_user(self, user, include_email=True):
-        fxa_uid = user.profile.fxa_uid
-        id_str = user.profile.zendesk_id
+    def _user_to_zendesk_user(self, user, include_email=True, ticket_fields=None):
```
akatsoulas, line 50: do we still the include_email? With the loginless option, maybe we can drop this flag completely and we can also remove the `email` field from the `if/else` and replace it with:

email = ticket_fields.get("email") or user.email in the ZendeskUser

Alternatively we can always have `email = ticket_fields.get("email") and populate the ticket_fields either from the form or from the user object.

kitsune/customercare/zendesk.py
```diff
@@ -7,14 +7,23 @@
...
     ("payment", _lazy("Payments & Billing")),
     ("accounts", _lazy("Accounts & Login")),
     ("technical", _lazy("Technical")),
     ("feedback", _lazy("Provide Feedback/Request Features")),
     ("not_listed", _lazy("Not listed")),
 ]
 
+CATEGORY_CHOICES_LOGINLESS = [
+    (None, _lazy("Select a reason for contacting")),
+    ("fxa-reset-password", _lazy("I forgot my password")),
```
akatsoulas, line 20: If these are placeholders, let's add an issue to update them when we have the final values

kitsune/customercare/forms.py
```diff
@@ -10,21 +11,34 @@ class ZendeskForm(forms.Form):
     """Form for submitting a ticket to Zendesk."""
 
     product = forms.CharField(disabled=True, widget=forms.HiddenInput)
-    category = forms.ChoiceField(
-        label=_lazy("What do you need help with?"), choices=CATEGORY_CHOICES
+    email = forms.EmailField(
+        label=_lazy("Email"), required=False, disabled=True, widget=forms.HiddenInput
```
akatsoulas, line 15: Since this field can be used in both states (loginless and loginfull) we can remove `required` and `disabled` flags. Or remove `disabled` and keep `required=True`

kitsune/customercare/forms.py
```diff
@@ -10,21 +11,34 @@ class ZendeskForm(forms.Form):
...
-    def __init__(self, *args, product, **kwargs):
+    def __init__(self, *args, product, user=None, categories=None, **kwargs):
         kwargs.update({"initial": {"product": product.slug}})
         super().__init__(*args, **kwargs)
+        self.fields["category"].choices = categories if categories is not None else []
+        if product.slug in settings.LOGINLESS_PRODUCTS and not user.is_authenticated:
+            self.fields["subject"].initial = "Loginless MA Support"
+            self.fields["subject"].widget = forms.HiddenInput()
+            self.fields["email"].widget = forms.EmailInput()
+            self.fields["email"].disabled = False
```
akatsoulas, line 35: Following the comment above, we can completely omit any manipulation of the email field here

kitsune/customercare/forms.py
```diff
@@ -10,21 +11,34 @@ class ZendeskForm(forms.Form):
...
+    description = forms.CharField(label=_lazy("Tell us more"), widget=forms.Textarea())
     country = forms.CharField(widget=forms.HiddenInput)
 
-    def __init__(self, *args, product, **kwargs):
+    def __init__(self, *args, product, user=None, categories=None, **kwargs):
         kwargs.update({"initial": {"product": product.slug}})
         super().__init__(*args, **kwargs)
+        self.fields["category"].choices = categories if categories is not None else []
+        if product.slug in settings.LOGINLESS_PRODUCTS and not user.is_authenticated:
+            self.fields["subject"].initial = "Loginless MA Support"
```
akatsoulas, line 32: Are we keeping the field from the loginfull experience after all?

kitsune/customercare/forms.py
```diff
@@ -10,21 +11,34 @@ class ZendeskForm(forms.Form):
...
     subject = forms.CharField(label=_lazy("Subject"), required=False)
-    description = forms.CharField(label=_lazy("Your message"), widget=forms.Textarea())
+    description = forms.CharField(label=_lazy("Tell us more"), widget=forms.Textarea())
     country = forms.CharField(widget=forms.HiddenInput)
 
-    def __init__(self, *args, product, **kwargs):
+    def __init__(self, *args, product, user=None, categories=None, **kwargs):
         kwargs.update({"initial": {"product": product.slug}})
         super().__init__(*args, **kwargs)
+        self.fields["category"].choices = categories if categories is not None else []
```
akatsoulas, line 30: Why do we need this?

akatsoulas (changes requested):
This is a first pass which includes a couple of questions regarding the alignment of the forms that we render for each user state

kitsune/settings.py
smithellis: That makes sense.

kitsune/customercare/forms.py
smithellis: We have to set the categories on the displayed form on the fly based on which flavor form we are showing - logged in or not logged in.  The `self.fields["category"]` piece has to come after the `super` in order to alter the field.

kitsune/customercare/zendesk.py
smithellis: Joe sent me what will be the "real" values.  Adding and updating.

kitsune/questions/jinja2/questions/includes/question_editing_frame.html
smithellis: I am unclear what space you mean.

kitsune/customercare/forms.py
smithellis: Pending a meeting....

kitsune/customercare/forms.py
smithellis: Pending meeting....

kitsune/customercare/zendesk.py
smithellis: Learned they all post to a different field, too - so changes incoming for that.

kitsune/questions/config.py
smithellis: It will be different now, I had to revert to `firefox_accounts` for the product.

kitsune/questions/views.py
smithellis: I think I've fulfilled this.

kitsune/questions/jinja2/questions/includes/question_editing_frame.html
akatsoulas: `You've found the right place.  Complete the form below`

Between the dot and `Complete` there are 2 spaces instead of one

kitsune/questions/views.py
```diff
@@ -533,8 +536,19 @@ def aaq(request, product_key=None, category_key=None, step=1):
 
             return HttpResponseRedirect(path)
 
-        if has_subscriptions:
-            zendesk_form = ZendeskForm(data=request.POST or None, product=product)
+        # If product allows loginless support
+        # and not logged in, use CATEGORY_CHOICES_LOGINLESS
+        if has_subscriptions or is_loginless:
```
akatsoulas, line 541: this could be also written:

```python
if has_subscriptions:
    categories = CATEGORY_CHOICES
    if not request.user.is_authenticated:
        categories = CATEGORIES_CHOICES_LOGINLESS
```

kitsune/questions/views.py
smithellis: Added `is_loginless`

kitsune/questions/views.py
smithellis: Got it.

kitsune/questions/jinja2/questions/includes/question_editing_frame.html
smithellis: Got it!

kitsune/questions/views.py
smithellis: Agreed.

akatsoulas (changes requested):

kitsune/sumo/jinja2/base.html
```diff
@@ -258,7 +258,7 @@ <h5 class="mzp-c-footer-heading">{{ _('Mozilla Accounts') }}</h5>
             <li><a href="https://www.mozilla.org/en-US/firefox/accounts/">{{ _('What is it?') }}</a></li>
             <li><a href="https://accounts.firefox.com/reset_password">{{ _('Reset Password') }}</a></li>
             <li><a href="{{ url('wiki.document', 'switching-devices') }}">{{ _('Sync My Data') }}</a></li>
-            <li><a href="{{ url('products.product', 'mozilla-accounts')}}">{{ _('Get Help') }}</a></li>
+            <li><a href="{{ url('products.product', 'firefox_accounts')}}">{{ _('Get Help') }}</a></li>
```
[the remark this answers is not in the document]
smithellis: It isn't changing in Zendesk and the URLs can't change.  So we can't send you to `products/mozilla-account` nor can we send a ticket to Zendesk with the product of `mozilla-accounts`. So it doesn't make sense to change it in SUMO at the moment - we'd have to rework things to translate the slug in a few places vs. just using what exists.

kitsune/questions/config.py
```diff
@@ -639,6 +639,17 @@
                 ),
             },
         ),
+        (
+            "firefox_accounts",
```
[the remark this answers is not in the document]
smithellis: Changing it back to firefox_accounts - see note above.

kitsune/questions/jinja2/questions/includes/question_editing_frame.html
```diff
@@ -90,12 +90,20 @@ <h1 class="sumo-page-heading">{% block headline %}{% endblock %}</h1>
               {% endif %}
 
               <div class="info card shade-bg highlight mb">
+              {% if is_loginless %}
```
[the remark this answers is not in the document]
smithellis: The `is_loginless` is very product specific text for a login problem. To abstract away form using the product slug, I sued this. But if you aren't logged in, and you aren't using an 'is_loginless' product, you shouldn't see text about your login problems. If you want only two variations, we'd have to go back to the business and ask for a revision on the design.

kitsune/questions/views.py
```diff
@@ -533,22 +536,33 @@ def aaq(request, product_key=None, category_key=None, step=1):
...
                         messages.SUCCESS,
                         _(
-                            "Done! Your message was sent to Mozilla Support, "
-                            "thank you for reaching out. "
-                            "We'll contact you via email as soon as possible."
-                        ),
+                            "Done! Thank you for reaching out Mozilla Support."
+                            " We've sent a confirmation email to "
+                        )
+                        + email
```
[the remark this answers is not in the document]
smithellis: Sorry, this was an artifact of troubleshooting.  I'll fix.

kitsune/questions/jinja2/questions/includes/question_editing_frame.html
```diff
@@ -115,19 +123,28 @@ <h1 class="sumo-page-heading">{% block headline %}{% endblock %}</h1>
...
                     {% if field.name == 'content' %}
                       {{ content_editor(field) }}
                     {% elif field.name == 'troubleshooting' %}
                       {{ troubleshooting_instructions(field) }}
+                    {% elif field.name == 'description' and is_loginless %}
+                      <label for="{{ field.id_for_label }}">
+                        <span class="mzp-c-fieldnote">
+                          {{ _(
+                          "Include details such as your account email or specifics about"
+                          " your sign-in issue to help us get you back into your account quicker."
```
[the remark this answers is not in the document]
smithellis: Ah, we don't.

kitsune/questions/jinja2/questions/includes/question_editing_frame.html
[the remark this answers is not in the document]
akatsoulas: I was referring more to the description that follows the if statement. Sorry for the confusion

kitsune/customercare/zendesk.py
```diff
@@ -7,14 +7,23 @@
...
+    ("fxa-emailverify-lockout", _lazy("I can't recover my account using email")),
+    ("fxa-Remove3rdprtyLogin", _lazy("I'm having issues signing in with my Google or Apple ID")),
```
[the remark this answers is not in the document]
smithellis: Will correct case, will need to revisit post Moz account rollout to address naming convention. The impact to support is too large at the moment.

kitsune/customercare/forms.py
```diff
@@ -1,30 +1,41 @@
...
     """Form for submitting a ticket to Zendesk."""
 
+    required_css_class = "required"
+
     product = forms.CharField(disabled=True, widget=forms.HiddenInput)
-    category = forms.ChoiceField(
-        label=_lazy("What do you need help with?"), choices=CATEGORY_CHOICES
-    )
+    email = forms.EmailField(label=_lazy("Contact Email"), required=True, widget=forms.HiddenInput)
+    category = forms.ChoiceField(label=_lazy("What do you need help with?"), required=True)
```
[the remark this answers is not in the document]
smithellis: This is a great suggestion. I'll move it all over.

kitsune/customercare/forms.py
smithellis: After a hard look at your prior comment, I'm changing this all out.

kitsune/customercare/zendesk.py
smithellis: Agreed - moving it to a single statement at the ZendeskUser call.

kitsune/sumo/jinja2/base.html
```diff
@@ -252,11 +252,13 @@ <h5 class="mzp-c-footer-heading">{{ _('Firefox for Developers') }}</h5>
         </section>
 
         <section class="mzp-c-footer-col">
-          <h5 class="mzp-c-footer-heading">{{ _('Firefox Accounts') }}</h5>
```
akatsoulas, line 255: I believe this is already part of the main branch 
https://github.com/mozilla/kitsune/blob/main/kitsune/sumo/jinja2/base.html#L255

akatsoulas (approved):
r+wc :rocket:

smithellis:
Closing in favor of https://github.com/mozilla/kitsune/pull/5694

PR #5671, opened 2023-10-10
fix popular searches

escattone:
mozilla/sumo#1486

This fixes popular searches. When https://github.com/mozilla/kitsune/pull/5621 was released, it changed the event that triggers searches from "keyup" to "input", but forgot to trigger the same event when popular searches are clicked.


escattone:
@smithellis I wasn't sure if I needed to also change this [spot](https://github.com/mozilla/kitsune/blob/b72f29829b152bbb335b81dc181ff20f2b611aa0/kitsune/sumo/static/sumo/js/instant_search.js#L257) to use the `'input'` instead of `'keyup'`? It seems like it could even be removed, because it seems like it's only used to change focus to the search box if the search icon is clicked in the top menu on mobile devices?

akatsoulas (approved):

PR #5686, opened 2023-10-19
Make the reminder dialog's generated ICS file work properly with Outlook

mikeconley:
This was working properly with iCal, but not with the Outlook desktop calendar application. This change in formatting appears to work for both.

Fixes mozilla/sumo#1514.


hannajones (approved):
Thanks!

PR #5689, opened 2023-10-23
Add an email field and calendar reminder button to Step 3 of the switching-devices wizard.

mikeconley:
Fixes mozilla/sumo#1505.


mikeconley:
Hey @hannajones - I think the only thing here that makes me raise an eyebrow is the new success-white.svg icon. Unfortunately, the web-at-large doesn't have context-properties yet, so we can't use CSS to tell the browser what stroke colour to use on the icon.

So I think we have two options:

1. Use a new icon with the hardcoded icon (that's in this patch)
2. Use CSS filtering to try to manipulate the original icon to look white. I think `filter: greyscale(100%) brightness(500%);` might do it, but seems pretty hacky.

What do you think?

hannajones (approved):
Couple of small visual suggestions, but overall this looks good to me. Thanks!

re: the checkmark SVG - I think sticking with the hardcoded fill color is fine for now. It definitely seems more straightforward than using a `filter` or something. If/when we need a the same icon with a different fill colour elsewhere.

akatsoulas (approved):

PR #5690, opened 2023-10-23
Moz Account - Topic Listing page ADD CTA

smithellis:
- Set up CTA on Featured Topics List page

### DIFF
--- a/kitsune/products/jinja2/products/documents.html
+++ b/kitsune/products/jinja2/products/documents.html
@@ -23,9 +23,10 @@
   {{ search_box(settings, id='support-search-sidebar', params=search_params) }}
 </div>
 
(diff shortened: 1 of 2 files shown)


akatsoulas (commented):
I would argue that we don't need this code. This PR changes the display of the CTA based on the product.slug aka when you are viewing the actual product. We already have code that does that through the subscription based attribute. We could argue that MA are not technically a subscription based product but we render the form regardless of the subscription status. Maybe all we need to do in another PR is change the `has_subscription` attribute to something more descriptive.

Eg https://support.allizom.org/en-US/products/mozilla-account
It renders the right CTA based on the ccodename field

My suggestion would be to completely disregard this PR

akatsoulas:
I opened PR https://github.com/mozilla/kitsune/pull/5691 to update the property name

akatsoulas (approved):

PR #5697, opened 2023-10-25
Attempt to prefill email address in switching devices wizard step 3

mikeconley:
This uses session storage to store the email address inputted by the user in Step 1, and retrieves it after the FxA redirect in Step 3.

In the event that the user never went through Step 1, this is a no-op.

Fixes mozilla/sumo#1520.


hannajones (approved):
Works well, thanks!

PR #5700, opened 2023-10-26
301 redirect to mozilla-accounts

akatsoulas:


kitsune/products/views.py
```diff
@@ -31,6 +31,9 @@ def _get_aaq_product_key(slug):
 @check_simple_wiki_locale
 def product_landing(request, slug):
     """The product landing page."""
+    if slug == "firefox-accounts":
+        return redirect(product_landing, slug="mozilla-accounts", permanent=True)
```
akatsoulas, line 35: Before merging we need to figure out if this `mozilla-account` or `mozilla-accounts`

smithellis: I have been using `mozilla-account` and have had that set up on ZD as well.

smithellis (approved):

PR #5708, opened 2023-10-27
Moz accounts styling changes for form

smithellis:
- update 'Required' styling
- add mzp-c-fieldnote style

### DIFF
--- a/kitsune/sumo/static/sumo/scss/base/forms/_fields.scss
+++ b/kitsune/sumo/static/sumo/scss/base/forms/_fields.scss
@@ -409,17 +409,8 @@ textarea {
 }
 
 label.required::after {
-  content: "\A Required";
+  content: "Required";
   white-space: pre;
-  display: block;
-  font-size: 14px;
-  font-size: .875rem;
-  font-weight: 400;
-  line-height: 1.5;
-  margin-bottom: 0;
-}
-
-.custom-fieldnote {
   display: block;
   font-size: 14px;
   font-size: .875rem;
--- a/kitsune/sumo/static/sumo/scss/base/forms/_fields.scss
+++ b/kitsune/sumo/static/sumo/scss/base/forms/_fields.scss
@@ -411,5 +411,19 @@ textarea {
 label.required::after {
   content: "\A Required";
   white-space: pre;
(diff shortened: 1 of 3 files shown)


kitsune/sumo/static/sumo/scss/base/forms/_fields.scss
```diff
@@ -411,5 +411,18 @@ textarea {
...
   white-space: pre;
-  color: var(--color-marketing-red-01);
+  display: block;
+  font-size: 14px;
+  font-size: .875rem;
+  font-weight: 400;
+  line-height: 1.5;
+  margin-bottom: 0;
+}
+.mzp-c-fieldnote {
```
escattone, line 421: I think we should rename this class because it's not the same as Protocol's `mzp-c-fieldnote`. I think it's only used in `kitsune/questions/jinja2/questions/includes/question_editing_frame.html`?

smithellis: I made custom.

escattone (commented):
@smithellis Oh, I see that we haven't included the Protocol form classes, which contain Protocol's `mzp-c-fieldnote`. It'd be nice if you could use Protocol's class, if possible. You can include it and experiment by un-commenting [this line](https://github.com/mozilla/kitsune/blob/7471cb72d7195dce249da33f21f11ee13e21c9f4/kitsune/sumo/static/sumo/scss/_protocol.scss#L28).

kitsune/sumo/static/sumo/scss/base/forms/_fields.scss
```diff
@@ -411,5 +411,10 @@ textarea {
 label.required::after {
   content: "\A Required";
```
escattone, line 412: The `\A` line break really adds a lot of space, just before the `Required` text, which to me looks off:
image

compared to the other fields:
image

I think it looks better with the `\A` removed, but because the `Required` text is added by this CSS, it's really difficult to position the same as the label.
image

smithellis: Thanks for pointing out - I've tightened it up a bit.

PR #5709, opened 2023-10-27
Moz account footer typo

smithellis:
- According to style guide, Mozilla Account is the correct header


escattone (approved):
r+ 
I had to re-read the Figma instructions around when to use `Mozilla Account` instead of `Mozilla account`! 😄

PR #5710, opened 2023-10-27
Moz account CTA rework

smithellis:
- All CTAs go to interstitial first EXCEPT cta on special wiki docs
- If you come to form from interstital or those docs, we let you in otherwise you go to interstitial

### DIFF
--- a/kitsune/questions/views.py
+++ b/kitsune/questions/views.py
@@ -636,8 +636,9 @@ def aaq_step3(request, product_key, category_key=None):
     # Since removing the @login_required decorator for MA form
     # need to catch unauthenticated, non-MA users here """
     is_loginless = product_key in settings.LOGIN_EXCEPTIONS
+    skip_auth = request.GET.get("skip_auth", False)
 
-    if not is_loginless and not request.user.is_authenticated:
+    if not ((is_loginless and skip_auth) or request.user.is_authenticated):
         return redirect_to_login(next=request.path, login_url=reverse("users.login"))
 
     return aaq(
--- a/kitsune/users/jinja2/users/auth.html
+++ b/kitsune/users/jinja2/users/auth.html
@@ -50,7 +50,8 @@
                   <div class="trouble-text">
                     <p class="help-text">
                       {{ _('Having trouble signing in? ')}}
-                      <a href="{{ url('questions.aaq_step3', product_key='mozilla-account') }}">{{ _('Contact support') }}</a>
+                      <a
+                        href="{{ url('questions.aaq_step3', product_key='mozilla-account') }}?skip_auth=true">{{ _('Contact support') }}</a>
                     </p>
                   </div>
                 </div>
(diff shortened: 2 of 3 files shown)


kitsune/questions/views.py
```diff
@@ -636,8 +636,9 @@ def aaq_step3(request, product_key, category_key=None):
     # Since removing the @login_required decorator for MA form
     # need to catch unauthenticated, non-MA users here """
     is_loginless = product_key in settings.LOGIN_EXCEPTIONS
+    skip_auth = request.GET.get("skip_auth", False)
 
-    if not is_loginless and not request.user.is_authenticated:
+    if not request.user.is_authenticated and not skip_auth:
```
escattone, line 641: As it stands, this is a loophole that would allow users to skip login for any product key, just by adding `?skip_auth=True` to `/{locale}/questions/new/{product-key}/form`, although the flow after skipping the login will probably raise an exception at some point since it'll expect a non-anonymous user with attributes like `email` defined. In any case, we should only respect `skip_auth` when `is_loginless` is true.

```suggestion
    if not ((is_loginless and skip_auth) or request.user.is_authenticated):
```

kitsune/users/jinja2/users/auth.html
```diff
@@ -50,7 +50,7 @@ <h1 class="sumo-page-subheading sumo-header-medium">{{ _('Continue with your Moz
                   <div class="trouble-text">
                     <p class="help-text">
                       {{ _('Having trouble signing in? ')}}
-                      <a href="{{ url('questions.aaq_step3', product_key='mozilla-account') }}">{{ _('Contact support') }}</a>
+                      <a href="{{ url('questions.aaq_step3', product_key='mozilla-account') }}?skip_auth=True">{{ _('Contact support') }}</a>
```
escattone, line 53: Nit. The case for the value given to `skip_auth` doesn't matter, in fact any non-empty string works, so I think it's simpler to just use lowercase.
```suggestion
                      <a href="{{ url('questions.aaq_step3', product_key='mozilla-account') }}?skip_auth=true">{{ _('Contact support') }}</a>
```

kitsune/wiki/jinja2/wiki/includes/document_macros.html
```diff
@@ -501,7 +501,7 @@ <h3 class="card--title">
         </p>
         {% set aaq_url=url('questions.aaq_step3', product_key=product_key) %}
         <a class="sumo-button primary-button button-lg"
-          href="{{ aaq_url }}" 
+          href="{{ aaq_url }}?skip_auth=True"
```
escattone, line 504: Nit. Same here.
```suggestion
          href="{{ aaq_url }}?skip_auth=true" 
```

escattone (changes requested):
@smithellis Sorry for my earlier lack of understanding, in our Slack discussions earlier today, of why this PR is needed!

kitsune/questions/views.py
smithellis: Thanks!

kitsune/users/jinja2/users/auth.html
smithellis: Sure thing.

akatsoulas:
closing in favor of #5717

PR #5712, opened 2023-10-27
Moz Account - Fix for lack of subject or description

smithellis:
- If we don't have a subject, use the category
- If we dont have a description, use the category


escattone (approved):
r+

PR #5716, opened 2023-10-30
Moz account - Show appropriate text on form based on whether user is logged in or not

smithellis:
- Only show 'Can't sign in..' text to logged out users


akatsoulas (approved):
looks good. Let's add a more descriptive commit message though. r+wc

PR #5720, opened 2023-10-30
Allow form submission from self

akatsoulas:

PR #5722, opened 2023-10-30
Moz account allow form submissions

smithellis:
When we altered `is_loginless` in a prior PR, we altered how the form submissions would flow. 

Issue is here:  https://github.com/mozilla/sumo/issues/1544

https://github.com/mozilla/kitsune/pull/5720 was the initial PR from @akatsoulas but was failing linting, so this solves that issue.

### DIFF
--- a/kitsune/questions/views.py
+++ b/kitsune/questions/views.py
@@ -646,6 +646,10 @@ def aaq_step3(request, product_key, category_key=None):
         ]
     )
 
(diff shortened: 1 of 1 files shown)


escattone (changes requested):
I think the logic for skipping auth and the logic for setting `is_loginless` have become confusingly intertwined. I'm proposing that we close this in favor of https://github.com/mozilla/kitsune/pull/5726.

PR #5723, opened 2023-10-30
Update copy for switching devices wizard.

mikeconley:
This is the latest iteration for the copy in the switching devices wizard. This also makes the ICS file download filename something that can be localized to something meaningful.

Fixes mozilla/sumo#1529.

### DIFF
--- a/kitsune/sumo/static/sumo/js/form-wizard-reminder-dialog.js
+++ b/kitsune/sumo/static/sumo/js/form-wizard-reminder-dialog.js
@@ -35,20 +35,20 @@ export class ReminderDialog extends HTMLDialogElement {
             <button id="close" class="mzp-c-button mzp-t-neutral" aria-label="${gettext("Close")}" data-event-category="device-migration-wizard" data-event-action="click" data-event-label="close-reminder-dialog"><img src="${closeIconURL}" aria-hidden="true"/></button>
           </div>
           <div class="vbox">
-            <div id="directions">${gettext("Save the download link to your calendar and finish the download whenever you’re ready.")}</div>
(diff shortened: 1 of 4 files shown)


kitsune/sumo/static/sumo/js/form-wizard-reminder-dialog.js
```diff
@@ -143,9 +131,9 @@ export class ReminderDialog extends HTMLDialogElement {
    *   The description for the event, including the linkURL.
    */
   #generateEventSummaryAndDescription(linkURL, format) {
-    let summary = gettext("Reminder to complete your Firefox backup");
+    let summary = gettext("Download and install Firefox on your new device");
     let description = interpolate(
-      gettext("Your Firefox data has been successfully backed up.\n\nFollow this link to start your download: %s"),
+      gettext(`Your Firefox data is successfully backed up. To get started on your new device, download Firefox and sign in to your account with the link below.\n\n<a href="%s">Download Firefox</a>`),
```
hannajones, line 136: It looks like for `.ics` we might be better off omitting the `<a>` tag since the calendar app on mac seems to be smart about linkifying things

Screenshot 2023-10-30 at 1 16 28 PM

hannajones (approved):
LGTM! Just one minor point about links in the ICS file that might be nice to address

kitsune/sumo/static/sumo/js/form-wizard-reminder-dialog.js
mikeconley: Good call. I'll special-case the iCal case, and just use the URL in that case.

PR #5725, opened 2023-10-30
Add more GA events for the switching-devices reminder dialog

mikeconley:
This adds GA events for when the reminder dialog is closed, regardless
of how it's closed (in case, for example, it's closed by the keyboard),
as well as an event that describes which kind of calendar was selected
from the dropdown.

Closes mozilla/sumo#1528.


hannajones (approved):
LGTM, thanks!

PR #5726, opened 2023-10-30
fix loginless form submission

escattone:
mozilla/sumo#1544

Replaces #5722


smithellis (approved):
I think this is clearer.

PR #5732, opened 2023-10-31
remove locale from loginless referer sources

escattone:
mozilla/sumo#1547

Here's what was happening prior to this PR:
- `/de/users/auth`
- click on `Contact support`
- redirects to `/de/questions/new/mozilla-account/form`
- redirects to `/en-US/questions/new/mozilla-account/form`
- the original referer of `/de/users/auth` is maintained, but it doesn't match `/en-US/users/auth` -- the value of `reverse("users.auth")`, so it's redirected to...
- `/en-US/users/login?next=/en-US/questions/new/mozilla-account/form`

So this PR simply removes the locale prefix from our referer sources when setting `is_loginless`, since the locale doesn't matter.

Also, I couldn't help it. I removed the `skip_auth` confusion as well.

### DIFF
--- a/kitsune/questions/views.py
+++ b/kitsune/questions/views.py
@@ -635,16 +635,17 @@ def aaq_step3(request, product_key, category_key=None):
 
     # Since removing the @login_required decorator for MA form
     # need to catch unauthenticated, non-MA users here """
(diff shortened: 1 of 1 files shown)


smithellis (approved):
Tested locally and it works, plus you get the satisfaction of dropping skip_auth.  Approved.

PR #5739, opened 2023-11-01
Update download links created for calendar events

mikeconley:
This uses the updated mzl.la link for the calendar events so that they can be properly attributed.

Fixes mozilla/sumo#1503


hannajones (commented):
This works well, but I wonder if it's worth adding some test coverage to validate that the link is getting included in the event description since the link is the most important part of the whole thing. At least for Google calendar we could probably do something like this:

```js
expect(params.get("details").includes(NEW_DEVICE_DOWNLOAD_REMINDER_URL)).to.be.true;
```

(there might be a better sinon matcher, not sure)

mikeconley:
I don't know of a better matcher here, so that works for me! I've gone ahead and added that check.

hannajones (approved):
Thanks!

PR #5740, opened 2023-11-01
Submit request for switching-devices email in the document language.

mikeconley:
The email request to Basket was originally using navigator.language, but this doesn't take into account users that might be viewing the switching-devices page in a language that doesn't match the browser default.

This patch makes it so that we can pull the language code off of the document element, and use that when we submit the newsletter request.

If for some reason, the language attribute isn't set, we fallback to using navigator.language.


hannajones (approved):
Thanks!

PR #5745, opened 2023-11-03
Don't let users (or tests!) spam Basket so much.

mikeconley:
mozilla/sumo#1562 was originally about disabling the submit button sooner in the form submission process so that quick clicks on the submission button wouldn't result in multiple XHRs while we waited for Basket to respond.

While writing a test for that, I also noticed that our mocha tests were actually sending real XHRs to the Basket endpoint, both when running in automation and when running locally. Hilarious! This patch fixes that too.

Fixes mozilla/sumo#1562

### DIFF
--- a/kitsune/sumo/static/sumo/js/tests/form-wizard-setup-device-tests.js
+++ b/kitsune/sumo/static/sumo/js/tests/form-wizard-setup-device-tests.js
@@ -3,6 +3,8 @@ import sinon from "sinon";
 import { SetupDeviceStep } from "sumo/js/form-wizard-setup-device-step";
 import { ReminderDialog } from "sumo/js/form-wizard-reminder-dialog";
 
+const TEST_EMAIL_ADDRESS = "invalid@invalid.invalid";
+
 function assertFormElements(form, expectedElements) {
   let expectations = Object.assign({}, expectedElements);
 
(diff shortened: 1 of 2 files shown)


kitsune/sumo/static/sumo/js/tests/form-wizard-setup-device-tests.js
```diff
@@ -107,26 +121,97 @@ describe("setup-device-step custom element", () => {
...
+        step.addEventListener(
+          "DeviceMigrationWizard:SetupDeviceStep:EmailSubmitted",
+          (e) => {
+            resolve();
+          },
+          { once: true }
+        );
+      });
+
+      emailField.value = "invalid@invalid.invalid";
```
hannajones, line 160:
```suggestion
      emailField.value = TEST_EMAIL_ADDRESS;
```

Maybe worth making this a constant so that it's more likely to get reused in future tests?

hannajones (approved):
Works well, thanks!

This seems worth bringing up to the sumo team - definitely out of scope here, but the fact that the tests have full network access is a pretty big potential footgun.

kitsune/sumo/static/sumo/js/tests/form-wizard-setup-device-tests.js
mikeconley: Thanks! Done.

mikeconley:
Thanks for the quick review!

> This seems worth bringing up to the sumo team - definitely out of scope here, but the fact that the tests have full network access is a pretty big potential footgun.

Agreed - I'll mention it to the team in their development channel.

PR #5747, opened 2023-11-06
Update libraries

akatsoulas:


escattone (approved):
Thanks @akatsoulas. I made a small tweak. We don't need the `zope-interface` in the `pyproject.toml` (since it'll be added automatically as a dependency), so I removed it.

PR #5768, opened 2023-11-14
Anonymous user doesn't overwrite username

smithellis:
* Added check for existing user by email ( get_user_by_email )
* Return that user if exists vs. updating to Anonymous user


akatsoulas:
Let's open a new PR against this branch when we are done with the review

PR #5769, opened 2023-11-14
Anonymous user doesn't overwrite username

smithellis:
* Added check for existing user by email ( get_user_by_email )
* Return that user if exists vs. updating to Anonymous user

### DIFF
--- a/kitsune/customercare/zendesk.py
+++ b/kitsune/customercare/zendesk.py
@@ -18,6 +18,14 @@ class ZendeskClient(object):
         self.client = Zenpy(**creds)
 
     def _user_to_zendesk_user(self, user, email):
+        """Given a Django user, return a Zendesk user."""
+        # If the user already exists in Zendesk return
+        # them instead of creating a new user
+        if self.filter_user_by_email(email):
+            return self.filter_user_by_email(email)
+
+        # If the user is not authenticated, we can't save anything to
+        # AnonymousUser Profile as it has none
         if not user.is_authenticated:
             name = "Anonymous User"
             locale = "en-US"
@@ -41,6 +49,11 @@ class ZendeskClient(object):
             user_fields=user_fields,
             external_id=external_id,
         )
+
+    def filter_user_by_email(self, email):
+        """Given an email, return a user from Zendesk."""
+        for user in self.client.search(type="user", query=f"email:{email}"):
+            return user
 
     def create_user(self, user, email=""):
         """Given a Django user, create a user in Zendesk."""
--- a/kitsune/customercare/zendesk.py
+++ b/kitsune/customercare/zendesk.py
@@ -20,10 +20,10 @@ class ZendeskClient(object):
     def _user_to_zendesk_user(self, user, email):
         """Given a Django user, return a Zendesk user."""
         # If the user already exists in Zendesk return
-        # them instead of creating a new user
-        if self.filter_user_by_email(email):
-            return self.filter_user_by_email(email)
-
+        # the Zendesk user object
+        # instead of creating a new one
+        if zuser := self.get_user_by_email(email):
+            return zuser
         # If the user is not authenticated, we can't save anything to
         # AnonymousUser Profile as it has none
         if not user.is_authenticated:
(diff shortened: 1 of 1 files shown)


kitsune/customercare/zendesk.py
```diff
@@ -42,6 +50,11 @@ def _user_to_zendesk_user(self, user, email):
             external_id=external_id,
         )
 
+    def get_user_by_email(self, email):
```
akatsoulas, line 53: this should be named to something like `fitler_user_by_email`

kitsune/customercare/zendesk.py
```diff
@@ -42,6 +50,11 @@ def _user_to_zendesk_user(self, user, email):
...
+        """Given an email, return a user from Zendesk."""
+        for user in self.client.search(type="user", query=f"email:{email}"):
```
akatsoulas, line 55: What happens when we get a response with multiple users?

kitsune/customercare/zendesk.py
```diff
@@ -18,6 +18,14 @@ def __init__(self, **kwargs):
         self.client = Zenpy(**creds)
 
     def _user_to_zendesk_user(self, user, email):
+        """Given a Django user, return a Zendesk user."""
+        # If the user already exists in Zendesk return
+        # them instead of creating a new user
+        if self.get_user_by_email(email):
+            return self.get_user_by_email(email)
```
akatsoulas, line 25: Does this match what we return through `ZendeskUser`? If not, we can end up having this helper method returning two different types of ZD users.

akatsoulas (changes requested):

kitsune/customercare/zendesk.py
smithellis: Ok.

kitsune/customercare/zendesk.py
smithellis: It shouldn't be possible, as Zendesk uses the email as a primary id for users.
https://support.zendesk.com/hc/en-us/articles/4408831446298-Can-user-accounts-share-an-email-address-

kitsune/customercare/zendesk.py
smithellis: This returns an actual ZD user, with whatever attributes they have in ZD. Since the goal of this PR is to _not_ update existing users it doesn't really matter if it is exactly the same, since its only purpose is to have a an id to log a ticket against.

smithellis: Let me be specific - if you have a ZD account, we don't want to do anything with that account. 
I don't see a reason to run updates on any of the other user fields like we do when we match a SUMO user to a ZD user - in this case you are anonymous to SUMO so we just care if you exist in ZD, and we want to leave you be because without auth at SUMO we can't be sure you are you. At least, that's my thought process.

smithellis:
@akatsoulas I think my last commit addresses what we discussed - if I confounded things, let me know. Basically:

* Given an email, this queries ZD and
  returns a generator object
* If that object has one and only one
  result, it is a valid Zendesk user object
  and we return it
* If there are more than one items returned,
  we throw a ValueError because something is broken
* If nothing is returned, we return None

kitsune/customercare/zendesk.py
```diff
@@ -42,6 +50,21 @@ def _user_to_zendesk_user(self, user, email):
...
 
+    def get_user_by_email(self, email):
+        """Given an email, return a user from Zendesk."""
+        # This returns a generator, but we only want/expect one user
+        # If it returns more than one, we should fail
+        # Otherwise return the Zendesk user object
+        search_results = self.client.search(type="user", query=f"email:{email}")
+
+        user_found = None
+        for user in search_results:
```
akatsoulas, line 61: How do you ensure here that there's only one user in the results?

akatsoulas (approved):
r+wc

kitsune/customercare/zendesk.py
```diff
@@ -42,6 +50,21 @@ def _user_to_zendesk_user(self, user, email):
...
+            if user_found is not None:
+                raise ValueError(f"Found more than one user with email {email}")
```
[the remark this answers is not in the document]
smithellis: The search feature only ever returns a generator. One result or 100, you get a generator. Since we are looking up an e-mail and that is a primary identifier, we should never expect more than one response - although the API would allow it! With no other way to return a search result, we have to live with the generator object.

kitsune/customercare/zendesk.py
```diff
@@ -42,6 +50,21 @@ def _user_to_zendesk_user(self, user, email):
...
+    def get_user_by_email(self, email):
+        """Given an email, return a user from Zendesk."""
+        # This returns a generator, but we only want/expect one user
+        # If it returns more than one, we should fail
+        # Otherwise return the Zendesk user object
+        search_results = self.client.search(type="user", query=f"email:{email}")
+
+        user_found = None
+        for user in search_results:
+            if user_found is not None:
```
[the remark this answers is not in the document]
smithellis: Trying to think about this - I am checking if user_found exists, but you are saying to check if it doesn't exist. I think you might mean `if user_found` since it that would be the same as `if user_found is not None`. I'm going to leave it, as this this little section is hard to follow and I think this helps it be readable. I set user_found to None, explicitly, then I check if it is not None, explicitly.

PR #5776, opened 2023-11-17
1556 - Subject and Description changes

smithellis:
* Update subject of Loginless form to fallback to product friendly name + "support" (ex. "Mozilla account support") if no subject provided
* Update description to read "No response provided." if description is left empty

### DIFF
--- a/kitsune/customercare/zendesk.py
+++ b/kitsune/customercare/zendesk.py
@@ -1,8 +1,11 @@
 from django.conf import settings
+from django.utils.translation import gettext_lazy as _lazy
 from zenpy import Zenpy
 from zenpy.lib.api_objects import Identity as ZendeskIdentity
 from zenpy.lib.api_objects import Ticket
 from zenpy.lib.api_objects import User as ZendeskUser
+
+NO_RESPONSE = _lazy("No response provided.")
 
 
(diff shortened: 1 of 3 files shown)


kitsune/customercare/zendesk.py
```diff
@@ -101,8 +101,8 @@ def create_ticket(self, user, ticket_fields):
                 ]
             )
         ticket = Ticket(
-            subject=ticket_fields.get("subject") or ticket_fields.get("category"),
-            comment={"body": ticket_fields.get("description") or ticket_fields.get("category")},
+            subject=ticket_fields.get("subject") or f"{product_config['name']} support",
+            comment={"body": ticket_fields.get("description") or "No response provided."},
```
akatsoulas, line 105: Let's add the `No response provided` in a var at the beginning of the file

akatsoulas (approved):
r+wc

akatsoulas:
When this is merged, let's cherry-pick the commit for the main branch. Alternatively we can just target `main` and release it on the 27th. Imho the latter is simpler but totally up to you

kitsune/customercare/zendesk.py
smithellis: Can do.

PR #5781, opened 2023-11-17
a simpler and clearer localization dashboard

escattone:
mozilla/sumo#1524

The issue above is happening because the "unreviewed" section is hidden when the page loads, so it can't be "jumped to" via `/ro/localization#unreviewed`. The "unreviewed" section only appears if the "unreviewed" tab is clicked, but it's not even clear there are tabs, what they control, and that they are clickable. The best approach, I think, is to get rid of the strange tabs, and always show and clearly delineate all of the collapsible sections, including the "unreviewed" section. This PR creates a simpler and I think much more understandable view of the localization dashboard. It does the following:

- Removes the strange tabs -- the `stats-tabs` `div` -- which is super confusing
- Organizes the page into four clear, collapsible "details" sections, each indicated by an `h4` header
- Each of the collapsible "details" sections is open when the page loads
- Each of the "details" sections retains its ability to be a URL hash target (e.g., `/ro/localization#unreviewed`)
- Removes the "Jump to" links from the end of the first section (but we could easily add them back if desired)

### All of the sections open
image

### All of the sections closed
image


kitsune/dashboards/jinja2/dashboards/localization.html
```diff
@@ -83,21 +83,8 @@ <h2>{{ _('Locale:') }} {{ current_locale_name }}</h2>
                                      (overview_rows['all'], True),
                                      (overview_rows['templates'], True))) }}
 
-      <div id="stats-tabs" data-ui-type="tabbed-view">
```
akatsoulas, line 86: We can replace this `div` with `<nav class="tabs">` from protocol and apply the proper classes (`tabs--list, tabs--item, tabs--link`) to the ul, li, a elements. This should probably give us a proper tabbed menu which will required small changes in css and js.

akatsoulas (commented):
Part of the functionality has been restored through https://github.com/mozilla/kitsune/pull/5777. We could apply the proposed change and have proper, easy to understand tabs for a shorter page. Thoughts?

kitsune/dashboards/jinja2/dashboards/localization.html
escattone: @akatsoulas I created #5786 as an alternative to consider.

akatsoulas:
Per our conversation, I closed this in favor of #5786

PR #5789, opened 2023-11-24
Adjust Makefile for mkdocs

akatsoulas:


escattone (commented):
When I ran `make docs`, I got the following error:

```
INFO    -  Doc file 'browser_permissions.md' contains an unrecognized relative link 'deployments', it was left as
           is. Did you mean 'deployments.md'?
```
which I fixed by changing `(deployments)` on line 7 of `docs/browser_permissions.md` to be `(deployments.md)` instead.

Also, after running `make servedocs`, I don't see anything at http://127.0.0.1:8000/:
image

Am I missing something there?

PR #5797, opened 2023-11-29
Remove no-doodle variant of the switching devices wizard.

mikeconley:
Fixes mozilla/sumo#1504.

### DIFF
--- a/kitsune/sumo/jinja2/base.html
+++ b/kitsune/sumo/jinja2/base.html
@@ -338,6 +338,5 @@
   <script src="https://pontoon.mozilla.org/pontoon.js"></script>
 {% endif %}
 
-  <script src="{{ url("wafflejs") }}"></script>
 </body>
 </html>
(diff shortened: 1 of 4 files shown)


hannajones:
This all looks good, but I'm curious if we also want to remove the script tag that loads waffle.js: https://github.com/mozilla/kitsune/blob/main/kitsune/sumo/jinja2/base.html#L341

IIRC the doodle was the only use case for waffle.js and I had to add that in when doing the initial doodle work. If we're not using waffle for anything else it seems worth removing it again.

mikeconley:
> This all looks good, but I'm curious if we also want to remove the script tag that loads waffle.js: https://github.com/mozilla/kitsune/blob/main/kitsune/sumo/jinja2/base.html#L341

Good catch! Removed that too in the latest update.

hannajones (approved):
LGTM thanks!

mikeconley:
Thanks!

PR #5799, opened 2023-11-30
1596 - Show useful announcement if translation in progress

smithellis:
* We weren't considering the case where you were currently on the `doc.current_revision`, but there is an outstanding unapproved revision
* Added a check for unapproved revisions - if there is one, we fail to the `translation_not_approved` message vs. `no_translation`
* Added test
* I noticed all the announcement tests prior to this were just checking for an announcement, not for the specific announcement being shown. So I modified the announcements to identify themselves with a unique id, and the tests that care which announcement you see can now check for that.

### DIFF
--- a/kitsune/wiki/views.py
+++ b/kitsune/wiki/views.py
@@ -139,30 +139,39 @@ def document(request, document_slug, document=None):
     if doc.locale != request.LANGUAGE_CODE:
         # The "doc" is the parent document.
         if doc.current_revision:
-            # There is no translation, so we'll fall back to its approved parent,
-            # unless we find an approved translation in a fallback locale.
-            fallback_reason = "no_translation"
+            # Check for unapproved revisions
+            unapproved_translation_exists = Revision.objects.filter(
+                based_on=doc.current_revision,
+                is_approved=False,
+                document__locale=request.LANGUAGE_CODE,
+            ).exists()
+            if unapproved_translation_exists:
+                fallback_reason = "translation_not_approved"
+            else:
+                # There is no translation, so we'll fall back to its approved parent,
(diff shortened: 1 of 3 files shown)


kitsune/wiki/views.py
```diff
@@ -139,30 +139,37 @@ def document(request, document_slug, document=None):
...
-                # For showing the fallback locale explanation message to the user
-                fallback_reason = "fallback_locale"
-                full_locale_name = {
-                    request.LANGUAGE_CODE: LOCALES[request.LANGUAGE_CODE].native,
-                    fallback_locale: LOCALES[fallback_locale].native,
-                }
+            # Check for unapproved revisions
+            unapproved_revs = Revision.objects.filter(
+                based_on=doc.current_revision, is_approved=False
+            ).exists()
```
escattone, line 145: I think we need to restrict this to revisions in the requested language, or am I missing something? For example, I'm thinking this should be:
```suggestion
            unapproved_revs = Revision.objects.filter(
                based_on=doc.current_revision, is_approved=False, document__locale=request.LANGUAGE_CODE
            ).exists()
```

escattone: I was thinking also that maybe `unapproved_translation_exists` might be a better name?

escattone: Sorry, one more thing. I think we should add a good test for this as well.

smithellis: Thanks @escattone ! I agree with all three points.

escattone (approved):
Looks great! Sorry for the delay, I got sidetracked by a "rabbit hole" :)

PR #5800, opened 2023-12-01
ensure nulls ordered last when desc in DRF viewsets

escattone:
mozilla/sumo#1599

## Notes
I believe the `QuestionViewSet` is the only one that includes an ordering field that could contain `null` values (`last_answer`), but I replaced all of the Django Rest Framework `OrderingFilter` backends with the new `kitsune.sumo.api_utils.OrderingFilter` backend just so we're safe if any new fields are added that could contain `null` values.


smithellis:
r+

PR #5806, opened 2023-12-08
Correct SUMO Docs

smithellis:
Docs reference run-dev.sh which doesn't exist 
* It was removed in /commit/ff1828981a30915fcebd9e7c2b72d7ece6e91023

### DIFF
--- a/docs/hacking_howto.md
+++ b/docs/hacking_howto.md
@@ -54,7 +54,7 @@ faster and easier if you need to do that:
 
 ```
 make runshell
-bin/run-dev.sh
+./manage.py runserver 0.0.0.0:8000
 ```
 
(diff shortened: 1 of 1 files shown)


docs/hacking_howto.md
```diff
@@ -54,7 +54,7 @@ faster and easier if you need to do that:
 
 ```
 make runshell
-bin/run-dev.sh
+./manage.py runserver 0.0.0.0:8000
```
akatsoulas, line 57: Let's replace this with `make run`

akatsoulas (approved):
r+wc

docs/hacking_howto.md
akatsoulas: all good points :)

PR #5813, opened 2023-12-13
Autofocus the sign-in-step email field for the switching-devices wizard

mikeconley:
Fixes mozilla/sumo#1614.


hannajones (approved):
Works for me, thanks!

One thing I thought of though - I know there are some [accessibility concerns](https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/autofocus#accessibility_considerations) related to `autofocus` (which I'm assuming also apply automatically focusing an input via JS). Might not be a big deal in this case, but it could be worth getting QA to check how this works with different screen readers.

mikeconley:
Thanks, @hannajones!

PR #5816, opened 2023-12-14
Move focus and validation logic to an 'activate' method

hannajones:
This PR makes a slight change to how we're autofocusing the email input, wrapping the focus in an `activate` method that also resets the visibility of the associated error message. It also changes the wizard to call `activate` whenever the active step changes. In theory this could be used for other form steps to implement logic that needs to run every time a step becomes visible.

### DIFF
--- a/kitsune/sumo/static/sumo/js/form-wizard.js
+++ b/kitsune/sumo/static/sumo/js/form-wizard.js
@@ -302,9 +302,9 @@ export class FormWizard extends HTMLElement {
       let name = step.getAttribute("name");
       if (name === activeStep) {
         step.slot = "active";
-        step.activate && step.activate();
       } else {
         step.slot = "";
+        step.deactivate && step.deactivate();
       }
     }
   }
@@ -423,6 +423,14 @@ export class BaseFormStep extends HTMLElement {
     let prevState = Object.assign({}, this.#state);
     this.#state = Object.assign(this.#state, nextState);
     this.render(prevState, this.#state);
+  }
+
+  /**
+   * Method that can be implemented to handle any clean up needed when a form
+   * step is no longer active e.g. clearing error states.
+   */
+  deactivate() {
+    // NOOP
   }
 
   /**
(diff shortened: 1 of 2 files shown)


kitsune/sumo/static/sumo/js/form-wizard.js
```diff
@@ -302,6 +302,7 @@ export class FormWizard extends HTMLElement {
       let name = step.getAttribute("name");
       if (name === activeStep) {
         step.slot = "active";
+        step.activate && step.activate();
       } else {
         step.slot = "";
```
mikeconley, line 307: I like this approach! But I'm going to suggest we switch this around, because that's going to work better with the patch I'm working on for only setting focus to the email field in particular circumstances.

Instead of having an `activate`, can we have a deactivate? And have the deactivate on the sign-in-step be something like this:

```js
    if (this.shadowRoot.activeElement == this.#emailEl) {
      this.#emailEl.blur();
      this.#emailErrorEl.classList.remove("visible");
    }
```

mikeconley: Also, maybe instead of checking for the presence of a `deactivate` method, can we make it a NOOP method on the base class that subclasses can override?

mikeconley (changes requested):
Thanks! See below.

kitsune/sumo/static/sumo/js/form-wizard.js
hannajones: this works well thanks for the suggestions! calling `blur` will make it so that the input is no longer focused when the user gets sent back to step 1, does that seem like an issue? I feel like if the step changes it should be pretty obvious that the form is interactive...

mikeconley: I agree. I think the only autofocus case we truly care about at this point is upon visiting the page and hitting Step 1.

mikeconley (approved):
Works wonderfully, thank you!

PR #5819, opened 2023-12-14
Improve accessibility of switching-devices wizard step 1 autofocus behaviour.

mikeconley:
Fixes mozilla/sumo#1618.

This adds aria attributes to make the dialog context clearer if autofocus is applied to the email input. We also don't focus the email input if focus has changed in the interim of waiting to verify that this is a valid version of Firefox Desktop to show the wizard on. Finally, we use a state attribute (sorry) on the `SignInStep` shadow DOM to only ever attempt to focus the email input once.


hannajones (approved):
Works for me, thanks!

PR #5836, opened 2024-01-10
Stop all watches for inactive users

smithellis:
* On signal `post_save` for User, if user `is_active=False` delete all user's watches
* On signal `post_delete' for User, delete all user's watches
* Added migration to capture and correct Users who deactivated in the past
* Added test


smithellis:
Should we also cron something, or run something in prod to catch existing cases?

escattone (commented):
@smithellis I don't know why, but when I test this locally and deactivate a user, the user's watches are not deactivated as expected, and the reason seems to be that  `_deactivate_users` is never called.

escattone:
@smithellis I don't think it's going to work. My sense is that you'll need to use Django's [signals](https://docs.djangoproject.com/en/5.0/ref/signals/) and register receivers for the `post-save` and `post-delete` signals for the `User` model. For the `post-save` signal, only if `instance.is_active` is false, while for the `post-delete` you'll always look for any watches for that user instance and deactivate them.

escattone (changes requested):
Other than some small changes, it looks good.

akatsoulas:
Closing this in favor of #5882 . No need to add functionality for this in the admin interface.

PR #5838, opened 2024-01-11
Update newrelic agent

escattone:
mozilla/sumo#1565

## Notes
This PR does the following:
- Upgrades our Python `newrelic` agent to the latest version (`9.4.0`)
- Converts all of the deprecated settings from the `newrelic.ini` file to use the recommended settings instead.
    - `capture_params = true` becomes `attributes.include = request.parameters.*`
    - `error_collector.ignore_errors` becomes `error_collector.ignore_classes`
- Disables capturing of package information during the `newrelic` agent startup. **This is important**, otherwise the deployment in Kubernetes will fail because of failed `liveness` probes due to a very slow startup. By disabling this, we lose the reporting of the package versions when clicking on the environment tab in New Relic, but that doesn't seem important, and perhaps is rarely if ever used. We would lose the vulnerability reporting as well, but we don't have that feature enabled anyway, as we shouldn't, since we're doing that within GitHub. So, in the end, we lose little to nothing by disabling.


newrelic.ini
```diff
@@ -52,7 +52,7 @@ monitor_mode = true
 # write out a log file, it is also possible to say "stderr" and
 # output to standard error output. This would normally result in
 # output appearing in your web server log.
-#log_file = /tmp/newrelic-python-agent.log
+#log_file = stdout
```
akatsoulas, line 55: Is this change intended?

escattone: Yes. I simply wanted a reminder that if we use the `log_file` for debugging -- like I did --- we can/should use `stdout` instead of a log file on the pod itself, because it's much easier to look at the log data either in GCP or via `kubectl logs -f <pod-name>`.

akatsoulas (approved):

PR #5839, opened 2024-01-11
Bump oidc lib to v4

akatsoulas:


escattone (approved):
r+

The `post_json_mock` within the `test_link_sumo_account_fxa` test was missing the `status_code` attribute, which is needed with this new version of `mozilla-django-oidc` due to https://github.com/mozilla/mozilla-django-oidc/blob/107805c7466b79cd5a024448e16ed8f4dd2b7d85/mozilla_django_oidc/auth.py#L252, so I added that.

PR #5845, opened 2024-01-17
Add 'Live?' and "Target Link" columns to inproduct admin

smithellis:
Add 'KB Article' to inproduct admin
* Clickable link to KB Article OR None if not kb artictle

### DIFF
--- a/kitsune/inproduct/admin.py
+++ b/kitsune/inproduct/admin.py
@@ -1,13 +1,77 @@
+from django.conf import settings
 from django.contrib import admin
+from django.db.models import Case, Exists, F, OuterRef, Value, When
+from django.db.models.functions import Substr
+from django.utils.html import format_html
+from django.utils.translation import gettext_lazy as _
 
 from kitsune.inproduct.models import Redirect
+from kitsune.wiki.models import Document
+
+
+class KBLinkFilter(admin.SimpleListFilter):
+    title = _("KB article")
+    parameter_name = "kb_article"
+
+    def lookups(self, request, model_admin):
+        return (
+            ("with_kb", _("KB Article Exists")),
+            ("without_kb", _("No KB Article Exists")),
+        )
+
+    def queryset(self, request, queryset):
+        qs = queryset.annotate(
+            kb_slug=Case(
+                When(target__startswith="kb/", then=Substr("target", 4)),
+                When(target__startswith="/kb/", then=Substr("target", 5)),
+                default=None,
+            ),
+            kb_locale=Case(
+                When(locale="", then=Value(settings.WIKI_DEFAULT_LANGUAGE)),
+                default=F("locale"),
+            ),
+        ).filter(kb_slug__isnull=False)
+        doc_exists = Exists(
+            Document.objects.filter(slug=OuterRef("kb_slug"), locale=OuterRef("kb_locale"))
+        )
+        if self.value() == "with_kb":
+            # Filter for objects with a KB link
+            return qs.filter(doc_exists)
+        if self.value() == "without_kb":
+            # Filter for KB objects without a KB link
+            return qs.exclude(doc_exists)
 
 
 class RedirectAdmin(admin.ModelAdmin):
-    list_display = ("product", "version", "platform", "locale", "topic", "target")
+    list_display = (
+        "product",
+        "version",
+        "platform",
+        "locale",
+        "topic",
+        "target",
+        "kb_link",
+    )
     list_display_links = ("topic", "target")
-    list_filter = ("product", "version", "platform", "locale")
+    list_filter = ("product", "version", "platform", "locale", KBLinkFilter)
     search_fields = ("topic", "target")
+
+    @admin.display(description="KB Article")
+    def kb_link(self, obj):
+        target = obj.target.lstrip("/")
+        if not target.startswith("kb/"):
+            return None
+
+        slug = target.split("/")[-1]
+        locale = obj.locale if obj.locale else settings.WIKI_DEFAULT_LANGUAGE
+
+        try:
+            doc = Document.objects.get(slug=slug, locale=locale)
+        except Document.DoesNotExist:
+            return None
+        else:
+            target_url = doc.get_absolute_url()
+        return format_html('<a href="{}">{}</a>', target_url, target_url)
 
 
 admin.site.register(Redirect, RedirectAdmin)


kitsune/inproduct/admin.py
```diff
@@ -1,13 +1,32 @@
...
     list_display_links = ("topic", "target")
     list_filter = ("product", "version", "platform", "locale")
     search_fields = ("topic", "target")
 
+    @admin.display(description="KB Article")
+    def target_link(self, obj):
+        if "kb" in obj.target:
+            target_url = reverse(
+                "wiki.document", kwargs={"document_slug": obj.target.split("/")[-1]}
+            )
```
escattone, line 27: If a `locale` was specified in the redirect, we should use that locale in the link to the KB article, otherwise it's OK to default to `None` (which means that `reverse` will use `en-US`).
```suggestion
            target_url = reverse(
                "wiki.document",
                kwargs={"document_slug": obj.target.split("/")[-1]},
                locale=obj.locale if obj.locale else None,
            )
```

escattone (changes requested):

escattone:
@smithellis According to https://github.com/mozilla/sumo/issues/1606, we should only add the link if the article exists. Should we add that as well?

kitsune/inproduct/admin.py
```diff
@@ -1,13 +1,32 @@
...
+            return format_html('<a href="%s">%s</a>' % (target_url, target_url))
```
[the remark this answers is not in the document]
smithellis: Wow!  Thanks, good catch.

kitsune/inproduct/admin.py
smithellis: Thanks!

kitsune/inproduct/admin.py
```diff
@@ -1,13 +1,84 @@
...
+                    locale = obj.locale
+                else:
+                    locale = settings.WIKI_DEFAULT_LANGUAGE
+                rev = Revision.objects.filter(
+                    document__slug=doc_slug, document__locale=locale
+                ).exists()
+                return rev
+            except Revision.DoesNotExist:
+                pass
+        return False
```
escattone, line 42: I really like the use of a `SimpleListFilter`! It'd be nice, however, to avoid the hundreds of `Revision` queries this would entail, one for each redirect. I think this can be done in a single query. I tried the following, and it worked:

```suggestion
    def queryset(self, request, queryset):
        qs = queryset.annotate(
            kb_slug=Case(
                When(target__startswith="kb/", then=Replace("target", Value("kb/"))),
                When(target__startswith="/kb/", then=Replace("target", Value("/kb/"))),
                default=None,
            ),
            kb_locale=Case(
                When(locale="", then=Value(settings.WIKI_DEFAULT_LANGUAGE)),
                default=F("locale"),
            ),
        ).filter(kb_slug__isnull=False)
        doc_exists = Exists(
            Document.objects.filter(slug=OuterRef("kb_slug"), locale=OuterRef("kb_locale"))
        )
        if self.value() == "with_kb":
            # Filter for objects with a KB link
            return qs.filter(doc_exists)
        if self.value() == "without_kb":
            # Filter for KB objects without a KB link
            return qs.exclude(doc_exists)
```

kitsune/inproduct/admin.py
```diff
@@ -1,13 +1,84 @@
...
+                        document__slug=obj.target.split("/")[-1],
+                        document__locale=settings.WIKI_DEFAULT_LANGUAGE,
+                    )
+                    or None
+                )
+            if rev:
+                target_url = rev.document.get_absolute_url()
+                return format_html('<a href="{}">{}</a>', target_url, target_url)
+        else:
+            return None
```
escattone, line 81: There are two things I think that would make this better. First, I'd make the initial check tighter, more consistent with the actual redirect code, which strips out a leading "/" if there is one. Second,  I'd make this more DRY, and I think it'd be fine, as well as simpler and faster, to search `Document.objects` instead.  

Maybe something like:
```suggestion
        target = obj.target.lstrip("/")

        if not target.startswith("kb/"):
            return None

        slug = target.split("/")[-1]
        locale = obj.locale if obj.locale else settings.WIKI_DEFAULT_LANGUAGE

        try:
            doc = Document.objects.get(slug=slug, locale=locale)
        except Document.DoesNotExist:
            return None
        else:
            target_url = doc.get_absolute_url()

        return format_html('<a href="{}">{}</a>', target_url, target_url)
```

escattone (changes requested):

kitsune/inproduct/admin.py
```diff
@@ -1,13 +1,84 @@
...
+
+
+class KBLinkFilter(admin.SimpleListFilter):
+    title = _("KB article")
+    parameter_name = "kb_article"
+
+    def lookups(self, request, model_admin):
+        return (
+            ("with_kb", _("Has KB Article")),
+            ("without_kb", _("No KB Article")),
```
escattone, line 17: Super Nit. I just realized that with my suggested changes, which always filter the `queryset` based for redirects targeted at KB articles, that perhaps these titles would be better as:
```suggestion
            ("with_kb", _("KB Article Exists")),
            ("without_kb", _("No KB Article Exists")),
```

kitsune/inproduct/admin.py
smithellis: Agreed.

kitsune/inproduct/admin.py
smithellis: Sure thing.

smithellis:
@escattone I accepted all your feedback - the queryset modifications with annotation are pretty cool. Thanks for that.

kitsune/inproduct/admin.py
```diff
@@ -1,13 +1,93 @@
...
+        slug = target.split("/")[-1]
+        locale = obj.locale if obj.locale else settings.WIKI_DEFAULT_LANGUAGE
+
+        try:
+            doc = Document.objects.get(slug=slug, locale=locale)
+        except Document.DoesNotExist:
+            return None
+        else:
+            target_url = doc.get_absolute_url()
+        return format_html('<a href="{}">{}</a>', target_url, target_url)
```
escattone, line 58: I think this `has_kb_link` method is no longer used, and can be deleted?

smithellis: yarg!

escattone (approved):
Thanks @smithellis!

PR #5854, opened 2024-01-23
User Clean Up : Part 1

smithellis:
* Drop all `is_fxa_migrated` users (via Profile)
* Drop the field from the database


smithellis:
Got ahead of myself.

PR #5856, opened 2024-01-23
is_fxa_migrated removal, user management process created

smithellis:
* Remove all references to is_fxa_migrated
* Implement management command that purges inactive users who don't have live content
* Implement a cron to run the management command monthly
* When a user deletes themselves, they get good and deleted


kitsune/users/models.py
```diff
@@ -158,7 +157,6 @@ def clear(self):
         self.people_mozilla_org = ""
         self.matrix_handle = ""
         self.city = ""
-        self.is_fxa_migrated = False
```
akatsoulas, line 161: this removal should be along with the migration

kitsune/users/migrations/0029_remove_profile_is_fxa_migrated.py
```diff
@@ -0,0 +1,20 @@
...
+    Profile = apps.get_model("users", "Profile")
+    Profile.objects.filter(is_fxa_migrated=False).delete()
+    
+class Migration(migrations.Migration):
+    dependencies = [
+        ("users", "0028_alter_profile_bio_and_upper_name_idx"),
+    ]
+
+    operations = [
+        migrations.RunPython(purge_is_fxa_migrated_false_users),
```
akatsoulas, line 15: the data migration needs a `noop` backwards method in order to be possible to go backwards if we ever need to

kitsune/users/management/commands/user_maintenance.py
```diff
@@ -0,0 +1,44 @@
+from django.core.management.base import BaseCommand
+
+from datetime import datetime, timedelta
+
+from django.contrib.auth.models import User
+from kitsune.users.utils import anonymize_user
+
+
+class Command(BaseCommand):
```
akatsoulas, line 9: Instead of using a command we can use a celery task. This will make the transition easier when we remove the APScheduler.

kitsune/users/management/commands/user_maintenance.py
```diff
@@ -0,0 +1,44 @@
...
+            is_active=False,
+            last_login__lte=datetime.now() - timedelta(days=365),
+            is_superuser=False,
+        )
+
+        # If these users don't have revisions or questions that are not archived,
+        # delete the user
+        for user in inactive_users:
+            # Get all the non archived content for the user
+            revisions = user.created_revisions.all().filter(document__is_archived=False)
```
akatsoulas, line 34: Is there a way to avoid the queries within the loop?

smithellis:
Have to change direction - model cascades complicate this.

PR #5861, opened 2024-01-26
Manage inactive users

smithellis:
* Delete users who have not logged in during the past year and who have no live revisions, questions or solutions

### DIFF
--- a/kitsune/users/migrations/0029_drop_inactive_users_no_content.py
+++ b/kitsune/users/migrations/0029_drop_inactive_users_no_content.py
@@ -1,0 +1,39 @@
+
+from datetime import datetime, timedelta
+
+from django.contrib.auth.models import User
+from django.db import migrations
+
+from kitsune.users.utils import anonymize_user
+
+def check_for_user_content(user):
+    """Check if a user has any content that is live,
+    not archived """
+    return ( 
+        user.revisions.filter(is_approved=True, is_archived=False).exists() or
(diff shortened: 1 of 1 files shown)


Anagirotti (approved):

kitsune/users/migrations/0029_drop_inactive_users_no_content.py
```diff
@@ -0,0 +1,39 @@
...
+def check_for_user_content(user):
+    """Check if a user has any content that is live,
+    not archived """
+    return ( 
+        user.revisions.filter(is_approved=True, is_archived=False).exists() or
+        user.questions.filter(is_archived=False).exists() or
+        user.questions.filter(solution__creator=user, is_spam=False).exists()
+    )
+
+def drop_inactive_users_no_content(apps, schema_editor):
```
akatsoulas, line 18: This feels that it should be a recurrent theme so it would be better suited for a periodic task. What we could though once is to drop all users that are not migrated to FxA and have no content until we have a good way forward on how to handle that.

akatsoulas (changes requested):

kitsune/users/migrations/0029_drop_inactive_users_no_content.py
yehiamitro111: Thankyou

akatsoulas:
The requirements have changed for this and it will be replaced by subsequent PRs

PR #5866, opened 2024-02-01
Sentry -- ignore events from the `request.summary` logger

escattone:
mozilla/sumo#1656


ahoneiser (approved):

akatsoulas:
I believe there is value in keeping track of those 404 errors. When there's a spike it's a way to know that something is not performing as it should (configuration error in the admin, bad links, broken in-product links etc). Completely removing this ability should not be the solution to reduce the number of events. We should do some sampling so that we can preserve the information but also reduce the number of registered events if this is a big problem. Identifying the number of 404 can provide it's own value.

akatsoulas (approved):

PR #5869, opened 2024-02-02
fix flicker in article restriction group select

escattone:
mozilla/sumo#1657

Second attempt to fix the flicker. Boy, the styling of the `<select>` element is difficult to control.


kitsune/wiki/forms.py
```diff
@@ -141,9 +141,9 @@ def __init__(self, *args, **kwargs):
     is_archived = forms.BooleanField(label=_lazy("Obsolete:"), required=False)
 
     restrict_to_groups = forms.ModelMultipleChoiceField(
-        label=_lazy("The document will be restricted to members of the selected group(s)."),
         required=False,
         queryset=Group.objects.order_by("name").all(),
+        label=_lazy("The document will be restricted to members of the selected group(s)."),
```
escattone, line 146: This is just moving the label line to a different spot.

PR #5881, opened 2024-02-08
filter event notifications by restrictions

escattone:
mozilla/sumo#1664

### DIFF
--- a/kitsune/kbforums/events.py
+++ b/kitsune/kbforums/events.py
@@ -5,6 +5,7 @@ from kitsune.kbforums.models import Thread
 from kitsune.sumo.email_utils import emails_with_users_and_watches
 from kitsune.sumo.templatetags.jinja_helpers import add_utm
 from kitsune.tidings.events import Event, EventUnion, InstanceEvent
+from kitsune.wiki.events import filter_by_unrestricted
 from kitsune.wiki.models import Document
 
 
@@ -73,6 +74,10 @@ class NewPostEvent(InstanceEvent):
             self, NewThreadEvent(self.reply), NewPostInLocaleEvent(self.reply)
         ).send_emails(exclude=exclude)
 
+    def _users_watching(self, **kwargs):
+        users_and_watches = super()._users_watching(**kwargs)
+        return filter_by_unrestricted(self.instance.document, users_and_watches)
+
     def _mails(self, users_and_watches):
         return new_post_mails(self.reply, users_and_watches)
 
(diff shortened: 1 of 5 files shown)


kitsune/kbforums/events.py
```diff
@@ -103,9 +108,11 @@ def __init__(self, post):
 
     def send_emails(self, exclude=None):
         """Notify watches of the document and of the locale."""
-        return EventUnion(
-            self, NewThreadEvent(self.post), NewThreadInLocaleEvent(self.post)
-        ).send_emails(exclude=exclude)
+        return EventUnion(self, NewThreadInLocaleEvent(self.post)).send_emails(exclude=exclude)
```
escattone, line 111: When working on this, I noticed that the second argument to the `EventUnion`  constructor -- `NewThreadEvent(self.post)` -- was a duplicate of the first -- `self` -- so I removed it.

kitsune/wiki/models.py
```diff
@@ -742,7 +743,7 @@ def gets_mozilla_account_cta(self):
             self.parent.slug if self.parent else self.slug
         ) in settings.MOZILLA_ACCOUNT_ARTICLES
 
-    @property
+    @cached_property
```
akatsoulas, line 746: Nice! How about adding a method to do the calculation and then set that method here as a cached property? This way we can force also the evaluation if ever needed without having logic in the property. Eg

```python
def get_is_restricted(self):
     """Is this document restricted in terms of visibility?"""
        if self.parent:
            # Translations follow their parent's restrictions.
            return self.parent.is_restricted
        return self.restrict_to_groups.exists()

@ cached_property
def is_restricted(self):
    return self.get_is_restricted()
```

kitsune/wiki/models.py
```diff
@@ -751,19 +752,28 @@ def is_restricted(self):
...
+            or (user.id in self.restrict_to_user_ids)
+        )
+
+    @cached_property
+    def restrict_to_user_ids(self):
+        """
+        Returns a set of user ids comprising all members of the groups specified in
+        "restrict_to_groups".
+        """
+        return set(
```
akatsoulas, line 775: I would move this query into the `is_unrestricted_for` method. We are adding a property that's not used anywhere else. Let's move it in the above method and if we can extract it if we ever need it in another place.

akatsoulas (approved):
r+wc :guitar:

kitsune/wiki/models.py
escattone: That's a great idea, thanks!

PR #5886, opened 2024-02-13
fix article helpfulness voting

escattone:
mozilla/sumo#1649

### DIFF
--- a/kitsune/sumo/static/sumo/js/ajaxvote.js
+++ b/kitsune/sumo/static/sumo/js/ajaxvote.js
@@ -73,7 +73,7 @@ export default function AjaxVote(form, options) {
             },
             error: function() {
               var msg = gettext('There was an error submitting your vote.');
-              self.showMessage(msg, $btn);
+              self.showMessage(msg, $btn, $form);
               $btns.prop("disabled", false);
               $form.removeClass('busy');
             }
--- a/kitsune/sumo/views.py
+++ b/kitsune/sumo/views.py
@@ -9,6 +9,7 @@ from django.shortcuts import render
 from django.utils import translation
 from django.utils.translation import gettext as _
 from django.views.decorators.cache import never_cache
+from django.views.decorators.http import require_GET
 
 from kitsune.lib.sumo_locales import LOCALES
 from kitsune.sumo.decorators import cors_enabled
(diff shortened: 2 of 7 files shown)


kitsune/wiki/views.py
```diff
@@ -1244,7 +1244,7 @@ def helpful_vote(request, document_slug):
             # Send a survey if flag is enabled and vote wasn't helpful.
             if "helpful" not in request.POST:
                 survey = render_to_string(
-                    "wiki/includes/unhelpful_survey.html", {"vote_id": vote.id}
+                    "wiki/includes/unhelpful_survey.html", {"vote_id": vote.id}, request
```
escattone, line 1247: Without passing `request` into the `render_to_string`, it has been failing to render `{% csrf_token %}` in the `unhelpful_survey.html` template, since the `csrf_token` tag relies on a request context, so this fixes that issue.

kitsune/sumo/views.py
```diff
@@ -26,6 +27,12 @@ def locales(request):
     return render(request, template, dict(next_url=get_next_url(request) or reverse("home")))
 
 
+@never_cache
+@require_GET
+def csrftoken(request):
+    return render(request, "sumo/csrftoken_for_form.html")
```
akatsoulas, line 33: no need to render the an html page here. We could use the `django.middleware.csrf.get_token()` here

```python
from django.middleware.csrf import get_token()
from django.http import JsonResponse

def csrf_token(request):
   return JsonResponse('csrfToken': get_token(request)
```

and use `fetch` directly on the url

escattone: I think we still need to use Django's `csrf_token` template tag here because it returns HTML meant to be embedded in a form, a hidden input element with the CSRF token, which is what we then replace the placeholders with:
```html
<input type="hidden" name="csrfmiddlewaretoken" value="7T1AMhLMl9SKaKnp6u0vIK5Pb3wEtMxVKzMgrRE5c5EjjaxZb2otyuEnp7Ti45RL">
```

escattone: Actually, I'll do what you suggested, in that I'll get the token, and simply replace the `value` attribute of any hidden `input` elements with `name="csrfmiddlewartoken"` within all forms.

kitsune/sumo/static/sumo/js/ajaxvote.js
```diff
@@ -73,7 +73,7 @@ export default function AjaxVote(form, options) {
             },
             error: function() {
               var msg = gettext('There was an error submitting your vote.');
-              self.showMessage(msg, $btn);
+              self.showMessage(msg, $btn, $form);
```
escattone, line 76: This fixes the JS error that caused the spinner to be displayed forever instead of displaying the error message when votiing, and was a symptom of the larger CSRF token error.

akatsoulas (approved):

PR #5887, opened 2024-02-15
make wiki links to articles/images case-insensitive

escattone:
mozilla/sumo#1605

This PR makes the title search for all Wiki links to articles, images and videos case-insensitive.

### DIFF
--- a/kitsune/sumo/parser.py
+++ b/kitsune/sumo/parser.py
@@ -61,53 +61,62 @@ def wiki_to_html(
     )
 
 
-def get_object_fallback(cls, title, locale, default=None, **kwargs):
-    """Return an instance of cls matching title and locale, or fall
-    back to the default locale.
-
-    When falling back to the default locale, follow any wiki redirects
-    internally.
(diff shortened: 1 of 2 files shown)


kitsune/sumo/parser.py
```diff
@@ -61,53 +61,62 @@ def wiki_to_html(
...
+                raise
+            # We only run the much more expensive case-insensitive match if the
+            # case-sensitive match failed.
+            return cls.objects.get(title__iexact=title, locale=locale, **kwargs)
 
-    """
     try:
-        return cls.objects.get(title=title, locale=locale, **kwargs)
-    except (cls.DoesNotExist, IOError):
+        return get(locale)
```
akatsoulas, line 89: This add a nested try/except. We should rewrite this in a way to avoid that.

akatsoulas (approved):
I am suspecting that we are getting many 404s b/c of this thus I am merging to unblock testing. We should change though that try/except nesting

PR #5895, opened 2024-02-21
Timeout decorator

akatsoulas:
mozilla/sumo#1677


escattone (approved):
r+wc

This all looks great, but I'm wondering about the following:

- The `match_regex_with_timeout` function does a lot of unnecessary work. It doesn't need to `findall` because if it finds any it should return `True`. So, it could be made much more efficient like this:
    ```python
    def match_regex_with_timeout(compiled_regex, data):
        return bool(compiled_regex.search(data))
    ```
- Maybe we don't need the `timeout` decorator at all if the function is much more efficient?

akatsoulas:
The problem with the timeout is not `findany` or `findall` but in how many ways the regex can match a pattern which can grow exponentially the larger the input is. Looking at the regex though it doesn't seem that can cause these kind of problems. I am dropping completely the timeout decorator and I am keeping `findall`. If we start seeing problems we can re-introduce the new decorator.

PR #5900, opened 2024-02-23
include locale in query within review-revision view

escattone:
mozilla/sumo#1680


akatsoulas (approved):
Code looks good. r+ with tests passing

PR #5905, opened 2024-02-27
only include docs without a translation or with a visible translation

escattone:
mozilla/sumo#1674 (without regressing mozilla/sumo#1678)

This resolves mozilla/sumo#1674 by only including documents that either have no translation or a translation that is visible to the current user. In other words, it excludes documents that could show a link that would 404 if clicked.

### DIFF
--- a/kitsune/dashboards/readouts.py
+++ b/kitsune/dashboards/readouts.py
@@ -74,19 +74,27 @@ def get_visits_subquery(document=OuterRef("pk"), period=LAST_30_DAYS):
     )
 
 
-def get_visible_or_no_translation_condition(user, locale):
+def visible_translation_exists(user, locale):
     """
-    Returns a queryset expression for Document queries that only includes documents
-    that either do not have a translation for the given locale, or have a translation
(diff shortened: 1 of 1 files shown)


kitsune/dashboards/readouts.py
```diff
@@ -74,6 +74,21 @@ def get_visits_subquery(document=OuterRef("pk"), period=LAST_30_DAYS):
     )
 
 
+def get_visible_or_no_translation_condition(user, locale):
+    """
+    Returns a queryset expression for Document queries that only includes documents
+    that either do not have a translation for the given locale, or have a translation
+    that is visible to the given user.
+    """
+    return ~Q(translations__locale=locale) | Exists(
```
akatsoulas, line 83: nit: I am wondering if it would be clearer here to create a couple of `Subquery` and combine them into another one.
Eg `get_visible_subquery`, `get_no_translation_subquery and combine those in the filter below

akatsoulas (approved):
r+wc

PR #5908, opened 2024-02-29
Allow Staff to private message all group members

smithellis:
* Added groups as PM targets
* Use new `Staff` group to check for perms
	* Added `is_staff` to `user.profile`
* Updated `InboxMessages` and `OutboxMessages` models to have
	`to_group`
* Updated message form to show To Group to empowered users
* Updated group profile page
* Consolidate `users.autocomplete.js` and `groups.autocomplete.js`
    into `object.autocomplete.js`
* Added 'To Group' to outbox view for empowered users
* General changes in scss to accommodate new column
* Modified `outbox.html` and scss to be more usable


akatsoulas (commented):
I did a first pass. Overall this looks good. Let me know when it's ready for another one.

kitsune/groups/views.py
```diff
@@ -237,3 +239,8 @@ def _user_can_manage_leaders(user, group_profile):
     """Can the given user add and remove leaders?"""
     # Limit to users with the change_groupprofile permission
     return user.has_perm("groups.change_groupprofile")
+
+
+def _user_can_message_group(user, group_profile):
```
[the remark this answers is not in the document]
smithellis: Per some slack chat - I'm adding a constant for 'PRIVILEGED_GROUP' and we can check for that in the Profile. So yes to your suggestion, just a little different post a few more chats.

kitsune/messages/views.py
```diff
@@ -76,16 +78,38 @@ def new_message(request):
             )
             return HttpResponseRedirect(reverse("messages.new"))
 
+    if to_group:
+        try:
+            for groupname in to_group.split(","):
```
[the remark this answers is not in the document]
smithellis: Moved.

kitsune/messages/views.py
```diff
@@ -76,16 +78,38 @@ def new_message(request):
...
+                Group.objects.get(name=groupname)
+        except Group.DoesNotExist:
+            contrib_messages.add_message(
```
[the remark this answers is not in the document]
smithellis: Let me see what I can do.

kitsune/messages/forms.py
```diff
@@ -1,21 +1,43 @@
...
 
 TO_PLACEHOLDER = _lazy("username1, username2,...")
+GROUP_PLACEHOLDER = _lazy("group1, group2,...")
 
 
 class MessageForm(forms.Form):
     """Form send a private message."""
 
+    def __init__(self, *args, **kwargs):
+        user = kwargs.pop("user", None)
```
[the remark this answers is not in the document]
smithellis: Yes.

kitsune/groups/templatetags/jinja_helpers.py
```diff
@@ -25,3 +27,13 @@ def group_link(group):
         return Markup(html)
     except GroupProfile.DoesNotExist:
         return group.name
+
+
+@library.global_function
+def private_message_group(group):
```
akatsoulas, line 33: As far as I can tell this is only used in the `profile.html`. If this is the case, why don't we add the html directly into the profile?

Even if we wanted to extract this, probably a macro would be a better candidate. For now I would suggest to just add it in the html and extract it only when/if needed.

kitsune/groups/views.py
```diff
@@ -47,6 +48,7 @@ def profile(request, group_slug, member_form=None, leader_form=None):
             "members": members,
             "user_can_edit": user_can_edit,
             "user_can_manage_leaders": user_can_manage_leaders,
+            "user_can_message_group": user_can_message_group,
```
akatsoulas, line 51: No need for the var. It's not used anywhere else
`"user_can_message_group": request.user.profile.is_staff`

kitsune/messages/forms.py
```diff
@@ -1,23 +1,43 @@
...
-            attrs={"placeholder": TO_PLACEHOLDER, "class": "user-autocomplete"}
-        ),
-    )
-    message = forms.CharField(label=_lazy("Message:"), max_length=10000, widget=forms.Textarea)
-    in_reply_to = forms.IntegerField(widget=forms.HiddenInput, required=False)
+    def __init__(self, *args, **kwargs):
+        self.user = kwargs.pop("user", None)
+        super(MessageForm, self).__init__(*args, **kwargs)
+
+        self.fields["to"] = MultiUsernameField(
```
akatsoulas, line 14: Let's move all the field declarations outside of the `__init__`

akatsoulas: Regarding field ordering, you can also use `field_order`
https://docs.djangoproject.com/en/5.0/ref/forms/api/#django.forms.Form.field_order

kitsune/messages/forms.py
```diff
@@ -1,23 +1,43 @@
...
+            label=_lazy("To:"),
+            widget=forms.TextInput(attrs={"class": "user-autocomplete"}),
+            required=not self.user.profile.is_staff,
+        )
+
+        self.fields["to_group"] = forms.CharField(widget=forms.HiddenInput, required=False)
```
akatsoulas, line 20: probably this needs to go?

kitsune/messages/forms.py
akatsoulas: I wonder if we could combine `to` and `to_group` into a single field and add autocompletion for groups only if the user is staff. It's going to be simpler for the user and the validation will be more straight-forward.

smithellis:
I believe I will close this in favor of https://github.com/mozilla/kitsune/pull/5922

smithellis:
Closing in favor of https://github.com/mozilla/kitsune/pull/5922

PR #5910, opened 2024-03-01
Fix group profile page

smithellis:
The Group Profile page was severely ugly/broken. I have triaged it to a usable state.

### DIFF
--- a/kitsune/groups/jinja2/groups/edit.html
+++ b/kitsune/groups/jinja2/groups/edit.html
@@ -4,7 +4,8 @@
 
 {% block content %}
   <article id="group-profile" class="main">
-    <h1>{{ _('Edit {group} profile information')|f(group=profile.group.name) }}</h1>
+  <h1>{{ _('Edit Group Description') }}</h1>
+    <strong>{{ _('Editing description for group {group}')|f(group=profile.group.name) }}</strong>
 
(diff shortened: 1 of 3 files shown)


kelimuttu:
Ooh, this is amazing. Thanks for working on this, Smith!

escattone (changes requested):
Thanks for updates @smithellis. Several new observations:
- It appears that we've lost the functionality to add new members?
- I'm not sure we need the `Edit group in admin` button, since I think we can do everything on the page that we could do in the group profile admin? However, if we'd prefer to keep it, since it's currently available, it seems like it should be moved out of the group profile description section, and up into the top area? I also prefer the text `Edit in admin` instead of`Edit group in admin`, just because it shorter.
- Some of the buttons seem too big to me, because of the text. What do you think of shortening `Change Group Avatar` and `Delete Group Avatar` to just `Change Avatar` and `Delete Avatar`?
- When removing members/leaders, the text on the confirmation page is within an `h1` tag, which looks really over-sized to me.

image

image

akatsoulas:
@smithellis is this PR still active and relevant?

akatsoulas:
bumping this

akatsoulas:
Closing this as stale. Please re-open if there's an update

PR #5916, opened 2024-03-12
Switching playwright reporting from HTML to Allure.

emilghittasv:
- Using allure.step() to define steps in allure report.
- Modifying the soft assertions to play nicely with allure reporting by interleaving check and step.
- Adding allure to poetry.
- Modifying the playwright.yml to exclude the HTML reporter & use Allure reporting instead.
- Building up the logic of attaching screencasts to Allure report on test failure.

### DIFF
--- a/.github/workflows/playwright.yml
+++ b/.github/workflows/playwright.yml
@@ -51,132 +51,135 @@ jobs:
         id: create-sessions
         working-directory: playwright_tests
         run: |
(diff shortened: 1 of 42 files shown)


akatsoulas (approved):
I just skimmed through the changes. Is there a way in the future to break PRs like this in smaller chunks?

PR #5917, opened 2024-03-12
Add hubs announcment

smithellis:
This adds an announcement about Hubs sunset to the product pages and wiki pages related to Hubs.

This is a temporary solution until we add Product based announcements to the admin.


escattone (approved):
r+

I think the intention is for this to be temporary, so there's no need to make it DRY, for example.

PR #5922, opened 2024-03-18
Allow Staff members to message groups

smithellis:
* API for messages to get users/groups
* Added custom form field type in `form_fields.py` to handle group AND user selection and validation
* Update form, models, urls and tests for messages
* Update group profile page to add button for messaging group
* Removed unnecessary validation from `view.py` for `new_messages`
* Rewrote `new_messages` to handle Groups
* New .js file for handling drop down for messages `messages.autocomplete.js`
* Updated webpack `entrypoints.js` to handle new file
* Updated `tokenizer.js`
* Updated `outbox.html`, `read-outbox.html` and `_inbox.scss`
* Created abstracted group in `settings.py` for our `Staff` group
* Added new cached property to `Profile` model for `in_staff_group`
* Added `in_staff_group` function to `sumo/utils.py` ( thanks @escattone ! )


kitsune/messages/api.py
```diff
@@ -0,0 +1,84 @@
...
+            suggestions.append(create_suggestion(group, "Group", False))
+
+    append_exact_match(pre, suggestions, is_staff)
+    return suggestions
+
+
+@login_required
+@require_GET
+@json_view
+def usernames_or_groups(request):
```
akatsoulas, line 78: Do we need this? Can't we used directly the `get_autocomplete_suggestions` and move the checks of this method into the former?

kitsune/messages/urls.py
```diff
@@ -1,9 +1,16 @@
-from django.urls import re_path
+from django.urls import include, re_path
 
-from kitsune.messages import views
+from kitsune.messages import api, views
+
+api_patterns = [
+    re_path(
+        r"^usernames_or_groups", api.usernames_or_groups, name="messages.api.usernames_or_groups"
```
akatsoulas, line 7: I would rephrase this to `autocomplete` or something like that. Similar comment about the method. I also added a comment in the api.py to use directly the `get_autocomplete_suggestions`. Personally I find it more descriptive and suitable to focus on the autocomplete functionality in the url and not on the type that's returned

kitsune/messages/forms.py
```diff
@@ -1,20 +1,20 @@
...
-
-TO_PLACEHOLDER = _lazy("username1, username2,...")
+from kitsune.sumo.form_fields import MultiUsernameOrGroupnameField
 
 
 class MessageForm(forms.Form):
     """Form send a private message."""
 
-    to = MultiUsernameField(
+    to = MultiUsernameOrGroupnameField(
```
akatsoulas, line 10: nit: `MultiUserGroupNameField` - feel free to skip

kitsune/messages/forms.py
```diff
@@ -1,20 +1,20 @@
...
         label=_lazy("To:"),
-        widget=forms.TextInput(
-            attrs={"placeholder": TO_PLACEHOLDER, "class": "user-autocomplete"}
```
akatsoulas, line 16: It would be nice to have a placeholder. You can pass in the form `__init__` the request.user and update dynamically the text based on whether they can select groups.

kitsune/users/models.py
```diff
@@ -227,6 +227,10 @@ def answer_helpfulness(self):
     def is_subscriber(self):
         return self.products.exists()
 
+    @property
+    def is_staff(self):
+        return self.user.groups.filter(name="Staff").exists()
```
akatsoulas, line 232: Let's add the name in the settings. Eg SUMO_STAFF="Staff" and use `settings.SUMO_STAFF` here

kitsune/users/tests/test_tasks.py
```diff
@@ -30,8 +30,8 @@ def test_process_delete_user(self):
         # Populate inboxes and outboxes with messages between the user and other users.
         other_users = UserFactory.create_batch(2)
         for sender in other_users:
-            send_message([user], "foo", sender=sender)
-        send_message(other_users, "bar", sender=user)
+            send_message([user], to_group="", text="foo", sender=sender)
```
akatsoulas, line 33: In `users/tests/test_views.py` you are using `to_group=None`. Let's be consistent and use either `""` or `None`

webpack/entrypoints.js
```diff
@@ -90,7 +90,7 @@ const entrypoints = {
     "sumo/js/reportabuse.js",
   ],
   messages: [
-    "sumo/js/users.autocomplete.js",
+    "sumo/js/users_or_groups.autocomplete.js",
```
akatsoulas, line 93: s/users_or_groups.autocomplete.js/messages.autocomplete.js/ ?

kitsune/messages/tests/test_templates.py
```diff
@@ -23,7 +23,7 @@ def test_send_message_page(self):
 
     def _test_send_message_to(self, to):
         # Post a new message and verify it was sent.
-        data = {"to": to, "message": "hi there"}
+        data = {"to": to, "to_group": "", "message": "hi there"}
```
akatsoulas, line 26: Similar comment with the other test. Let's user either an empty string or None but not both

kitsune/messages/forms.py
```diff
@@ -1,20 +1,20 @@
...
-        ),
+        widget=forms.TextInput(attrs={"class": "user-autocomplete"}),
+    )
+    to_group = forms.CharField(
```
akatsoulas, line 14: Is this used anywhere? Do we need it?

kitsune/messages/forms.py
```diff
@@ -24,5 +24,10 @@ class ReplyForm(forms.Form):
     """Form to reply to a private message."""
 
     to = forms.CharField(widget=forms.HiddenInput)
+    to_group = forms.CharField(
```
akatsoulas, line 27: Same comment: Is this used/needed?

kitsune/sumo/form_fields.py
```diff
@@ -83,6 +83,41 @@ def to_python(self, value):
...
+    def to_python(self, value):
+        if not value:
+            if self.required:
+                raise forms.ValidationError(_("To field is required."))
+            else:
+                return []
+
+        users_and_groups = []
+        for user_or_group in value.split(","):
+            if ":" in user_or_group:
```
akatsoulas, line 100: nit: you can skip the `if` check by using
```python
_type, *names = user_or_group.split(":")
```

akatsoulas (changes requested):

kitsune/groups/jinja2/groups/profile.html
```diff
@@ -18,7 +18,9 @@
         <a class="edit" href="{{ url('admin:groups_groupprofile_change', profile.id) }}">{{ _('Edit in admin') }}</a>
       {% endif %}
       <h1>{{ profile.group.name }}</h1>
-
+      {% if user.profile.is_staff %}
+        <p class="pm"><a class="sumo-button primary-button button-lg" href="{{ url('messages.new')|urlparams(to=profile.group.name) }}">{{ _('Private message group members') }}</a></p>
```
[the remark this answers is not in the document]
smithellis: Got it.

kitsune/messages/api.py
```diff
@@ -0,0 +1,113 @@
...
+from kitsune.access.decorators import login_required
+from kitsune.sumo.decorators import json_view
+from kitsune.sumo.utils import webpack_static
+from kitsune.users.models import Profile
+from kitsune.users.templatetags.jinja_helpers import profile_avatar
+
+
+def display_name_or_none(user):
+    try:
+        return user.profile.name
```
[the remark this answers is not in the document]
smithellis: Interesting - this was repeated from `users/api.py` so I assumed it was solving a problem. I'll remove and test. Thanks!

kitsune/messages/utils.py
```diff
@@ -1,25 +1,32 @@
...
     """
     if sender:
-        msg = OutboxMessage.objects.create(sender=sender, message=text)
-        msg.to.add(*to)
-    for user in to:
-        im = InboxMessage.objects.create(sender=sender, to=user, message=text)
-        if Setting.get_for_user(user, "email_private_messages"):
-            email_private_message(inbox_message_id=im.id)
+        obm = OutboxMessage.objects.create(sender=sender, message=text)
+        obm.to.set(to)
```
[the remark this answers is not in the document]
smithellis: `.set` forces uniqueness so we don't have to worry about spamming people.

kitsune/messages/api.py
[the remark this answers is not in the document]
smithellis: If a user doesn't have a **profile** (which should never happen, but happens for me locally when I don't create a profile for a test user) - you don't get `None`, you get ye ole `Traceback....RelatedObjectDoesNotExist`. 

I think removing this *should* be fine in prod? Is there any case where we have production users without profiles? Even an invalid case would be bad if I remove this as it will break the return of suggestions by producing a 500 on the api call.

kitsune/messages/forms.py
smithellis: The placeholder wasn't functional in production in the first place! But, because I aim to please, I have patched `jquery.tokeninput.js` and the form `__init__` to get the thing back on the rails.

kitsune/users/models.py
smithellis: Ok. It seems weird to me to have SUMO_STAFF be a placeholder for Staff. But the idea of that group being more configurable makes sense. Done.

kitsune/messages/utils.py
```diff
@@ -1,25 +1,32 @@
...
+        if to_group:
+            obm.to_group.set(to_group)
+    for obj in to:
+        if isinstance(obj, User):
+            im = InboxMessage.objects.create(sender=sender, to=obj, message=text)
+            if to_group:
```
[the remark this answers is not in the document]
smithellis: I tried to make this clearer in new code - the idea is if we have a `User`, we make an `InboxMessage` for them, and then we add the groups it was sent to as well so we can later see them in our inbox email.

kitsune/messages/forms.py
smithellis: Nope.

kitsune/messages/forms.py
smithellis: And nope again.

kitsune/messages/api.py
smithellis: I was trying to break the code down into more palatable blocks - maybe I went too far? Re-integrating.

kitsune/messages/urls.py
smithellis: Agreed.

kitsune/messages/api.py
```diff
@@ -0,0 +1,75 @@
...
+
+        for user in users:
+            suggestions.append(create_suggestion(user, "User"))
+
+        if is_staff:
+            groups = Group.objects.filter(name__istartswith=pre)[:10]
+            for group in groups:
+                suggestions.append(create_suggestion(group, "Group", False))
+
+        append_exact_match(pre, suggestions, is_staff)
```
escattone, line 72: I noticed that when `pre` is an exact match -- for example I had a group named `Beatles` and typed `Beatles` in the `To:` search bar -- that the group `Beatles` was returned twice, because it was found and appended first by the `istartswith` search and then again by the `iexact` search. It seems to me that `append_exact_match()` isn't needed, unless perhaps for ordering the results such that the exact match is placed first in the list of returned suggestions? In any case, the suggestions shouldn't include duplicate results.

escattone (changes requested):

kitsune/messages/api.py
smithellis: Thanks @escattone - I wondered about this code, as it existed in the original selector. I think it would make more sense if the exact match were highlighted green or somehow called out - but as it is I agree with you that this is not ideal.

smithellis: I have dropped that code!

kitsune/sumo/form_fields.py
smithellis: Altered functionality a good bit - this is no longer relevant.

kitsune/sumo/form_fields.py
```diff
@@ -83,6 +83,41 @@ def to_python(self, value):
...
+                _type, name = user_or_group.split(":")
+                name = name.strip()
+            else:
+                name = user_or_group.strip()
+            msg = ""
+            if name:
```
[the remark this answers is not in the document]
smithellis: I've abandoned this completely - no longer relevant.

kitsune/messages/api.py
```diff
@@ -0,0 +1,57 @@
...
+                "avatar": profile_avatar(item, 24)
+                if is_user
+                else webpack_static(settings.DEFAULT_AVATAR),
+            }
+
+        suggestions = []
+        user_criteria = Q(username__istartswith=pre) | Q(profile__name__istartswith=pre)
+        users = (
+            User.objects.filter(user_criteria, is_active=True)
+            .exclude(profile__is_fxa_migrated=False)
```
escattone, line 43: I checked, and there are 189 active users in production that do not have a profile. Although 189 out of over 4.2 million is very small, perhaps this query could change to ensure that we're not including any users without a profile. I think using `.exclude(profile__is_fxa_migrated=False)` won't exclude users without a profile, but I think using `.filter(user_criteria, is_active=True, profile__is_fxa_migrated=True)` will.

kitsune/messages/api.py
```diff
@@ -0,0 +1,57 @@
...
+    if not pre or not request.user.is_authenticated:
+        return []
+
+    def get_suggestions(pre, is_staff=False):
+        """Generate autocomplete suggestions for users and, if applicable, groups."""
+
+        def create_suggestion(item, item_type, is_user=True):
+            """Create a dictionary object for the autocomplete suggestion."""
+            return {
+                "type": item_type,
```
escattone, line 28: Nits.
- Instead of passing in `item_type`, you could use something like `item.__class__.__name__` or `"User" if isinstance(item, User) else "Group"`.
- Instead of passing in `is_user`, you could use `isinstance(item, User)`.

kitsune/messages/forms.py
```diff
@@ -1,24 +1,30 @@
...
     )
     message = forms.CharField(label=_lazy("Message:"), max_length=10000, widget=forms.Textarea)
     in_reply_to = forms.IntegerField(widget=forms.HiddenInput, required=False)
 
+    def __init__(self, *args, **kwargs):
+        # Grab the user
+        self.user = kwargs.pop("user")
+        super(MessageForm, self).__init__(*args, **kwargs)
+
+        # If the user is_staff, the placholder text needs updated
```
escattone, line 24: Typo. `to be updated` instead of `updated`

kitsune/settings.py
```diff
@@ -728,6 +730,10 @@ def JINJA_CONFIG():
     "SESSION_SERIALIZER", default="django.contrib.sessions.serializers.PickleSerializer"
 )
 
+
+# Our group with permissions to see hidden docs and message groups
+SUMO_STAFF = "Staff"
```
escattone, line 735: Could we use `STAFF_GROUP` instead of `SUMO_STAFF`? I'll change https://github.com/mozilla/kitsune/pull/5947 to use whatever you use here.

escattone (changes requested):

kitsune/groups/jinja2/groups/profile.html
```diff
@@ -18,10 +18,12 @@
         <a class="edit" href="{{ url('admin:groups_groupprofile_change', profile.id) }}">{{ _('Edit in admin') }}</a>
       {% endif %}
       <h1>{{ profile.group.name }}</h1>
-
-      {% if user_can_edit %}
-        <a class="edit" href="{{ url('groups.edit', profile.slug) }}">{{ _('Edit group profile') }}</a>
-      {% endif %}
+        {% if user.profile.is_staff %}
```
escattone, line 21: This will not work when the user is not authenticated, since `user.profile` will be `None`. I just realized that putting `is_staff` -- which should be renamed as mentioned earlier -- on the profile means that we'll also have to constantly check if the user is authenticated or not, so now we'll have to use something like `user.is_authenticated and user.profile.is_staff` in nearly every place where we currently use `user.is_staff`.

It makes me wonder if we should abandon the idea of putting the `is_staff` on the profile, and instead go with the function idea I use within https://github.com/mozilla/kitsune/pull/5947. Within the file `kitsune/sumo/utils.py`, I created the function `is_staff_group_member`, which takes care of ensuring the user is defined and authenticated before making the DB call, and also caches the result on the user instance.

```python
def is_staff_group_member(user: User | None) -> bool:
    """
    Returns true only if the given user is a member of settings.STAFF_GROUP. If the user
    is authenticated, it also caches the result on the user instance so subsequent calls
    don't need to repeat the database query.
    """
    if not (user and user.is_authenticated):
        return False
    if not hasattr(user, "is_staff_group_member"):
        user.is_staff_group_member = user.groups.filter(name=settings.STAFF_GROUP).exists()
    return user.is_staff_group_member
```

So, with the function approach, we can replace `user.is_staff` with `is_staff_group_member(user)`, instead of `user.is_authenticated and user.profile.is_staff`.

escattone (changes requested):
@smithellis With this latest realization, I've changed my mind and no longer like the idea of putting `is_staff` (or whatever we rename it to) on the profile. I'd vote for using the function approach instead.

kitsune/groups/jinja2/groups/profile.html
smithellis: @escattone Could we meet in the middle and have the `Group` check live on as a `Profile` property and still use your helper function - but have your function check the `Profile` for `is_staff`? That way, we have:
1 - Explicit separation of the idea of authentication and group membership
2 - Consistent usage of the `Profile` as a place we store `User` properties (until we go to a custom `User` model)

Your function might combine concerns that should be separate - for example, a use case where I want to check if a non-authenticated `User` has group membership, won't this just return` False`? So we'd be using the longer `user.groups.filter(name=settings.STAFF_GROUP).exists()` more for those cases....

Thoughts?

escattone: NOTE: We had a video discussion on what name to use other than `is_staff` and settled on `in_staff_group` for now.

Since any user instance that isn't an `AnonymousUser` instance is guaranteed to have `is_authenticated == True`, there really can't be any separation between authentication and the possibility of group membership. A user has to be authenticated -- in other words, not an `AnonymousUser` -- in order to be a member of one or more groups or have a profile. So we'd always have to check if the user is authenticated before trying `user.profile.in_staff_group`.

As you said, we could still meet in the middle, and assuming we make `in_staff_group` a `cached_property`, we could add the following to `kitsune/sumo/utils.py`:

```python
def in_staff_group(user: User | None) -> bool:
    """
    Returns true only if the given user is a member of settings.STAFF_GROUP.
    """
    return bool(user and user.is_authenticated and user.profile.is_staff_group_member)
```

and then also add a Jinja helper function within `kitsune/sumo/templatetags/jinja_helpers.py`:

```python
from kitsune.sumo.utils import in_staff_group
...
library.global_function(in_staff_group)
```

kitsune/users/models.py
```diff
@@ -227,6 +227,10 @@ def answer_helpfulness(self):
     def is_subscriber(self):
         return self.products.exists()
 
+    @property
```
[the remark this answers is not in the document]
smithellis: Agreed.

kitsune/messages/api.py
```diff
@@ -0,0 +1,57 @@
...
+@require_GET
+@json_view
+def get_autocomplete_suggestions(request):
+    """An API to provide auto-complete data for user names or groups."""
+    pre = request.GET.get("term", "") or request.GET.get("query", "")
+    if not pre or not request.user.is_authenticated:
+        return []
+
+    def get_suggestions(pre, is_staff=False):
+        """Generate autocomplete suggestions for users and, if applicable, groups."""
```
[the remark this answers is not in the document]
smithellis: Agreed.

kitsune/messages/api.py
smithellis: Agreed, changed.

kitsune/settings.py
smithellis: You got it.

kitsune/groups/jinja2/groups/profile.html
smithellis: I've implemented `in_staff_group`, added `cached_property` and added both the `utils.py` and `jinja_helpers.py` changes.

kitsune/messages/api.py
```diff
@@ -0,0 +1,51 @@
...
+        }
+
+    suggestions = []
+    user_criteria = Q(username__istartswith=pre) | Q(profile__name__istartswith=pre)
+    users = User.objects.filter(
+        user_criteria, is_active=True, profile__is_fxa_migrated=True
+    ).select_related("profile")[:10]
+
+    for user in users:
+        suggestions.append(create_suggestion(user))
```
escattone, line 44: What you have is perfect, of course, but just for the fun of sharing it, you could also do:
```python
    suggestions.extend(create_suggestion(user) for user in users)
```

kitsune/messages/api.py
```diff
@@ -0,0 +1,51 @@
...
+
+    if request.user.profile.in_staff_group:
+        groups = Group.objects.filter(name__istartswith=pre)[:10]
+        for group in groups:
+            suggestions.append(create_suggestion(group))
```
escattone, line 49: Again, just for the fun of sharing it, you could also do:
```python
        suggestions.extend(create_suggestion(group) for group in groups)
```

kitsune/sumo/form_fields.py
```diff
@@ -83,6 +83,41 @@ def to_python(self, value):
         return users
 
 
+class MultiUsernameOrGroupnameField(forms.Field):
+    """Form field that takes a comma-separated list of usernames or groupnames
+    and validates that users/groups exist for each one, and returns the list of
+    users/gorups."""
```
escattone, line 89: Typo: `groups` instead of `gorups`

escattone (approved):
r+wc

Just a typo change.

kitsune/messages/api.py
smithellis: Got it.

PR #5924, opened 2024-03-20
Update playwright workflow file (reduce steps, limit the scheduled execution occurence & introduce workflow_dispatch)

emilghittasv:
Updated the playwright.yml file to:
- Enable workflow_dispatch so that we can manually trigger the playwright tests when needed.
- Reduced the full functional recurrent test run from daily (Monday-Friday) execution to Monday & Friday only.
- Reduced the number of steps for running all test suites.

### DIFF
--- a/.github/workflows/playwright.yml
+++ b/.github/workflows/playwright.yml
@@ -1,8 +1,18 @@
 name: Playwright Tests
 on:
     schedule:
-        # Firefox tests running on Monday, Wednesday and Friday. Chrome runs Tuesday and Thursday.
-        - cron: '0 5 * * 1,2,3,4,5'
+        # Playwright tests are running automatically in Firefox on each Monday & in Chrome on each Friday.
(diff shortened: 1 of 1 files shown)


.github/workflows/playwright.yml
```diff
@@ -1,8 +1,18 @@
...
-        # Firefox tests running on Monday, Wednesday and Friday. Chrome runs Tuesday and Thursday.
-        - cron: '0 5 * * 1,2,3,4,5'
+        # Playwright tests are running automatically in Firefox on each Monday & in Chrome on each Friday.
+        - cron: '0 5 * * 1,5'
+    workflow_dispatch:
+        inputs:
+            Browsers:
+                description: 'Browsers'
+                required: true
+                default: 'firefox'
```
escattone, line 11: Nit. I don't think you need the single quotes around `'Browsers'` and `'firefox'`?

escattone (approved):
r+wc

PR #5934, opened 2024-04-02
Drop version from compose file.

akatsoulas:


escattone (approved):
Nice. I didn't realize it was optional until now (https://docs.docker.com/compose/compose-file/04-version-and-name/#version-top-level-element-optional).

PR #5943, opened 2024-04-04
Update docs workflow to build directly to deployment

janbrasna:
This moves the whole build and deploy into actions workflow, skipping committing to `gh-pages` branch (that could eventually be removed), and publishing straight to the deployment environment.

It also triggers only for `docs/**` and related changes, so saves unnecessary runs when there's nothing to publish.

_(Usually needs manual switch in repo Settings from branch deploy to action deploy to work, otherwise the env is protected from the classic runs…)_

### DIFF
--- a/.github/workflows/publish-docs.yml
+++ b/.github/workflows/publish-docs.yml
@@ -1,28 +1,52 @@
 name: Publish Docs in GitHub Pages
+
 on:
+  workflow_dispatch: # allow manual runs
   push:
     branches:
       - main
+    paths: # avoid extra builds
+      - docs/**
+      - mkdocs.yml
+      - .github/workflows/publish-docs.yml
+
 permissions:
-  contents: write
+  contents: read
+  pages: write
+  id-token: write
+
+concurrency:
+  group: github-pages
+  cancel-in-progress: false # skip any intermediate builds but let finish
+
 jobs:
-  deploy:
+  build:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v4
-      - name: Configure Git Credentials
-        run: |
-          git config user.name github-actions[bot]
-          git config user.email 41898282+github-actions[bot]@users.noreply.github.com
-      - uses: actions/setup-python@v4
+      - uses: actions/setup-python@v5
         with:
           python-version: 3.x
+      - id: pages
+        uses: actions/configure-pages@v4
       - run: echo "cache_id=$(date --utc '+%V')" >> $GITHUB_ENV 
-      - uses: actions/cache@v3
+      - uses: actions/cache@v4
         with:
           key: mkdocs-material-${{ env.cache_id }}
           path: .cache
           restore-keys: |
             mkdocs-material-
       - run: pip install mkdocs-material 
-      - run: mkdocs gh-deploy --force
+      - run: mkdocs build -d dist
+      - uses: actions/upload-pages-artifact@v3
+        with:
+          path: dist
+  deploy:
+    environment:
+      name: github-pages
+      url: ${{ steps.deployment.outputs.page_url }}
+    runs-on: ubuntu-latest
+    needs: build
+    steps:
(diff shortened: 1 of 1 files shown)


janbrasna:
See example run for [_1c84e14_ here](https://github.com/janbrasna/kitsune/actions/runs/8562216217/job/23465007346)

.github/workflows/publish-docs.yml
```diff
@@ -1,28 +1,52 @@
...
         with:
           python-version: 3.x
+      - id: pages
+        uses: actions/configure-pages@v4
       - run: echo "cache_id=$(date --utc '+%V')" >> $GITHUB_ENV 
-      - uses: actions/cache@v3
+      - uses: actions/cache@v4
         with:
           key: mkdocs-material-${{ env.cache_id }}
           path: .cache
```
janbrasna, line 36: FYI the only thing I'm not exactly satisfied with is the cache. I haven't touched that bit, but TL;DR is doesn't work at all.

I see the caching key here is week number, and the attempt to store `.cache` is common for various package manager resolves. Unfortunately here the cache action stores `/home/runner/work/kitsune/kitsune/.cache` (empty), whereas the built python wheels that could be worth caching are stored in `/home/runner/.cache/pip/wheels` out of reach. These _could_ shave a few seconds off the clock. But with the frequency this is gonna be run it doesn't really make that much difference TBH…

akatsoulas:
Thanks @janbrasna for this PR. I will take a look early next week.

.github/workflows/publish-docs.yml
```diff
@@ -1,28 +1,52 @@
...
+    paths: # avoid extra builds
+      - docs/**
+      - mkdocs.yml
+      - .github/workflows/publish-docs.yml
+
 permissions:
-  contents: write
+  contents: read
+  pages: write
+  id-token: write
```
akatsoulas, line 16: @janbrasna can you provide some context as to why is this needed?

akatsoulas:
@janbrasna Overall this looks good and it's much better compared to the existing workflow! I have just a question about the `id-token` before merging this.

.github/workflows/publish-docs.yml
janbrasna: Ah yea, that's the case: [Deploying GitHub Pages artifacts](https://docs.github.com/en/pages/getting-started-with-github-pages/using-custom-workflows-with-github-pages#deploying-github-pages-artifacts)

> The deploy-pages action handles the necessary setup for deploying artifacts. To ensure proper functionality, the following requirements should be met:
> 
> The job must have a minimum of pages: write and id-token: write permissions.

akatsoulas (approved):

akatsoulas:
Thanks again  @janbrasna :rocket:

.github/workflows/publish-docs.yml
```diff
@@ -1,28 +1,52 @@
...
+
 on:
+  workflow_dispatch: # allow manual runs
   push:
     branches:
       - main
+    paths: # avoid extra builds
+      - docs/**
+      - mkdocs.yml
+      - .github/workflows/publish-docs.yml
```
janbrasna, line 11: FYI I'm not completely sure it's necessary to explicitly mention the action itself in the paths, but I'm including it for extra visibility into what triggers it, even if it may be superfluous.

janbrasna (commented):
@akatsoulas Pleasure! I see it builds and deploys alright so the environment protections are set up correctly, the logs are happy, and I've added a trivial fix #5991 to double check the changes are being deployed as needed.

After some time you can delete the `gh-pages` branch completely as it won't be needed anymore, and just maybe keep an eye on the paths trigger, to verify the docs only build on docs changes, and not with any other commit — and that you're happy with the setup. (This is what works for me, it should cover the structure here too, maybe you'd want to add some lockfiles/requirements too, later, if you see it fit the workflow need.)

PR #5944, opened 2024-04-04
increase debounce threshold for instant searches

escattone:
mozilla/sumo#1717

I discovered that we're already "debouncing" changes to the instant search input box, but the pause threshold is far too low, so we're seeing too many fractured search terms in GA4. This PR increases the pause threshold from 200ms to 600ms, which should remove a lot of search-term fracturing.

### Alternative
Another option would be to make our "instant" searches behave like our searches made on the [search page](https://support.mozilla.org/en-US/search/). Those searches are only triggered once the user submits the search term via the "enter" key, so we're capturing the user's intention rather than all of the intermediate steps.


akatsoulas (approved):
Let's try first this approach. The downside with the alternative you proposed is that many users may won't hit the enter button b/c of the search-as-you-type functionality

PR #5947, opened 2024-04-08
check staff group rather than is_staff for elevated perms

escattone:
mozilla/sumo#1683

### DIFF
--- a/kitsune/sumo/templatetags/jinja_helpers.py
+++ b/kitsune/sumo/templatetags/jinja_helpers.py
@@ -28,9 +28,7 @@ import wikimarkup.parser
 from kitsune.products.models import Product
 from kitsune.sumo import parser
 from kitsune.sumo.urlresolvers import reverse
-from kitsune.sumo.utils import is_trusted_user as is_trusted_user_func
-from kitsune.sumo.utils import webpack_static as webpack_static_func
-from kitsune.sumo.utils import in_staff_group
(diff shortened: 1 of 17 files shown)


smithellis (approved):
I think you caught them all, @escattone !

kitsune/sumo/templatetags/jinja_helpers.py
```diff
@@ -467,11 +465,6 @@ def static(path):
         return ""
 
 
-@library.global_function
```
akatsoulas, line 470: let's revert this, it's not relevant to this PR

akatsoulas: On second thought let's keep it to clean things up a bit

kitsune/sumo/templatetags/jinja_helpers.py
```diff
@@ -565,9 +558,6 @@ def show_header_fx_download(context):
         return True
 
 
-@library.global_function
-def is_trusted_user(user):
-    return is_trusted_user_func(user)
-
-
+library.global_function(webpack_static)
```
akatsoulas, line 561: I would move all the Jinja function registrations to the top of the file

akatsoulas (approved):
Looks good to me! r+wc

akatsoulas:
For completeness we could also remove the underscore from https://github.com/mozilla/kitsune/blob/main/kitsune/messages/forms.py#L24

PR #5957, opened 2024-04-16
Updates for 5922 - Staff can send Group messages

smithellis:
* Fixes privelege escalation - 1721
* Fixes bolded message text - 1726
* Removes unneeded Avatars - 1724
* Don't show a To if there is no To - 1723
* Attempts to resolve 502 issue - 1720

### DIFF
--- a/kitsune/messages/utils.py
+++ b/kitsune/messages/utils.py
@@ -1,3 +1,5 @@
+from django.contrib.auth.models import Group
+
 from kitsune.messages.models import InboxMessage, OutboxMessage
 from kitsune.users.models import User
 from kitsune.messages.signals import message_sent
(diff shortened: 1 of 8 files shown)


AcharayuAiyara (approved):

kitsune/messages/utils.py
```diff
@@ -18,22 +19,37 @@ def send_message(to, to_group=None, text=None, sender=None):
 
     # Assuming OutboxMessage doesn't necessarily need to be created for every recipient
     if to:
+        # This is users from the To field
+        users = {user for user in to if isinstance(user, User)}
+        # This is groups from the To field
+        groups = [group for group in to if isinstance(group, Group)]
```
escattone, line 25: Is there a reason why `groups` should not be a set like `users`?

kitsune/messages/utils.py
```diff
@@ -18,22 +19,37 @@ def send_message(to, to_group=None, text=None, sender=None):
...
+        # Lets get all the data for groups that we need
+        prefetched_groups = Group.objects.filter(
+            pk__in=[group.pk for group in groups]
+        ).prefetch_related("user_set")
+        # This is who is actually getting a message
+        # - A combo of users from To field and users from groups
+        receivers = users | {user for group in prefetched_groups for user in group.user_set.all()}
+
```
escattone, line 33: I think directly getting the users might be a better way to go? In other words, something like:

```python
        receivers = users | set(User.objects.filter(groups__in=groups).distinct())
```

escattone (changes requested):

kitsune/messages/utils.py
```diff
@@ -18,22 +19,37 @@ def send_message(to, to_group=None, text=None, sender=None):
...
+            # Add the groups from the To field to the message
+            outbox_message.to_group.set(groups)
+
+    for recipient in receivers:
+        inbox_message = InboxMessage.objects.create(sender=sender, to=recipient, message=text)
+        # If we had a user, and we made them an inbox message,
+        # we should also add the groups to their message as well
+        if groups:
+            inbox_message.to_group.set(groups)
+        if Setting.get_for_user(recipient, "email_private_messages"):
```
escattone, line 49: This is a query for every user in the `receivers`, so could be 10's to 100's of queries. Could we make a single bulk query for this before we enter the loop? Maybe something like:

```python
set_of_user_pks_to_email_private_message = set(Setting.objects.filter(user__in=receivers, name="email_private_messages").values_list("user", flat=True))
```

and then, within the loop, check if the recipient's PK is within that set?

kitsune/messages/utils.py
smithellis: No, no reason at all.

kitsune/messages/utils.py
smithellis: I'll try this - I thought `.distinct()` would be an expensive choice, but maybe this combination makes that a wash.

kitsune/messages/utils.py
```diff
@@ -1,39 +1,56 @@
...
+    outbox_message.to.set(users)
+
+    if groups:
+        # Add the groups from the To field to the message
+        outbox_message.to_group.set(groups)
+
+    set_of_user_pks_to_email_private_message = set(
+        Setting.objects.filter(
+            user__in=receivers, name="email_private_messages", value=True
+        ).values_list("user__pk", flat=True)
```
escattone, line 41: Just for the sake of knowledge sharing, `.values_list("user", flat=True)` is equivalent to `.values_list("user__pk", flat=True)`.

escattone (approved):

kitsune/messages/utils.py
escattone: Good point. I was debating that as well, whether we should drop the `.distinct()` and just let the `set()` take care of it. I was thinking that the `distinct` call in Postgres might be faster, but I really don't know.

kitsune/messages/utils.py
smithellis: Ah - yes, I left that when troubleshooting why this was failing. It was the `value=True` bit that was missing.

kitsune/messages/utils.py
escattone: Good catch! I was wondering if the `value` had to be checked as well.

kitsune/messages/utils.py
escattone: Good catch on the `value=True`!

PR #5959, opened 2024-04-16
Move send_message to celery task

smithellis:
To resolve ongoing 502 issues in Staging after multiple attempts to improve database calls, we will move send_message into a celery task.

Further, to ensure integrity of these database transactions, I've added `transaction.atomic` to the database activity.


escattone (approved):
r+

kitsune/messages/tasks.py
```diff
@@ -61,3 +64,55 @@ def _send_mail(locale):
         locale = settings.WIKI_DEFAULT_LANGUAGE
 
     _send_mail(locale)
+
+
+@shared_task
+def send_message(to, text=None, sender=None):
```
akatsoulas, line 70: For a celery task it would be more appropriate to pass the IDs instead of the objects and fetch a fresh copy from database when the task is running. If you pass an object, things might have changed by the time you run the task.

akatsoulas (changes requested):
I am not sure if this is the best approach. The task will still have to perform the queries, the only difference is that it won't break the user experience with a 502. It will still timeout though and the message won't go through.

Have you found which queries are causing the error?

akatsoulas:
Closing this for now, since current effort is focusing on optimizing the queries. We will re-open if needed

PR #5963, opened 2024-04-17
502 remediation

smithellis:
Optimize db queries for 502

In `send_message` in `utils.py`:
* Reduce number of queries
*  Prefetch group members
*  One query for users_to_email

In `MultiUsernameorGroupnameField` in `form_fields.py`
* Get users and groups in one query each
* A single pass through to collect `found_names`

Refactor `_add_recipients`
* Mainly readability


kitsune/messages/utils.py
```diff
@@ -18,38 +18,42 @@ def send_message(to, text=None, sender=None):
...
-    users = {user for user in to if isinstance(user, User)}
-    # This is groups from the To field
-    groups = {group for group in to if isinstance(group, Group)}
-    # This is all the users that are going to receive the message
-    receivers = users | set(User.objects.filter(groups__in=groups).distinct())
+    # Prepare querysets for users and groups directly and efficiently
+    users_qs = User.objects.filter(
+        username__in=[obj.username for obj in to if isinstance(obj, User)]
+    )
+    groups_qs = Group.objects.filter(name__in=[obj.name for obj in to if isinstance(obj, Group)])
```
akatsoulas, line 25: This could be further optimized by populating a list of `usernames` and `group_names` in one go:

```python
username = []
group_names = []

for obj in to:
    if isinstance(obj, User):
        usernames.add(obj)
    else:
         group_names.add(obj)

group_qs = Group.objects.filter(name__in=group_names)
user_qs = User.objects.filter(username__in=usernames)
```
It would be even better when you construct the `to` iterable to make it something that you can manipulate easier. Eg a dictionary where you could have:
```python 
to = {
    'users': [list_of_usernames],
    'groups': [list_of_group_names]
```
this way you have essentially a flat list of strings instead of objects that's already separated - like using `values_list` instead of the whole object.

akatsoulas: any of these approaches will affect the following code too.

kitsune/messages/jinja2/messages/outbox.html
```diff
@@ -75,7 +74,10 @@ <h1 class="sumo-page-heading">{{ title }}</h1>
                       {% if message.to_group %}
                         {% if message.to_groups_count > 0 -%}
                           {%- for group in message.to_group.all()[:3] -%}
-                            {{ group_link(group) }}
+                            {%- for profile in group.profile.all() -%}
```
akatsoulas, line 77: this feels too expensive for a big group, why are we adding this?

akatsoulas (changes requested):

kitsune/messages/jinja2/messages/outbox.html
smithellis: `group_link(group)` is a global jinja helper, but it makes a db hit for every group passed to it. This change takes advantage of having prefetched the `to_group` in the view. I didn't want to change the global helper function since it is also used in the `users` app.

kitsune/messages/utils.py
```diff
@@ -18,38 +18,42 @@ def send_message(to, text=None, sender=None):
...
-    for recipient in receivers:
-        inbox_message = InboxMessage.objects.create(sender=sender, to=recipient, message=text)
-        # If we had a user, and we made them an inbox message,
-        # we should also add the groups to their message as well
-        if groups:
-            inbox_message.to_group.set(groups)
-        if recipient.pk in set_of_user_pks_to_email_private_message:
+    # Create inbox messages and handle emails
+    for user in all_recipients_of_message:
+        inbox_message = InboxMessage.objects.create(sender=sender, to=user, message=text)
```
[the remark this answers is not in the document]
smithellis: Added.

kitsune/messages/views.py
```diff
@@ -183,15 +200,9 @@ def _add_recipients(msg):
     msg.to_groups_count = len(groups)
 
     # Assign the recipient based on the number of recipients
-    if msg.recipients == 1:
-        msg.recipient = recipients[0]
-    else:
-        msg.recipient = None
+    msg.recipient = recipients[0] if len(recipients) == 1 else None
```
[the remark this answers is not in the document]
smithellis: Just trying to minimize database interactions. If I gather both `msg.to.all().count()` and `msg.to_group.all().count()`, then that is two database interactions that only get me back count. Later, depending on the counts, I'll need the first member of the `to` group - so I'd have to ask for that item from the database. Then I would later have to fetch all groups if there are more than one, again hitting the database. 
If I just dump everything into lists once, I can `len()`, compare and access till my heart is content with only two database hits.

kitsune/messages/views.py
```diff
@@ -183,15 +200,9 @@ def _add_recipients(msg):
     msg.to_groups_count = len(groups)
```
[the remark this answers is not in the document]
smithellis: This is used in several places - I don't know that I see efficiency in altering it. Can you explain a little more?

kitsune/messages/utils.py
```diff
@@ -18,38 +18,48 @@ def send_message(to, text=None, sender=None):
...
-        if groups:
-            inbox_message.to_group.set(groups)
-        if recipient.pk in set_of_user_pks_to_email_private_message:
-            email_private_message(inbox_message_id=inbox_message.id)
+    # Create inbox messages and handle emails
+
+    inbox_messages = [
+        InboxMessage(sender=sender, to=user, message=text) for user in all_recipients_of_message
+    ]
+    messages = InboxMessage.objects.bulk_create(inbox_messages)
```
escattone, line 55: I like the `bulk_create()` change! I noticed however that you're not setting the `to_group` field on each of the `InboxMessage`'s like you used to. Are we getting rid of `to_group` on the `InboxMessage`? If not, I wonder if we could do a `bulk_update()` of the `to_group` field? That would be nice, but not sure if the `bulk_update` works with many-to-many fields.

kitsune/messages/utils.py
```diff
@@ -18,38 +18,48 @@ def send_message(to, text=None, sender=None):
...
+
+    # Send emails to users who want them
+    for message in messages:
+        if message.to.id in users_to_email:
+            email_private_message(inbox_message_id=message.id)
```
escattone, line 60: For large groups like `Contributors` which have over 2k users that have asked for private emails in addition to the message, I still think that doing over 2k `email_private_message()` calls could also cause a 502, so perhaps we should send the email asynchronously via our Celery task queue? So make the call with `email_private_message.delay(...)` instead of `email_private_message(...)`.

escattone (changes requested):

kitsune/messages/utils.py
```diff
@@ -18,38 +17,58 @@ def send_message(to, text=None, sender=None):
...
+    group_names = []
 
-    outbox_message = OutboxMessage.objects.create(sender=sender, message=text)
-    # Add the users from the To field to the outbox message, not all the users
-    # that are going to receive the message - this way we don't overwhelm the
-    # message UI
-    outbox_message.to.set(users)
+    for obj in to:
+        if isinstance(obj, User):
+            usernames.append(obj)
```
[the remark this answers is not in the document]
escattone: Since we want the actual `User` objects later for setting the `to` field of the `OutboxMessage`, can't we just collect those here -- in addition to the `usernames` above -- by creating another list called `users` and doing a `users.append(obj)` here?

kitsune/messages/utils.py
```diff
@@ -18,38 +17,58 @@ def send_message(to, text=None, sender=None):
...
+    group_users_qs = User.objects.filter(groups__in=groups_qs).values_list("id", flat=True)
 
-    if groups:
-        # Add the groups from the To field to the message
-        outbox_message.to_group.set(groups)
+    # Resolve all unique user ids, including those in groups
+    # but keep up with users from groups
+    to_users = set(users_qs)
+    group_users = set(group_users_qs)
+    all_recipients_of_message = to_users.union(group_users)
```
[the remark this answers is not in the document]
smithellis: We have to keep up with the users that came from the `to` field and the users that came from the `groups` from the to field. So capturing the pks in separate sets will matter later. We don't want to set the `to` in an outbox message to every user; just to the users that were entered individually.

kitsune/messages/utils.py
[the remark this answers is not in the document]
smithellis: I think this is bad naming on behalf of a cut and paste - these are objects being collected, not names. Even my comment is broken here - I'll fix this up. But we get the objects separated here.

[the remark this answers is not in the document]
smithellis: For the users specified individually in the `to` line, we want the objects. We also want the objects for the groups. The users that we retrieve from `groups` will just be pks per our chat earlier.

kitsune/messages/utils.py
```diff
@@ -18,38 +17,58 @@ def send_message(to, text=None, sender=None):
...
-    outbox_message = OutboxMessage.objects.create(sender=sender, message=text)
-    # Add the users from the To field to the outbox message, not all the users
-    # that are going to receive the message - this way we don't overwhelm the
-    # message UI
-    outbox_message.to.set(users)
+    for obj in to:
+        if isinstance(obj, User):
+            usernames.append(obj)
+        else:
+            group_names.append(obj)
```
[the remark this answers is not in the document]
smithellis: Yes - see note above. Bad naming...I'll fix.

kitsune/messages/utils.py
```diff
@@ -18,38 +17,58 @@ def send_message(to, text=None, sender=None):
...
+    outbox_message.to_group.set(groups_qs)
+
+    # Fetch settings for email notifications in one go
+    # For now, we only send emails to users who were individually
+    # listed in 'to' field, not to users who are from groups
+    users_to_email = set(
         Setting.objects.filter(
-            user__in=receivers, name="email_private_messages", value=True
-        ).values_list("user__pk", flat=True)
+            user__in=to_users, name="email_private_messages", value=True
```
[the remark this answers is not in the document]
smithellis: We are disabling e-mail (not messaging, just email) for users in Groups for now, pending some follow up.

kitsune/messages/utils.py
```diff
@@ -18,38 +17,58 @@ def send_message(to, text=None, sender=None):
...
+        for user_id in all_recipients_of_message
+    ]
+
+    InboxMessage.objects.bulk_create(inbox_messages)
+
+    # Assuming messages are fetched again with IDs if necessary
+    for user_id in users_to_email:
+        inbox_messages = InboxMessage.objects.filter(sender=sender, to__id=user_id, message=text)
+        for message in inbox_messages:
+            email_private_message.delay(inbox_message_id=message.id)
```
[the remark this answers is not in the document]
smithellis: This was sketchy in the docs vs. info from other sources. Good to know!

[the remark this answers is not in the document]
smithellis: Implemented!

kitsune/messages/utils.py
smithellis: Done

kitsune/messages/jinja2/messages/outbox.html
```diff
@@ -52,40 +51,31 @@ <h1 class="sumo-page-heading">{{ title }}</h1>
...
-                    {%- endif %}
+
+                  {% for user in message.to.all()[:1] -%}
+                    <a rel="nofollow" href="{{ profile_url(user) }}">
+                      {%- if name -%}
+                        {{ name }}
+                      {%- else -%}
+                        {{ user.profile.display_name }}
+                      {%- endif -%}
+                    {%- if message.recipients_count>1 -%}, ...{% endif %}
```
escattone, line 62: Nit. Add spaces around `>`.

kitsune/messages/jinja2/messages/outbox.html
```diff
@@ -52,40 +51,31 @@ <h1 class="sumo-page-heading">{{ title }}</h1>
...
-                          ...
-                          {% endif %}
-                        {% endif %}
-                        {% if message.to_groups_count < 1 -%}
+                        {%- for group in message.to_groups[:1] -%}
+                          {%- for profile in group.profile.all() -%}
+                            {% set group_slug = profile.slug %}
+                            <a href="{{ url('groups.profile', group_slug) }}">{{ group }}</a>
+                          {%- endfor -%}
+                          {% if message.to_groups_count>1 %}, ...{% endif %}
```
escattone, line 75: Nit. Add spaces around `>`.

kitsune/messages/jinja2/messages/outbox.html
```diff
@@ -52,40 +51,31 @@ <h1 class="sumo-page-heading">{{ title }}</h1>
...
-                          {% endfor %}
-                          {%- if message.to_groups_count > 3 -%}
-                          ...
-                          {% endif %}
-                        {% endif %}
-                        {% if message.to_groups_count < 1 -%}
+                        {%- for group in message.to_groups[:1] -%}
+                          {%- for profile in group.profile.all() -%}
+                            {% set group_slug = profile.slug %}
+                            <a href="{{ url('groups.profile', group_slug) }}">{{ group }}</a>
```
escattone, line 73: Nit.
```suggestion
                            <a href="{{ url('groups.profile', profile.slug) }}">{{ group }}</a>
```

kitsune/messages/utils.py
```diff
@@ -18,38 +16,52 @@ def send_message(to, text=None, sender=None):
...
-
-    set_of_user_pks_to_email_private_message = set(
+    # Fetch settings for email notifications in one go
+    # For now, we only send emails to users who were individually
+    # listed in 'to' field, not to users who are from groups
+    users_to_email = set(
         Setting.objects.filter(
-            user__in=receivers, name="email_private_messages", value=True
-        ).values_list("user__pk", flat=True)
+            user__in=users, name="email_private_messages", value=True
```
escattone, line 45: Just wanted to record that I'd prefer this to use `all_recipients_of_message` instead of just the `users`, but that's for another discussion.

escattone (approved):
r+wc

kitsune/messages/utils.py
smithellis: Got  it.

kitsune/messages/jinja2/messages/outbox.html
smithellis: Thanks!

kitsune/messages/jinja2/messages/outbox.html
smithellis: Got it.

kitsune/messages/jinja2/messages/outbox.html
smithellis: Got it.

PR #5967, opened 2024-04-21
vscode launch.json upgrade

smithellis:
* Replace "type": "python" with "type": "debugpy"
* Add common test patterns for debugging 
- All tests
- Wiki tests
- Messages tests

### DIFF
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -6,7 +6,7 @@
     "configurations": [
         {
             "name": "Kitsune",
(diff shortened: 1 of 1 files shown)


escattone (approved):
I'm not very knowledgeable on the vscode dev containers, but this seems fine to me. I tried it locally, and the one thing I had to do was install `debugpy` in the dev container the first time I ran the Wiki tests via the configuration.

PR #5968, opened 2024-04-21
Add users and groups as part of generatedata.py

smithellis:
Creates group with int number of users
* staff: 15
* small_group: 10
* medium_group: 15
* large_group: 1000
* xlarge_group: 5000

All usernames will start with group name for easy testing Usernames are a concatenation of [group name]+User+count
* Example: small_groupUser1, medium_groupUser10, etc.


akatsoulas:
@smithellis is this still relevant/valid?

smithellis:
I think this was a casual Friday bit of work. I like it and find it useful, but if you don't, feel free to close. Stemmed from having the messaging work and needing to test with various group sizes.

akatsoulas:
In that case, personally I would use the test factories for User and Groups. Please also move this under the `scripts`  folder under the root level of the repo. Let me know when it's ready for a final pass

smithellis:
I moved this to scripts - not sure what you were saying about the factories? They are in use. Did you mean to modify factories or add to factories or...?

scripts/generate_users_and_groups.py
```diff
@@ -0,0 +1,24 @@
...
+    """Generate user and group sample data."""
+    for group_name, user_count in GROUPS.items():
+        g = GroupFactory(name=group_name)
+        g.save()
+        print(f"Created group {group_name}")
+
+        # Make the users for this group
+        for _ in range(user_count):
+            user_name = f"{group_name}User{_}"
+            u = UserFactory(username=user_name, groups=[g])
```
akatsoulas, line 22: This should be enough to create the user without the need to call the `save()` method

scripts/generate_users_and_groups.py
```diff
@@ -0,0 +1,24 @@
...
+    "medium_group": 15,
+    "large_group": 1000,
+    "xlarge_group": 5000,
+}
+
+
+def generate_sampledata(options):
+    """Generate user and group sample data."""
+    for group_name, user_count in GROUPS.items():
+        g = GroupFactory(name=group_name)
```
akatsoulas, line 15: This should be enough to create the group without the need to call the `save()` method

akatsoulas: GroupFactory.create(args)

scripts/generate_users_and_groups.py
```diff
@@ -0,0 +1,24 @@
...
+GROUPS = {
+    "staff": 15,
+    "small_group": 10,
+    "medium_group": 15,
+    "large_group": 1000,
+    "xlarge_group": 5000,
+}
+
+
+def generate_sampledata(options):
```
akatsoulas, line 12: is the `options` used anywhere?

scripts/generate_users_and_groups.py
```diff
@@ -0,0 +1,24 @@
...
+    """Generate user and group sample data."""
+    for group_name, user_count in GROUPS.items():
+        g = GroupFactory(name=group_name)
+        g.save()
+        print(f"Created group {group_name}")
+
+        # Make the users for this group
+        for _ in range(user_count):
```
akatsoulas, line 20: You can use UserFactory.create_batch(size, ...

akatsoulas (changes requested):

akatsoulas:
Closing this as stale. Please reopen if it's still relevant.

PR #5971, opened 2024-04-23
1742 : Refactor to eliminate User and Group name collisions

smithellis:
* Make sure we catch users and groups with same name
* Move validation to `form.clean()'
* User custom field in `form_fields.py` to create data structure

### DIFF
--- a/kitsune/messages/forms.py
+++ b/kitsune/messages/forms.py
@@ -25,6 +25,17 @@ class MessageForm(forms.Form):
         if self.user and self.user.profile.in_staff_group:
             self.fields["to"].widget.attrs["placeholder"] = "Search for Users or Groups"
 
+    def clean_to(self):
+        """Ensure that all usernames and group names are valid."""
+        to = self.cleaned_data["to"]
+
+        if not to["users"] and not to["groups"]:
+            raise forms.ValidationError("Please select at least one user or group.")
+        if to["groups"] and not self.user.profile.in_staff_group:
+            raise forms.ValidationError("You are not allowed to send messages to groups.")
+
+        return to
+
 
 class ReplyForm(forms.Form):
     """Form to reply to a private message."""
--- a/kitsune/messages/forms.py
+++ b/kitsune/messages/forms.py
@@ -27,12 +27,23 @@ class MessageForm(forms.Form):
 
     def clean_to(self):
         """Ensure that all usernames and group names are valid."""
-        to = self.cleaned_data["to"]
+        to = self.cleaned_data.get("to", {})
 
-        if not to["users"] and not to["groups"]:
+        # Check if there are valid users or groups selected.
+        if not to.get("users") and not to.get("groups"):
             raise forms.ValidationError("Please select at least one user or group.")
-        if to["groups"] and not self.user.profile.in_staff_group:
-            raise forms.ValidationError("You are not allowed to send messages to groups.")
+
+        # Check for group messages permissions.
+        if to.get("groups"):
+            # If the user is not a member of the staff group,
+            # they are not allowed to send messages to groups.
+            if not self.user.profile.in_staff_group:
+                raise forms.ValidationError("You are not allowed to send messages to groups.")
+            # If group.profile is None, it means we don't allow messages to it.
+            if not all(group.profile for group in to.get("groups")):
+                raise forms.ValidationError(
+                    "You are not allowed to send messages to groups without profiles."
+                )
 
         return to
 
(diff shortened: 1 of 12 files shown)


kitsune/messages/utils.py
```diff
@@ -1,34 +1,31 @@
...
     # Resolve all unique user ids, including those in groups
-    # We need to keep group_users separate for email notifications
-    all_recipients_of_message = set(user.id for user in users) | set(
+    # We need to keep group users separate for email notifications
+
+    to_users = to.get("users", [])
+    to_groups = to.get("groups", [])
+
+    users = User.objects.filter(username__in=to_users) if to_users else User.objects.none()
+    groups = Group.objects.filter(name__in=to_groups) if to_groups else Group.objects.none()
```
escattone, line 26: Nit. You don't need the `if-else` since `User.objects.filter(username__in=[])` will effectively be the same thing as `User.objects.none()` -- and the same goes for the `groups`.

escattone (changes requested):
I also found a change that should be made to the line within `kitsune/messages/views._add_recipients` where you assign a value to `msg.to_groups`. The current code will return a queryset when there's only one group, and a list when there are none or more than one. It can be simplified to:

```python
msg.to_groups = list(msg.to_group.prefetch_related("profile"))
```

kitsune/sumo/form_fields.py
```diff
@@ -95,29 +95,26 @@ def to_python(self, value):
...
+        for key, value in key_value_pairs:
+            grouped.setdefault(key, []).append(value)
 
-        if missing_names:
-            raise ValidationError(
-                _("The following are not valid usernames or group names: {names}").format(
-                    names=", ".join(missing_names)
-                )
-            )
+        to_objects = {"users": grouped.get("User", []), "groups": grouped.get("Group", [])}
```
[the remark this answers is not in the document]
smithellis: This required changes to the form `clean_to()` method, but I implemented both.

kitsune/sumo/form_fields.py
```diff
@@ -95,29 +95,24 @@ def to_python(self, value):
...
-
-        # Check if all names were found
-        found_names = {
-            obj.username if isinstance(obj, User) else obj.name for obj in users_and_groups
-        }
-        missing_names = names - found_names
+        # Split the parts into key-value pairs
+        # * ["User: user1", "Group: group1", "User: user2"] ->
+        #   [("User", "user1"), ("Group", "group1"), ("User", "user2")]
+        key_value_pairs = (tuple(map(str.strip, part.split(":"))) for part in parts)
```
escattone, line 106: On further review, I realized that this whole chunk could be simplified, I think, to something like:

```python
        key_value_pairs = (tuple(map(str.strip, part.split(":"))) for part in value.split(",") if part.strip())
```

and adding a new comment to fit.

kitsune/sumo/form_fields.py
```diff
@@ -95,29 +95,24 @@ def to_python(self, value):
...
 
-        if missing_names:
-            raise ValidationError(
-                _("The following are not valid usernames or group names: {names}").format(
-                    names=", ".join(missing_names)
-                )
-            )
+        # Crete data structure to hold grouped items
+        # * [("User", "user1"), ("Group", "group1"), ("User", "user2")] ->
+        #   {"User": ["user1", "user2"], "Group": ["group1"]}
```
escattone, line 110: The comment here needs to be adjusted to fit the new code.

escattone (changes requested):
Just some minor changes, and it should be good to go.

kitsune/sumo/form_fields.py
smithellis: This is trading off readability I think, but I used it and just commented it thoroughly.

escattone (changes requested):

kitsune/sumo/form_fields.py
escattone: Good point. It might be too much in one line. Feel free to go back to the two-step approach if you think it's more readable.

kitsune/messages/forms.py
```diff
@@ -25,6 +25,28 @@ def __init__(self, *args, **kwargs):
...
+            raise forms.ValidationError("Please select at least one user or group.")
+
+        # Check for group messages permissions.
+        if to.get("groups"):
+            # If the user is not a member of the staff group,
+            # they are not allowed to send messages to groups.
+            if not self.user.profile.in_staff_group:
+                raise forms.ValidationError("You are not allowed to send messages to groups.")
+            # If group.profile is None, it means we don't allow messages to it.
+            if not all(group.profile for group in to.get("groups")):
```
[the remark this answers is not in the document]
smithellis: I think we still need the check here. I don't think there is a better place to validate this. I'll adopt your code.

escattone (approved):

PR #5976, opened 2024-04-24
Resolve CSS Regression

smithellis:
* New _messages.scss for inbox, outbox
* New inbox.html, outbox.html

### DIFF
--- a/kitsune/messages/jinja2/messages/inbox.html
+++ b/kitsune/messages/jinja2/messages/inbox.html
@@ -16,7 +16,6 @@
     <form action="{{ url('messages.bulk_action') }}" method="POST">
       <div class="sumo-button-wrap">
         <input class="sumo-button" type="submit" name="mark_read" value="{{ _('Mark Selected as Read') }}">
-        <input class="sumo-button" type="submit" name="delete" value="{{ _('Delete Selected') }}">
       </div>
       <div class="inbox-table">
             <div class="email-row header">
@@ -24,7 +23,7 @@
                     <input type="checkbox" aria-label="Select all">
                 </div>
                 <div class="email-cell avatar"></div>
-                <div class="email-cell from">{{ _('From User') }}</div>
+                <div class="email-cell from">{{ _('From') }}</div>
                 <div class="email-cell date">{{ _('Received On') }}</div>
                 <div class="email-cell excerpt">{{ _('Excerpt') }}</div>
                 <div class="email-cell delete">{{ _('Delete?') }}</div>
--- a/kitsune/messages/jinja2/messages/inbox.html
+++ b/kitsune/messages/jinja2/messages/inbox.html
@@ -17,8 +17,8 @@
       <div class="sumo-button-wrap">
         <input class="sumo-button" type="submit" name="mark_read" value="{{ _('Mark Selected as Read') }}">
       </div>
-      <div class="inbox-table">
-            <div class="email-row header">
+      <ol class="inbox-table">
+            <li class="email-row header">
                 <div class="email-cell checkbox">
                     <input type="checkbox" aria-label="Select all">
                 </div>
@@ -27,10 +27,10 @@
                 <div class="email-cell date">{{ _('Received On') }}</div>
                 <div class="email-cell excerpt">{{ _('Excerpt') }}</div>
                 <div class="email-cell delete">{{ _('Delete?') }}</div>
-            </div>
+            </li>
             <!-- Repeat the following div for each email message -->
           {% for message in msgs.object_list %}
-            <div class="email-row">
+            <li class="email-row{% if message.read %} read{% endif %}">
                 <div class="email-cell check">
                     <input class="field checkbox no-label" type="checkbox" aria-label="Select this message" name="id" value="{{ message.id }}" id="id_checkbox_{{ message.id }}">
                 </div>
@@ -51,10 +51,10 @@
                 <div class="email-cell delete">
                   <a class="delete" href="{{ url('messages.delete', message.id) }}" title="{{ _('Delete message') }}">&#x2716;</a>
                 </div>
-            </div>
+            </li>
           {% endfor %}
             <!-- Additional rows go here -->
-        </div>
+        </ol>
           {% csrf_token %}
         </form>
         {{ msgs|paginator }}
--- a/kitsune/messages/jinja2/messages/inbox.html
+++ b/kitsune/messages/jinja2/messages/inbox.html
@@ -1,53 +1,61 @@
 {% extends "messages/base.html" %}
 {% from "messages/includes/macros.html" import avatar_link, name_link %}
 {% set title = _('Inbox') %}
-{% set crumbs = [(None, title)] %}
+{% set crumbs = [(url('messages.inbox'), _('Messages')),
+                 (None, title)] %}
 {% set active_tab = 'inbox' %}
 
 {% block content %}
-  <div class="sumo-page-section">
+<div class="sumo-page-section">
     <article id="inbox" class="main">
       <h1 class="sumo-page-heading">{{ title }}</h1>
       <div class="actions sumo-button-wrap">
         <a class="sumo-button primary-button" href="{{ url('messages.new') }}">{{ _('New Message') }}</a>
       </div>
-
-      {% if msgs %}
-        <form action="{{ url('messages.bulk_action') }}" method="POST">
-          <div class="sumo-button-wrap">
-            <input class="sumo-button" type="submit" name="mark_read" value="{{ _('Mark Selected as Read') }}">
-            <input class="sumo-button" type="submit" name="delete" value="{{ _('Delete Selected') }}">
-          </div>
-
-          <ol class="message-list">
-            {% for message in msgs.object_list %}
-              <li
-                class="message-list--item
-                  {% if message.unread %}unread{% endif %}
-                  {% if message.replied %}replied{% endif %}"
-              >
-                <div class="field checkbox no-label">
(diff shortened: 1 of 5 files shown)


kitsune/messages/jinja2/messages/inbox.html
```diff
@@ -1,53 +1,61 @@
...
+        <input class="sumo-button" type="submit" name="mark_read" value="{{ _('Mark Selected as Read') }}">
+        <input class="sumo-button" type="submit" name="delete" value="{{ _('Delete Selected') }}">
+      </div>
+      <div class="inbox-table">
+            <div class="email-row header">
+                <div class="email-cell checkbox">
+                    <input type="checkbox" aria-label="Select all">
+                </div>
+                <div class="email-cell avatar"></div>
+                <div class="email-cell from">{{ _('From User') }}</div>
```
escattone, line 27: Since this will always be a user, I'm wondering if just `From` might be OK?

kitsune/messages/jinja2/messages/outbox.html
```diff
@@ -61,34 +46,33 @@ <h1 class="sumo-page-heading">{{ title }}</h1>
                       {%- endif -%}
                     {%- if message.recipients_count > 1 -%}, ...{% endif %}
```
escattone, line 47: I think we should move the `, ...` outside of the link for the single user we're showing here, the same way you're doing it for the group below.

kitsune/sumo/static/sumo/scss/components/_inbox.scss
```diff
@@ -1,15 +1,17 @@
...
   border-top: 1px solid var(--color-border);
 
   &--item {
+
     display: grid;
     grid-gap: p.$spacing-xs 10px;
     grid-template-columns: 36px 1fr 30px;
     grid-template-areas:
-      "checkbox avatar delete"
+      "checkbox avatar  delete"
```
escattone, line 14: Is the extra space a typo?

kitsune/messages/jinja2/messages/inbox.html
```diff
@@ -1,53 +1,61 @@
...
-                  {% if message.replied %}replied{% endif %}"
-              >
-                <div class="field checkbox no-label">
-                  <input type="checkbox" name="id" value="{{ message.id }}" id="id_checkbox_{{ message.id }}">
-                  <label for="id_checkbox_{{ message.id }}"></label>
+    {% if msgs %}
+    <form action="{{ url('messages.bulk_action') }}" method="POST">
+      <div class="sumo-button-wrap">
+        <input class="sumo-button" type="submit" name="mark_read" value="{{ _('Mark Selected as Read') }}">
+        <input class="sumo-button" type="submit" name="delete" value="{{ _('Delete Selected') }}">
```
escattone, line 19: Do we still need the `Delete Selected` button? I'm thinking that with the ability to delete individual inbox messages, and also since it'd be less clicks to use the individual delete buttons rather than selecting each row and then pushing the `Delete Selected` button, we can get rid of it?

Now we have the ability to delete individual inbox messages with the new delete button for each row, and also

kitsune/messages/jinja2/messages/outbox.html
```diff
@@ -17,42 +17,27 @@ <h1 class="sumo-page-heading">{{ title }}</h1>
...
-                </span>
-                </div>
-              </li>
+                <div class="email-cell sent">{{ _('Sent On') }}</div>
+                <div class="email-cell to">{{ _('To Users') }}</div>
+                {% if in_staff_group(request.user) %}
+                    <div class="email-cell to-groups">{{ _('To Groups') }}</div>
+                {% endif %}
+                <div class="email-cell excerpt">{{ _('Excerpt') }}</div>
+            </div>
```
escattone, line 31: For consistency with the inbox, can we add an individual delete column here? If so, could we then remove the `Delete Selected` button above?

escattone (changes requested):
This is definitely better, but I have some questions and one request for a change.

kitsune/messages/jinja2/messages/inbox.html
smithellis: I had some earlier feedback from @emilghittasv about consistency in the column headers, and I carried that over to the inbox from the outbox. Your point is good, though - I think I'll drop the "User" piece.

kitsune/sumo/static/sumo/scss/components/_inbox.scss
smithellis: Yarg!

kitsune/messages/jinja2/messages/inbox.html
smithellis: Agreed.

kitsune/messages/jinja2/messages/outbox.html
```diff
@@ -14,81 +14,60 @@ <h1 class="sumo-page-heading">{{ title }}</h1>
...
-                  <span class="to user">
-
-                  {% for user in message.to.all()[:1] -%}
+            <div class="email-row">
+                <div class="email-cell sent">{{ datetimeformat(message.created) }}</div>
+                <div class="email-cell to">
+                {% for user in message.to.all()[:1] -%}
                     <a rel="nofollow" href="{{ profile_url(user) }}">
                       {%- if name -%}
                         {{ name }}
```
escattone, line 35: As far as I can tell, `name` is never defined, so this chunk of code will never be executed.

kitsune/messages/jinja2/messages/outbox.html
```diff
@@ -14,81 +14,60 @@ <h1 class="sumo-page-heading">{{ title }}</h1>
...
+                    {% endfor %}
+                    </div>
+                {% endif %}
+                <div class="email-cell excerpt">
+                    <a class="read message text-body-sm" href="{{ url('messages.read_outbox', message.id) }}">
+                        {{ message.content_parsed|striptags|truncate(length=160) }}
+                    </a>
+                </div>
+                <div class="email-cell delete">
+                  <a class="delete" href="{{ url('messages.delete', message.id) }}" title="{{ _('Delete message') }}">&#x2716;</a>
```
escattone, line 64: This should be `'message.delete_outbox'`.

kitsune/messages/jinja2/messages/outbox.html
```diff
@@ -14,81 +14,60 @@ <h1 class="sumo-page-heading">{{ title }}</h1>
...
-                </span>
+                {%- endfor %}
                 </div>
-              </li>
+                {% if in_staff_group(request.user) %}
+                    <div class="email-cell to-groups">
+                    {%- for group in message.to_groups[:1] -%}
+                        {%- for profile in group.profile.all() -%}
+                        {% set group_slug = profile.slug %}
+                        <a href="{{ url('groups.profile', group_slug) }}">{{ group }}</a>
```
escattone, line 50: No links will be created here for groups without a profile, even though they can be added to the `to_group` field when creating the message. Should we disallow sending messages to groups without a profile? If we do want to allow them, we should handle that case.

escattone (changes requested):

escattone (commented):
Oh, one more thing. What do you think about just leaving the `To Users` or `To Groups` columns blank instead of inserting `None` when they are no users or groups? I think I'd prefer that.

kitsune/messages/jinja2/messages/inbox.html
```diff
@@ -1,53 +1,60 @@
...
-      {% if msgs %}
-        <form action="{{ url('messages.bulk_action') }}" method="POST">
-          <div class="sumo-button-wrap">
-            <input class="sumo-button" type="submit" name="mark_read" value="{{ _('Mark Selected as Read') }}">
-            <input class="sumo-button" type="submit" name="delete" value="{{ _('Delete Selected') }}">
-          </div>
-
-          <ol class="message-list">
-            {% for message in msgs.object_list %}
-              <li
```
akatsoulas, line 24: Why are you replacing the list with a set of divs? Isn't the structure that the list provides a better suited approach for an inbox? How about accessibility?

kitsune/messages/jinja2/messages/outbox.html
smithellis: Resolving this in https://github.com/mozilla/kitsune/pull/5971

kitsune/sumo/static/sumo/scss/components/_messages.scss
```diff
@@ -0,0 +1,87 @@
+@use '../config' as c;
+@use 'protocol/css/includes/lib' as p;
+
+.inbox-table {
+    display: grid;
```
akatsoulas, line 5: I understand that the html will be updated to use a list so css will be changed but fwiw I would stick with flexbox for this and I wouldn't use grid

kitsune/messages/jinja2/messages/outbox.html
smithellis: `messages.delete_outbox` - Fixed!

kitsune/messages/jinja2/messages/inbox.html
smithellis: Great catch on the accessibility shift - reverting to list.

kitsune/sumo/static/sumo/scss/components/_messages.scss
smithellis: Converted.

escattone: I'm definitely not a CSS expert, so I'm genuinely wondering why you'd recommend against using `grid`? Is there a drawback to using it?

PR #5978, opened 2024-04-26
Urls redux

smithellis:
Eliminate URL warning/error spam at Django startup.


smithellis:
This involved a lot of test changes, so warrants pretty close inspection.

akatsoulas (approved):
This looks good! Let's though add an issue about it and add it in the current sprint so it can be thoroughly tested

akatsoulas:
Let's squash it also in a single commit

kitsune/settings.py
```diff
@@ -463,6 +463,7 @@ def immutable_file_test(path, url):
     },
 ]
 
+APPEND_SLASH = True
```
akatsoulas, line 466: this is not needed

akatsoulas:
Please do not merge - I am having another pass

kitsune/dashboards/tests/test_models.py
```diff
@@ -38,14 +38,14 @@ def test_visit_count_from_analytics(self, _build_request, close_old_connections)
 PAGEVIEWS_BY_DOCUMENT_RESPONSE = {
     "kind": "analytics#gaData",
     "rows": [
-        ["/en-US/kb/hellỗ", "27"],
```
akatsoulas, line 41: Let's update the urls per app. One PR for one app and avoid adding multiple commits per app

kitsune/wiki/tests/test_parser.py
```diff
@@ -192,11 +192,11 @@ def test_internal_links(self):
         # Both internal links should link to the same article
         self.assertEqual(
             p.parse("[[%s]]" % doc.title),
-            '<p><a href="/en-US/kb/%s">%s</a>\n</p>' % (doc.slug, doc.title),
```
akatsoulas, line 195: why does this need changing?

akatsoulas (changes requested):

kitsune/dashboards/tests/test_models.py
smithellis: I'd ask you to reconsider - this is a test in `dashboards` that fails because of a setting change in the `wiki` urls . 

This PR is for `wiki` url changes, and fixing what those changes break, which is tests in a four other apps. You need the `wiki/urls.py` changes in place so your changed tests will work.

kitsune/wiki/tests/test_parser.py
smithellis: Because
`'<p><a href="/en-US/kb/real-article/">Real article</a>\n</p>' != '<p><a href="/en-US/kb/real-article">Real article</a>\n</p>'
- <p><a href="/en-US/kb/real-article/">Real article</a>`

PR #5979, opened 2024-04-26
Stop logging warnings about missing Locales

smithellis:
Logging a warning here doesn't seem to have any benefit at the moment and creates noise during testing.

I propose we remove the logging.


escattone (approved):
❤️ I'm so happy to see these removed!

PR #5994, opened 2024-05-06
Introduce Wagtail and Wagtail-Localize

escattone:
mozilla/sumo#1681

## Notes
- This lays the foundation for all of our following Wagtail and Wagtail-Localize work.
- The model changes and associated database migrations are necessary to avoid `related_name` clashes with Wagtail.
- The Wagtail admin is rendered properly and its UI will be localized with the user's preferred locale. Currently the user's preferred locale must be specified in the user's Wagtail account profile, but as @smithellis noted, we need to decide what we want to do with the Wagtail profile in light of the SuMO profile.
- The Wagtail admin is temporarily behind LDAP.
- Jinja2 templates play nicely with Django templates and both work with Wagtail.
- This PR also provides the ability to conditionally serve Wagtail pages from the `/{locale}/wagtail/` root URL. Wagtail pages are served from that root URL only if the `wagtail_experiments` Waffle flag exists and is active for the given request.


kitsune/sumo/middleware.py
```diff
@@ -78,15 +78,18 @@ def __init__(self, get_response=None):
...
             if backend and backend.split(".")[-1] != "SumoOIDCAuthBackend":
                 logout(request)
-                messages.error(request, "OIDC login required for admin access")
-                return HttpResponseRedirect("/admin/login/")
+                console = "CMS" if is_cms_request else "admin"
+                messages.error(request, f"OIDC login required for {console} access")
+                return HttpResponseRedirect(f"/{console.lower()}/login/")
 
-            return super(SUMORefreshIDTokenAdminMiddleware, self).process_request(request)
+        return super(SUMORefreshIDTokenAdminMiddleware, self).process_request(request)
```
smithellis, line 92: I believe this needs indented.

escattone: Good catch!

akatsoulas:
@escattone overall this looks good. Please re-arrange a bit the commits in a way that can be reversible and group functionality per context (eg installation  and initial configuration, waffle, class conflicts, customized login etc) and I will have one final pass.

escattone:
@akatsoulas I re-organized the commits, and it's ready for another review.

akatsoulas (approved):
This is great! Thanks @escattone :rocket:

PR #5997, opened 2024-05-08
Hide delete col on mobile instead of checkboxes

smithellis:
Per QA feedback, hiding the final column vs. first to preserve "Delete Selected" functionality.


escattone (approved):
Works great, thanks @smithellis.

PR #5998, opened 2024-05-08
This proposes a fix for 1774

smithellis:
Uses the `get_autocomplete_suggestions` method from `api.py` when we receive a GET request that contains `to`

- This ensures that what is sent as 'to' is handled like entered text
- `api.py` returns confirmed User or Group in the right format 
- We populate the field with the returned items

Caveat : if a name is passed in 'to' that is both a User and a Group, both will be returned and populated (if the user has
 permissions). The end user can remove one if they'd like. This shouldn't be a problem as the bulk of this usage will be from people clicking "Private Message" on a profile page.

### DIFF
--- a/kitsune/messages/utils.py
+++ b/kitsune/messages/utils.py
@@ -1,9 +1,12 @@
+from django.conf import settings
 from django.contrib.auth.models import Group, User
-
+from django.db.models import Q
 from kitsune.messages.models import InboxMessage, OutboxMessage
 from kitsune.messages.signals import message_sent
 from kitsune.messages.tasks import email_private_message
+from kitsune.sumo.utils import webpack_static
 from kitsune.users.models import Setting
+from kitsune.users.templatetags.jinja_helpers import profile_avatar
 
 
 def send_message(to, text=None, sender=None):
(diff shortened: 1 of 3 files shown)


kitsune/messages/views.py
```diff
@@ -85,7 +86,26 @@ def outbox(request):
...
+        # anything passed in to populate the 'to' field correctly when we
+        # don't know if the 'to' is a user or a group or none.
+        # If a name is passed in to 'to' that is the name of both a group
+        # and a user, both will be added if the user has group messaging
+        # permissions. The user can then remove the one they don't want.
+        # Without separating the to field into to_users and to_groups,
+        # we can't know which is which, so we do a best fit.
+        to = request.GET.get("to")
+        if to:
+            suggestions = json.loads(get_autocomplete_suggestions(request).content)
```
escattone, line 103: This works fine, but I don't like that we're treating it like a web request -- converting to and from JSON, passing the `request` object, etc. -- when it's really just a function call. I think we should extract the core of `get_autocomplete_suggestions` into a non-web function that accepts a single name string as its argument, and returns a list of suggestions, and then use that function here and within `get_autocomplete_suggestions`.

escattone (changes requested):
This works great, but I would recommend an organizational change.

kitsune/messages/utils.py
```diff
@@ -65,3 +68,41 @@ def send_message(to, text=None, sender=None):
...
+        "name": item.username if is_user else item.name,
+        "type_and_name": f"User: {item.username}" if is_user else f"Group: {item.name}",
+        "display_name": item.profile.name if is_user else item.name,
+        "avatar": (
+            profile_avatar(item, 24) if is_user else webpack_static(settings.DEFAULT_AVATAR)
+        ),
+    }
+
+
+def find_users_and_groups_by_search(pre, show_groups=False, exact=False):
```
escattone, line 90: Since both current use cases -- the API and the `new_message` view -- need suggestions rather than DB objects, and there's no current use case for DB objects, I would have this function return a list of suggestions by default, and perhaps add a keyword argument that allows the caller to ask for a list of DB objects instead.

Do you think the limit of users or groups -- in this case 10 -- should be configurable via a keyword argument as well?

escattone (changes requested):

akatsoulas:
I opened PR #6001 as an alternative suggestion to populate the username by default. It is assumed that's always a user since it's populated by the username of the profile. Thoughts?

smithellis:
In the original work I was told there was no reason to grab the 'to' from a GET request, so I removed that functionality. We now found this use case for it. Thinking to the future, if our form allows users and groups in the `to` field, it could be problematic to some later use case - and the dev will have to go looking around to figure out why this form seemingly accepts a `to` but not in all the same cases he/she/they might expect. I was honestly not willing to assume I should only fix for this one case, because as sure as I did that I'd get a ticket about how groups can't be sent as a `to` param but are allowed on the form.
If that isn't something to be worried about, then https://github.com/mozilla/kitsune/pull/6001 is an eloquent solution.
Let me know your thoughts...We need to decide on:
1 - Just your proposed fix OR
2 - Refactor plus your proposed fix OR
3 - Continue with `group` inclusive fix and refactor

akatsoulas:
Since groups target many users, I would argue that it's best to explicitly type the name in the form instead of prepopulating the field (now and in the future) to avoid spamming by mistake. Moreover, since this is a GET request passed as a parameter I believe it's good hygiene to default to the `User:` instance for the same reasons - to avoid spamming groups

Regarding the refactor, it's totally up to you. For what is worth I don't see a use case right now for refactoring the code. The refactor won't make it more DRY than it is right now. FWIW, I would opt for just getting the fix in.

smithellis:
Closing in favor of https://github.com/mozilla/kitsune/pull/6001

PR #6001, opened 2024-05-13
Populate username by default

akatsoulas:


kitsune/sumo/form_fields.py
```diff
@@ -92,22 +92,24 @@ def to_python(self, value):
...
+            for part in value.split(",")
+            if part.strip()
         )
 
         # Create data structure to hold values grouped by keys
         to_objects = {}
         for key, value in key_value_pairs:
+            # check if the value is a valid username in the database
+            if not User.objects.filter(username=value).exists():
+                raise ValidationError(_(f"{value} is not a valud username."))
```
smithellis, line 112: Typo - "valid"

smithellis (changes requested):

kitsune/sumo/form_fields.py
akatsoulas: updated!

smithellis:
Assuming that was a test hiccup, I'm good with this. I'll drop https://github.com/mozilla/kitsune/pull/5998

PR #6016, opened 2024-05-17
Update navigation menus

akatsoulas:


kitsune/products/views.py
```diff
@@ -68,26 +70,53 @@ def product_landing(request, slug):
...
+        del topic_kw["slug"]
+        topic_kw["slug__in"] = NAVIGATION_TOPICS[topic_title]
+        topic_list = Topic.objects.filter(title__in=NAVIGATION_TOPICS.keys())
+        topic_subquery = topic_list.filter(slug=OuterRef("slug")).order_by("id").values("id")[:1]
+        topic_list = Topic.objects.filter(id__in=Subquery(topic_subquery))
+    else:
+        topics = topics_for(request.user, product=product, parent=None)
+
+    topics = Topic.objects.filter(**topic_kw)
+    if not (topic := topics.exists()):
```
escattone, line 108: I don't think the walrus operator is needed here?

```python
    if not topics.exists():
```

kitsune/wiki/facets.py
```diff
@@ -108,12 +108,14 @@ def _documents_for(user, locale, topics=None, products=None):
     # speed up query by removing any ordering, since we're doing it in python:
     qs = qs.select_related("current_revision", "parent").order_by()
 
+    topic_ids = [t.id for t in topics]
     for topic in topics or []:
         # we need to filter against parent topics for localized articles
-        qs = qs.filter(Q(topics=topic) | Q(parent__topics=topic))
+        qs = qs.filter(Q(topics__in=topic_ids) | Q(parent__topics__in=topic_ids))
```
escattone, line 114: With this approach, we don't need the loop, and since `topics` can be `None`, maybe something like the following instead?

```python
    if topics:
        topic_ids = [t.id for t in topics]
        # we need to filter against parent topics for localized articles
        qs = qs.filter(Q(topics__in=topic_ids) | Q(parent__topics__in=topic_ids))
```

kitsune/questions/jinja2/questions/question_list.html
```diff
@@ -295,6 +304,22 @@ <h3 class="sidebar-subheading">{{ _('Filter by') }}</h3>
       {% for f, desc in filters.items() %}
       <li {{ filter|class_selected(f) }}><a href="{{ questions_url(filter=f, page=None) }}">{{ desc }}</a></li>
       {% endfor %}
+      <li class="sidebar-subheading sidebar-nav--heading-item">{{ _('Topic') }}</li>
```
escattone, line 307: This seems to me like it should be moved under the `{% if topic_list %}` clause immediately below, because it looks odd on the page when there is no `topic_list`.

image

kitsune/gallery/jinja2/gallery/gallery.html
```diff
@@ -17,29 +18,31 @@ <h1 class="sumo-page-heading no-logo">
         </h1>
       </div>
       <div class="sumo-l-two-col--sidebar">
-        {% if user.is_authenticated %}
-          <a id="btn-upload" class="sumo-button primary-button" href="#">{{ _('Upload a New Media File') }}</a>
-        {% endif %}
+        <ul class="sidebar-nav sidebar-folding">
```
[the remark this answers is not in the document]
akatsoulas: updated

kitsune/products/jinja2/products/documents.html
```diff
@@ -1,24 +1,34 @@
...
+  {% set ga_products = "/" + product.slug + "/" %}
+  {% set page_title = pgettext('DB: products.Topic.title', topic.title) + ' | ' +  _('{product} Help')|f(product=pgettext('DB: products.Product.title', product.title)) %}
 {% else %}
-  {% set ga_content_group = "product-and-topic-home" %}
-  {% set crumbs = crumbs + [(None, pgettext('DB: products.Topic.title', topic.title))] %}
-  {% set canonical_url = canonicalize(model_url=topic.get_absolute_url()) %}
+  {% set crumbs = [(url('products.topic_documents', topic_slug=topic.slug), pgettext('DB: products.Topic.title', topic.title))] %}
+  {% set search_params = {'topic': topic.slug} %}
+  {% set page_title = pgettext('DB: products.Topic.title', topic.title) %}
+    {% set ga_content_group = "topic-home" %}
```
escattone, line 26: Nit. Dedent.

escattone (approved):
Wow, this was a lot of work, thanks @akatsoulas! r+wc

PR #6022, opened 2024-05-21
Playwright custom xpath cleanup

emilghittasv:
- No changes or additions, just deleting some unnecessary variables and applying the xpath directly to the driver functions.


escattone (approved):
This is cleaner, nice!

PR #6024, opened 2024-05-21
KB Document metadata

smithellis:
Enable meta data for KB articles and KB article listing pages:
- Add row below title that indicates product, last updated date,  and number of helpful votes
- Various scss changes to support the layout 
- I implemented the `django_jinja.contrib._humanize` app in order to give us friendly last updated times instead of reinventing the wheel.
- Changed `view.py` and `facets.py` files to support passing data needed

Screenshot 2024-05-21 at 7 58 59 PM

seScreenshot 2024-05-23 094456


escattone (commented):
I'm still reviewing, but wanted to make an early comment. Could you provide a screenshot of the desired metadata and its layout? I was given the following shot from Figma, which doesn't include the product but instead includes the read time, but maybe that's the wrong one?

image

kitsune/messages/migrations/0002_inboxmessage_to_group_outboxmessage_to_group.py
```diff
@@ -1,23 +0,0 @@
-# Generated by Django 4.2.10 on 2024-03-18 11:23
```
akatsoulas, line 1: This shouldn't belong here.

kitsune/messages/migrations/0003_remove_inboxmessage_to_group.py
```diff
@@ -1,16 +0,0 @@
-# Generated by Django 4.2.10 on 2024-04-18 12:20
```
akatsoulas, line 1: same

kitsune/wiki/jinja2/wiki/includes/document_macros.html
```diff
@@ -12,6 +12,20 @@
   <h1 class="sumo-page-heading">{{ document.title }}</h1>
 {%- endmacro %}
 
+{% macro document_meta(document, helpful_votes, product) -%}
```
akatsoulas, line 15: s/document_meta/document_metadata

kitsune/sumo/static/sumo/scss/layout/_document.scss
```diff
@@ -305,3 +305,89 @@
     }
   }
 }
+
+
+.sumo-kb-article-header {
```
akatsoulas, line 310: I would revert this to the way it was before and adjust only for the KB. The `sumo-article-header` is the sumo version of the Protocol Article and not specific to the KB.

akatsoulas (changes requested):

kitsune/sumo/static/sumo/scss/layout/_document.scss
smithellis: Ok.

kitsune/wiki/jinja2/wiki/includes/document_macros.html
```diff
@@ -12,6 +12,20 @@
   <h1 class="sumo-page-heading">{{ document.title }}</h1>
 {%- endmacro %}
 
+{% macro document_metadata(document, helpful_votes, product) -%}
```
escattone, line 15: Nit. Indent (2 spaces I think) everything under this macro.

kitsune/wiki/jinja2/wiki/includes/document_macros.html
```diff
@@ -12,6 +12,20 @@
...
+<div id="document_meta" class="has-border-bottom">
+  <span class="product">{{ product }}</span>
+  <span class="last-updated">
+    <img class="pencil" src="{{ static('/sumo/img/pencil.svg') }}" /> <strong>Last updated:</strong> <span class="time">{{ document.current_revision.created|naturaltime }}</span>
```
escattone, line 19: This should be `webpack_static('sumo/img/pencil.svg')` -- note the lack of an initial forward slash in the path -- instead of `static('/sumo/img/pencil.svg')` in order to pick up the hashed version created by webpack.

kitsune/wiki/jinja2/wiki/includes/document_macros.html
```diff
@@ -12,6 +12,20 @@
...
+  </span>
+  {% if helpful_votes %}
+  <span class="helpful-info">
+    <img class="thumbsup" src="{{ static('/sumo/img/thumbs-up.svg') }}"/><span class="helpful-count">{{ helpful_votes }}%</span> of users found this helpful
```
escattone, line 23: Likewise here. This should be `webpack_static('sumo/img/thumbs-up.svg')` -- note the lack of an initial forward slash in the path -- instead of `static('/sumo/img/thumbs-up.svg')` in order to pick up the hashed version created by webpack.

kitsune/sumo/static/sumo/scss/layout/_document.scss
```diff
@@ -305,3 +305,54 @@
...
 }
+
+#document_meta {
+  display: flex;
+  align-items: center;
+  margin-top: 16px;
+  padding-bottom: 16px;
+  margin-bottom: 32px;
+  font-family: 'Inter', sans-serif;
+  font-size: 14px;
```
escattone, line 316: I was thinking that it might be better to replace this with `1rem`, and in general as many of the `px` values with Protocol constants that match.

escattone (changes requested):

escattone (changes requested):
The only other thing I'd consider is that when you release this code -- which has a new data structure for the `votes_dict`cached under the `votes_cache_key` -- the prior cached value will clash with your current code. It might be worth considering how to handle that so it doesn't cause errors during the period until the cached content expires.

kitsune/wiki/facets.py
```diff
@@ -122,32 +122,62 @@ def _documents_for(user, locale, topics=None, products=None):
...
         #       field, and the "HelpfulVote" query will be massively slower.
+        # Extended votes_query
         votes_query = (
             HelpfulVote.objects.filter(
-                revision_id__in=qs.values_list("current_revision_id", flat=True),
-                created__range=(datetime.now() - timedelta(days=30), Now()),
-                helpful=True,
+                revision_id__in=qs.values_list("current_revision_id", flat=True)
             )
-            .values("revision_id")
```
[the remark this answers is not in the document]
smithellis: This is a great catch. I literally thought it was redundant, but now I understand its role. Thanks for this.

kitsune/wiki/facets.py
```diff
@@ -122,32 +122,62 @@ def _documents_for(user, locale, topics=None, products=None):
...
         )
-        votes_dict = {row["revision_id"]: row["count"] for row in votes_query}
+
+        # Convert the query results into a dictionary
+        votes_dict = {
+            row["revision_id"]: {
+                "helpful_count": row["helpful_count"],
+                "total_count": row["total_count"],
+                "time_limited_count": row["time_limited_count"],
+            }
```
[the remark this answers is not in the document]
smithellis: Another great catch. Thanks.

kitsune/wiki/facets.py
```diff
@@ -122,32 +122,62 @@ def _documents_for(user, locale, topics=None, products=None):
         # NOTE: It's important to use "created__range" rather than "created__gt"
         #       with Postgres, otherwise it won't use the index on the "created"
         #       field, and the "HelpfulVote" query will be massively slower.
+        # Extended votes_query
         votes_query = (
             HelpfulVote.objects.filter(
-                revision_id__in=qs.values_list("current_revision_id", flat=True),
-                created__range=(datetime.now() - timedelta(days=30), Now()),
```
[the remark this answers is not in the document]
escattone: Hm, the current revision of more than a third of all of the active KB articles is older than 5 years.

```python
In [37]: Document.objects.filter(current_revision__created__lte=date(2019, 1, 1), is_archived=False, is_template=Fa
    ...: lse).count()
Out[37]: 13696

In [38]: Document.objects.filter(current_revision__isnull=False, is_archived=False, is_template=False).count()
Out[38]: 37062
```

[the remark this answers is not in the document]
escattone: We could reduce the size of the `HelpfulVote` table by about 25% by removing votes on revisions of documents that have been archived.

```python
In [40]: HelpfulVote.objects.filter(revision__document__is_archived=True).count()
Out[40]: 14709060

In [41]: HelpfulVote.objects.count()
Out[41]: 56457295
```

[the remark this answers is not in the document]
smithellis: Worth more discussion. I'm not sure what the right answer is in the short term - I can't produce a % of votes without the total votes. I've excluded archived documents in the query.
If this turns out to be too intensive we can look at batch processing for totals and storing them somewhere.

kitsune/products/jinja2/products/documents.html
```diff
@@ -115,6 +115,17 @@ <h2 class="sumo-card-heading">
...
+                <span class="product">{% for product in document['products'] %}{{ product }}{% endfor %}</span>
+                <span class="last-updated">
+                  <img class="pencil" src="{{ webpack_static('sumo/img/pencil.svg') }}" /> <strong>Last updated:</strong> <span class="time">{{ document['created']|naturaltime }}</span>
+                </span>
+                {% if document['percent_helpful']>0 %}
+                <span class="helpful-info">
+                  <img class="thumbsup" src="{{ webpack_static('sumo/img/thumbs-up.svg') }}"/><span class="helpful-count">{{ document['percent_helpful'] }}%</span> of users found this helpful
+                </span>
+                {% endif %}
+              </div>
```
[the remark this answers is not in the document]
smithellis: Initially, I had talked myself out of that since it lives in a different app. I have now altered it and implemented it.

kitsune/wiki/facets.py
```diff
@@ -122,32 +122,54 @@ def _documents_for(user, locale, topics=None, products=None):
...
         #       with Postgres, otherwise it won't use the index on the "created"
         #       field, and the "HelpfulVote" query will be massively slower.
+        # Extended votes_query
         votes_query = (
-            HelpfulVote.objects.filter(
-                revision_id__in=qs.values_list("current_revision_id", flat=True),
-                created__range=(datetime.now() - timedelta(days=30), Now()),
-                helpful=True,
-            )
+            HelpfulVote.objects.filter(revision__document__is_archived=False)
```
escattone, line 127: I don't think this `.filter(revision__document__is_archived=False)` is necessary, given that we're already limiting the revisions to those within the `qs` and the `qs` is already filtering out the revisions of archived documents. It may even cause the `votes_query` to be a bit more costly since now it has to join to both the `Revision` and `Document` tables, so I would recommend removing this.

kitsune/wiki/facets.py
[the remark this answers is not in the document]
escattone: Perhaps for now, we shouldn't worry about it? We're already caching the results for 24 hours, so only a single request every 24 hours may experience the "cold cache" delay.

escattone (changes requested):
I'm thinking that in addition to removing `.filter(revision__document__is_archived=False)` from the `votes_query`, you probably should retrieve/store the `votes_dict` from/in a different key -- maybe just version the key -- so that the old data in the cache doesn't cause failures for the initial period of up to 24 hours (the cache timeout) after the release of this new code. Maybe something like:

```python
...
votes_cache_key = f"votes_for_v2:{cache_key}"
if votes_dict is None:
...
```

escattone:
After a SuMO platform team discussion, we decided to not show the helpful votes metadata for the "list" of documents case in order to avoid the performance issue for now. We will show the helpful votes metadata for the single document case. We also decided to truncate the products metadata -- a comma-separated list of product titles associated with the document -- to 50 characters, although when the helpful votes metadata is included, I've reduced it to 25 characters in order to avoid overflow when the last updated metadata is long for cases like `Last updated: 4 months, 1 week ago`.

PR #6038, opened 2024-05-31
Improving playwright kb restricted visibility tests

emilghittasv:
- Removing some repetitive steps inside the "KB - Restricted Visibility" tests & adding a helper function for the restricted visibility group removal flow.


playwright_tests/flows/explore_articles_flows/article_flows/edit_article_meta_flow.py
```diff
@@ -87,3 +87,13 @@ def edit_article_metadata(self, title=None,
                 super()._click_needs_change_checkbox()
 
         super()._click_on_save_changes_button()
+
+    def _remove_a_restricted_visibility_group(self, group_name='', remove_all=False):
+        if KBArticleRevision.KB_EDIT_METADATA not in super()._get_current_page_url():
+            super()._click_on_edit_article_metadata()
+
+        if remove_all:
```
akatsoulas, line 95: No need to change anything here but it looks like that you can omit one of the params. If the group_name is empty you can assume that you can delete all. Otherwise limit the deletion to specific group.

Even better you can handle this logic to the `_delete_all_restricted_visibility_groups_metadata()` function. Just pass it a group_name. If it's empty remove them all. This way the responsibility for the deletion of the metadata belongs solely to the function responsible for doing so.

akatsoulas (approved):

playwright_tests/flows/explore_articles_flows/article_flows/edit_article_meta_flow.py
emilghittasv: Yes! Thank you @akatsoulas!

emilghittasv:
@akatsoulas I've updated the code based on your suggestion. 

Thanks for the feedback!  It's definitely cleaner.

PR #6044, opened 2024-06-04
Created vs Last Updated

smithellis:
If we only have a single revision, show "Created" as the label in document metadata vs. "Last updated"

### DIFF
--- a/kitsune/products/views.py
+++ b/kitsune/products/views.py
@@ -9,6 +9,7 @@ from product_details import product_details
 from kitsune.products.models import Product, Topic
 from kitsune.questions import config as aaq_config
 from kitsune.wiki.decorators import check_simple_wiki_locale
+from kitsune.wiki.models import Revision
 from kitsune.wiki.facets import documents_for, topics_for
 from kitsune.wiki.utils import get_featured_articles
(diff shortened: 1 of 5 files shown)


kitsune/products/views.py
```diff
@@ -96,6 +97,9 @@ def document_listing(request, product_slug, topic_slug, subtopic_slug=None):
     thirty_days_ago = datetime.now() - timedelta(days=30)
     for document in documents:
         document["is_past_thirty_days"] = document["created"] < thirty_days_ago
+        document["is_first_revision"] = (
+            Revision.objects.filter(document=document["id"]).count() == 1
```
escattone, line 101: I was testing the PR and realized that simply creating a new revision that hasn't been approved yet changed `Created` to `Last updated`, which seems wrong. So I'm thinking this should change slightly to count the number of approved revisions.

escattone (changes requested):

kitsune/products/views.py
smithellis: Great catch. Updated.

escattone (approved):
Looks good!

PR #6056, opened 2024-06-10
Add localization note to clarify usage

smithellis:
Add a L10N comment to clarify for translators.

### DIFF
--- a/kitsune/wiki/jinja2/wiki/includes/document_macros.html
+++ b/kitsune/wiki/jinja2/wiki/includes/document_macros.html
@@ -26,7 +26,7 @@
     {% endif %}
     {% if helpful_votes %}
     <span class="helpful-info">
-      <img class="thumbsup" src="{{ webpack_static('sumo/img/thumbs-up.svg') }}"/>{{ _('<span class="{0}">{1}{2}</span> of users voted this helpful')|fe("helpful-count", helpful_votes, "%") }}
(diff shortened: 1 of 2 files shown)


kitsune/wiki/jinja2/wiki/includes/document_macros.html
```diff
@@ -26,7 +26,7 @@ <h1 class="sumo-page-heading">{{ document.title }}</h1>
     {% endif %}
     {% if helpful_votes %}
     <span class="helpful-info">
-      <img class="thumbsup" src="{{ webpack_static('sumo/img/thumbs-up.svg') }}"/>{{ _('<span class="{0}">{1}{2}</span> of users voted this helpful')|fe("helpful-count", helpful_votes, "%") }}
+	    <img class="thumbsup" src="{{ webpack_static('sumo/img/thumbs-up.svg') }}"/>{# L10N: {0} is 'helpful-count', {1} is the actual percentage and {2} is the percent sign. #}{{ _('<span class="{0}">{1}{2}</span> of users voted this helpful')|fe("helpful-count", helpful_votes, "%") }}
```
escattone, line 29: I think the case matters, but I'm not certain, based on https://github.com/mozilla/kitsune/blob/715338dbc502b491464aaa79f9eb073102471bca/kitsune/sumo/management/commands/extract.py#L34

...so I think it should be `L10n:` instead of `L10N:`?

smithellis: Got it.

akatsoulas (approved):
Please squash before merging

PR #6057, opened 2024-06-10
fix article & question weekly pageview crons

escattone:
mozilla/sumo#1819


akatsoulas (approved):

foxrider598 (commented):
kitsune/questions/views.py

PR #6064, opened 2024-06-13
Make `products/` configurable and served in WT

smithellis:
* Register Product model as Wagtail snippet
* Add panels and preview template for Products
* Add a custom `StructBlock` for Products that lets user choose products
* Add ability to preview block in `/cms/`
* Add a new page type "LandingPage" that user can add to site and configure
* Allow user to preview new page in `/cms/`

NOTE: To use this locally, you need to enable wagtail admin, and set up a Site in the admin that has a root at /products. This URL approach is one that works, but I'm looking at other options.


smithellis:
I'm going to close this, reset Dev to main begin using Dev for this work.

PR #6070, opened 2024-06-17
Expanding playwright localization coverage

emilghittasv:
1. Verifying that SUMO redirects to a fallback locale when trying to access an unsupported one.
2. Verifying SUMO's fallback mechanism for the fy, zh-Hans and zh-Hant locales. (currently skipped due to a known failure).
3. Verifying that accessing all supported SUMO locales returns status code 200
4. Verifying SUMO locale priority:
- SUMO should redirect to the locale specified inside the user profile if no locale is specified in the URL.
- Verifying that SUMO should respect the ?lang query parameter and redirect to that particular locale regardless of the locale specified inside the user profile.
- Verifying that SUMO redirects to the locale specified inside the Accept-Language header if no preferred language is set at profile level.
- Verifying the fallback mechanism of the Accept-Language header if the first specified language is not supported or invalid.
- Verifying that selecting all the available options inside the "Language" dropdown menu displayed inside the footer section redirects the user to the chosen locale successfully.
5. Adding the required env variables for decouple in order to use the SUMO_LANGUAGES, FALLBACK_LANGUAGES, NON_SUPPORTED_LOCALES dictionaries in tests.
6. Fixing the Recent Revisions Dashboard test failures.
7. Adding playwright capability of setting/manipulating the http header.

### DIFF
--- a/.github/workflows/playwright.yml
+++ b/.github/workflows/playwright.yml
@@ -49,6 +49,10 @@ env:
     TEST_ACCOUNTS_PS: ${{secrets.AUTOMATION_ACCOUNTS_PASSWORD}}
     TEST_ACCOUNT_MODERATOR: ${{secrets.AUTOMATION_MODERATOR_ACCOUNT}}
     BROWSER: ${{secrets.BROWSER}}
+    DATABASE_URL: postgres://kitsune:kitsune@postgres:5432/kitsune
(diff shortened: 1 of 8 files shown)


escattone (approved):
❤️ Awesome! I love that you're testing all of these locale-related features.

PR #6074, opened 2024-06-18
Expanding playwright coverage over premium products ticket submission

emilghittasv:
- Updating the aaq submission flow to work with premium and loginless forms.
- Adding new locators for the premium and loginless forms and helper functions.
- Expanding playwright coverage for the following cases:
1. Verifying that the premium tickets can be successfully submitted for all premium products. Currently, we are sending one ticket /each premium product and choosing a random topic & os from the dropdowns. We are also making use of the whitelisted subject & body keyword so that Zendesk automatically closes those tickets to avoid spam.
2. Verifying that 3 loginless tickets can be successfully submitted before hitting the ratelimit. (making use of the whitelisted subject & body keyword in order to avoid spamming Zendesk).

### DIFF
--- a/playwright_tests/core/basepage.py
+++ b/playwright_tests/core/basepage.py
@@ -1,3 +1,4 @@
+import random
 from playwright.sync_api import Page, ElementHandle, Locator, TimeoutError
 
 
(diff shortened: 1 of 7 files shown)


escattone (approved):
🚀

PR #6075, opened 2024-06-18
Make products/ configurable and served in WT

smithellis:
* Register Product model as Wagtail snippet
* Add panels and preview template for Products
* Add a custom StructBlock for Products that lets user choose products
* Add ability to preview block in `/cms/`
* Add a new page type "LandingPage" that user can add to site and configure
* Allow user to preview new page in `/cms/`
* wagtail_urls conditionally added
* WAGTAIL_ENABLE setting added

This enables Products to be wagtail_snippets, and thus editable inside the Wagtail admin. Further, it lets you construct a page of type `LandingPage` out of blocks that represent product cards.

The `wagtail_urls` are appended if `WAGTAIL_ENABLE` is true in settings.

To adequately test this, you will need to enable `WAGTAIL_ENABLE` along with `WAGTAIL_ADMIN_ENABLE` in settings. Then log in, then go to `/cms/` and add a page.

What I need most eyes on is the URL flow - is it working like I think, am I cached, etc. I will continue to test locally as well - just wanted to note that I'm no yet 100% that URL are working like I believe they are.

### DIFF
--- a/kitsune/products/jinja2/products/blocks/product_snippet_block.html
+++ b/kitsune/products/jinja2/products/blocks/product_snippet_block.html
@@ -1,0 +1,17 @@
+<div class="card card--product zoom-on-hover"> 
+  {% set product = value.product %}
+  <img class="card--icon" src="{{ product.image_url }}" alt="{{ product.title }}" />
+  <div class="card--details">
+    <h3 class="card--title">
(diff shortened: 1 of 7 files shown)


smithellis:
Changing approach. Again.

PR #6093, opened 2024-06-27
do not cache document responses when varying on accept-language header

escattone:
mozilla/sumo#1500


smithellis (approved):
Thanks @escattone - this looks nice.

PR #6095, opened 2024-07-01
Ignore products in topic listing

akatsoulas:


akatsoulas:
This is the same solution that we are already using in products.

PR #6098, opened 2024-07-03
Allow filtering on is_archived

akatsoulas:


akatsoulas:
the changes are localized to the admin pages of Product and Topic and can be verified through testing.

PR #6100, opened 2024-07-04
Do not override the deafult manager

akatsoulas:
* Introduce active manager for Topic and Product


akatsoulas:
This changes back the default manager to `models.Manager()` and introduces the `active` manager to solve for the problem that the changelist in the admin interface is missing the archived entries which in turns produces the ID error.

Most of the codebase has been converted to use the `active` manager. Exceptions to this are the data migrations and all the code responsible to calculate metrics. 

I am merging this PR to allow some QA time before Monday and maybe @escattone can have a pass on Monday in case anything is out of place or wrong.

escattone (commented):
This all looks good to me, including the spots where you kept `Product.objects` instead of changing to `Product.active`.

PR #6112, opened 2024-07-10
Hide showfor macro for non visible articles

akatsoulas:


akatsoulas:
The only change in this PR is this line
```jinja
{% if products|selectattr("visible")|list %}
{% endif %}
```

smithellis (approved):

PR #6116, opened 2024-07-11
Query subtopics for redirects

akatsoulas:


kitsune/products/views.py
```diff
@@ -85,19 +85,21 @@ def document_listing(request, topic_slug, product_slug=None, subtopic_slug=None)
...
             }
 
             if product_slug:
                 redirect_params["product_slug"] = product_slug
-            if subtopic_slug:
-                redirect_params["subtopic_slug"] = subtopic_slug
 
             return redirect(document_listing, **redirect_params)
         except TopicSlugHistory.DoesNotExist:
             ...
```
smithellis, line 100: Is this deferred handling on purpose?

kitsune/products/views.py
```diff
@@ -85,19 +85,21 @@ def document_listing(request, topic_slug, product_slug=None, subtopic_slug=None)
...
+        except TopicSlugHistory.MultipleObjectsReturned:
+            ...
```
smithellis, line 102: Same as above - legit?

kitsune/products/views.py
akatsoulas: yes - just moves along

kitsune/products/views.py
akatsoulas: same `...` is the same with `pass` in Python

smithellis: I know - just making sure you didn't intend to do something with those exceptions.

smithellis (approved):

PR #6125, opened 2024-07-17
Create Page models and new folder structure

smithellis:
- pages.py 
- Wagtail specific additions for: 
   - SumoPlaceholderPage - a page type for pages needed to fill in WT tree 
   - SingleProductIndexPage - a page that is the homepage for a chosen product
   - Various blocks used to build pages:
      - ProductSnippetBlock - allows product selection from model 
      - SearchBlock - include a search feature
      - CTABlock - a block for CTA inclusion
      - FeaturedArticleBlock - a block for adding a featured article section 
      - FrequestTopicsBlock - a block for adding frequent topics section
- wagtail_hooks.py is where we now register Products as a wagtail_snippet
It also includes various templates for blocks and WT-served pages.


smithellis:
Ran into an interesting test issue where the Wagtail `register_snippet` decorator was creating a failure for some tests because the `Product` model was getting re-registered, and WT would say `Product already registered` and wouldn't fail gracefully. I smote this issue after spending too much time looking at it today. We should probably register any existing SUMO models as WT snippets in a `wagtail_hooks.py` file from now on. This lets us keep WT and SUMO separate, even as we combine them.

smithellis:
Closing; to be replaced.

PR #6126, opened 2024-07-17
URL changes required WT work

smithellis:
Adding the decorator to products
Adding catch-all wagtail pattern at end (likely not needed now that we are using decorator)
Proposing we convert from re_path to path where it makes sense


smithellis:
Closing this to create more atomic PR. This includes decorator work that exists separately.

PR #6130, opened 2024-07-19
Taxonomy topics

akatsoulas:


escattone (approved):
Really nicely done! Just a few topic title nits, and it looks like https://github.com/mozilla/kitsune/blob/f9c7d4275f319850d5e4ffc702076985f8ea3808/kitsune/wiki/widgets.py#L14

needs to change to something like `topics_and_subtopics = Topic.active.filter(product__isnull=False).all()` in order to resolve the test failures.

PR #6131, opened 2024-07-21
Fix banner macro CTA l10n

janbrasna:
https://github.com/mozilla/sumo/issues/1887


akatsoulas (approved):
Thanks @janbrasna! :rocket:

PR #6134, opened 2024-07-23
Expanding playwright coverage over the Explore by topic page

emilghittasv:
- Adding a new helper function in BasePage to extract the list of text content for a given locator.
- Adding the https://support.allizom.org/en-US/kb/get-community-support?exit_aaq=1 page to test data.
- Adding the AAQ widget text which is displayed when the "All Products" filter is applied to AAQWidgetMessage for further usage in tests.
- Expanding the locators & actions for the ExploreByTopicsPage class.
- Adding playwright coverage for the "Explore By Topic" page:
  - Verifying that the correct explore by topic page header is displayed when changing the topic via the "All Topics" side navbar.
  - Verifying that the "Filter by product" successfully updates the displayed list of articles with articles that belong to the chosen product from the "Filter by product" dropdown (by verifying the displayed metadata for each listed article)
  - Verifying that the correct AAQ widget text is displayed for each filtered product.
  - Verifying the correct AAQ widget redirect for each filtered product.
  - Added playwright coverage for the https://github.com/mozilla/sumo/issues/1876 case
- Adding the new test suite to playwright.yml

### DIFF
--- a/.github/workflows/playwright.yml
+++ b/.github/workflows/playwright.yml
@@ -35,6 +35,7 @@ on:
                     - KB Article Translation
                     - recentRevisionsDashboard
                     - kbDashboard
+                    - exploreByTopics
 
 env:
     TEST_ACCOUNT_12: ${{secrets.AUTOMATION_TEST_ACCOUNT_12}}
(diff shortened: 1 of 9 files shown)


escattone (approved):
Nice!

PR #6137, opened 2024-07-24
fix products.js and convert to vanilla JS

escattone:
mozilla/sumo#1716

This PR is a fix for this issue, but at the immediate, surface level. It fixes `products.js` to handle the cases when things don't exist in the DOM, and also converts it to vanilla JS (removes its reliance on jQuery). The PR that addresses the deeper problem that this issue manifests is https://github.com/mozilla/kitsune/pull/6136. Only one of the two PR's should be merged. This PR should be favored over the other one if we'd like to remain open to making the `Refresh Firefox` button (the one included by the `download_firefox()` Jinja macro) work as intended sometime in the future.

### DIFF
--- a/kitsune/sumo/static/sumo/js/products.js
+++ b/kitsune/sumo/static/sumo/js/products.js
@@ -1,17 +1,22 @@
 import UITour from "./libs/uitour";
 import compareVersions from "./compare_versions";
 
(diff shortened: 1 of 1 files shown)


akatsoulas:
Closing in favor of #6136

PR #6139, opened 2024-07-25
[l10n] Fix the "Cancel" button from the /users/edit page which is missing localization

emilghittasv:


akatsoulas (approved):

emilghittasv:
Fixes https://github.com/mozilla/sumo/issues/1905

PR #6142, opened 2024-07-25
Display inline products m2m field.

akatsoulas:
* Default the product and topic queries to non archived.


escattone (approved):
Thanks @akatsoulas!

PR #6145, opened 2024-07-25
[l10n] Fix the unlocalized "More" option from the product topic listing page

emilghittasv:
Fix the "More" unlocalized string from the topics listing page (example: https://support.allizom.org/de/products/firefox/troubleshooting). 
image


akatsoulas (approved):

escattone (approved):
Thanks @emilghittasv!

PR #6150, opened 2024-07-29
Filter by product in topic navigation.

akatsoulas:


escattone (changes requested):
I'm getting an error when I select one of the products in the product dropdown, and it's because the `Topic.get_absolute_url()` method is still designed for the old world of a topic always existing under a single product.

image

Otherwise, this looks good to me.

akatsoulas:
Thanks for the prompt review. Since this code is looking good, I am merging this and I am going to tackle the error in https://github.com/mozilla/sumo/issues/1917 which affects all articles.

2000400:
```

```

2000400 (changes requested):
[]()[]()

PR #6153, opened 2024-07-30
Update Topic get_absolute_url method.

akatsoulas:
* Simplify breadcrumb construction in wiki.

### DIFF
--- a/kitsune/products/models.py
+++ b/kitsune/products/models.py
@@ -163,25 +163,18 @@ class Topic(BaseProductTopic):
         query.update(kwargs)
         return Document.objects.visible(**query)
 
-    def get_absolute_url(self):
-        if self.parent is None:
-            return reverse(
-                "products.documents",
-                kwargs={
-                    "product_slug": self.product.slug,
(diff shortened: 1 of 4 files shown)


escattone (approved):
r+wc

kitsune/products/models.py
```diff
@@ -164,25 +164,18 @@ def documents(self, user=None, **kwargs):
...
-            return reverse(
-                "products.subtopics",
-                kwargs={
-                    "product_slug": self.product.slug,
-                    "topic_slug": self.parent.slug,
-                    "subtopic_slug": self.slug,
-                },
-            )
+            kwargs.update({"topic_slug": self.parent.slug, "subtopic_slug": self.slug})
+            named_url = "products.subtopics"
```
escattone, line 177: When the topic has a parent, that path assumes that a `product_slug` has been provided. Should we `assert` that?

akatsoulas: I moved all the block within the `if product_slug` clause.

PR #6163, opened 2024-08-05
Switch topic queries to m2m.

akatsoulas:


kitsune/products/api.py
```diff
@@ -115,7 +111,7 @@ class TopicList(LocaleNegotiationMixin, generics.ListAPIView):
     serializer_class = RootTopicSerializer
 
     def get_queryset(self):
-        queryset = self.queryset.filter(product__slug=self.kwargs["product"])
+        queryset = self.queryset.filter(products__slug=self.kwargs["product"], is_archived=False)
```
escattone, line 114: Nit. I don't think the `is_archived=False` is needed because we're using the `active` manager above when defining the `queryset`?

kitsune/products/tests/test_api.py
```diff
@@ -28,8 +28,8 @@ def test_it_works(self):
 
     def test_expected_output(self):
         p = ProductFactory()
-        t1 = TopicFactory(product=p, visible=True, display_order=1)
-        t2 = TopicFactory(product=p, visible=True, display_order=2)
+        t1 = TopicFactory(products=[p], visible=True, is_archived=False, display_order=1)
+        t2 = TopicFactory(products=[p], visible=True, is_archived=False, display_order=2)
```
escattone, line 32: Super nit. You could drop the `is_archived=False` since that's the default.

escattone (approved):
r+wc

Nice! Thanks @akatsoulas!

PR #6166, opened 2024-08-05
Drop product FK from Topic

akatsoulas:


escattone (approved):

escattone (commented):
Oops, I missed a couple of things in my first review:

- The `Topic.documents` method uses `self.product` (https://github.com/mozilla/kitsune/blob/87448ca01c2388060fa532848472effce594e7ad/kitsune/products/models.py#L159). I'm thinking that line should be `"products__in": self.products,` instead?
- The `Topic.__str__` method also uses `self.product` (https://github.com/mozilla/kitsune/blob/87448ca01c2388060fa532848472effce594e7ad/kitsune/products/models.py#L133). I'm thinking lines 133-134 can be removed?

PR #6178, opened 2024-08-08
improve differentiation of read/unread inbox messages

escattone:
mozilla/sumo#1929

This PR changes the way read and unread inbox messages are distinguished. With this PR, unread inbox messages are given a bold font, while read inbox messages get a normal font. This is a common approach used, for example, by most email clients, like Gmail.

image


cajinajr:
This PR has the UX seal of approval 👍 🏆

akatsoulas (approved):

PR #6184, opened 2024-08-16
Return old taxonomy topics.

akatsoulas:


kitsune/products/models.py
```diff
@@ -151,6 +151,21 @@ def documents(self, user=None, **kwargs):
...
+        """Return topics that are not migrated to the new Taxonomy but have active documents."""
+        migrated_topic_slugs = [
+            item
+            for topics in TopicSlugHistory.objects.all().values_list("slug", "topic__slug")
+            for item in topics
+        ]
+
+        return (
+            Topic.active.exclude(slug__in=migrated_topic_slugs)
+            .annotate(doc_count=models.Count("document"))
```
escattone, line 165: Should we filter the `models.Count` to avoid counting archived documents?
```suggestion
            .annotate(doc_count=models.Count("document", filter=Q(document__is_archived=False))
```

PR #6185, opened 2024-08-16
Use django-silk for profiling

akatsoulas:


akatsoulas:
approved in person by @smithellis

PR #6186, opened 2024-08-17
new relevant products/topics selection 

escattone:
mozilla/sumo#1945

image

image

## Notes
- Since  the `Set.intersection` method has only been available since June 2024, and @babel doesn't yet provide a polyfill for it, I'm using a custom polyfill (the `intersection` function) that works within older browser versions. We can switch to `Set.intersection` at some later date.
- If we decide to allow contributor topics that aren't associated with any product or products, we'll have to adjust the logic that enables/disables products/topics based on current selections, since that logic assumes that every topic is associated with one or more products, and vice versa.


akatsoulas (approved):
This looks great! I opened #6221 which adds a couple of small changes in the topic widget

PR #6216, opened 2024-09-06
Wagtail - block templates

smithellis:
- cta_block template
- featured_article_block - template for single article block
- featured_articles_block - template for all featured articles
- frequent_topics_block - template for all frequent topics for product
- product_snippet_block - allows viewing and selecting of product models once registered as snippets
- search_block - search block with parent page product context


kitsune/products/jinja2/products/blocks/cta_block.html
```diff
@@ -0,0 +1,12 @@
+{% if value.type=='Community' %}
```
akatsoulas, line 1: Is this check needed since we don't have another type at the moment?

kitsune/products/jinja2/products/blocks/featured_articles_block.html
```diff
@@ -0,0 +1,18 @@
+  {% if value.featured %}
```
akatsoulas, line 1: similar comment as above, is this check needed?

kitsune/products/jinja2/products/blocks/frequent_topics_block.html
```diff
@@ -0,0 +1,55 @@
+<div class="mzp-l-content sumo-page-section--inner">
```
akatsoulas, line 1: this needs indentation. 2 spaces for html

akatsoulas (commented):
There's a lot of repetition with the Django templates. Can we make this a bit more DRY?

kitsune/products/jinja2/products/blocks/featured_articles_block.html
smithellis: Dropped.

kitsune/products/jinja2/products/blocks/frequent_topics_block.html
smithellis: I'm assuming you mean the whole block, so I did that.

kitsune/products/jinja2/products/blocks/cta_block.html
smithellis: Dropped

akatsoulas (approved):
Before merging, I would check whether we can use base templates to extract the common code for the Django and WT versions. This way it's going to be much clearer what changes and we will avoid mistakes and regressions. The `search_block` is something that could benefit for this approach.

The templates need indentation and please squash the commits before merging. 
r+wc

akatsoulas:
Closing per [comment](https://github.com/mozilla/sumo/issues/1966#issuecomment-2410512887)

PR #6219, opened 2024-09-06
Wagtail settings changes to support RichTextArea

smithellis:
Flesh out the WAGTAILADMIN_RICH_TEXT_EDITORS setting to have a sane default for RichTextAreas

### DIFF
--- a/kitsune/settings.py
+++ b/kitsune/settings.py
@@ -1361,4 +1361,12 @@ WAGTAILADMIN_PERMITTED_LANGUAGES = [
 WAGTAIL_SITE_NAME = config("WAGTAIL_SITE_NAME", default="Mozilla Support CMS")
 WAGTAILADMIN_BASE_URL = config("WAGTAILADMIN_BASE_URL", default="")
 WAGTAILIMAGES_MAX_UPLOAD_SIZE = IMAGE_MAX_FILESIZE
+WAGTAILADMIN_RICH_TEXT_EDITORS = {
+    "default": {
+        "WIDGET": "wagtail.admin.rich_text.DraftailRichTextArea",
+        "OPTIONS": {
+            "features": ["h2", "bold", "italic", "link", "document-link", "image", "embed", "code"]
+        },
+    },
+}
 WAGTAILDOCS_DOCUMENT_MODEL = "sumo.WagtailDocument"


kitsune/settings.py
```diff
@@ -1361,4 +1361,12 @@ def filter_exceptions(event, hint):
 WAGTAIL_SITE_NAME = config("WAGTAIL_SITE_NAME", default="Mozilla Support CMS")
 WAGTAILADMIN_BASE_URL = config("WAGTAILADMIN_BASE_URL", default="")
 WAGTAILIMAGES_MAX_UPLOAD_SIZE = IMAGE_MAX_FILESIZE
+WAGTAILADMIN_RICH_TEXT_EDITORS = {
+    "default": {
```
akatsoulas, line 1365: should we use a different editor here like `sumo` in order to avoid overriding the default?

kitsune/settings.py
```diff
@@ -1361,4 +1361,12 @@ def filter_exceptions(event, hint):
 WAGTAIL_SITE_NAME = config("WAGTAIL_SITE_NAME", default="Mozilla Support CMS")
 WAGTAILADMIN_BASE_URL = config("WAGTAILADMIN_BASE_URL", default="")
 WAGTAILIMAGES_MAX_UPLOAD_SIZE = IMAGE_MAX_FILESIZE
+WAGTAILADMIN_RICH_TEXT_EDITORS = {
```
akatsoulas, line 1364: Why are we overriding this? To add the `code` feature? Where do we need it?

kitsune/settings.py
smithellis: Ah, I was looking at this like defining a sane default. It looks like the out-of-the-box defaults would include:
```

h2, h3, h4 - heading elements
bold, italic - bold / italic text
ol, ul - ordered / unordered lists
hr - horizontal rules
link - page, external and email links
document-link - links to documents
image - embedded images
embed - embedded media 
```

Which is actually pretty ok for us. So. I'll alter or remove this.

kitsune/settings.py
smithellis: See above.

akatsoulas:
I am closing this PR per the reply in the comment. Wagtail defaults are OK at least for now

PR #6223, opened 2024-09-12
Remove waffle serving URL path

smithellis:
This early pattern is no longer needed.

### DIFF
--- a/kitsune/urls.py
+++ b/kitsune/urls.py
@@ -4,12 +4,9 @@ from django.views.decorators.csrf import csrf_exempt
 from django.views.generic.base import RedirectView
 from django.views.static import serve as servestatic
 from graphene_django.views import GraphQLView
-from waffle.decorators import waffle_flag
 from waffle.views import wafflejs
 from wagtail.admin.urls import urlpatterns as wagtail_admin_urlpatterns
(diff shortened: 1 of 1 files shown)


akatsoulas:
Why isn't this needed?

smithellis:
I'm not sure why it was ever there - I think the idea early on was to test some WT stuff at /wagtail, but we have not been doing that. When I asked @escattone about it a while back he agreed that it wasn't really serving a purpose right now.

akatsoulas:
From what I recall, it was an extra mechanism to control whether WT is enabled in production. If this is the case, I would argue to keep it around until we are ready to go live

smithellis:
Just to note my thoughts here - we aren't using this right now, so I'd argue it doesn't need to be here. It only allows us to serve things at `/wagtail` and then only when a waffle flag is set. As far as I know, we are not planning to serve anything at `/wagtail` at the moment, nor do we use the waffle flag to control wagtail on/off. The login and/or CMS admin is not at `/wagtail` nor are any pages we are serving.
In another PR I have a more general setting `WAGTAIL_ENABLE` which does this. 
I don't think it makes sense to have unused code (with no known future use) hanging around.

akatsoulas (approved):

PR #6230, opened 2024-09-18
Add back wagtail_urls to the urls.py

smithellis:
- This is important for the localization work
- Regardless, the addition is recommended by WT and is the common practice
- I think being the final i18n item is the right place, as we want wt pages processed with locales

https://docs.wagtail.org/en/stable/advanced_topics/i18n.html#adding-a-language-prefix-to-urls


smithellis:
I need to make this production friendly.

akatsoulas:
Let me know when you want a review on this

akatsoulas (approved):

PR #6234, opened 2024-09-20
Moderate topics for forum questions

akatsoulas:


kitsune/flagit/jinja2/flagit/includes/flagged_question.html
```diff
@@ -1,27 +1,44 @@
...
+
+  {% if object.content_object.updated_by %}
+    <h3 class="flagged-content__subheading">{{ _('Updated:') }}</h3>
+    <p>{{ date_by_user(object.content_object.updated, object.content_object.updated_by) }}</p>
+  {% endif %}
+
+  <h3 class="flagged-content__subheading">{{ _('Flagged:') }}</h3>
+  <p>{{ date_by_user(object.created, object.creator) }}</p>
+
+  <h3 class="flagged-content__subheading">{{ _('Take Action:') }}</h3>
```
escattone, line 20: I think the `View`, `Edit`, and `Delete` buttons belong under the `Take Action:` header right? If so, I think they should be moved within the `div.flagged-content` element. Otherwise, it looks off to me (see below:

image

escattone (changes requested):
In addition to the comments below, I noticed that the look of a flagged answer is different than the look of a flagged question:

- Should the lighter gray area be limited to just the `Content:` section like it is for flagged answers?
- I noticed that flagged answers have large empty gaps at the top and bottom.

### Question

image

### Answer

image

kitsune/flagit/jinja2/flagit/includes/flagged_question.html
akatsoulas: Good catch, in this case we need to remove the Take Action section completely. We discussed this with Josh and the decision was to have the 3 buttons in the outer box.

akatsoulas:
@escattone PR updated. I added a forth commit to fix a bug in the flag_form. It was using the same ID of "report-abuse" for every entry in the page. As a result, JS was picking the first occurrence and was assigning the wrong flags

kitsune/flagit/jinja2/flagit/includes/flagged_answer.html
```diff
@@ -1,27 +1 @@
...
-<div class="actions sumo-button-wrap">
-  <a class="sumo-button button-sm" href="{{ object.content_object.get_absolute_url() }}">{{ _('View') }}</a>
-  {% if user.has_perm('questions.change_answer') %}
-    <a class="sumo-button button-sm edit" href="{{ url('questions.edit_answer', object.content_object.question.id, object.content_object.id) }}">{{ _('Edit') }}</a>
-  {% endif %}
-  {% if user.has_perm('questions.delete_question') %}
-    <a class="sumo-button button-sm delete" href="{{ url('questions.delete_answer', object.content_object.question.id, object.content_object.id) }}">{{ _('Delete') }}</a>
-  {% endif %}
-</div>
+{% include 'flagit/includes/flagged_question.html' %}
```
escattone, line 1: ❤️

kitsune/flagit/jinja2/flagit/includes/flagged_question.html
```diff
@@ -1,27 +1,55 @@
...
-</p>
-<h3 class="sumo-page-intro">{{ _('Take Action:') }}</h3>
-<div class="actions sumo-button-wrap">
-  <a class="sumo-button button-sm" href="{{ object.content_object.get_absolute_url() }}">{{ _('View') }}</a>
-  <a class="sumo-button button-sm edit" href="{{ url('questions.edit_question', object.content_object.id) }}">{{ _('Edit') }}</a>
-  {% if user.has_perm('questions.delete_question') %}
-    <a class="sumo-button button-sm delete" href="{{ url('questions.delete', object.content_object.id) }}">{{ _('Delete') }}</a>
+<div class="flagged-content">
+  <h3 class="flagged-content__heading">{{ _('Title:') }}</h3>
+  <p class="flagged-content__title">{{ object.content_object.title }}</p>
```
escattone, line 4: We shouldn't show the title for answers, right? I wonder if we could simply wrap this with a `{% if object.content_object.title %}...{% endif %}`?

image

escattone (commented):
Nit. When showing the `Take Action:` section for flagged questions, which includes the ability to change the topic, it looks to me that there's a bit more space than usual between the `Take Action:` header and the `Current topic:` header immediately below it.

image

escattone (approved):

PR #6240, opened 2024-09-23
Validate topics in the clean method

akatsoulas:


kitsune/wiki/forms.py
```diff
@@ -158,20 +153,34 @@ def clean_slug(self):
...
-        c = super(DocumentForm, self).clean()
-        locale = c.get("locale")
+        cdata = super(DocumentForm, self).clean()
+        locale = cdata.get("locale")
 
         # Products are required for en-US
-        products = c.get("products")
-        if locale == settings.WIKI_DEFAULT_LANGUAGE and (not products or len(products) < 1):
+        product_ids = cdata.get("products")
+        products = Product.objects.filter(id__in=[int(product_id) for product_id in product_ids])
```
escattone, line 161: This needs to take into account that `product_ids` can be `None`, and I think we should use `Product.active.filter(...)` here?

kitsune/wiki/forms.py
```diff
@@ -158,20 +153,34 @@ def clean_slug(self):
...
+        if locale == settings.WIKI_DEFAULT_LANGUAGE and (not product_ids or len(product_ids) < 1):
             raise forms.ValidationError(PRODUCT_REQUIRED)
 
         # Topics are required for en-US
-        topics = c.get("topics")
-        if locale == settings.WIKI_DEFAULT_LANGUAGE and (not topics or len(topics) < 1):
+        topic_ids = cdata.get("topics")
+        topics = Topic.objects.filter(id__in=[int(topic_id) for topic_id in topic_ids])
```
escattone, line 167: Same here for `topic_ids`. It needs to take into account that `topic_ids` can be `None`, and I think we should use `Topic.active.filter(...)` here as well?

kitsune/wiki/forms.py
```diff
@@ -158,20 +153,34 @@ def clean_slug(self):
...
+        if locale == settings.WIKI_DEFAULT_LANGUAGE and (not topic_ids or len(topic_ids) < 1):
             raise forms.ValidationError(TOPIC_REQUIRED)
 
-        return c
+        invalid_topics = []
+        for topic in topics:
+            associated_products = topic.products.all()
+            if not set(products).intersection(associated_products):
```
escattone, line 174: I think `associated_products` has to be a set here?

escattone (changes requested):
Thanks for this @akatsoulas ! I should have done this as part of https://github.com/mozilla/kitsune/pull/6186.

In addition to the comments below -- which when addressed resolve at least one of the unit test failures -- I think one or two of the other unit test failures still need to be addressed as well.

kitsune/wiki/forms.py
akatsoulas: Good catch. I removed the `set()` completely in order to use Django's `intersection()` method since we are operating on QuerySets.

kitsune/wiki/forms.py
```diff
@@ -158,20 +153,33 @@ def clean_slug(self):
...
-        topics = c.get("topics")
-        if locale == settings.WIKI_DEFAULT_LANGUAGE and (not topics or len(topics) < 1):
+        topic_ids = cdata.get("topics", [])
+        topics = Topic.active.filter(id__in=[int(topic_id) for topic_id in topic_ids if topic_id])
+        if locale == settings.WIKI_DEFAULT_LANGUAGE and (not topic_ids or len(topic_ids) < 1):
             raise forms.ValidationError(TOPIC_REQUIRED)
 
-        return c
+        associated_topics = Topic.active.filter(products__in=products).distinct()
+        invalid_topics = topics.difference(associated_topics)
```
escattone, line 174: ❤️ Nice! I've overlooked this feature, so thanks for enlightening me!

escattone (approved):
Thanks @akatsoulas!

PR #6242, opened 2024-09-23
Playwright refactor flows & classes belonging to AAQ

emilghittasv:
- Mark the aaq_form_page, questions_page, ways_to_contribute_pages page classes functions as public and replace super() with self.
- Use composition instead of inheritance for the aaq_flow.py class.
- Updating modified function calls in tests.


escattone (approved):
Nice changes!

PR #6247, opened 2024-09-23
change the location for the statement setup-buildx-actions

Basma1912:


.github/workflows/build-and-push.yml
```diff
@@ -67,7 +66,7 @@ jobs:
 
           echo TAG ="dev-$(git describe --tags --abbrev=7)" >> "$GITHUB_ENV"
 
-          # Updates to the main branch are deployed to stage.
+          # Updates to the main branch are deployed to stage.git
```
bkochendorfer, line 69: Why'd you add the `.git` here?

bkochendorfer (approved):

.github/workflows/build-and-push.yml
Basma1912: that must be a typo

PR #6253, opened 2024-09-25
Validate both topics and products

akatsoulas:


escattone (approved):
❤️ Really nicely done!

PR #6275, opened 2024-10-02
Update `prefer_cms` decorator and `remove_locale` func

smithellis:
* `prefer_cms` will now check if WT page exists in the requested locale instead of just at base path
* `remove_locale` now returns the path and the locale_code

### DIFF
--- a/kitsune/sumo/decorators.py
+++ b/kitsune/sumo/decorators.py
@@ -10,6 +10,7 @@ from django.http import Http404
 
 from kitsune.sumo.utils import is_ratelimited
 
+from wagtail.models.i18n import Locale
 from wagtail.views import serve as wagtail_serve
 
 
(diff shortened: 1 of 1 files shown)


kitsune/sumo/decorators.py
```diff
@@ -162,21 +164,42 @@ def wrapped(*args, **kwargs):
...
 
 def prefer_cms(view_func):
     @wraps(view_func)
     def _wrapped_view(request, *args, **kwargs):
-        path = remove_locale(request.path_info)
+        path, locale_code = remove_locale(request.path_info)
+
+        # Set the language code in the request and activate the translation
+        request.LANGUAGE_CODE = locale_code
+        translation.activate(locale_code)
```
escattone, line 186: I don't think you need these lines, since our [LocaleMiddleware](https://github.com/mozilla/kitsune/blob/e08d26f04028fdbbb24340a3eda62cd64ec40e6c/kitsune/sumo/middleware.py#L131) will take care of that before the request reaches the view function.

akatsoulas:
Per our discussion last week in the team meeting this is no longer relevant. Closing

PR #6276, opened 2024-10-03
Management command (create_wt_locales) that:

smithellis:
* Creates and enables Locales in wt admin
* Synchronizes all new locale trees with en-us tree


akatsoulas:
Per our discussion last week in the team meeting this is no longer relevant. Closing

PR #6280, opened 2024-10-08
Default moderation queue filtering to spam

akatsoulas:


escattone (changes requested):
With this change, the `/{locale}/flagged` page doesn't work when selecting `All reasons` for the reason filter. In that case, the JS does a fetch with no reason added to the URL (since the value for the `All reasons` case is the empty string) and since the view function defaults to a reason of "spam" in that case, you only see the spam moderation items instead of all moderation items.

I'm thinking we'll either have to make the "spam" reason have a value of the empty string as well as give the `All reasons` selection a value of "all" or something like that, or else maybe we have to give the `All reasons` selection a value of "all" and keep the value of "spam"  as well?

escattone (commented):
Something is still off. When I load the page initially, it shows spam only, but when I then change the reason filter to "All reasons", nothing changes. The reason remains "spam".

kitsune/sumo/static/sumo/js/flagit.js
```diff
@@ -1,15 +1,16 @@
 document.addEventListener('DOMContentLoaded', () => {
-    const { reasonFilter, flaggedQueue, topicDropdown, updateStatusButton } = {
+    let isInitialLoad = true;
```
escattone, line 2: When I remove this line and its related code below, this PR works great.

kitsune/sumo/static/sumo/js/flagit.js
```diff
@@ -50,25 +62,27 @@ document.addEventListener('DOMContentLoaded', () => {
...
-            const data = await response.text();
-            const parser = new DOMParser();
-            const doc = parser.parseFromString(data, 'text/html');
-            flaggedQueue.innerHTML = doc.querySelector('#flagged-queue').innerHTML;
-            disableUpdateStatusButtons();
-            handleDropdownChange();
+        if (isInitialLoad) {
+            isInitialLoad = false;
+            return;
         }
```
escattone, line 82: When I remove this code related to `isInitialLoad`, this PR works great.

escattone (approved):
r+wc 

I don't think the `isInitialLoad`-related code is needed. When I remove it, this PR works great for me.

kitsune/sumo/static/sumo/js/flagit.js
Jeff9049: I want to see if I can get some help with this money by simply replying to get that from my account details for the use of the 6 different wallets I've been thinking about what I have been in the future for the use of this amount of money for the payment of money for delivery to my bank account details for my self and others who are interested in the bank account details for my self bank of Queensland BSB 124047 acc22513293

kitsune/sumo/static/sumo/js/flagit.js
Jeff9049: > When I remove this code related to `isInitialLoad`, this PR works great.

I'll try to get that money for delivery on all my details for the payment prize money for delivery on all my life and I have been looking for a few more wallets to get the money that is in the future for some time to send it to me and I will be able to send it to my bank account details for the payment of money for delivery on all my details for the crypto currency interest rate payment method cash on delivery of this winnings for selling stock will be able to get the latest product of paid vacation to get that money for delivery on all my account details from the play store and different wallets I've attached my resume to this video is not forgetton to find out more information about the money to pay the balance in my bank account details

PR #6282, opened 2024-10-09
Normalize locale in the API

akatsoulas:


escattone (approved):
Good catch, thanks @akatsoulas!

PR #6289, opened 2024-10-11
Display spam status in the all questions

akatsoulas:


escattone (approved):
Nice, thanks @akatsoulas!

PR #6291, opened 2024-10-14
Remove wagtail from the code-base

akatsoulas:
This PR removes almost all of Wagtail references from the code-base and drops the database tables. Once this is merged and deployed we can remove the remaining references from the settings.py and the pyproject.toml files. Since our database has no wagtail data at the moment, it makes sense to completely remove Wagtail and revert this PR when we are ready to resume work.


escattone (approved):
This looks great, thanks @akatsoulas!

PR #6295, opened 2024-10-16
Squash Wagtail migrations

akatsoulas:


escattone (commented):
This ran without error for me locally, but I see that the DB migration is failing when run in CircleCI at the start of the unit tests. When I look at the squashed migration for the `products` app, I see that it's including some intermediate steps in the operations that, in the end, are removed. For example, for the `Topic` model, the `product` field is created, altered, and then finally removed -- likewise for the `unique_together` constraint and the `ordering` for that model. Perhaps we should manually adjust the operations to only include the final values for each model?

escattone (approved):
Good idea on the fix!

PR #6323, opened 2024-10-30
Replace selectisize tom select

akatsoulas:


escattone (commented):
❤️ it! Nice find! It looks great except for one thing. For me,  the `x` of the clear button isn't visible (see below):

image

It looks to me like the SCSS for the clear and remove button plugins is being added to the `screen.scss`, so I'm confused why it isn't visible.

kitsune/sumo/static/sumo/js/wiki.editor.js
```diff
@@ -173,34 +162,49 @@ function clearSelected(name) {
...
-  ).addEventListener("click", function(event) {
+  ).addEventListener("click", function (event) {
     event.preventDefault();
     clearSelected("products");
   });
+
+  new TomSelect("select[id='id_restrict_to_groups']", {
+    closeAfterSelect: true,
+    plugins: {
+      clear_button: {
```
akatsoulas, line 199: @escattone `tom-select` doesn't have an `x` for the clear button but rather a checkbox. 
See https://tom-select.js.org/plugins/clear-button/

We could override this but unless there are strong objections I would prefer to tackle this in a different PR/issue to avoid scope creep. Thoughts?

escattone: Yes, let's tackle that in another PR/issue!

However, I do see a small `x` both in our HTML and on their website.
image

escattone (approved):
r+wc

kitsune/sumo/static/sumo/js/wiki.editor.js
akatsoulas: that's for Bootstrap 4/5. If you select the `Default` from the drop down it's the checkbox..

emilghittasv:
> ❤️ it! Nice find! It looks great except for one thing. For me, the `x` of the clear button isn't visible (see below):
> image
> 
> It looks to me like the SCSS for the clear and remove button plugins is being added to the `screen.scss`, so I'm confused why it isn't visible.

I've also encountered this issue during testing in `stage`. Filled under https://github.com/mozilla/sumo/issues/2029

PR #6330, opened 2024-11-01
add kitsune.l10n app for handling content localization

escattone:


akatsoulas:
We have pivoted from the initial implementation as defined here but parts of this work have been already integrated. Great work nevertheless.

PR #6341, opened 2024-11-07
[2018] Stop 500 error on page changes

smithellis:
Updated wiki.js to retain form data between page changes
Also re-updates the revision list if there is data 
Stops the 500 error


kitsune/sumo/static/sumo/js/wiki.js
```diff
@@ -627,99 +627,87 @@ import collapsibleAccordionInit from "sumo/js/protocol-details-init";
...
-      history.replaceState({}, '', url);
-      $('#revisions-fragment').css('opacity', 0);
-
-      $.get(url + '&fragment=1', function (data) {
-        $('.loading').hide();
-        $('#revisions-fragment').html(data).css('opacity', 1);
-      });
+    // Retrieve and safely parse form data from localStorage
+    let formData = {};
+    const savedFormData = localStorage.getItem('revision-list-filter');
```
akatsoulas, line 637: Probably a better option for the scenario at hand is the `sessionStorage`. Using `sessionStorage` will ensure that we won't have any stale data and we won't also exceed the capacity.

kitsune/sumo/static/sumo/js/wiki.js
```diff
@@ -627,99 +627,87 @@ import collapsibleAccordionInit from "sumo/js/protocol-details-init";
...
-    $form.find('.btn, .button').remove();
+    // Disable form submission via Enter key
     $form.on('keydown', function (e) {
-      // 13 is enter.
-      if (e.which === 13) {
-        return false;
-      }
+        if (e.which === 13) {
+            e.preventDefault();
+            return false;
```
akatsoulas, line 706: you can skip the return statement here

akatsoulas (approved):
Overall let's remove the debugging statements from the code. r+wc

PR #6342, opened 2024-11-08
2021 - 500 Error after submission with no file

smithellis:
* Check for file, if none, return error message

### DIFF
--- a/kitsune/groups/forms.py
+++ b/kitsune/groups/forms.py
@@ -29,18 +29,24 @@ class GroupAvatarForm(forms.ModelForm):
             size=settings.AVATAR_SIZE
         )
 
-    class Meta(object):
+    class Meta:
         model = GroupProfile
         fields = ["avatar"]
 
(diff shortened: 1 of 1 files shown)


DevBeatDad:
2021
500
6342

Sent from Proton Mail Android

-------- Original Message --------

kitsune/groups/views.py
```diff
@@ -92,6 +92,11 @@ def edit_avatar(request, group_slug):
         old_avatar_path = prof.avatar.name
 
     if request.method == "POST" and form.is_valid():
+        # If no avatar was uploaded, let's error out.
+        if "avatar" not in request.FILES:
```
akatsoulas, line 96: This should be moved to the clean method of the form. This way the `form.is_valid()` check will be erroneous and we won' have to add the next couple of lines.

akatsoulas (changes requested):

akatsoulas (approved):

PR #6351, opened 2024-11-12
2019 - Update 'MultiUsernameField' to handle display names

smithellis:


kitsune/sumo/form_fields.py
```diff
@@ -71,9 +72,13 @@ def to_python(self, value):
             if username:
                 try:
                     user = User.objects.get(username=username)
-                    users.append(user)
                 except User.DoesNotExist:
-                    msg = _("{username} is not a valid username.")
+                    try:
```
akatsoulas, line 76: In order to avoid the nesting of the `try/except` clause you could use something like:
```python
user = User.objects.filter(Q(username=username) | Q(profile__name=username).first()
```

akatsoulas (approved):
Looks good! Please squash the 2 commits before merging.

PR #6354, opened 2024-11-13
Extract moderation tool to a separate url

akatsoulas:


kitsune/questions/models.py
```diff
@@ -209,7 +209,7 @@ def save(self, update=False, *args, **kwargs):
                     object_id=self.id,
                     creator=self.creator,
```
escattone, line 210: Nit. It seems odd in this case, that the creator of the flagged object is the creator of the question. Maybe we should consider creating a new bot user that we use as the creator of these auto-flagged objects?

kitsune/flagit/views.py
```diff
@@ -97,6 +104,35 @@ def flagged_queue(request):
     )
 
 
+@login_required
+@permission_required("flagit.can_moderate")
+def moderate_content(request):
+    """Display flagged content that needs moderation."""
```
escattone, line 110: Maybe you're planning this in a subsequent PR, but I think we should either create a new permission for topic moderation that we give to users who understand the taxonomy, or else add another check here that the user is a member of a specific group that we define for that purpose.

escattone (approved):
r+wc

According to Abby's PRD, once the content-moderation flagged object has been moderated, it should be locked so that users are no longer able to add/delete topic tags. Is that something planned in a separate PR as well?

akatsoulas:
Thanks @escattone! That PR was finalized after the submission of the PR so I will be following with another once I go through the PRD thoroughly.

kitsune/flagit/views.py
akatsoulas: I am going to combine this with the locking PR

akatsoulas:
Thanks @escattone for the review!

PR #6358, opened 2024-11-15
Custom Tag model and Zendesk segmentation tags.

akatsoulas:


escattone (approved):
Wow, this PR kind of blew my mind, nice work! Thanks @akatsoulas!

PR #6361, opened 2024-11-20
Improve user lookup form field query

smithellis:
* Remove from loop
* Get all responses, not just first

Attempting to correct Stage issues with https://github.com/mozilla/sumo/issues/2019 - I think the issue is related related to speed. Improving this query might resolve.

### DIFF
--- a/kitsune/sumo/form_fields.py
+++ b/kitsune/sumo/form_fields.py
@@ -55,9 +55,11 @@ class TypedMultipleChoiceField(forms.MultipleChoiceField):
 
 
 class MultiUsernameField(forms.Field):
-    """Form field that takes a comma-separated list of usernames OR profile
+    """
+    Form field that takes a comma-separated list of usernames OR profile
     names (display names) as input, validates that users exist for each one,
-    and returns the list of users."""
+    and returns the list of users.
+    """
 
     def to_python(self, value):
         if not value:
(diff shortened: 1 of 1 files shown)


kitsune/sumo/form_fields.py
```diff
@@ -67,15 +67,14 @@ def to_python(self, value):
...
-        for username in value.split(","):
-            username = username.strip()
-            if username:
-                user = User.objects.filter(
-                    Q(username=username) | Q(profile__name=username)
-                ).first()
-                if user:
-                    if user.is_active:
-                        users.append(user)
+        usernames = [u.strip() for u in value.split(",") if u.strip()]
```
akatsoulas, line 70: Good idea to remove the `for` loop but isn't his already a list of usernames? Additionally why is `if u.strip()` needed?

smithellis: Lordy. That last u.strip is not needed.

smithellis: It isn't a list of usernames originally - it's whatever the value of the form field is as the user types. So the flow is get that value, convert to a clean list of strings, get a queryset of any User instance that matches on username or profile.name, validate that the user is active then return the list of User objects.

akatsoulas: I mean why are you validating the u.strip. This should be 
```python
usernames = [u.strip() for u in values.split() if u]
```

akatsoulas: I would also go for something more descriptive that `u`

akatsoulas (approved):
r+wc

kitsune/sumo/form_fields.py
smithellis: Oh - got it. Thanks!

PR #6364, opened 2024-11-20
Do not edit moderated questions

akatsoulas:


kitsune/questions/models.py
```diff
@@ -392,6 +392,14 @@ def is_solved(self):
     def is_offtopic(self):
         return config.OFFTOPIC_TAG_NAME in [t.name for t in self.my_tags]
 
+    @property
```
escattone, line 395: Nit (since it's only used once so far). Maybe use `@cached_property` (from `functools`) instead, since this performs a DB query?

escattone (approved):
r+wc

kitsune/questions/models.py
akatsoulas: great idea!

PR #6366, opened 2024-11-20
Extend debounce and improve query

smithellis:
Trying to resolve inconsistencies on staging that are likely caused by DB calls and timing issues, and/or broken user look-ups.

This code:
* Makes the JS call to fetch the revision data wait a little longer between polls; move debounce from 200 to 500
* Further optimizes query and code

### DIFF
--- a/kitsune/sumo/form_fields.py
+++ b/kitsune/sumo/form_fields.py
@@ -68,17 +68,13 @@ class MultiUsernameField(forms.Field):
             else:
                 return []
 
-        users = []
-        usernames = [name.strip() for name in value.split(",") if name]
+        usernames = [name.strip() for name in value.split(",") if name.strip()]
         if usernames:
             all_users = User.objects.filter(
-                Q(username__in=usernames) | Q(profile__name__in=usernames)
-            )
-            for user in all_users:
-                if user and user.is_active:
-                    users.append(user)
+                Q(profile__name__in=usernames) | Q(username__in=usernames), is_active=True
+            ).distinct()
 
-        return users
+        return list(all_users)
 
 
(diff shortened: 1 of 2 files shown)


kitsune/sumo/form_fields.py
```diff
@@ -68,17 +68,13 @@ def to_python(self, value):
             else:
                 return []
 
-        users = []
-        usernames = [name.strip() for name in value.split(",") if name]
+        usernames = [name.strip() for name in value.split(",") if name.strip()]
```
akatsoulas, line 71: Why are you changing this? I don't see how the check on the strip() call will help.

kitsune/sumo/form_fields.py
```diff
@@ -68,17 +68,13 @@ def to_python(self, value):
...
-                Q(username__in=usernames) | Q(profile__name__in=usernames)
-            )
-            for user in all_users:
-                if user and user.is_active:
-                    users.append(user)
+                Q(profile__name__in=usernames) | Q(username__in=usernames), is_active=True
+            ).distinct()
 
-        return users
+        return list(all_users)
```
akatsoulas, line 77: I don't see how forcing the evaluation will help here.

akatsoulas (commented):
I am not sure how these changes will solve any problems in staging.

kitsune/sumo/form_fields.py
smithellis: When I test this in shell, if I pass something like `"Smith, smith, ,, smith1"` as the value, that comprehension will return a username of `" "` that then has to be looked up. I don't think it's vital either way, but why wouldn't I catch it here instead of letting through?

PR #6367, opened 2024-11-20
Add back updateRevisionList to ensure list populates with saved data

smithellis:


kitsune/sumo/static/sumo/js/wiki.js
```diff
@@ -653,6 +653,8 @@ import collapsibleAccordionInit from "sumo/js/protocol-details-init";
         }
     }
 
+    updateRevisionList(); // Initial update
```
akatsoulas, line 656: That was removed in a previous PR. Is this needed? If yes, where is it used. Staging seems to be working without this call.

smithellis: This call was used in my other PR that ensures the revision list pre-populates with sessionStorage data; so when you got back to the revision list, if there data, we want the revision list to populate vs. show the fields filled but have standard data. I removed this because I was thinking it may be part of the issue in with the revision list lookup having intermittent/strange failures, but realized that it isn't related and needs to go back.

akatsoulas (approved):

PR #6372, opened 2024-11-25
Update topic tag on moderation

akatsoulas:


escattone (commented):
When I update a question's topic via the moderation view, I still see the old topic tag when I view the question at `/en-US/questions/{question_id}`. It appears to be due to the redis-based caching of the question's `my_tags` property. Maybe clear that cache entry when the topic is updated, or else just cache the tags for the lifetime of the question instance by using `cached_property` instead?

escattone (approved):

PR #6373, opened 2024-11-25
Allow link posts by moderators and trusted users

smithellis:


akatsoulas (commented):
This PR will just allow moderators and trusted users to post a link. Is this the intention here? Reading the linked issue, it feels that we want to allow links for everyone

PR #6381, opened 2024-11-29
Replace custom widget with tom-select

akatsoulas:


kitsune/questions/views.py
```diff
@@ -455,6 +455,7 @@ def question_details(
         {
             "all_products": products,
             "all_topics": topics,
+            "all_tags": SumoTag.objects.filter(is_archived=False),
```
escattone, line 458: Would you prefer to convert this to `SumoTag.objects.active()` now that https://github.com/mozilla/kitsune/pull/6383 has merged?

escattone (approved):
❤️ Awesome, thanks @akatsoulas! Just the one comment.

PR #6388, opened 2024-12-03
Limit moderation access to specific group

akatsoulas:


kitsune/access/decorators.py
```diff
@@ -101,3 +100,28 @@ def deny_func(user):
...
+                raise Http404
+
+            if only_active:
+                if not (
+                    request.user.is_active and request.user.groups.filter(name=group_name).exists()
+                ):
+                    raise Http404
+            else:
+                if not request.user.groups.filter(name=group_name).exists():
+                    raise Http404
```
escattone, line 121: Nit. What do you think of the following instead?
```
            if not (
                (!only_active or request.user.is_active) and request.user.groups.filter(name=group_name).exists()
            ):
                raise Http404
```

escattone (approved):
Nicely done, thanks @akatsoulas! I just have a small comment.

PR #6390, opened 2024-12-04
Add moderation timestamp to Question

akatsoulas:


kitsune/questions/models.py
```diff
@@ -147,6 +147,7 @@ class Question(AAQBase):
     tags_cache_key = "question:tags:%s"
     images_cache_key = "question:images:%s"
     contributors_cache_key = "question:contributors:%s"
+    moderation_timestamp = models.DateTimeField(default=None, null=True)
```
escattone, line 150: Nit. Maybe use `moderated` instead of `moderation_timestamp` to match the style of the other date-time fields like `created` and `updated`?

escattone (approved):
r+wc

kitsune/questions/models.py
akatsoulas: yeah I spent some time on this. I had moderated but then I switched to more verbose naming schema which avoids confussion. When I am thinking of `moderated` I default to whether the question is moderated or not and I am not thinking timestamp.

PR #6404, opened 2024-12-10
Optimize topics hierarchy helper

akatsoulas:


escattone (commented):
This approach still has the problem of making N queries -- where N is the total number of questions in the moderation queue -- when no product is provided in the URL. I think the caching should be done based on the `question.product.slug` rather than the product slug provided in the URL. I'll add a commit that does that.

PR #6408, opened 2024-12-12
Improved topic cards now show three articles, with new design

smithellis:
* Alter the help_topics macro to support showing three articles
* Add show more section with sectional line above
* Add logic to the product_landing view to return docs and counts
* Styling

Note - right now this just shows three (the first three) documents in a topic. This should be updated to choose most popular or newest pending discussions with Content.

### DIFF
--- a/kitsune/products/views.py
+++ b/kitsune/products/views.py
@@ -59,13 +59,15 @@ def product_landing(request, slug: str) -> HttpResponse:
             latest_version = 0
     topics = topics_for(request.user, product=product, parent=None)
     # Create a dictionary of topics and three documents
-    # TODO: get popular documents instead of just the first three
     topics_with_urls = []
     for topic in topics:
         docs, _ = documents_for(
             request.user, request.LANGUAGE_CODE, topics=[topic], products=[product]
         )
         total_articles = len(docs)
+        # Leverage get_featured_articles to get active articles
+        docs = get_featured_articles(product, locale=request.LANGUAGE_CODE, topic=topic)
+
         # Create a dict with topic data including the URL
         topic_data = {
             "topic": topic,
--- a/kitsune/products/views.py
+++ b/kitsune/products/views.py
@@ -4,6 +4,7 @@ from django.conf import settings
 from django.db.models import Exists, OuterRef, Q
 from django.http import Http404, HttpResponse
 from django.shortcuts import get_object_or_404, redirect, render
+from django.urls import reverse
 from product_details import product_details
 
 from kitsune.products.models import Product, Topic, TopicSlugHistory
@@ -23,8 +24,19 @@ def product_list(request):
 
 
 @check_simple_wiki_locale
-def product_landing(request, slug):
-    """The product landing page."""
+def product_landing(request, slug: str) -> HttpResponse:
+    """The product landing page.
+
+    Args:
+        request: The HTTP request
+        slug: Product slug identifier
+
+    Returns:
+        Rendered product landing page
+
+    Raises:
+        Http404: If product not found
+    """
     if slug == "firefox-accounts":
         return redirect(product_landing, slug="mozilla-account", permanent=True)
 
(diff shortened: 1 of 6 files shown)


smithellis:
https://github.com/mozilla/sumo/issues/2042

kitsune/products/views.py
```diff
@@ -45,18 +57,39 @@ def product_landing(request, slug):
...
+            request.user, request.LANGUAGE_CODE, topics=[topic], products=[product]
+        )
+        total_articles = len(docs)
+        # Create a dict with topic data including the URL
+        topic_data = {
+            "topic": topic,
+            "topic_url": reverse("products.documents", args=[product.slug, topic.slug]),
+            "title": topic.title,
+            "total_articles": total_articles,
+            "image_url": getattr(topic, "image_url", ""),
```
escattone, line 76: This can be `topic.image_url` instead, since the `image_url` is a property that always returns either the topic image or the topic placeholder image.

kitsune/products/views.py
```diff
@@ -45,18 +57,39 @@ def product_landing(request, slug):
...
+        }
+        topics_with_urls.append(topic_data)
+        documents[topic] = docs[:3]
```
escattone, line 79: It seems that you could toss the list of documents into the `topic_data` dict, and then you wouldn't have to pass the `documents` dict into the template context nor the `help_topics()` macro.

kitsune/sumo/static/sumo/scss/layout/_products.scss
```diff
@@ -32,3 +32,21 @@
...
 }
+
+
+// SCSS for Topics Section Layout
+
+.topics-section {
+  padding: 40px 20px;
+  h2 {
+    font-size: 24px;
+    margin-bottom: 30px;
```
escattone, line 43: I would vote for this bottom margin to be zero. In my opinion, the current value introduces too much space below the `Topics` header:

image

escattone (commented):
Overall, this looks pretty good! I have some suggestions below, and also I'm thinking that the topics grid should collapse to a width of 1 or 2 cards for mobile devices in order to avoid the need for horizontal scrolling (but I'll defer to Josh for that).

kitsune/products/views.py
smithellis: Good catch. Thanks!

kitsune/products/views.py
smithellis: Thanks - totally obvious now.

kitsune/sumo/static/sumo/scss/layout/_products.scss
smithellis: Thanks - I think I agree.

escattone (changes requested):
I think this needs two more adjustments:
- I think the topics card grid should be responsive to screen size, so I think we should collapse to 2 columns or 1 column (depending on the device screen size) in order to avoid the need for horizontal scrolling on mobile and other smaller-screen devices.
- It seems to me that we should delete the `Frequent Topics` header, which is redundant with the `Topics` header (see image below).

image

smithellis:
@escattone I spoke with Josh and we are in agreement about simplifying the headers. The scrolling is an issue that I think we have a case for elsewhere - not just for this use case but for other cases that use cards. The idea of stacking the cards makes the horizontal scroll too much, but we definitely could improve the other scrolling options.

escattone (approved):
Looks good, thanks @smithellis!

kitsune/products/views.py
```diff
@@ -45,14 +57,33 @@ def product_landing(request, slug):
             latest_version = versions[0].min_version
         else:
             latest_version = 0
+    topics = topics_for(request.user, product=product, parent=None)
+    # Create a dictionary of topics and three documents
+    # TODO: get popular documents instead of just the first three
```
akatsoulas, line 62: Can't we use the `get_featured_articles` from wiki/utils (with small changes) to get the most popular docs for a specific topic and product combo?

smithellis: Being super atomic, this was the look/feel until we had the details for exactly what to show from content. I don't mind adding the top docs as the default though.

smithellis: Done. I modified `get_featured_articles` to accept a topic.

akatsoulas (approved):

PR #6411, opened 2024-12-13
Use locale_picker_url in footer

janbrasna:
Link to `sumo.locales` instead of hardcoded _bedrock_ locale page.

Resolves https://github.com/mozilla/sumo/issues/2100


akatsoulas (approved):
Thanks @janbrasna!

PR #6413, opened 2024-12-15
Update mozorg footer links

janbrasna:
- removes hardcoded locales to leave that to content negotiation
- changes repetitive destination URLs with links to specific content


akatsoulas (approved):
Thanks @janbrasna! :rocket:

PR #6419, opened 2024-12-19
Exclude pagination from sessionStorage

akatsoulas:


escattone (approved):
Nice! The code looks good and works well. I have one comment to consider (see below).

PR #6420, opened 2024-12-19
fix update of GA4 pageview data for questions

escattone:
mozilla/sumo#1913

# Notes
- The `Reload Question Traffic Stats` weekly cron has been broken for over 5 months. This PR fixes it.
- There are two problems with the currently released code:
    - The `kitsune.sumo.googleanalytics.pageviews_by_question()` function returns multiple rows for each question ID -- one for each locale that a question was viewed in during the past 365 days -- and that means that more than 700K rows can be returned in `prod`. In this PR, it returns a `dict` with one entry for each question ID where the value of that entry is the total pageviews of the question across all locales, which means over 300K entries in `prod`.
    - The `QuestionVisits.reload_from_analytics()` method was causing an OOM error because it was building the `instance_by_question_id` `dict` for all questions at once. In this PR, that's now done in batches of 30K.
- A very minor variation of the code in this PR has already been successfully tested within the `prod` environment.

### DIFF
--- a/kitsune/questions/models.py
+++ b/kitsune/questions/models.py
@@ -758,29 +758,15 @@ class QuestionVisits(ModelBase):
         from kitsune.sumo import googleanalytics
 
         with transaction.atomic():
-            # First, let's gather some data we need. We're sacrificing memory
-            # here in order to reduce the number of database queries later on.
             if verbose:
                 log.info("Gathering pageviews per question from GA4 data API...")
 
-            instance_by_question_id = {}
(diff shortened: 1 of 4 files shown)


smithellis (commented):
This is nice! It has a few hard to read spots (your clever use of enumerate, for instance, felt a bit inception-like -- but I couldn't think of a better/easier to read way of doing that). 

I didn't run this locally (don't think it would work unless I mocked GA or something), but it looks good to me, and the tests work. Approved!

akatsoulas (approved):

PR #6423, opened 2024-12-20
Update KB helpful survey

akatsoulas:


escattone (changes requested):
- When the user is not authenticated but has already voted anonymously, the voting form is still displayed when the article is viewed again, and if another vote is attempted, the response from the vote POST is a redirect to the KB article, which causes a strange article-within-an-article view of the page.

image

- The font size and weight of the `Yes` and `No` of the voting form should match those of its `Was this article helpful?` text.
    
image

- When the user is authenticated, the voting form is pushed way down on the page.

image

- The `Cancel` and `Submit` buttons of both the "yes" and "no" surveys are stacked instead of side-by-side.

image

- When submitting the survey, the response is `Thanks for making us better!` instead of `Thanks for your feedback!` as the Figma indicates. Also, the text is not bold as is shown in Figma.

akatsoulas:
> The use of `htmx` and `alpine.js` is pretty cool!
> 
>     * The GA4 tracking events are not being generated (see comments in code below).
> 
>     * When the user is not authenticated but has already voted anonymously, the voting form is still displayed when the article is viewed again, and if another vote is attempted, the response from the vote POST is a redirect to the KB article, which causes a strange article-within-an-article view of the page.
> 
> 
> image
> 
>     * The font size and weight of the `Yes` and `No` of the voting form should match those of its `Was this article helpful?` text.
> 
> 
> image

I am not sure about this. We treat the `Was this article helpful` as a header. I will follow up with Josh on this.
> 
>     * When the user is authenticated, the voting form is pushed way down on the page.
> 
It's actually the same space with the logged out view. The difference is that the editing tools push down the whole section
> 
>     * The `Cancel` and `Submit` buttons of both the "yes" and "no" surveys are stacked instead of side-by-side.

Have you tried a fresh build? It seems to be working for me
image

escattone:
> It's actually the same space with the logged out view. The difference is that the editing tools push down the whole section

Ah, I see that now, thanks. Ignore that comment! 😄 

Regarding the cancel and submit buttons of the survey, I did have a fresh build, but they're still always stacked for me. However, I noticed that your survey is wider than mine, so I guess the width of my survey is what is causing the buttons to stack. Hm, I'm not sure why my survey has a smaller width. Anyway, ignore that one too.

image

akatsoulas:
> > It's actually the same space with the logged out view. The difference is that the editing tools push down the whole section
> 
> Ah, I see that now, thanks. Ignore that comment! 😄
> 
> Regarding the cancel and submit buttons of the survey, I did have a fresh build, but they're still always stacked for me. However, I noticed that your survey is wider than mine, so I guess the width of my survey is what is causing the buttons to stack. Hm, I'm not sure why my survey has a smaller width. Anyway, ignore that one too.

I will have another look to make sure that it scales properly to different screen sizes

escattone (commented):
Overall, I really like the GA4-event-related changes, and when I tested them locally, the GA4 events are generated properly (except for the first comment below). I have two comments:
- After submitting a survey response, a final pair of `survey_opened` and `survey_closed` GA4 events are generated, because `survey_form.html` is rendered and returned for the `response_message` of `Thanks for your feedback!`. Both of those survey events include an empty `survey_type` parameter. See the image below. I'm thinking we shouldn't generate survey events in this case, because it'll misleadingly inflate the GA4 survey events.

image

- For the survey events, maybe we should use `article_survey_...` instead of `survey_...` to distinguish these events from any possible future survey events?

kitsune/wiki/jinja2/wiki/includes/document_macros.html
```diff
@@ -351,16 +351,25 @@ <h3 class="{% if close_button %}sumo-card-heading{% else %}sidebar-subheading fo
...
-            <button type="submit" class="btn helpful-button" name="helpful"><span class="hidden-text">{{ _('Yes') }}</span>👍</button>
-            <button type="submit" class="btn not-helpful-button" name="not-helpful"><span class="hidden-text">{{ _('No') }}</span>👎</button>
+            <button type="submit" class="btn helpful-button" name="helpful"
+                    data-event-name="article_vote"
+                    data-event-parameters='{"vote": "helpful"}'
+            >{{ _('Yes') }}👍</button>
+            <button type="submit" class="btn not-helpful-button" name="not-helpful"
+                    data-event-name="article_vote"
+                    data-event-parameters='{"vote": "not-helpful"}'
+            >{{ _('No') }}👎</button>
```
escattone, line 372: ❤️

kitsune/wiki/jinja2/wiki/includes/survey_form.html
```diff
@@ -0,0 +1,89 @@
...
+             if (surveyType) {
+                 trackEvent('article_survey_closed', { survey_type: surveyType });
+             }
+             document.querySelector('.document-vote').remove();
+         }
+     }" x-init="
+        const surveyType = '{{ survey_type }}'.trim();
+        if (surveyType) {
+            trackEvent('article_survey_opened', { survey_type: surveyType });
+        }
```
escattone, line 32: Nit. The indentation is off.

escattone (approved):
The GA4 events look good now. Just a indentation nit, and there's still the issue of the text of the `Yes👍` and `No👎` buttons looking different than the Figma designs (but as you said, you can discuss that with Josh). Thanks @akatsoulas!

PR #6432, opened 2025-01-07
restore GA4 logging which is used for testing on stage

escattone:
https://github.com/mozilla/kitsune/pull/6418 removed the GA4 logging infrastructure, but it's our only way to test GA4 events within the `stage` environment (it's disabled in all other environments), so this PR restores it.

### DIFF
--- a/kitsune/sumo/static/sumo/js/analytics.js
+++ b/kitsune/sumo/static/sumo/js/analytics.js
@@ -1,5 +1,11 @@
 export default function trackEvent(name, parameters) {
   if (window.gtag) {
+    if (window.gaConsoleLogging) {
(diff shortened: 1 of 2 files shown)


akatsoulas (approved):
Would something like DebugView could be useful instead of logging into the console?

escattone:
> Would something like DebugView could be useful instead of logging into the console?

In the past, even when using GA4's DebugView, we still found that we often needed to confirm whether or not the `trackEvent` call was even made by the JS, so we decided to keep GA4 event console logging on for stage only. Currently, given that the `trackEvent` calls have proven trustworthy, Emil just uses the GA4 console logging when QA'ing GA4 events.

PR #6433, opened 2025-01-08
Update robots view to disallow non-canonical

smithellis:
This should disallow all crawls at any domain that isn't support.mozilla.org .

Our `robots.txt` file is dynamically generated, so this just adds another test to determine if we serve the detailed template or the 'disallow everything' code.

### DIFF
--- a/kitsune/sumo/views.py
+++ b/kitsune/sumo/views.py
@@ -105,7 +105,12 @@ def robots(request):
     if not settings.ENGAGE_ROBOTS:
         template = "User-Agent: *\nDisallow: /"
     else:
-        template = render(request, "sumo/robots.html")
+        if not request.build_absolute_uri("/").startswith("https://support.mozilla.org"):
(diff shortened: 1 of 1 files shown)


kitsune/sumo/views.py
```diff
@@ -105,7 +105,12 @@ def robots(request):
     if not settings.ENGAGE_ROBOTS:
         template = "User-Agent: *\nDisallow: /"
     else:
-        template = render(request, "sumo/robots.html")
+        if not request.build_absolute_uri("/").startswith("https://support.mozilla.org"):
```
escattone, line 108: I would recommend using `request.get_host()` instead of `request.build_absolute_uri("/")`, since we only care about the requested host.

smithellis: Ah, thanks.

escattone (approved):
Thanks @smithellis!

PR #6443, opened 2025-01-13
Allow unsafe-eval for AlpineJS

akatsoulas:


escattone:
We're going to merge this for exploration on stage while we discuss it more.

PR #6446, opened 2025-01-13
fix cancel of helpfulness survey

escattone:
mozilla/sumo#2115


escattone:
Closing because I'm going to fix this within mozilla/sumo#6448

PR #6448, opened 2025-01-13
Make AlpineJS CSP friendly

escattone:
This is a revert of a revert of https://github.com/mozilla/kitsune/pull/6442, so @akatsoulas is the real creator of this PR. The only thing I added was a second commit that is a small fix for the survey `Cancel` button.


escattone:
Closing in favor of https://github.com/mozilla/kitsune/pull/6453

PR #6456, opened 2025-01-16
This updates the topic cards on Product landing and AAQ

smithellis:
* A utility function builds the topics_context for the cards
* Article and Topic links are corrected
* This addresses:
    * https://github.com/mozilla/sumo/issues/2128 - 500 error
    * https://github.com/mozilla/sumo/issues/2134 - Links wrong
    * https://github.com/mozilla/sumo/issues/2130 - Articles not listed
    * https://github.com/mozilla/sumo/issues/2133 - Number of articles displayed

### DIFF
--- a/kitsune/questions/jinja2/questions/includes/aaq_macros.html
+++ b/kitsune/questions/jinja2/questions/includes/aaq_macros.html
@@ -1,4 +1,4 @@
-{% from 'products/includes/topic_macros.html' import help_topics with context %}
+{% from 'products/includes/topic_macros.html' import help_topics %}
 {% from 'includes/common_macros.html' import featured_articles, scam_banner %}
 
 {% macro select_product(products) -%}
--- a/kitsune/questions/jinja2/questions/includes/aaq_macros.html
+++ b/kitsune/questions/jinja2/questions/includes/aaq_macros.html
@@ -1,4 +1,4 @@
-{% from 'products/includes/topic_macros.html' import help_topics %}
+{% from 'products/includes/topic_macros.html' import help_topics with context %}
 {% from 'includes/common_macros.html' import featured_articles, scam_banner %}
 
 {% macro select_product(products) -%}
--- a/kitsune/questions/views.py
+++ b/kitsune/questions/views.py
@@ -70,6 +70,7 @@ from kitsune.tidings.models import Watch
 from kitsune.upload.models import ImageAttachment
 from kitsune.users.models import Setting
 from kitsune.wiki.facets import topics_for
+from kitsune.wiki.utils import build_topics_data
 
 log = logging.getLogger("k.questions")
 
@@ -568,8 +569,11 @@ def aaq(request, product_slug=None, step=1, is_loginless=False):
         context["ga_products"] = f"/{product_slug}/"
 
     if step == 2:
+        topics = topics_for(request.user, product, parent=None)
+        topics_data = build_topics_data(request, product, topics)
+
         context["featured"] = get_featured_articles(product, locale=request.LANGUAGE_CODE)
-        context["topics"] = topics_for(request.user, product, parent=None)
+        context["topics"] = topics_data
 
     elif step == 3:
         context["cancel_url"] = get_next_url(request) or (
--- a/kitsune/questions/views.py
+++ b/kitsune/questions/views.py
@@ -570,10 +570,9 @@ def aaq(request, product_slug=None, step=1, is_loginless=False):
 
     if step == 2:
         topics = topics_for(request.user, product, parent=None)
-        topics_data = build_topics_data(request, product, topics)
 
         context["featured"] = get_featured_articles(product, locale=request.LANGUAGE_CODE)
-        context["topics"] = topics_data
+        context["topics"] = build_topics_data(request, product, topics)
 
     elif step == 3:
         context["cancel_url"] = get_next_url(request) or (
--- a/kitsune/wiki/utils.py
+++ b/kitsune/wiki/utils.py
@@ -205,31 +205,46 @@ def get_visible_revision_or_404(user, **kwargs):
 def build_topics_data(request, product, topics):
     """Build topics_data for use in topic cards"""
     topics_data = []
+
+    # Get all documents for all topics up front
+    all_docs, _ = documents_for(
+        request.user, request.LANGUAGE_CODE, topics=list(topics), products=[product]
+    )
+
+    # Convert all docs to Document objects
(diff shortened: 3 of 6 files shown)


kitsune/wiki/utils.py
```diff
@@ -198,3 +200,36 @@ def get_visible_document_or_404(
...
+
+
+def build_topics_data(request, product, topics):
+    """Build topics_data for use in topic cards"""
+    topics_data = []
+    for topic in topics:
+        docs, _ = documents_for(
+            request.user, request.LANGUAGE_CODE, topics=[topic], products=[product]
+        )
+        docs = [Document.objects.get(id=doc_dict["id"]) for doc_dict in docs]
```
akatsoulas, line 212:
```python
docs = Document.objects.filter(id__in=[doc.id for doc in docs]
```

kitsune/wiki/utils.py
```diff
@@ -198,3 +200,36 @@ def get_visible_document_or_404(
...
+
+        total_articles = len(docs)
```
akatsoulas, line 214: `total_articles = docs.count()`

kitsune/wiki/utils.py
```diff
@@ -198,3 +200,36 @@ def get_visible_document_or_404(
...
+        # Leverage get_featured_articles to get active articles
+        featured_articles = get_featured_articles(
+            product, locale=request.LANGUAGE_CODE, topic=topic
+        )
+
+        if len(featured_articles) >= 3:
```
akatsoulas, line 220: Can you explain the use case for this check?

kitsune/wiki/utils.py
```diff
@@ -198,3 +200,36 @@ def get_visible_document_or_404(
...
+            docs = featured_articles
+        else:
+            docs = set(featured_articles + docs)
+
+        # Create a dict with topic data including the URL
+        topic_data = {
+            "topic": topic,
+            "topic_url": reverse("products.documents", args=[product.slug, topic.slug]),
+            "title": topic.title,
+            "total_articles": total_articles,
```
akatsoulas, line 230: you can also use here directly `docs.count()`. If I am not mistaken, `total_articles` is not used anywhere else

akatsoulas (changes requested):

kitsune/wiki/utils.py
smithellis: Ah - this is good. Better database performance. Thanks for this.

kitsune/questions/jinja2/questions/includes/aaq_macros.html
```diff
@@ -1,4 +1,4 @@
-{% from 'products/includes/topic_macros.html' import help_topics %}
+{% from 'products/includes/topic_macros.html' import help_topics with context %}
```
akatsoulas, line 1: Is this needed?

kitsune/questions/views.py
```diff
@@ -568,8 +569,11 @@ def aaq(request, product_slug=None, step=1, is_loginless=False):
         context["ga_products"] = f"/{product_slug}/"
 
     if step == 2:
+        topics = topics_for(request.user, product, parent=None)
+        topics_data = build_topics_data(request, product, topics)
+
         context["featured"] = get_featured_articles(product, locale=request.LANGUAGE_CODE)
-        context["topics"] = topics_for(request.user, product, parent=None)
+        context["topics"] = topics_data
```
akatsoulas, line 576: nit: context['topics'] = build_topics_data(request, product, topics)

kitsune/wiki/utils.py
```diff
@@ -198,3 +200,36 @@ def get_visible_document_or_404(
...
+    """Build topics_data for use in topic cards"""
+    topics_data = []
+    for topic in topics:
+        # Get featured articles first (returns up to 4)
+        featured_articles = get_featured_articles(
+            product, locale=request.LANGUAGE_CODE, topic=topic
+        )
+
+        # Always get total count of available documents
+        docs, _ = documents_for(
```
akatsoulas, line 215: Ideally this should be moved outside of the for loop by fetching documents for all topics and then organize them per topic. This way you can loop over data in memory instead of querying the database on each loop iteration

akatsoulas (changes requested):

kitsune/questions/jinja2/questions/includes/aaq_macros.html
smithellis: no

kitsune/wiki/utils.py
```diff
@@ -198,3 +200,51 @@ def get_visible_document_or_404(
...
+            product, locale=request.LANGUAGE_CODE, topic=topic
+        )
+
+        # Get docs for this topic from documents
+        topic_docs = [doc for doc in documents if topic.id in getattr(doc, "topics_list", [])]
+
+        # Only query for additional docs if we need more to reach 3
+        if len(featured_articles) < 3:
+            # If we don't have enough featured articles, append more from topic_docs
+            remaining_docs = [doc for doc in topic_docs if doc not in featured_articles]
```
escattone, line 237: Nit. Since you are only going to use the first 3 documents of `featured_articles`, I would just add whatever remaining documents you need to get to 3.

Another way to do this might be to use `itertools.chain` like `chain(featured_articles, remaining_docs)[:3]` where you could make `remaining_docs` a generator instead of a list by wrapping it with `(...)` instead of `[...]`. That way you don't even need the `len(featured_articles) < 3` check. The `chain` only grabs documents from `remaining_docs` if needed.

escattone (changes requested):

kitsune/wiki/utils.py
```diff
@@ -198,3 +200,51 @@ def get_visible_document_or_404(
...
+    # Convert all docs to Document objects
+    doc_ids = [doc["id"] for doc in all_docs]
+    documents = Document.objects.filter(id__in=doc_ids)
+
+    # Create a mapping of Document objects to their topics from all_docs
+    for doc_dict in all_docs:
+        for doc in documents:
+            if doc.id == doc_dict["id"]:
+                doc.topics_list = doc_dict.get("topics", [])
+                break
```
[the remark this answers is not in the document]
smithellis: Yes; I think this will work great.

kitsune/wiki/utils.py
smithellis: This is a great idea. Thanks.

smithellis:
Closing in favor of new consolidated PR https://github.com/mozilla/kitsune/pull/6471

PR #6457, opened 2025-01-16
provide flagged-object queues per user

escattone:
mozilla/sumo#2124

## Notes
I'm really happy with this. I think it provides a simple, solid way for each agent to manage their own queue of unique flags.

- I removed the `(Reason: content_moderation)` since all of the flags under `/moderate` are the same (see below).
- I added an indicator of the assignee for each flag (see below).

### The main queue remains the same except with the ability to filter by assignee
image

### Users can manage their own queue (assign/unassign items)
image

### Users can view (but not manage) any other's queue 
image


kitsune/flagit/models.py
```diff
@@ -54,6 +54,11 @@ class FlaggedObject(ModelBase):
     handled = models.DateTimeField(default=datetime.now, db_index=True)
     handled_by = models.ForeignKey(User, on_delete=models.CASCADE, null=True)
 
+    assignee = models.ForeignKey(
+        User, on_delete=models.SET_NULL, null=True, related_name="assigned_flags"
+    )
+    assigned = models.DateTimeField(default=None, null=True)
```
akatsoulas, line 60: nit: this can be confused with a boolean flag. Maybe rename it to something like `assigned_timestamp` or anything else that would clearly indicate that's a `DateTimeField`?

kitsune/flagit/jinja2/flagit/content_moderation.html
```diff
@@ -57,6 +66,29 @@ <h3 class="sumo-card-heading"><br>{{ _('Update Status:') }}</h3>
...
+      label='Filter by assignee:',
+      name='assignee',
+      default_option='All assignees',
+      options=assignees,
+      selected_filter=selected_assignee
+  ) }}
+  <div id="my-queue-tools" data-current-username="{{ current_username }}"
+       {% if not (selected_assignee and (selected_assignee == current_username)) %} hidden{% endif %}>
+    <label>{{ _('Manage my queue') }}:</label>
+    <form id="my-queue-assign" hx-post="" hx-target="#flagged-queue" hx-select="#flagged-queue">
```
akatsoulas, line 81: I believe this ID `id` is not used anywhere.

kitsune/flagit/views.py
```diff
@@ -163,9 +171,26 @@ def build_hierarchy(parent_id=None, level=0):
...
+    assignee = request.GET.get("assignee")
+
+    if request.method == "POST":
+        if not (assignee and (request.user.username == assignee)):
+            return HttpResponseForbidden()
+        action = request.POST.get("action")
+        if not (action and (action in ("assign", "unassign"))):
+            return HttpResponseBadRequest()
+    else:
+        if (
```
akatsoulas, line 187: You can move this above the check if the request is `POST` to avoid the nesting. This will also help you to have a single check of 
```python
if request.method == "POST":
```

akatsoulas (changes requested):

kitsune/flagit/jinja2/flagit/content_moderation.html
```diff
@@ -57,6 +66,29 @@ <h3 class="sumo-card-heading"><br>{{ _('Update Status:') }}</h3>
...
+      {% csrf_token %}
+      <input type="hidden" name="action" value="assign">
+      <input class="sumo-button secondary-button button-lg btn" type="submit" value="{{ _('Assign more') }}" />
```
akatsoulas, line 84: For buttons like `Submit` and `Cancel` we are using `primary-button` for `Submit` and `secondary` for Cancel. I wonder if we should follow the same paradigm here by making this `primary`

kitsune/flagit/jinja2/flagit/content_moderation.html
escattone: Good catch!

kitsune/flagit/views.py
```diff
@@ -163,9 +171,26 @@ def build_hierarchy(parent_id=None, level=0):
...
+
+    if (
+        assignee
+        and not User.objects.filter(
+            is_active=True, username=assignee, groups__name="Content Moderators"
+        ).exists()
+    ):
+        return HttpResponseNotFound()
+
+    if request.method == "POST":
```
akatsoulas, line 188: This can be merged with the other check whether the request.method equals to POST.

akatsoulas (approved):
r+wc

kitsune/flagit/views.py
escattone: Yeah, that's cleaner. I was trying to avoid the bit of work involved with getting the `content_type` and setting up the query for `objects` if the POST was invalid.

PR #6460, opened 2025-01-18
make moderation submit async

escattone:
mozilla/sumo#2122

## Notes
I'm not sure if we just want to remove the flagged item once it's been updated, or do something else, but I just kept this simple and removed the flagged item.


akatsoulas (approved):
nice!

PR #6462, opened 2025-01-21
Remove containers from other tabs if user has voted

akatsoulas:


kitsune/wiki/jinja2/wiki/includes/survey_form.html
```diff
@@ -1,4 +1,4 @@
-<div class="survey-container" x-data="surveyForm" data-survey-type="{{ survey_type }}">
+<div class="survey-container" x-data="surveyForm" data-survey-type="{{ survey_type }}" data-page-url="{{ page_url }}">
```
escattone, line 1: I don't think this is needed, right? Maybe a leftover from an earlier iteration?

escattone (approved):
r+wc

This is brilliant! ❤️ I didn't even know about `BroadcastChannel` prior to this, but I'll remember it going forward.

I just had the one comment about something that seems to me like it could be removed.

PR #6468, opened 2025-01-22
Removed documentation of The search suggest API (/api/2/search/sugges…

tanii1125:
#2104

AS suggest API as it is no longer present. It should not be there in the documents as it will make unnecessary confusions.
So I removed:
 - [x] It's Endpoint
 - [x] Required arguments
 - [x] Optional arguments
 - [x] Responses
 - [x] Examples


akatsoulas (approved):
Thanks @tanii1125! :rocket:

PR #6471, opened 2025-01-23
Topic cards revised to show three articles

smithellis:
* Improved topic card now show three top articles
* Alter the help_topics macro to support showing three articles
* Add show more section with sectional line above
* Add logic to the product_landing view to return docs and counts
* A utility function builds the topics_context for the cards
* Fix card scroll for topic cards on mobile
* Topics can only wrap once before being cut off with ellipses
* Use query annoation in `utils.py` to gather topics
* Use `chain` and `islice` to ensure proper doc order and count

This creates a new card for topics that contains three article titles. The articles are the three most frequently viewed for each topic and of that product, falling back three regular documents if we don't get enough documents from frequently viewed data.

The article titles are allowed to wrap once if they are lengthy, but then are cut off with an ellipse if they exceed two lines. 

The cards are in three columns, using a flex layout.

The cards scroll left to right on mobile, where they have fixed heights and widths to keep the experience optimal.

### DIFF
--- a/kitsune/products/views.py
+++ b/kitsune/products/views.py
@@ -11,7 +11,7 @@ from kitsune.sumo.utils import has_aaq_config, set_aaq_context
 from kitsune.wiki.decorators import check_simple_wiki_locale
 from kitsune.wiki.facets import documents_for, topics_for
 from kitsune.wiki.models import Document, Revision
-from kitsune.wiki.utils import get_featured_articles
+from kitsune.wiki.utils import build_topics_data, get_featured_articles
 
 
 @check_simple_wiki_locale
@@ -23,8 +23,19 @@ def product_list(request):
 
 
 @check_simple_wiki_locale
-def product_landing(request, slug):
-    """The product landing page."""
+def product_landing(request, slug: str) -> HttpResponse:
+    """The product landing page.
+
+    Args:
+        request: The HTTP request
+        slug: Product slug identifier
+
+    Returns:
+        Rendered product landing page
+
+    Raises:
+        Http404: If product not found
+    """
     if slug == "firefox-accounts":
         return redirect(product_landing, slug="mozilla-account", permanent=True)
 
@@ -45,6 +56,7 @@ def product_landing(request, slug):
             latest_version = versions[0].min_version
         else:
             latest_version = 0
+    topics = topics_for(request.user, product=product, parent=None)
 
     return render(
         request,
@@ -52,7 +64,7 @@ def product_landing(request, slug):
         {
             "product": product,
             "products": Product.active.filter(visible=True),
-            "topics": topics_for(request.user, product=product, parent=None),
+            "topics": build_topics_data(request, product, topics),
             "search_params": {"product": slug},
             "latest_version": latest_version,
             "featured": get_featured_articles(product, locale=request.LANGUAGE_CODE),
--- a/kitsune/products/views.py
+++ b/kitsune/products/views.py
@@ -2,7 +2,7 @@ import json
 
 from django.conf import settings
 from django.db.models import Exists, OuterRef, Q
-from django.http import Http404, HttpResponse
+from django.http import Http404, HttpRequest, HttpResponse
 from django.shortcuts import get_object_or_404, redirect, render
 from product_details import product_details
 
@@ -23,7 +23,7 @@ def product_list(request):
 
 
 @check_simple_wiki_locale
-def product_landing(request, slug: str) -> HttpResponse:
+def product_landing(request: HttpRequest, slug: str) -> HttpResponse:
     """The product landing page.
 
     Args:
--- a/kitsune/sumo/static/sumo/scss/components/_card.scss
+++ b/kitsune/sumo/static/sumo/scss/components/_card.scss
@@ -185,18 +185,18 @@
     flex-direction: column;
     justify-content: space-between;
     height: 100%;
-    background-color: #fff;
+    background-color: p.$color-white;
     border: 1px solid #ddd;
-    border-radius: 8px;
-    padding: 16px;
-    box-shadow: 0 1px 3px rgba(0, 0, 0, 0.1);
+    border-radius: p.$border-radius-md;
+    padding: p.$spacing-md;
+    box-shadow: p.$box-shadow-sm;
     box-sizing: border-box;
 
     .topic-header {
       display: flex;
       align-items: flex-start;
       gap: 10px;
-      margin-bottom: 16px;
+      margin-bottom: p.$spacing-md;
     }
 
     .card--icon {
@@ -216,7 +216,7 @@
 
     .article-list {
       flex-grow: 1;
-      margin: 0 0 16px;
+      margin: 0 0 p.$spacing-md;
       padding: 0;
       list-style: none;
 
@@ -244,7 +244,7 @@
     .view-all-link {
       border-top: 1px solid #ddd;
       padding-top: 10px;
-      margin-top: 16px;
+      margin-top: p.$spacing-md;
       display: inline-block;
       font-size: 14px;
       color: #000000;
--- a/kitsune/sumo/static/sumo/scss/components/_card.scss
+++ b/kitsune/sumo/static/sumo/scss/components/_card.scss
@@ -181,21 +181,79 @@
   }
 
   &--topic {
-    @include c.elevation-01;
     display: flex;
-    align-items: center;
+    flex-direction: column;
+    justify-content: space-between;
+    height: 100%;
+    background-color: #fff;
+    border: 1px solid #ddd;
+    border-radius: 8px;
+    padding: 16px;
+    box-shadow: 0 1px 3px rgba(0, 0, 0, 0.1);
+    box-sizing: border-box;
 
-    .card-title {
-      margin: 0;
+    .topic-header {
+      display: flex;
+      align-items: flex-start;
+      gap: 10px;
+      margin-bottom: 16px;
     }
 
     .card--icon {
-      width: p.$spacing-lg;
-      height: p.$spacing-lg;
-      object-fit: contain;
-      font-size: 9px;
-      line-height: 1;
-      flex: 0 0 auto;
+      flex-shrink: 0;
+      width: 18px;
+      height: 18px;
+    }
+
+    .card--title {
+      font-family: Inter;
+      font-size: 16px;
+      font-weight: 700;
+      margin: 0;
+      line-height: 1.2;
+      flex-grow: 1;
+    }
+
+    .article-list {
+      flex-grow: 1;
+      margin: 0 0 16px;
+      padding: 0;
+      list-style: none;
+
+      li {
+        margin-bottom: 8px;
+        line-height: 1.5;
+        
+        a {
+          color: black;
+          font-size: 14px;
+          text-decoration: underline;
+          display: -webkit-box;
+          -webkit-line-clamp: 2;
+          line-clamp: 2;
+          -webkit-box-orient: vertical;
+          overflow: hidden;
+
+          &:hover {
+            text-decoration: underline;
+          }
+        }
+      }
+    }
+
+    .view-all-link {
+      border-top: 1px solid #ddd;
+      padding-top: 10px;
+      margin-top: 16px;
+      display: inline-block;
+      font-size: 14px;
+      color: #000000;
+      text-decoration: underline;
+      font-weight: normal;
+
+      &:hover {
+        text-decoration: underline;
+      }
     }
   }
 
--- a/kitsune/sumo/static/sumo/scss/layout/_products.scss
+++ b/kitsune/sumo/static/sumo/scss/layout/_products.scss
@@ -1,7 +1,6 @@
 @use '../config' as c;
 @use 'protocol/css/includes/lib' as p;
 
-
 .sumo-page-subheader {
   display: flex;
   flex-direction: column-reverse;
@@ -33,9 +32,6 @@
   }
 }
 
-
-// SCSS for Topics Section Layout
-
 .topics-section {
   padding: 40px 20px;
   h2 {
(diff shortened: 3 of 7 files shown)


kitsune/products/views.py
```diff
@@ -23,8 +23,19 @@ def product_list(request):
 
 
 @check_simple_wiki_locale
-def product_landing(request, slug):
-    """The product landing page."""
+def product_landing(request, slug: str) -> HttpResponse:
```
akatsoulas, line 26: Nice! Let's add also a type annotation for the request too, `request: HttpRequest`

kitsune/sumo/static/sumo/scss/components/_card.scss
```diff
@@ -181,21 +181,79 @@
...
   &--topic {
-    @include c.elevation-01;
     display: flex;
-    align-items: center;
+    flex-direction: column;
+    justify-content: space-between;
+    height: 100%;
+    background-color: #fff;
+    border: 1px solid #ddd;
+    border-radius: 8px;
```
akatsoulas, line 190: Let's use [design tokens](https://protocol.mozilla.org/docs/fundamentals/design-tokens) from protocol for consistency with the rest of the codebase.

kitsune/sumo/static/sumo/scss/layout/_products.scss
```diff
@@ -32,3 +32,62 @@
     margin: 0 p.$spacing-xs;
   }
 }
+
+
+// SCSS for Topics Section Layout
+
+.topics-section {
+  padding: 40px 20px;
```
akatsoulas, line 40: Similar comment for design tokens where applicable.

kitsune/sumo/static/sumo/scss/layout/_products.scss
```diff
@@ -32,3 +32,62 @@
...
+    > .card--topic {
+      flex: 0 0 280px; // Fixed width for each item
+      height: 280px; // Fixed height
+      scroll-snap-align: start;
+      
+      // Force all content inside the card to respect the height
+      display: flex;
+      flex-direction: column;
+      
+      // Prevent content from overflowing
```
akatsoulas, line 84: There are comments that are redundant throughout your CSS. In this case the overflow property is self explanatory about what it does.

akatsoulas: This applies to most comments in CSS.

kitsune/wiki/utils.py
```diff
@@ -103,10 +109,11 @@ def _active_contributors_id(from_date, to_date, locale, product):
     return set(list(editors) + list(reviewers))
 
 
-def get_featured_articles(product=None, locale=settings.WIKI_DEFAULT_LANGUAGE):
+def get_featured_articles(product=None, locale=settings.WIKI_DEFAULT_LANGUAGE, topic=None):
```
akatsoulas, line 112: I would change the signature to  `product, topic, locale`when it comes to the order of args

kitsune/wiki/utils.py
```diff
@@ -119,11 +126,22 @@ def get_featured_articles(product=None, locale=settings.WIKI_DEFAULT_LANGUAGE):
         .exclude(document__is_template=True)
         .order_by("-visits")
         .select_related("document")
+        .annotate(
+            topic_ids=StringAgg(
```
akatsoulas, line 130: This section could benefit from a comment

kitsune/wiki/utils.py
```diff
@@ -119,11 +126,22 @@ def get_featured_articles(product=None, locale=settings.WIKI_DEFAULT_LANGUAGE):
...
+                ordering="document__topics__id",
+                default="",
+            )
+        )
     )
 
     if product:
         visits = visits.filter(document__products__in=[product.id])
 
+    if topic:
```
akatsoulas, line 142: nit: This section could be optimized for efficiency
```python
filters = {}
if product:
    filters['document__products__in'] = [product.id]
if topic:
    filters['document__topics__in'] = [topic.id]

for filter_name, filter_value in filters.items():
    visits = visits.filter(**{filter_name: filter_value})
```

kitsune/wiki/utils.py
```diff
@@ -194,3 +212,51 @@ def get_visible_document_or_404(
...
+        featured_articles = get_featured_articles(
+            product, locale=request.LANGUAGE_CODE, topic=topic
+        )
+
+        # Get docs for this topic from documents
+        topic_docs = [
+            doc for doc in documents if doc.topic_ids and str(topic.id) in doc.topic_ids.split(",")
+        ]
+
+        # Create a generator for remaining docs
```
akatsoulas, line 249: Similar comment with CSS, the comment is not needed here. This applies to all submitted code.

kitsune/wiki/utils.py
```diff
@@ -194,3 +212,51 @@ def get_visible_document_or_404(
 
 def get_visible_revision_or_404(user, **kwargs):
     return get_object_or_404(Revision.objects.visible(user, **kwargs))
+
+
+def build_topics_data(request, product, topics):
```
akatsoulas, line 217: Let's add type hints similar to what you did with the product view.

kitsune/wiki/utils.py
```diff
@@ -194,3 +212,51 @@ def get_visible_document_or_404(
...
+    """Build topics_data for use in topic cards"""
```
akatsoulas, line 218: For consistency let's document the input and output of this function similar to the product view

kitsune/wiki/utils.py
```diff
@@ -194,3 +212,51 @@ def get_visible_document_or_404(
...
+    topics_data = []
+
+    # Get all documents for all topics up front
+    all_docs, _ = documents_for(
+        request.user, request.LANGUAGE_CODE, topics=list(topics), products=[product]
```
akatsoulas, line 223: Why is the `list()` conversion needed? It looks like `_documents_for` expects an iterable and not a list specifically. This will also force the evaluation of the queryset.

kitsune/wiki/utils.py
```diff
@@ -194,3 +212,51 @@ def get_visible_document_or_404(
...
+    )
+
+    # Convert all docs to Document objects
+    doc_ids = [doc["id"] for doc in all_docs]
+    documents = (
+        Document.objects.filter(id__in=doc_ids)
+        .prefetch_related("topics")
+        .annotate(
+            topic_ids=StringAgg(
```
akatsoulas, line 232: nice!

kitsune/wiki/utils.py
```diff
@@ -194,3 +212,51 @@ def get_visible_document_or_404(
...
+                Cast("topics__id", TextField()), delimiter=",", ordering="topics__id", default=""
```
akatsoulas, line 233: from what I can tell, the ordering is not needed here. The way you are comparing the strings in the `topic_docs` below shouldn't be affected by removing the `ordering`.

kitsune/wiki/utils.py
```diff
@@ -194,3 +212,51 @@ def get_visible_document_or_404(
...
+        # Create a generator for remaining docs
+        remaining_docs = (doc for doc in topic_docs if doc not in featured_articles)
+
+        topic_data = {
+            "topic": topic,
+            "topic_url": reverse("products.documents", args=[product.slug, topic.slug]),
+            "title": topic.title,
+            "total_articles": len(topic_docs),
+            "image_url": topic.image_url,
+            "documents": list(islice(chain(featured_articles, remaining_docs), 3)),
```
akatsoulas, line 258: this needs a comment :)

kitsune/wiki/utils.py
akatsoulas: The topic arg can be an iterable. This way you can have a single query and separate the articles in memory, similarly to what you did to the `build_topics_data` function

akatsoulas (changes requested):

akatsoulas:
Nice work so far!

kitsune/products/views.py
smithellis: Done.

kitsune/sumo/static/sumo/scss/components/_card.scss
smithellis: Got it.

kitsune/sumo/static/sumo/scss/layout/_products.scss
smithellis: Got it.

kitsune/sumo/static/sumo/scss/layout/_products.scss
smithellis: Fixed.

kitsune/wiki/utils.py
smithellis: Done.

kitsune/wiki/utils.py
smithellis: I have made some changes that i think improve things.

kitsune/wiki/utils.py
smithellis: Thanks!

kitsune/wiki/utils.py
smithellis: Done.

kitsune/wiki/utils.py
smithellis: Trimmed out some of the more obvious items.

kitsune/wiki/utils.py
smithellis: Added.

kitsune/wiki/utils.py
```diff
@@ -103,52 +110,65 @@ def _active_contributors_id(from_date, to_date, locale, product):
     return set(list(editors) + list(reviewers))
 
 
-def get_featured_articles(product=None, locale=settings.WIKI_DEFAULT_LANGUAGE):
-    """Returns 4 random articles from the most visited.
-
-    If a product is passed, it returns 4 random highly visited articles.
-    """
+def get_featured_articles(product=None, topic=None, locale=settings.WIKI_DEFAULT_LANGUAGE):
```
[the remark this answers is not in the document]
smithellis: Fixed. `get_featured_articles` now accepts an iterable and I've minimized the database interactions.

kitsune/wiki/utils.py
```diff
@@ -103,56 +108,68 @@ def _active_contributors_id(from_date, to_date, locale, product):
...
-        .order_by("-visits")
-        .select_related("document")
     )
 
+    # Add product and topics filters to the database query
     if product:
-        visits = visits.filter(document__products__in=[product.id])
+        visits = visits.filter(document__products=product)
+    if topics:
+        visits = visits.filter(document__topics__in=topics).distinct()
```
akatsoulas, line 153: Why is distinct needed here? I assume that we do want the same article to appear in different topics if it's the most visited

smithellis: My thought was that the document we need is singular, but the results we'd get here would have duplicates because a document can have multiple topics. The document itself, once retrieved, will still have it's connection to the topics - so nothing is lost, but we have a smaller result to deal with. 

I guess technically it could be dropped if you know a good reason - is distinct() expensive for the lack of value? Or am I wrong that document visits don't change regardless of topics/products?

kitsune/wiki/utils.py
```diff
@@ -103,56 +108,68 @@ def _active_contributors_id(from_date, to_date, locale, product):
...
 
-    visits = visits[:10]
-    documents = []
+    # Order by visits but don't limit - we need enough documents for all topics
+    visits = visits.order_by("-visits")
+
+    # Convert to list to avoid additional queries
+    visits = list(visits)
```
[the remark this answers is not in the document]
smithellis: That's an artifact - you are right. Since no further evaluations are happening there is no need for this.

kitsune/wiki/utils.py
```diff
@@ -103,56 +108,68 @@ def _active_contributors_id(from_date, to_date, locale, product):
...
-                continue
-            documents.append(translation)
+            translation = next(iter(visit.document.translations.all()), None)
+            if translation:
+                documents.append(translation)
 
-    if len(documents) <= 4:
-        return documents
-    return random.sample(documents, 4)
+    # Return all documents - the filtering per topic will happen in build_topics_data
```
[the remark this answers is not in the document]
smithellis: I'll remove.

kitsune/wiki/utils.py
```diff
@@ -194,3 +211,60 @@ def get_visible_document_or_404(
...
+
+    # Convert all docs to Document objects
+    doc_ids = [doc["id"] for doc in all_docs]
+    documents = (
+        Document.objects.filter(id__in=doc_ids)
+        .prefetch_related("topics")
+        .annotate(topic_ids=StringAgg(Cast("topics__id", TextField()), delimiter=",", default=""))
+    )
+
+    # Get featured articles for all topics in one query
```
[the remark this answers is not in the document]
smithellis: Will remove.

kitsune/wiki/utils.py
```diff
@@ -103,56 +108,68 @@ def _active_contributors_id(from_date, to_date, locale, product):
...
-            documents.append(translation)
+            translation = next(iter(visit.document.translations.all()), None)
+            if translation:
+                documents.append(translation)
 
-    if len(documents) <= 4:
-        return documents
-    return random.sample(documents, 4)
+    # Return all documents - the filtering per topic will happen in build_topics_data
+    return documents
```
akatsoulas, line 172: this no longer returns 4 articles like it used to. It will probably work for the topic cards but it will break functionality elsewhere

kitsune/wiki/utils.py
akatsoulas: We are querying `WikiDocumentVisits` and not `Document`. This means that each visit record related to one document b/c of the FK. I believe we can drop this without having any duplicates or affecting the query in any way

kitsune/wiki/utils.py
smithellis: You are right - this will break the homepage. I've reverted.

akatsoulas (approved):
Nice work. Please squash those commits into one before merging

PR #6480, opened 2025-01-30
provide SMTP email backend with Sentry capture

escattone:
Helps with mozilla/sumo#1993

This is intended for use in `prod` only -- since `dev` and `stage` use the `django-email-bandit` backend -- by changing the `EMAIL_LOGGING_REAL_BACKEND` setting from `"django.core.mail.backends.smtp.EmailBackend"` to `"kitsune.lib.email.SMTPEmailBackendWithSentryCapture"` in K8s.

### DIFF
--- a/kitsune/lib/email.py
+++ b/kitsune/lib/email.py
@@ -1,8 +1,11 @@
 import logging
+import smtplib
 import time
 
 from django.conf import settings
+from django.core.mail.backends import smtp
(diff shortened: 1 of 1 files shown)


smithellis (approved):
This will definitely help us monitor and manage email issues. I don't think this can be tested great locally, but the logic here seems solid.

akatsoulas (approved):

PR #6481, opened 2025-01-31
Exclude AAQ macro on the get-community-support page

smithellis:
If the `default_slug` is set to `get-community-support` we won't show the aaq sidebar button. `default_slug` is already being set in the template, it represents the parent slug if there is a parent, or the current doc slug if there isn't.

### DIFF
--- a/kitsune/wiki/jinja2/wiki/document.html
+++ b/kitsune/wiki/jinja2/wiki/document.html
@@ -56,7 +56,9 @@
     {{ search_box(settings, id='support-search-sidebar', params=search_params) }}
   </div>
 
-  {{ aaq_widget(request) }}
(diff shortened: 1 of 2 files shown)


kitsune/wiki/jinja2/wiki/document.html
```diff
@@ -56,7 +56,9 @@
     {{ search_box(settings, id='support-search-sidebar', params=search_params) }}
   </div>
 
-  {{ aaq_widget(request) }}
+  {% if default_slug != 'get-community-support' %}
```
akatsoulas, line 59: I would probably move the check inside the view and expose a variable like `show_aaq_widget` in the context. Up to your discretion if you would like to go down this path.

akatsoulas (approved):
r+wc

PR #6482, opened 2025-01-31
rate limit emails

escattone:
mozilla/sumo#1993

Do not merge this until after we discover more from https://github.com/mozilla/kitsune/pull/6480.


akatsoulas (approved):
Really nice work! :rocket: 

nit: I wonder if it makes sense to add some jitter in the wait function to distribute the load and avoid multiple processes running at the same time. Up to your discretion.

PR #5962, opened 2024-04-17
Add a hint to relogin

kokobhara:
Fixes #5446

PR #6483, opened 2025-02-01
Hint on required re-logging to see changed avatar

tanii1125:


kitsune/users/jinja2/users/edit_profile.html
```diff
@@ -29,6 +29,10 @@ <h1 class="sumo-page-heading">{{ _('Your Account') }}</h1>
     <p>
       {% trans a_open='<a href="https://support.mozilla.org/en-US/kb/change-primary-email-address-firefox-accounts">'|safe, a_close='</a>'|safe %}
         To change your email or avatar, visit the Mozilla accounts page. {{ a_open }} Learn more. {{ a_close }}
+        <br>
+        <small>After changing your avatar, please re-login on
+          <a href="https://support.mozilla.org/en-US/users/auth">support.mozilla.org</a> to see the changes to apply.
+        </small>
```
janbrasna, line 35: BTW this would trigger new translations of the whole string, incl. the already existing one, and the added markup also doesn't really handle well in localization.

A more successful path could be as shown in [#5962](https://github.com/mozilla/kitsune/pull/5962/files) — as a new separate string (also with no links etc. complicating it more than necessary…)

PR #6503, opened 2025-02-11
Ensure view gets a locale

smithellis:
In the case of leaving the revision page, then returning via the browser back button, the locale wasn't being set in any way. This change ensures the locale is set.

This should resolve https://github.com/mozilla/sumo/issues/2018

### DIFF
--- a/kitsune/wiki/views.py
+++ b/kitsune/wiki/views.py
@@ -1697,10 +1697,10 @@ def recent_revisions(request):
     form.is_valid()
 
     filters = {}
+    locale = None
     if hasattr(form, "cleaned_data"):
         if form.cleaned_data.get("locale"):
             filters.update(document__locale=form.cleaned_data["locale"])
-
         # Only apply user filter if there are valid users
         if form.cleaned_data.get("users"):
             filters.update(creator__in=form.cleaned_data["users"])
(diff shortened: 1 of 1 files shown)


kitsune/wiki/views.py
```diff
@@ -1697,10 +1697,10 @@ def recent_revisions(request):
     form.is_valid()
 
     filters = {}
+    locale = None
```
escattone, line 1700: Why is this is needed?

kitsune/wiki/views.py
```diff
@@ -1717,7 +1717,7 @@ def recent_revisions(request):
     c = {
         "revisions": revs,
         "form": form,
-        "locale": request.GET.get("locale", request.LANGUAGE_CODE),
+        "locale": locale or request.LANGUAGE_CODE,  # Ensure locale is always set
```
escattone, line 1720: I don't understand this change as well. Why are you no longer getting the locale from the URL's query parameters?

akatsoulas: What basically this does is equivalent to `"locale": request.LANGUAGE_CODE`. You are setting `locale=None` which will force the short circuit operation to evaluate always to `request.LANGUAGE_CODE`

The problem here is the get() method. It defaults to `request.LANGUAGE_CODE` if the "locale" dictionary key is not present. It doesn't check the value of the key which is the empty string. 

This should be:
```python
"locale": request.GET.get("locale") or request.LANGUAGE_CODE

akatsoulas (changes requested):
Please merge after the release.

kitsune/wiki/views.py
smithellis: Correcting.

kitsune/wiki/views.py
smithellis: You are right of course; correcting.

akatsoulas (approved):

PR #6507, opened 2025-02-13
Introduce handlers for AAQ content moderation

akatsoulas:


kitsune/questions/handlers.py
```diff
@@ -0,0 +1,85 @@
...
+        }
+
+        for handler in self.handlers:
+            handler.handle_account(data)
+
+        # Re-assign remaining questions and answers to SumoBot.
+        try:
+            sumo_bot = User.objects.get(username=settings.SUMO_BOT_USERNAME)
+        except User.DoesNotExist:
+            raise ValueError("SumoBot user not found")
```
akatsoulas, line 82: This section could/should change when we introduce the system account/sumobot. For now, it just ensures that it will fail if we don't have the account to reassign

kitsune/sumo/handlers.py
```diff
@@ -0,0 +1,42 @@
...
+class AbstractHandler(ABC):
+    """Abstract base class for all handlers."""
+
+    @abstractmethod
+    def handle(self) -> None:
+        pass
+
+
+@dataclass
+class AbstractChain(Generic[T]):
```
escattone, line 18: I don't fully understand all of `typing` stuff, so I'm probably missing something, but I would have expected `AbstractHandler` instead `T` within this class.

escattone (approved):
r+wc

kitsune/questions/handlers.py
```diff
@@ -0,0 +1,85 @@
...
+        answers = data["answers"]
+
+        archived_questions = questions.filter(product__is_archived=True).values_list(
+            "id", flat=True
+        )
+        archived_answers = answers.filter(question__in=archived_questions).values_list(
+            "id", flat=True
+        )
+        Question.objects.filter(id__in=archived_questions).delete()
+        Answer.objects.filter(id__in=archived_answers).delete()
```
escattone, line 39: I think these should be switched in terms of order because of the dependency of `archived_answers` on `archived_questions`?

akatsoulas:
closing in favor of #6509

PR #6509, opened 2025-02-14
Account deletions

akatsoulas:


kitsune/questions/handlers.py
```diff
@@ -0,0 +1,95 @@
...
+        answers = data["answers"]
+
+        archived_questions = questions.filter(product__is_archived=True).values_list(
+            "id", flat=True
+        )
+        archived_answers = answers.filter(question__in=archived_questions).values_list(
+            "id", flat=True
+        )
+        Question.objects.filter(id__in=archived_questions).delete()
+        Answer.objects.filter(id__in=archived_answers).delete()
```
escattone, line 40: I think these two should be switched in order of execution due to the dependency of `archived_answers` on `archived_questions`? Otherwise, I think if you delete the questions first, the query for the `archived_answers` won't return anything so no answers will ever be deleted?

Oh, actually, when I noticed that you had tests that run successfully for this, I then realized that it works only because of the cascade-delete [here](https://github.com/mozilla/kitsune/blob/c10de5dd2360f47860d0552e3a0e5abdfec54002/kitsune/questions/models.py#L873). So I don't think you even need to delete the answers, because they'll be cascade-deleted when the questions are deleted.

kitsune/questions/handlers.py
```diff
@@ -0,0 +1,95 @@
...
+
+class OrphanedQuestionAAQHandler(AccountHandler):
+    """Handler for orphaned question cleanup."""
+
+    def handle_account(self, data: dict) -> None:
+        """Handler to delete all questions without answers."""
+        questions = data["questions"]
+
+        orphaned_questions = questions.filter(num_answers=0).values_list("id", flat=True)
+        Question.objects.filter(id__in=orphaned_questions).delete()
```
escattone, line 51: Likewise, I think this could be simpler?
```python
        questions.filter(num_answers=0).delete()
```

kitsune/sumo/handlers.py
```diff
@@ -0,0 +1,42 @@
...
+
+    _handlers: list[T] = field(default_factory=list)
+
+    @property
+    def handlers(self) -> Sequence[T]:
+        return self._handlers
+
+    def add_handler(self, handler: T) -> None:
+        """Add a handler to the chain."""
+        self._handlers.append(handler)
```
escattone, line 29: Within this class, why not use `AbstractHandler` instead of `T`? I can't see that it would ever be anything other than an `AbstractHandler`.

kitsune/settings.py
```diff
@@ -928,6 +928,7 @@ def JINJA_CONFIG():
     # TODO: Make sure that db number is respected
     "default": config("REDIS_DEFAULT_URL"),
     "helpfulvotes": config("REDIS_HELPFULVOTES_URL"),
+    "account_events": config("REDIS_ACCOUNT_EVENTS_URL"),
```
escattone, line 931: FYI. I did check and any new Redis DB will be automatically created upon initial connection. So there's nothing to worry about there.

kitsune/users/tasks.py
```diff
@@ -20,10 +19,9 @@
 def process_event_delete_user(event_id):
     event = AccountEvent.objects.get(id=event_id)
 
-    anonymize_user(event.profile.user)
-
-    event.status = AccountEvent.PROCESSED
-    event.save()
+    if delete_user_pipeline.delay(event.profile.user):
```
escattone, line 22: Is the `.delay(...)` leftover from an earlier iteration? The `delete_user_pipeline` function is not a task. Or perhaps it needs to be pulled into this file and made into a Celery task?

kitsune/users/handlers.py
```diff
@@ -0,0 +1,56 @@
...
+    """
+
+    user: User
+    listeners: list[UserDeletionListener] = field(default_factory=list)
+    _redis: Any = field(default=None, init=False, repr=False)
+    redis_key: str = field(init=False)
+
+    def __post_init__(self):
+        if UserDeletionPublisher._redis is None:
+            UserDeletionPublisher._redis = redis_client("account_events")
```
escattone, line 38: I don't understand the need for Redis. As far as I can tell, all of the subscribed listeners are run sequentially when `publish()` is called, and none of the listeners that have been defined in this PR do anything asynchronously (like submit their work to a Celery task) within their `on_user_deletion` method. So when the `publish()` completes, it seems to me that all of the work has already been completed. I can't see any need to check `completed_processing()`.

kitsune/users/utils.py
```diff
@@ -159,3 +165,25 @@ def user_is_contributor(user):
...
+    publisher.register_listener(PostListener())
+    publisher.register_listener(KBThreadListener())
+    publisher.register_listener(KBPostListener())
+    publisher.register_listener(AAQListener())
+    publisher.register_listener(DocumentListener())
+
+    publisher.publish()
+
+    # if all listeners have completed processing, delete the user
+    if publisher.completed_processing():
```
escattone, line 188: Per my comment above, I don't see any need for this call to `completed_processing()`. As far as I can tell, none of the work of any of the listeners called within `publish()` is ever done asynchronously.

escattone (commented):
Thanks @akatsoulas, there is lots of good stuff here, but I have some questions.

kitsune/sumo/handlers.py
akatsoulas: True but the problem lies in the signature of `AccountHandler`. This defines an extra method `handle_account`. I did that in order to keep the `AbstractHandler` generic for future use and create a specific handler for accounts. Because of this, mypy will raise an error for the `AAQChain(AbstractChain[AccountHandler])` definition if the `AbstractChain` is specific to `AbstractHandler`

kitsune/messages/models.py
```diff
@@ -5,7 +5,6 @@
 
 from kitsune.sumo.models import ModelBase
 
-
```
escattone, line 8: Super nit. I think we do want this to be two empty lines after the imports?

escattone (approved):
This looks great @akatsoulas! I have one microscopic nit for you to consider.

kitsune/messages/models.py
akatsoulas: This is done automatically by `black`. It's going to introduce two lines after the imports and before the class definition but it won't change it for variable definitions.

PR #6511, opened 2025-02-14
record failed deflections

escattone:
mozilla/sumo#2147

## Notes
- I did confirm that when a user transitions from anonymous to signed-in, Django retains the session data from the anonymous session within the new signed-in session.
- Provides deflection reporting in GA4 via the `question_submit` event with its new `is_failed_deflection` event parameter, which can be `"true"` or `"false"`. This covers both Zendesk tickets and community-forum questions.
- Also provides reporting via the `is_failed_deflection` boolean field on question instances. This only covers community-forum questions. Do we want to track failed deflections for Zendesk tickets in the database as well? If we go with this PR, we should remember to add the `is_failed_deflection` boolean field to the `kitsune_questions_plus` table in BQ as a follow-up task.


kitsune/questions/views.py
```diff
@@ -680,6 +691,22 @@ def aaq(request, product_slug=None, step=1, is_loginless=False):
             content_type=user_ct,
         ).order_by("-id")[:IMG_LIMIT]
 
+        if form.is_bound and (topic := form.cleaned_data.get("category")):
+            # We've got invalid POST data, but the topic has been provided.
+            # Let's set the proper GA4 event parameters on the submit button.
+            context["submit_event_parameters"] = json.dumps(
+                {
+                    "topics": f"/{topic.slug}/",
```
akatsoulas, line 699: You could use a helper function to get the GA4 params in all 3 places that's needed.

kitsune/questions/models.py
```diff
@@ -122,6 +122,7 @@ class Question(AAQBase):
     )
     is_locked = models.BooleanField(default=False)
     is_archived = models.BooleanField(default=False, null=True)
+    is_failed_deflection = models.BooleanField(default=False)
```
akatsoulas, line 125: We could avoid this by having something like a `@property` and store a key about failed deflections in the question metadata. We can use the add_metadata method for this.

akatsoulas: We should also make it clear somehow (in the variable name, GA4 events, something else?) that this is related to the first submissions of the question and it's measured against the knowledge base. Question threads also deflect. A thread that has this attribute set to True, can potentially deflect many users from submitting more questions.

kitsune/wiki/utils.py
```diff
@@ -290,3 +294,58 @@ def build_topics_data(request: HttpRequest, product: Product, topics: list[Topic
         topics_data.append(topic_data)
 
     return topics_data
+
```
akatsoulas, line 297: Let's add annotations to these methods.

kitsune/wiki/utils.py
```diff
@@ -290,3 +294,58 @@ def build_topics_data(request: HttpRequest, product: Product, topics: list[Topic
...
+    now = time.time()
+
+    # Remove any slugs that have expired.
+    if expired_slugs := [slug for slug, ts in kb_visited.items() if (now - ts) > ttl]:
+        for slug in expired_slugs:
+            del kb_visited[slug]
+        session.modified = True
+
+    return kb_visited
```
akatsoulas, line 315: Probably it's cleaner if this method doesn't return anything and it just deletes (any) entries

kitsune/wiki/utils.py
```diff
@@ -290,3 +296,70 @@ def build_topics_data(request: HttpRequest, product: Product, topics: list[Topic
         topics_data.append(topic_data)
 
     return topics_data
+
+
+def clean_kb_visited(session: SessionBase, ttl: int = KB_VISITED_DEFAULT_TTL) -> None:
```
akatsoulas, line 301: should this be renamed to something that will indicate that it cleans expired slugs? I was confused a couple of times by the call of this function and I had to check the code to understand that it doesn't clean all slugs

kitsune/wiki/utils.py
```diff
@@ -290,3 +296,70 @@ def build_topics_data(request: HttpRequest, product: Product, topics: list[Topic
...
+    Cleans and then updates the "kb-visited" dictionary within the given session.
+    """
+    if not session:
+        return
+
+    if "kb-visited" in session:
+        clean_kb_visited(session, ttl=ttl)
+        kb_visited = session["kb-visited"]
+    else:
+        kb_visited = session["kb-visited"] = {}
```
akatsoulas, line 332: this could be greatly simplified by using:

```python
clean_kb_visited(session, ttl=ttl)
kb_visited = session.get("kb-visited", {})
```

kitsune/wiki/utils.py
```diff
@@ -290,3 +296,70 @@ def build_topics_data(request: HttpRequest, product: Product, topics: list[Topic
...
+    Does the given session indicate that the user has visited at least one
+    KB article associated with the given product and, if provided, topic,
+    within the given TTL.
+    """
+    if not (session and ("kb-visited" in session)):
+        return []
+
+    clean_kb_visited(session, ttl=ttl)
+
+    if not (visited_slugs := tuple(session["kb-visited"].keys())):
```
akatsoulas, line 355: why are you using `tuple` here?

kitsune/wiki/utils.py
```diff
@@ -290,3 +296,70 @@ def build_topics_data(request: HttpRequest, product: Product, topics: list[Topic
...
+        return []
+
+    qs = Document.objects.filter(
```
akatsoulas, line 358: Since we have the slugs in the session, we could just store them in the metadata without querying the database. We can query when and only if we need them. Returning the list of slugs as strings should be enough?

kitsune/wiki/utils.py
```diff
@@ -290,3 +296,70 @@ def build_topics_data(request: HttpRequest, product: Product, topics: list[Topic
...
+        return
+
+    if "kb-visited" in session:
+        clean_kb_visited(session, ttl=ttl)
+        kb_visited = session["kb-visited"]
+    else:
+        kb_visited = session["kb-visited"] = {}
+
+    # Note that we have visited the given KB article.
+    kb_visited[doc.original.slug] = time.time()
```
akatsoulas, line 335: suggestion: We could have the kb_visited dictionary to hold list of slugs per topic.

`kb_visited {'topic' : [slug1, slug2]`

kitsune/wiki/utils.py
escattone: For no reason other than I don't need the mutability of a list.

akatsoulas (approved):
r+wc

kitsune/wiki/utils.py
escattone: Good point. I went with the following because I need to insert the key/value in the session if the key isn't present.
```python
    remove_expired_from_kb_visited(session, ttl=ttl)
    kb_visited = session.setdefault("kb-visited", {})
```

kitsune/wiki/utils.py
escattone: I renamed to `remove_expired_from_kb_visited`.

kitsune/wiki/utils.py
escattone: I've modified this to avoid the need for any DB query.

kitsune/wiki/utils.py
escattone: I've modified this considerably with something in the same vein as you're suggesting (but I had to include the products as well).

PR #6514, opened 2025-02-18
99 - Long user watch list causes page loading issues

smithellis:
* Added pagination to view and template


escattone (approved):
r+wc

Pagination depends on a consistent ordering of the individual items, but since the ordering is currently by `content_type` only, it's possible that the individual items aren't consistently ordered -- since for example a user could be watching hundreds of the question content type and there would be no consistent ordering of the hundreds of individual questions. I would recommend changing the `order_by("content_type")` to `order_by("content_type", "id")` so that ordering is guaranteed at the level of individual items.

akatsoulas (approved):

PR #6516, opened 2025-02-20
Improve database queries to prevent 500 error on /users/watches

smithellis:
* Use `select_related` and `prefetch_related` to avoid extra queries

### DIFF
--- a/kitsune/users/views.py
+++ b/kitsune/users/views.py
@@ -297,19 +297,22 @@ def edit_watch_list(request):
         .order_by("content_type", "id")
     )
 
-    watch_list = [
-        item
-        for item in watches
-        if item.content_object is not None
-        and (item.content_type.name != "question" or not item.content_object.is_archived)
-    ]
+    watch_list = []
+    for item in watches:
+        if item.content_object is None:
+            continue
+        if item.content_type.name == "question":
+            if not item.content_object.is_archived:
+                watch_list.append(item)
+        else:
+            watch_list.append(item)
 
     watch_list = paginate(request, watch_list)
 
     if request.method == "POST":
         for item in watch_list:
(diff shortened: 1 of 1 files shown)


kitsune/users/views.py
```diff
@@ -290,17 +290,19 @@ def edit_contribution_area(request):
...
-                if not item.content_object.is_archived:
-                    watch_list.append(item)
-            else:
-                watch_list.append(item)
+    watches = (
+        Watch.objects.filter(user=request.user)
+        .select_related("content_type")
+        .prefetch_related("content_object")
+        .order_by("content_type", "id")
+    )
```
akatsoulas, line 298: Probably is more efficient to exclude from fetching items that you don't care about.

```python
watches = (
    Watch.objects.filter(
        user=request.user
    )
    .exclude(content_object__isnull=True) 
    .exclude(Q(content_type__model="question") & Q(content_object__is_archived=True))
    .select_related("content_type")
    .prefetch_related("content_object")
    .order_by("content_type", "id")
)

watch_list = paginate(request, watches)
```

akatsoulas (approved):
r+wc

kitsune/users/views.py
akatsoulas: the reason for the Q object is to ensure that it will filter out `is_archived` attributes only for questions. Otherwise it will potentially raise AttributeError for content_types that don't have that attribute defined in their model

smithellis: That's a good catch. Thanks @akatsoulas

smithellis: The only requested optimization I could keep was adding the is_null exclusion and removing the later check. I also noticed a potential issue with the POST content making multiple saves vs. a single bulk update, and altered that.

akatsoulas (approved):

PR #6523, opened 2025-02-23
improved email rate limiting

escattone:
mozilla/sumo#2215

### DIFF
--- a/kitsune/lib/email.py
+++ b/kitsune/lib/email.py
@@ -7,8 +7,6 @@ from django.core.mail.backends import base, smtp
 from django.utils.module_loading import import_string
 from sentry_sdk import capture_exception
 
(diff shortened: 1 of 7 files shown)


akatsoulas:
I haven't reviewed this yet but I did come across the following libraries that may be of interest to us
* https://github.com/ui/django-post_office
* https://github.com/django-ses/django-ses

escattone:
@akatsoulas I love the email backend of `django_ses`. I think it's much more reliable than the vanilla Django SMTP backend. However, the rate-limiting functionality of `django_ses.EmailBackend` won't work in an environment like ours where emails can be simultaneously sent from multiple Celery workers, because it assumes that only one instance of `django_ses.EmailBackend` is sending emails at a time. So, we still need this PR no matter what backend we decide upon.

PR #6536, opened 2025-03-03
2046 - Mulitproduct AAQ CTA URL

smithellis:
Addresses https://github.com/mozilla/sumo/issues/2046

The AAQ CTA in the sidebar would send the user to the new question page for the first product on documents where there were multiple products. 

Now we set `multiple_products` in the `aaq_context` and if there are multiple products we will send the user to `/questions/new` so they can select a product.

### DIFF
--- a/kitsune/wiki/views.py
+++ b/kitsune/wiki/views.py
@@ -238,7 +238,7 @@ def document(request, document_slug, document=None):
         product = products.first()
 
     # Set the AAQ context for the widget
-    set_aaq_context(request, product, multiple_products=True if len(products) > 1 else False)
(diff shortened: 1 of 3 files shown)


escattone (approved):
r+wc

kitsune/wiki/views.py
```diff
@@ -238,7 +238,7 @@ def maybe_vary_on_accept_language(response):
         product = products.first()
 
     # Set the AAQ context for the widget
-    set_aaq_context(request, product)
+    set_aaq_context(request, product, multiple_products=True if len(products) > 1 else False)
```
escattone, line 241: This can just be `multiple_products=len(products) > 1`.

PR #6537, opened 2025-03-04
Change SumoBot to deleted_user

smithellis:
SumoBot isn't a meaningful name in the messaging system for a user who is deleted, and could cause confusion.

- Messages from or to a deleted user will now be from or to `deleted user`.

https://github.com/mozilla/sumo/issues/2225

### DIFF
--- a/kitsune/messages/handlers.py
+++ b/kitsune/messages/handlers.py
@@ -13,8 +13,16 @@ class MessageListener(UserDeletionListener):
         When a user is deleted:
         - Delete their outbox messages
         - Keep inbox messages for other users and reassign them to SumoBot
+        - Update outbox message recipients to SumoBot where the deleted user was a recipient
         """
         sumo_bot = Profile.get_sumo_bot()
         InboxMessage.objects.filter(to=user).delete()
         InboxMessage.objects.filter(sender=user).update(sender=sumo_bot)
+
+        # Update outbox messages where the deleted user was a recipient
+        outbox_messages = OutboxMessage.objects.filter(to=user)
+        for message in outbox_messages:
+            message.to.remove(user)
+            message.to.add(sumo_bot)
+
         OutboxMessage.objects.filter(sender=user).delete()
--- a/kitsune/messages/jinja2/messages/inbox.html
+++ b/kitsune/messages/jinja2/messages/inbox.html
@@ -1,5 +1,5 @@
 {% extends "messages/base.html" %}
-{% from "messages/includes/macros.html" import avatar_link, name_link %}
+{% from "messages/includes/macros.html" import avatar_link, display_user %}
 {% set title = _('Inbox') %}
 {% set crumbs = [(url('messages.inbox'), _('Messages')),
                  (None, title)] %}
@@ -37,7 +37,7 @@
                     {{ avatar_link(message.sender, default_avatar) }}
                 </div>
                 <div class="email-cell from">
-                    {{ name_link(message.sender) }}
+                    {{ display_user(message.sender) }}
                 </div>
                 <div class="email-cell date">
                     {{ datetimeformat(message.created) }}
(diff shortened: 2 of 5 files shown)


kitsune/messages/jinja2/messages/inbox.html
```diff
@@ -37,7 +37,11 @@ <h1 class="sumo-page-heading">{{ title }}</h1>
                     {{ avatar_link(message.sender, default_avatar) }}
                 </div>
                 <div class="email-cell from">
+                  {% if message.sender.username == 'SumoBot' %}
```
akatsoulas, line 40: I would move the logic in the view and return a flag like `is_bot_profile` or something similar. The template could be just a boolean check.

akatsoulas: It looks like you can also pass a name to the `name_link` macro which can be helpful in this case or you can update the macros used to incorporate this logic

smithellis: I think catching it here in the main templates vs. the macro is best - the macro sets up links to user profiles, and we don't want to do that with `deleted user`. Moving the logic there just pollutes that macro, IMHO.

I'll look at sending a flag from the view vs. using the username here.

smithellis:
I moved the logic to the view, and appended `is_bot_user` to each user to designate the bot user(s).

akatsoulas:
@smithellis with the #6559 merged, this needs an update

kitsune/messages/views.py
```diff
@@ -179,13 +197,37 @@ def preview_async(request):
...
+    """
     # Set the counts based on the lists
     msg.recipients_count = msg.to.all().count()
     msg.to_groups_count = msg.to_group.all().count()
 
     # Assign the recipient based on the number of recipients
     msg.recipient = msg.to.all()[0] if msg.recipients_count == 1 else None
 
+    # Set is_bot_user flag only if there's a recipient with a username
+    if msg.recipient and hasattr(msg.recipient, "username"):
```
akatsoulas, line 228: There is already a property in the `Profille` model that checks for system accounts, `is_system_account`

akatsoulas (changes requested):

kitsune/messages/handlers.py
```diff
@@ -13,8 +13,16 @@ def on_user_deletion(self, user: User) -> None:
...
         """
         sumo_bot = Profile.get_sumo_bot()
         InboxMessage.objects.filter(to=user).delete()
         InboxMessage.objects.filter(sender=user).update(sender=sumo_bot)
+
+        # Update outbox messages where the deleted user was a recipient
+        outbox_messages = OutboxMessage.objects.filter(to=user)
+        for message in outbox_messages:
+            message.to.remove(user)
+            message.to.add(sumo_bot)
```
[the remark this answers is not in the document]
smithellis: If userA sent a message to userB and userB gets deleted, the message in userA's Outbox needs to have a 'To' person - thus the reassignment.

kitsune/messages/jinja2/messages/includes/macros.html
```diff
@@ -12,7 +12,11 @@
 
 {% macro name_link(user=None, name=None) -%}
   {% if user -%}
-    <a rel="nofollow" href="{{ profile_url(user) }}">{% if name %}{{ name }}{% else %}{{ display_name(user) }}{% endif %}</a>
+    {% if user.profile.is_system_account %}
```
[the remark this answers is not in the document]
smithellis: Extracted!

kitsune/messages/jinja2/messages/outbox.html
```diff
@@ -38,9 +38,13 @@ <h1 class="sumo-page-heading">{{ title }}</h1>
                 <div class="email-cell sent">{{ datetimeformat(message.created) }}</div>
                 <div class="email-cell to">
                 {% for user in message.to.all()[:1] -%}
-                    <a rel="nofollow" href="{{ profile_url(user) }}">
+                    {% if user.profile.is_system_account %}
```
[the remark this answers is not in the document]
smithellis: Changed.

akatsoulas (approved):
Looks good! Please squash into a single commit before merging

PR #6549, opened 2025-03-10
Fix n+1 query

smithellis:
In this part of the code:

`for obj in objects:
    question = obj.content_object
    obj.available_topics = get_hierarchical_topics(question.product)
    obj.available_tags = available_tags
    obj.saved_tags = question.tags.values_list("id", flat=True)`

We had `obj.saved_tags` executing a query for each questions tags. Now we `prefetch_related` in the initial query and hopefully avoid.

### DIFF
--- a/kitsune/flagit/views.py
+++ b/kitsune/flagit/views.py
@@ -193,7 +193,7 @@ def moderate_content(request):
             product_slug=product_slug,
         )
         .select_related("content_type", "creator", "assignee")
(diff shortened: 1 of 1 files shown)


akatsoulas (approved):

miorahmadhidayat:
👍 thanks for your concern, thanks for your help, thank for everything nice
to me. Thanks friend, bye bye MIOR

PR #6550, opened 2025-03-10
Skip write op in read only

smithellis:
Add `@skip_if_read_only_mode` to `_rebuild_kb_chunk` task

Prevents task from executing during read-only operations like document deletion,
avoiding transaction errors.

### DIFF
--- a/kitsune/wiki/tasks.py
+++ b/kitsune/wiki/tasks.py
@@ -214,6 +214,7 @@ def rebuild_kb():
 
 
 @shared_task(rate_limit="5/m")
+@skip_if_read_only_mode
 def _rebuild_kb_chunk(data):
     """Re-render a chunk of documents.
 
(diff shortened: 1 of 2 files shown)


kitsune/wiki/tasks.py
```diff
@@ -214,6 +214,7 @@ def rebuild_kb():
 
 
 @shared_task(rate_limit="5/m")
+@skip_if_read_only_mode
```
akatsoulas, line 217: this is called by `rebuild_kb` task which already has the decorator. Maybe that's not the cause of the sentry error?

smithellis: Thanks - sorry, I had pulled this back into In Progress status for this very reason. Revision added.

akatsoulas (approved):

PR #6557, opened 2025-03-12
Optmize email normalization

akatsoulas:
Avoids to call the normalzie_gmail function twice

### DIFF
--- a/kitsune/sumo/tests/test_email_utils.py
+++ b/kitsune/sumo/tests/test_email_utils.py
@@ -3,8 +3,8 @@ from unittest.mock import patch
 from django.conf import settings
 from django.contrib.sites.models import Site
 from django.core.mail import EmailMultiAlternatives
-from django.utils.functional import lazy
 from django.utils import translation
+from django.utils.functional import lazy
 from django.utils.translation import get_language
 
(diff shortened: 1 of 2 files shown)


kitsune/sumo/tests/test_email_utils.py
```diff
@@ -156,3 +156,20 @@ def test_send_messages(self, mock_mail):
...
+                "Testing",
+                from_email,
+                ["valid@example.com", "invalid@email", "another.valid@example.com"],
+            ),
+        ]
+        send_messages(messages)
+        send_messages_mock = mock_mail.get_connection().__enter__().send_messages
+        send_messages_mock.assert_called_once_with(messages)
+        # Check that the invalid email was removed
+        self.assertEqual(messages[0].to, ["valid@example.com", "another.valid@example.com"])
```
akatsoulas, line 175: `another.valid@example.com` is a valid email address. `invalid.@example.com` is not. The need for the gmail normalization is because for gmail addresses with dots (.) are the same. This is not true for other systems

escattone: FWIW, the original test above did test an invalid email address ("ringo").

PR #6558, opened 2025-03-12
only save when needed within `is_taken`

escattone:
This should resolve the continuous loop of saves and consequent ES re-indexing caused by #6555.

### DIFF
--- a/kitsune/questions/models.py
+++ b/kitsune/questions/models.py
@@ -708,10 +708,10 @@ class Question(AAQBase):
         this will reset the database fields and return False.
         """
         if self.taken_by is None or self.taken_until is None or self.taken_until < datetime.now():
-
-            self.taken_by = None
-            self.taken_until = None
-            self.save()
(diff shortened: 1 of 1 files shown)


kitsune/questions/models.py
```diff
@@ -708,10 +708,10 @@ def is_taken(self):
         this will reset the database fields and return False.
         """
         if self.taken_by is None or self.taken_until is None or self.taken_until < datetime.now():
-
-            self.taken_by = None
-            self.taken_until = None
-            self.save()
+            if (self.taken_by is not None) or (self.taken_until is not None):
```
smithellis, line 711: I see your are dropping the datetime.now() check - is that not needed?

smithellis (changes requested):

smithellis (approved):

belikin05:
Copyright © 2025 Apple Inc. Todos los derechos reservado

PR #6559, opened 2025-03-13
System account

akatsoulas:


escattone (changes requested):
This is beautifully done! ❤️ 🎉

I have a few comments sprinkled in with the code, but I'll also add some here as well.

- My understanding is that the profile of a system account should be viewable. So the `kitsune.users.views.profile` view needs to be modified such the user and profile queries use the `all_users` and `all_profiles` managers instead of the default ones.

- The users form field (`MultiUsernameField`) of the `kitsune.wiki.forms.RevisionFilterForm` needs to filter based on the `all_users` and `all_profiles` managers, since the system account user will show up on the [recent revisions page](https://support.allizom.org/en-US/kb/revisions), and we should be able to filter it like any other user or user profile. Can we add an option/setting to `kitsune.sumo.form_fields.MultiUsernameField` that determines whether it'll use the `all_users` and `all_profiles` managers instead of the default ones.

- I think we can roll out this PR with its DB migration in one go? At least it seems that way to me. The DB migration will be completed prior to the code deployment, and I don't think the DB migration will cause any prior code running in the soon-to-be-terminated pods to fail?

- While reviewing this PR, I realized that we also need to update a revision's `reviewer` with the system account within the `kitsune.wiki.handlers.DocumentListener` class.

Than3395 (approved):

Than3395 (approved):

akatsoulas:
@escattone can you have another pass? I haven't updated the manger for `MultiUsernameField` yet. My hope is that we won't end up with SumoBot owning any revisions. If this is not the case, we can update it in a separate PR

PR #6560, opened 2025-03-13
Deleting a user modal won't close and page won't redirect

smithellis:
When users attempted to delete their account, the form would submit but the modal would only shrink and empty instead of properly redirecting to the /users/close_account URL. This was caused by an interaction between the modal system(s) and the form submission process.

This fix:
- Handles the delete button click event directly
- Ensures both modal systems (Mzp.Modal and $.kbox) are properly closed
- Adds a small delay before programmatically submitting the form to allow modal closing to complete
- Prevents the default button submission behavior to take control of the submission flow

### DIFF
--- a/kitsune/sumo/static/sumo/js/users.js
+++ b/kitsune/sumo/static/sumo/js/users.js
@@ -25,8 +25,33 @@
     });
   }
 
+  function handleAccountDeletion() {
+    // Handle the delete account button click 
+    $('#delete-profile-button').on('click', function(e) {
+      e.preventDefault();
+      
+      var $form = $(this).closest('form');
+      
+      // Close modals - keep both systems for compatibility
+      if (typeof Mzp !== 'undefined' && Mzp.Modal) {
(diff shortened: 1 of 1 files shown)


Than3395 (approved):

escattone:
@smithellis I'm finally getting to this review. It seems that we no longer use the `#delete-profile-button` button?

escattone (commented):
@smithellis I don't like the way this worked in the old flow. It seems to me that there are too many moving parts/dependencies for what it does. I'm wondering if we could replace the entire old stack -- the jQuery-based JS, the protocol modal dependency -- with a simpler `htmx`-based flow?

smithellis:
@escattone I don't care much for the existing flow either, but this PR addresses the issue as defined. If we want to consider a different solution altogether I think that will need to be separate work/separate ticket.

escattone (approved):
I think the entire flow involved in closing an account is overly complicated, and I'd love to see it changed, but this PR solves the immediate problem.

PR #6565, opened 2025-03-14
improved PostListener

escattone:
mozilla/sumo#2246


escattone:
Closing this, since we decided that the current behavior of setting `updated_by` to `NULL` is the preferred approach.

PR #6567, opened 2025-03-17
Delete non-approved revisions on user deletion

akatsoulas:
* Delete docs without revisions on user deletion

### DIFF
--- a/kitsune/wiki/handlers.py
+++ b/kitsune/wiki/handlers.py
@@ -20,6 +20,14 @@ class DocumentListener(UserDeletionListener):
                 document.contributors.add(*content_group.user_set.all())
             document.contributors.remove(user)
 
+        non_approved_revisions = Revision.objects.filter(creator=user, is_approved=False)
+        # Delete documents with no approved revisions
+        Document.objects.filter(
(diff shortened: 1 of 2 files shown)


kitsune/wiki/handlers.py
```diff
@@ -19,5 +19,15 @@ def on_user_deletion(self, user: User) -> None:
                 document.contributors.add(*content_group.user_set.all())
             document.contributors.remove(user)
 
+        non_approved_revisions = Revision.objects.filter(creator=user, is_approved=False)
+
+        for revision in non_approved_revisions:
```
escattone, line 24: Could this loop be replace with the following instead?
```python
# First delete all of the docs that have no current revision but one or more of the users' unapproved revisions.
Document.objects.filter(current_revision__isnull=True, revisions__in=non_approved_revisions).delete()
# Next, delete all of the users' unapproved revisions that were not cascade deleted above.
non_approved_revisions.delete()
```

escattone (changes requested):

akatsoulas:
@escattone PR updated

escattone (approved):

PR #6574, opened 2025-03-19
Do not delete articles with approved revisions

akatsoulas:


escattone:
@akatsoulas I'm closing this only because your original work was fine. The root of this problem was [something else](https://github.com/mozilla/sumo/issues/2262#issuecomment-2738521734).

PR #6575, opened 2025-03-19
Lock a question when submitter is deleted

akatsoulas:


kitsune/questions/handlers.py
```diff
@@ -78,11 +79,28 @@ def run(self, user: User) -> None:
         for handler in self.handlers:
             handler.handle_account(data)
 
-        # Re-assign remaining questions and answers to SumoBot.
         sumo_bot = Profile.get_sumo_bot()
-        Question.objects.filter(creator=user).update(creator=sumo_bot)
+
+        questions_to_update = list(Question.objects.filter(creator=user))
+
+        Question.objects.filter(creator=user).update(creator=sumo_bot, is_locked=True)
```
escattone, line 86: Nit. I think you could simplify this with:
```suggestion
        questions_to_update = Question.objects.filter(creator=user)

        questions_to_update.update(creator=sumo_bot, is_locked=True)
```
because the loop on `questions_to_update` below will still work fine too.

escattone (approved):
r+wc

kitsune/questions/handlers.py
akatsoulas: True, I spent a few minutes thinking about this.  Given the update that we are doing, it's going to work fine but there's subtle bug lying around if we ever change the update to something that's not compatible  with the initial data because without the `list()` the query will be evaluated after the update. 

I will update but I will also add a comment for future reference.

akatsoulas: Actually it will not, `questions_to_update` will not be evaluated until the for loop. In the meantime, the `update()` op will have update directly in the DB the entries. When the questions_to_update will be evaluated in the `bulk_create` it will be an empty queryset b/c the data have been already re-assigned to sumobt

PR #6576, opened 2025-03-19
Add cleanup of old account events

smithellis:
This commit implements a data cleanup feature for account events:
- Adds migration to perform initial cleanup of account events older than 2 years
- Adds management command 'cleanup_old_account_events' for periodic cleanup
- Can be scheduled via crontab to run daily

The cleanup can be run manually with:
python manage.py cleanup_old_account_events

This `.delete()[0]` is an interesting thing I learned - delete returns a tuple. 

- The first item (index [0]) is the total number of objects deleted
- The second item (index [1]) is a dictionary that breaks down the number of deletions per object type


kitsune/users/migrations/0034_delete_old_account_events.py
```diff
@@ -0,0 +1,20 @@
+from django.db import migrations
+from django.utils import timezone
+from datetime import timedelta
+
+
+def delete_old_account_events(apps, schema_editor):
+    AccountEvent = apps.get_model('users', 'AccountEvent')
```
akatsoulas, line 7: In order to keep this code DRY and have a single place where we delete accounts,  you can use the `call_command` function to call the management command that you wrote

akatsoulas (approved):
Let's also add the cron entry to run once per week. r+wc

kitsune/users/migrations/0034_delete_old_account_events.py
```diff
@@ -0,0 +1,20 @@
+from django.db import migrations
+from django.utils import timezone
+from datetime import timedelta
+
+
+def delete_old_account_events(apps, schema_editor):
```
akatsoulas, line 6: let's match the name with the one of the management command

kitsune/users/migrations/0034_delete_old_account_events.py
smithellis: Done

kitsune/users/migrations/0034_delete_old_account_events.py
smithellis: Done

PR #6577, opened 2025-03-20
Properly handle non-approved revision deletions

akatsoulas:


kitsune/wiki/handlers.py
```diff
@@ -11,30 +12,41 @@ class DocumentListener(UserDeletionListener):
...
+        sumo_bot = Profile.get_sumo_bot()
         content_group = Group.objects.get(name=settings.SUMO_CONTENT_GROUP)
+
+        # Handle approved revisions first to avoid cascade deletion issues
+        Revision.objects.filter(creator=user, is_approved=True).update(creator=sumo_bot)
+
+        # Update reviewer and readied_for_localization_by fields
+        Revision.objects.filter(Q(reviewer=user) | Q(readied_for_localization_by=user)).update(
+            reviewer=sumo_bot, readied_for_localization_by=sumo_bot
+        )
```
escattone, line 32: These have to be separate updates. There are many revisions where the `reviewer` field is non-NULL, but the `readied_for_localization_by` is NULL. So, in those cases, if the `reviewer` is the deleted user, the `readied_for_localization_by` field will be incorrectly updated to be the sumo bot as well, when it should have remained NULL.

kitsune/wiki/handlers.py
```diff
@@ -11,30 +12,41 @@ class DocumentListener(UserDeletionListener):
...
+        # Anonymize the user's votes
         HelpfulVote.objects.filter(creator=user).update(
             creator=None, anonymous_id=AnonymousIdentity().anonymous_id
         )
+
+        # Finally, delete non-approved revisions
+        Document.objects.filter(
+            current_revision__isnull=True,
+            revisions__creator=user,
+            revisions__is_approved=False,
```
escattone, line 50: This `revisions__is_approved=False` is not needed, because it's already covered by the `current_revision__isnull=True` -- or, we could keep it, and get rid of the `current_revision__isnull=True`.

kitsune/wiki/handlers.py
```diff
@@ -11,30 +12,41 @@ class DocumentListener(UserDeletionListener):
...
+        ).exclude(revisions__creator__in=User.objects.exclude(id=user.id)).delete()
```
escattone, line 51: Nice! This is the line that fixes mozilla/sumo#2264.

kitsune/wiki/handlers.py
escattone: I think we should add another case to the `test_revision_reviewer_replacement` test that covers the above, either in this PR or I can add it in another PR. Something like:

```python
def test_revision_reviewer_replacement(self):
    ...
    # This revision has a reviewer, but no "readied_for_localization_by".
    rev4 = ApprovedRevisionFactory(creator=contributor, reviewer=self.user)
    ...
    self.assertEqual(rev4.creator.username, contributor.username)
    self.assertEqual(rev4.reviewer.username, settings.SUMO_BOT_USERNAME)
    # The "readied_for_localization_by" field should remain None.
    self.assertIsNone(rev4.readied_for_localization_by)
```

escattone (changes requested):

escattone:
This doesn't fix mozilla/sumo#2262, because that one is due to cascade deletions via the `based_on` field. I'll create a PR to resolve that.

escattone:
@akatsoulas I'll make the final tweak I suggested above and merge this.

escattone (approved):

PR #6578, opened 2025-03-20
Delete threads with no replies

akatsoulas:


escattone (approved):
Looks great, thanks @akatsoulas!

PR #6582, opened 2025-03-21
First pass of deletions for non-fxa-migrated users

smithellis:
Delete inactive non-FxA-migrated user accounts

This migration removes user accounts that meet both criteria:
1. Have not been migrated to Firefox Accounts (is_fxa_migrated=False)
2. Have no content or system relationships (no posts, votes, messages, etc.)

This is a one-way migration as deleted accounts cannot be restored.

### DIFF
--- a/kitsune/users/migrations/0033_batch_delete_non_migrated_users.py
+++ b/kitsune/users/migrations/0033_batch_delete_non_migrated_users.py
@@ -1,0 +1,55 @@
+from django.db import migrations
+from django.db.models import Q
+
+
+def delete_non_migrated_users(apps, schema_editor):
+    """
+    Delete users where is_fxa_migrated is False and who aren't creators/owners/users of
+    any content or relationships in the system
+    """
+    User = apps.get_model('auth', 'User')
+    
+    users_to_delete = User.objects.filter(
+        profile__is_fxa_migrated=False
+    ).exclude(
+        Q(answer_votes__isnull=False) |
+        Q(answers__isnull=False) |
+        Q(award_creator__isnull=False) |
+        Q(badge__isnull=False) |
+        Q(created_revisions__isnull=False) |
+        Q(gallery_images__isnull=False) |
+        Q(gallery_videos__isnull=False) |
+        Q(inboxmessage__isnull=False) |
+        Q(outbox__isnull=False) |
+        Q(poll_votes__isnull=False) |
+        Q(post__isnull=False) |
+        Q(question_votes__isnull=False) |
+        Q(questions__isnull=False) |
+        Q(readied_for_l10n_revisions__isnull=False) |
+        Q(reviewed_revisions__isnull=False) |
+        Q(thread__isnull=False) |
+        Q(wiki_post_set__isnull=False) |
+        Q(wiki_thread_set__isnull=False)
+    )
+    
+    users_to_delete.delete()
+
+
+def reverse_migration(apps, schema_editor):
+    """
+    No reverse migration possible since deletion cannot be undone
+    """
+    pass
+
+
+class Migration(migrations.Migration):
+    dependencies = [
+        ('users', '0032_profile_account_type_alter_profile_user'),
+    ]
+
+    operations = [
+        migrations.RunPython(
+            delete_non_migrated_users,
+            reverse_migration,
+        ),
+    ] 
--- a/kitsune/users/migrations/0033_batch_delete_non_migrated_users.py
+++ b/kitsune/users/migrations/0033_batch_delete_non_migrated_users.py
@@ -5,7 +5,7 @@ from django.db.models import Q
 def delete_non_migrated_users(apps, schema_editor):
     """
     Delete users where is_fxa_migrated is False and who aren't creators/owners/users of
-    any content or relationships in the system
+    any important content
     """
     User = apps.get_model('auth', 'User')
     
--- a/kitsune/users/tests/test_migration_0033.py
+++ b/kitsune/users/tests/test_migration_0033.py
@@ -1,0 +1,110 @@
+from django.contrib.auth.models import User
+from django.test import TestCase
+from django.db.models import Q
+
+from kitsune.questions.tests import (
+    AnswerFactory,
+    QuestionFactory,
+    QuestionVoteFactory,
+    AnswerVoteFactory,
+)
+from kitsune.users.tests import ProfileFactory
+from kitsune.wiki.tests import RevisionFactory
+from kitsune.messages.models import InboxMessage, OutboxMessage
+from kitsune.gallery.tests import ImageFactory, VideoFactory
+from kitsune.forums.tests import ThreadFactory, PostFactory
+
+
+class TestDeleteNonMigratedUsersMigrationQuery(TestCase):
+    """Test the migration that deletes non-migrated users."""
+
+    def setUp(self):
+        # Create users with different combinations of migration status and content creation
+
+        # Case 1: Non-migrated user with no content (should be deleted)
+        ProfileFactory(user__username="user1", is_fxa_migrated=False)
+
+        # Case 2: Non-migrated user with a question (should be kept)
+        user2 = ProfileFactory(user__username="user2", is_fxa_migrated=False).user
+        QuestionFactory(creator=user2)
+
+        # Case 3: Non-migrated user with an answer (should be kept)
+        user3 = ProfileFactory(user__username="user3", is_fxa_migrated=False).user
+        AnswerFactory(creator=user3)
+
+        # Case 4: Non-migrated user with a revision (should be kept)
+        user4 = ProfileFactory(user__username="user4", is_fxa_migrated=False).user
+        RevisionFactory(creator=user4)
+
+        # Case 5: Migrated user with no content (should be kept)
+        ProfileFactory(user__username="user5", is_fxa_migrated=True)
+
+        # Case 6: Non-migrated user with a question vote (should be kept)
+        user6 = ProfileFactory(user__username="user6", is_fxa_migrated=False).user
+        QuestionVoteFactory(creator=user6)
+
+        # Case 7: Non-migrated user with an answer vote (should be kept)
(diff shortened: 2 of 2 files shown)


kitsune/users/migrations/0033_batch_delete_non_migrated_users.py
```diff
@@ -0,0 +1,46 @@
...
+
+
+def delete_non_migrated_users(apps, schema_editor):
+    """
+    Delete users where is_fxa_migrated is False and who aren't creators of 
+    any revision, question, or answer, and who don't have any votes or messages
+    """
+    User = apps.get_model('auth', 'User')
+    
+    with transaction.atomic():
```
escattone, line 12: This is not needed. Django migrations are performed atomically by default.

kitsune/users/migrations/0033_batch_delete_non_migrated_users.py
```diff
@@ -0,0 +1,46 @@
...
+        users_to_delete = User.objects.filter(
+            profile__is_fxa_migrated=False
+        ).exclude(
```
escattone, line 15: You're missing a lot of other related foreign-key relationships here, like `reviewed_revisions` for example. If you don't exclude them as well, you could end up deleting a lot of content via their `on_delete=models.CASCADE`. You'll need to thoroughly review the code for all foreign-key relationships to the user model, and determine which ones need to be added.

escattone (changes requested):

kitsune/users/migrations/0033_batch_delete_non_migrated_users.py
smithellis: Thanks!

kitsune/users/migrations/0033_batch_delete_non_migrated_users.py
escattone: You just want to make sure that the foreign-key relationships with `on_delete=models.CASCADE` are taken care of, so deleting the user won't trigger the deletion of content that we don't want to be deleted. By "taken care of", I mean either:
- Don't delete the user if the relationship exists.
- Or, replaced the user with something else that makes sense, like `None` or the Sumo Bot or maybe something else that makes more sense.

escattone: I guess in this case, we're looking for users that we can just cleanly delete without any re-assignment involved, so I guess my second bullet point above doesn't apply.

kitsune/users/tests/test_migration_0033.py
```diff
@@ -0,0 +1,173 @@
...
+from kitsune.questions.tests import AnswerFactory, QuestionFactory
+from kitsune.questions.models import QuestionVote, AnswerVote
+from kitsune.users.tests import ProfileFactory
+from kitsune.wiki.tests import RevisionFactory, DocumentFactory
+from kitsune.messages.models import InboxMessage, OutboxMessage
+from kitsune.gallery.tests import ImageFactory, VideoFactory
+from kitsune.forums.tests import ThreadFactory, PostFactory
+
+
+class TestDeleteNonMigratedUsersMigration(TestCase):
```
escattone, line 14: I'm thinking that we should change/modify the name to suggest that we're not really testing the migration exactly as it is, but instead we're testing the query used within the migration.

kitsune/users/tests/test_migration_0033.py
```diff
@@ -0,0 +1,173 @@
...
+
+        # All other users should NOT be in this queryset
+        for i in range(2, 16):
+            username = f"user{i}"
+            self.assertFalse(
+                users_to_delete.filter(username=username).exists(),
+                f"User {username} should not be in the deletion queryset",
+            )
+
+    def test_migration_through_apps(self):
```
escattone, line 136: I don't think we need this test, since we're really just verifying the logic of the query that we use within the data migration.

escattone (changes requested):
@smithellis Your migration looks great to me, and your tests cover all of the cases, but I would recommend that some of the test code be simplified or removed (see notes below).

Once this is merged and then later applied in production, I think we could delete `test_migration_0033.py`, since it will have served its purpose.

kitsune/users/migrations/0033_batch_delete_non_migrated_users.py
```diff
@@ -0,0 +1,55 @@
+from django.db import migrations
+from django.db.models import Q
+
+
+def delete_non_migrated_users(apps, schema_editor):
+    """
+    Delete users where is_fxa_migrated is False and who aren't creators/owners/users of
+    any content or relationships in the system
```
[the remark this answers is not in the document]
smithellis: Done.

kitsune/users/tests/test_migration_0033.py
```diff
@@ -0,0 +1,173 @@
...
+        # Case 3: Non-migrated user with an answer (should be kept)
+        self.user3 = User.objects.create(username="user3")
+        ProfileFactory(user=self.user3, is_fxa_migrated=False)
+        AnswerFactory(creator=self.user3)
+
+        # Case 4: Non-migrated user with a revision (should be kept)
+        self.user4 = User.objects.create(username="user4")
+        ProfileFactory(user=self.user4, is_fxa_migrated=False)
+        doc = DocumentFactory()
+        RevisionFactory(document=doc, creator=self.user4)
```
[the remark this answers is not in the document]
smithellis: Fixed.

kitsune/users/tests/test_migration_0033.py
```diff
@@ -0,0 +1,173 @@
...
+
+        # Case 5: Migrated user with no content (should be kept)
+        self.user5 = User.objects.create(username="user5")
+        ProfileFactory(user=self.user5, is_fxa_migrated=True)
+
+        # Case 6: Non-migrated user with a question vote (should be kept)
+        self.user6 = User.objects.create(username="user6")
+        ProfileFactory(user=self.user6, is_fxa_migrated=False)
+        q = QuestionFactory()
+        QuestionVote.objects.create(question=q, creator=self.user6, anonymous_id="")
```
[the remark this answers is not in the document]
smithellis: Fixed.

kitsune/users/tests/test_migration_0033.py
```diff
@@ -0,0 +1,173 @@
...
+        ProfileFactory(user=self.user14, is_fxa_migrated=False)
+        thread = ThreadFactory()
+        PostFactory(thread=thread, author=self.user14)
+
+        # Case 15: Non-migrated user who reviewed a revision (should be kept)
+        self.user15 = User.objects.create(username="user15")
+        ProfileFactory(user=self.user15, is_fxa_migrated=False)
+        revision = RevisionFactory()
+        revision.reviewer = self.user15
+        revision.save()
```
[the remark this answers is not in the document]
smithellis: Fixed.

kitsune/users/tests/test_migration_0033.py
```diff
@@ -0,0 +1,173 @@
...
+        self.assertEqual(users_to_delete.count(), 1)
+        self.assertEqual(users_to_delete.first().username, "user1")
+
+        # All other users should NOT be in this queryset
+        for i in range(2, 16):
+            username = f"user{i}"
+            self.assertFalse(
+                users_to_delete.filter(username=username).exists(),
+                f"User {username} should not be in the deletion queryset",
+            )
```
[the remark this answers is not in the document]
smithellis: You are correct - fixed!

kitsune/users/tests/test_migration_0033.py
smithellis: I think I agree - dropped.

kitsune/users/tests/test_migration_0033.py
```diff
@@ -0,0 +1,110 @@
...
+        AnswerVoteFactory(creator=user7)
+
+        # Case 8: Non-migrated user who is a sender of inbox messages (should be kept)
+        user8 = ProfileFactory(user__username="user8", is_fxa_migrated=False).user
+        InboxMessage.objects.create(to=user2, sender=user8, message="test")
+
+        # Case 9: Non-migrated user with an outbox message (should be kept)
+        user9 = ProfileFactory(user__username="user9", is_fxa_migrated=False).user
+        outbox_msg = OutboxMessage.objects.create(sender=user9, message="test")
+        outbox_msg.to.add(user2)
```
escattone, line 57: Nit. I think you could just add `to=user2` to the `OutboxMessage.objects.create(...)` call?

kitsune/users/tests/test_migration_0033.py
```diff
@@ -0,0 +1,110 @@
...
+        VideoFactory(creator=user12)
+
+        # Case 13: Non-migrated user with a forum thread (should be kept)
+        user13 = ProfileFactory(user__username="user13", is_fxa_migrated=False).user
+        ThreadFactory(creator=user13)
+
+        # Case 14: Non-migrated user with a forum post (should be kept)
+        user14 = ProfileFactory(user__username="user14", is_fxa_migrated=False).user
+        thread = ThreadFactory()
+        PostFactory(thread=thread, author=user14)
```
escattone, line 78: Nit. You can just use `PostFactory(author=user14)` because the `PostFactory` will create the thread for you.

escattone (approved):
r+wc

kitsune/users/migrations/0033_batch_delete_non_migrated_users.py
```diff
@@ -0,0 +1,55 @@
...
+
+class Migration(migrations.Migration):
+    dependencies = [
+        ('users', '0032_profile_account_type_alter_profile_user'),
+    ]
+
+    operations = [
+        migrations.RunPython(
+            delete_non_migrated_users,
+            reverse_migration,
```
escattone, line 53: Oh, I forgot one more thing. Since this is not reversible, I think you should use `migrations.RunPython.noop` here instead.

akatsoulas (approved):

akatsoulas:
closing this in favor of https://github.com/mozilla/kitsune/pull/6611

PR #6587, opened 2025-03-24
Create a GroupProfile for the Staff Content Team

smithellis:
Add a GroupProfile for the Staff Content Team group.

### DIFF
--- a/kitsune/groups/migrations/0004_add_staff_content_team_profile.py
+++ b/kitsune/groups/migrations/0004_add_staff_content_team_profile.py
@@ -1,0 +1,42 @@
+from django.db import migrations
+
+
+def create_staff_content_team_profile(apps, schema_editor):
+    Group = apps.get_model('auth', 'Group')
+    GroupProfile = apps.get_model('groups', 'GroupProfile')
+
+    try:
+        group = Group.objects.get(name='Staff Content Team')
+        # Create the group profile if it doesn't exist
+        if not GroupProfile.objects.filter(group=group).exists():
+            GroupProfile.objects.create(
+                group=group,
+                information='Staff Content Team group for content management.',
+                information_html='Staff Content Team group for content management.'
+            )
+    except Group.DoesNotExist:
+        raise Exception("Staff Content Team group must exist before running this migration")
+
+
+def remove_staff_content_team_profile(apps, schema_editor):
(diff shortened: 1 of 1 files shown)


kitsune/groups/migrations/0004_add_staff_content_team_profile.py
```diff
@@ -0,0 +1,42 @@
...
+        group = Group.objects.get(name='Staff Content Team')
+        # Create the group profile if it doesn't exist
+        if not GroupProfile.objects.filter(group=group).exists():
+            GroupProfile.objects.create(
+                group=group,
+                information='Staff Content Team group for content management.',
+                information_html='Staff Content Team group for content management.'
+            )
+    except Group.DoesNotExist:
+        raise Exception("Staff Content Team group must exist before running this migration")
```
escattone, line 18: Nit. It's not a big deal, but in general, it's preferred to organized the code such that only the code that could trigger the exception clause is within the `try`, and everything else is under `else`. So something like:
```python
    try:
        group = Group.objects.get(name='Staff Content Team')
    except Group.DoesNotExist:
        raise Exception("Staff Content Team group must exist before running this migration")
    else:
        # Create the group profile if it doesn't exist
        if not GroupProfile.objects.filter(group=group).exists():
            GroupProfile.objects.create(
                group=group,
                information='Staff Content Team group for content management.',
                information_html='Staff Content Team group for content management.'
            )
```

kitsune/groups/migrations/0004_add_staff_content_team_profile.py
```diff
@@ -0,0 +1,42 @@
...
+
+
+def remove_staff_content_team_profile(apps, schema_editor):
+    Group = apps.get_model('auth', 'Group')
+    GroupProfile = apps.get_model('groups', 'GroupProfile')
+
+    try:
+        group = Group.objects.get(name='Staff Content Team')
+        GroupProfile.objects.filter(group=group).delete()
```
escattone, line 27: Same here.

escattone (approved):
r+wc

kitsune/groups/migrations/0004_add_staff_content_team_profile.py
smithellis: Thanks for this, @escattone ! Makes a ton of sense.

PR #6590, opened 2025-03-25
Automate user curation

smithellis:
* Delete users who last logged in over X days ago, as defined in `settings.USER_EXPIRATION_DAYS` via a management command `cleanup_expired_users.py`
* Run this weekly via cron
* Added DMS call and DMS to settings


kitsune/settings.py
```diff
@@ -1331,3 +1332,5 @@ def filter_exceptions(event, hint):
 
 SUMO_BOT_USERNAME = config("SUMO_BOT_USERNAME", default="SumoBot")
 SUMO_CONTENT_GROUP = config("SUMO_CONTENT_GROUP", default="Staff Content Team")
+
+USER_EXPIRATION_DAYS = config("USER_EXPIRATION_DAYS", default=1095, cast=int)
```
akatsoulas, line 1336: Maybe `USER_INACTIVITY_DAYS`?

kitsune/users/management/commands/cleanup_expired_users.py
```diff
@@ -0,0 +1,27 @@
...
+
+from kitsune.users.utils import delete_user_pipeline
+
+
+class Command(BaseCommand):
+    help = "Delete users who haven't logged in for more than settings.USER_EXPIRATION_DAYS days"
+
+    def handle(self, *args, **options):
+        User = get_user_model()
+        expiration_date = timezone.now() - timedelta(days=settings.USER_EXPIRATION_DAYS)
```
akatsoulas, line 16: timezone.now() returns a datetime object. Maybe we should convert this to `timezone.now().date()`

scripts/cron.py
```diff
@@ -399,6 +399,22 @@ def job_cleanup_old_account_events():
...
+    day="*",
+    hour="03",
+    minute="00",
+    day_of_week=0,
+    max_instances=1,
+    coalesce=True,
+    skip=settings.READ_ONLY,
+)
+@babis.decorator(ping_after=settings.DMS_CLEANUP_EXPIRED_USERS)
+def job_cleanup_expired_users():
```
akatsoulas, line 414: Let's add a waffle flag/switch to control when this is enabled through admin.

akatsoulas (approved):
r+wc

kitsune/settings.py
smithellis: More descriptive - done!

scripts/cron.py
smithellis: Added.

kitsune/users/management/commands/cleanup_expired_users.py
akatsoulas: No the reason for this nit was the naming of the variable. By converting to date() it just removes the time component which as you said it's not relevant for this kind of functionality. This is just nitpicking so treat it as such

PR #6592, opened 2025-03-25
Add a property to user Profile

smithellis:
* has_content returns a boolean designating whether a user has content they created or contributed

### DIFF
--- a/kitsune/users/models.py
+++ b/kitsune/users/models.py
@@ -259,27 +259,41 @@ class Profile(ModelBase):
     @property
     def has_content(self):
         """Returns True if the user has created any content on the site."""
-        content_types = [
-            self.user.answer_votes,
-            self.user.answers,
-            self.user.award_creator,
-            self.user.badge_set,
-            self.user.created_revisions,
-            self.user.gallery_images,
-            self.user.gallery_videos,
-            self.user.inboxmessage_set,
-            self.user.outbox,
-            self.user.poll_votes,
-            self.user.post_set,
-            self.user.question_votes,
-            self.user.questions,
-            self.user.readied_for_l10n_revisions,
-            self.user.reviewed_revisions,
-            self.user.thread_set,
-            self.user.wiki_post_set,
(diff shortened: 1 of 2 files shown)


kitsune/users/models.py
```diff
@@ -256,6 +256,31 @@ def last_contribution_date(self):
...
+            self.user.post_set,
+            self.user.question_votes,
+            self.user.questions,
+            self.user.readied_for_l10n_revisions,
+            self.user.reviewed_revisions,
+            self.user.thread_set,
+            self.user.wiki_post_set,
+            self.user.wiki_thread_set,
+        ]
+        return any(queryset.exists() for queryset in content_types)
```
escattone, line 282: You're making up to 18 queries here, when you could always accomplish this with a single query.

escattone (changes requested):

kitsune/users/models.py
```diff
@@ -256,6 +256,31 @@ def last_contribution_date(self):
     def settings(self):
         return self.user.settings
 
+    @property
```
escattone, line 259: Also, since this method will be making a DB query, I would use `functools.cached_property` instead.

kitsune/users/models.py
smithellis: Ahhhh. Thanks.

kitsune/users/models.py
smithellis: I couldn't find a good way to make this change to cached_property exist with tests, which made me think maybe we shouldn't use cached_property here - user content seems like something that is apt to change regularly, in real time. I think the caching is only for the lifetime of the object, so maybe wouldn't be a problem outside of tests, but maybe it would? I know I could always just handle the cache in the tests (so, not building a solution to meet tests - just inspired to ask the question once I saw that issue during testing). Thoughts?

akatsoulas:
What's the point of this query? Can you provide a scenario where this is needed?

PR #6593, opened 2025-03-25
include system accounts in recent revision filtering

escattone:
mozilla/sumo#2272


akatsoulas (approved):
Probably we need to update `RevisionFilterForm` too under `wiki/forms.py`. r+wc

PR #6594, opened 2025-03-26
improve the gathering of names in `MultiUsernameFilterField`

escattone:
mozilla/sumo#2277

### DIFF
--- a/kitsune/sumo/form_fields.py
+++ b/kitsune/sumo/form_fields.py
@@ -90,7 +90,7 @@ class MultiUsernameFilterField(forms.Field):
 
     def to_python(self, value):
         if value:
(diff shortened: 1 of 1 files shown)


smithellis (changes requested):
I think this exposes a bug elsewhere in the lookup process - you can now search for `SumoBot` but not `SuMo Bot`.

escattone:
> I think this exposes a bug elsewhere in the lookup process - you can now search for `SumoBot` but not `SuMo Bot`.

Hm, I did test `SuMo Bot` and it worked. In fact, that's why @emilghittasv created https://github.com/mozilla/sumo/issues/2277.

smithellis (approved):
Testing works, this is perfect. Thanks!

PR #6595, opened 2025-03-26
Change close_account to utilize the deletion pipeline

smithellis:
I just changed `close_account` to use the `delete_user_pipeline` and modified the tests to just make sure the user is gone.

### DIFF
--- a/kitsune/users/views.py
+++ b/kitsune/users/views.py
@@ -53,7 +53,6 @@ from kitsune.users.tasks import (
 )
 from kitsune.users.templatetags.jinja_helpers import profile_url
 from kitsune.users.utils import (
-    anonymize_user,
     deactivate_user,
     delete_user_pipeline,
     get_oidc_fxa_setting,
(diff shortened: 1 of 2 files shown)


kitsune/users/views.py
```diff
@@ -152,7 +151,7 @@ def profile(request, username):
 @login_required
 @require_POST
 def close_account(request):
-    anonymize_user(request.user)
+    delete_user_pipeline(request.user)
 
     # Log the user out
     auth.logout(request)
```
akatsoulas, line 157: I believe you are automatically redirect and logged out on user deletion which means that the `auth.logout` is redundant but it's worth confirming it.

akatsoulas (approved):
r+wc

kitsune/users/views.py
smithellis: As far as I can tell the logout and the redirect are not automatic, and have to be purposefully called.

PR #6596, opened 2025-03-27
Remove any references to fxa migration status from users app

smithellis:
* drop property from Profile model
* remove from admin
* remove from `auth.py`
* remove from all templates
* remove from factories
* remove from tests


akatsoulas (approved):
This looks good!. Let's squash this to a single commit but do not merge it. We need to first release everything else and then this PR

PR #6610, opened 2025-04-02
Implement tom-select for related documents

smithellis:
- Don't allow addition of article to it's own related
- Don't show old checkbox list on edit or new document pages
- On document view pages, don't show restricted articles in the related documents section if the user lacks permissions

This should fix [2259](https://github.com/mozilla/sumo/issues/2259), [2260](https://github.com/mozilla/sumo/issues/2260), [2261](https://github.com/mozilla/sumo/issues/2261)

### DIFF
--- a/kitsune/sumo/static/sumo/js/wiki_search.js
+++ b/kitsune/sumo/static/sumo/js/wiki_search.js
@@ -65,7 +65,7 @@ document.addEventListener("DOMContentLoaded", function() {
             };
           })
           .filter(item => item.id)
-          .filter(item => !currentDocId || item.id != currentDocId);
+          .filter(item => !currentDocId || String(item.id) !== String(currentDocId));
 
         callback(formattedResults);
       });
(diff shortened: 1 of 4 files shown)


kitsune/sumo/static/sumo/js/wiki_search.js
```diff
@@ -1,82 +1,116 @@
...
+              }
+            }
+            return {
+              id: id,
+              title: item.title,
+              url: item.url
+            };
+          })
+          .filter(item => item.id)
+          .filter(item => !currentDocId || item.id != currentDocId);
```
akatsoulas: Shouldn't this be `!==`

akatsoulas (approved):
r+wc

kitsune/sumo/static/sumo/js/wiki_search.js
smithellis: This lets there be a type mismatch without failing the comparison - so if we compare string "123" to numeric id 123, we get success. I think !== demands type consistency, but != does coercion first. 
Maybe the even better answer is to force both types to be Strings, thus ensuring there no issue. So I'll do that.

PR #6619, opened 2025-04-08
Fix document slug translation handling in get_visible_document_or_404

smithellis:
When a user tries to access a document using a translated slug in the default locale (e.g. /en-US/kb/flash-113-german), the function now properly finds and redirects to the English document (/en-US/kb/flash-113) by:

1. Looking for any document with the given slug in non-default locales
2. Finding its parent document
3. Using the parent's slug to locate the English document

This completes the bidirectional translation handling, ensuring proper redirects in both directions:
- English -> Translated: /en-US/kb/flash-113 -> /de/kb/flash-113-german
- Translated -> English: /de/kb/flash-113-german -> /en-US/kb/flash-113
- English with translated slug -> English: /en-US/kb/flash-113-german -> /en-US/kb/flash-113

### DIFF
--- a/kitsune/wiki/utils.py
+++ b/kitsune/wiki/utils.py
@@ -201,54 +201,9 @@ def get_visible_document_or_404(
         except Document.DoesNotExist:
             pass
 
-    # If we're in the default locale or couldn't find a parent, try to find a document
-    # that has a translation with the given slug
-    if locale == settings.WIKI_DEFAULT_LANGUAGE:
-        # Look for a document in any locale that has a translation in the default locale
-        # with the given slug
-        translation = Document.objects.filter(
-            translations__slug=kwargs.get("slug"),
-            translations__locale=settings.WIKI_DEFAULT_LANGUAGE,
-        ).first()
-        if translation and translation.is_visible_for(user):
-            return translation
-    else:
-        # Look for a document in the requested locale with the given slug
-        # and then get its parent
-        translation = Document.objects.filter(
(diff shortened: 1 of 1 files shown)


escattone (changes requested):
@smithellis I don't think we should go with this approach, mainly because it introduces something we've never done before, which is redirect from one KB slug to another within the same language. For example, if we have an English article at `/en-US/kb/best-article` whose Spanish translation is available at `/es/kb/articulo-mejor`, we'll redirect from `/en-US/kb/articulo-mejor` to `/en-US/kb/best-article`. To me, that sends a confusing message, and it's not worth making such a fundamental change just to support the language switcher. I'd rather change the JS of the language switcher.

Also, this approach still doesn't work in general. For example, if we're viewing the German article at `/de/kb/bester-artikel`, and we use the language selector in the footer to switch to Spanish, we get a 404. It doesn't find `/es/kb/articulo-mejor`.

I'd prefer an approach that doesn't alter something so fundamental as the `get_visible_document_or_404()` function. For example, I think something like the following might work?
- Stop using the JS of the Protocol language switcher.
- When a language switch is requested, we can gather the locale and slug of the current page -- which we know works, and which we can use to get the parent article (if it isn't already the parent) -- as well as the locale that we want to switch to, and send that data to an endpoint that returns the slug of the requested locale or nothing if one doesn't exist.

smithellis:
Closing for https://github.com/mozilla/kitsune/pull/6628

PR #6628, opened 2025-04-14
Replace the locale switcher to handle locale specific slugs

smithellis:
Replace locale switcher JS with code that uses a new endpoint to determine the appropriate document to return based on locale, slug, and requested locale. This also should continue to function properly on pages that are not kb/ related.

### DIFF
--- a/kitsune/sumo/urls.py
+++ b/kitsune/sumo/urls.py
@@ -11,4 +11,5 @@ urlpatterns = [
     path("manifest.json", views.manifest, name="manifest.json"),
     path("geoip-suggestion", views.geoip_suggestion, name="sumo.geoip_suggestion"),
     path("csrftoken", views.csrftoken, name="sumo.csrftoken"),
+    path("translate-url", views.translate_url, name="sumo.translate_url"),
 ]
--- a/kitsune/sumo/urls.py
+++ b/kitsune/sumo/urls.py
@@ -11,5 +11,4 @@ urlpatterns = [
     path("manifest.json", views.manifest, name="manifest.json"),
     path("geoip-suggestion", views.geoip_suggestion, name="sumo.geoip_suggestion"),
     path("csrftoken", views.csrftoken, name="sumo.csrftoken"),
-    path("translate-url", views.translate_url, name="sumo.translate_url"),
 ]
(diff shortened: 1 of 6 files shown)


kitsune/sumo/urls.py
```diff
@@ -11,4 +11,5 @@
     path("manifest.json", views.manifest, name="manifest.json"),
     path("geoip-suggestion", views.geoip_suggestion, name="sumo.geoip_suggestion"),
     path("csrftoken", views.csrftoken, name="sumo.csrftoken"),
+    path("translate-url", views.translate_url, name="sumo.translate_url"),
```
escattone, line 14: I think this should be moved into the `wiki` app.

kitsune/sumo/views.py
```diff
@@ -156,3 +157,35 @@ def serve_cors(*args, **kwargs):
...
+    """
+    current_slug = request.GET.get("current_slug")
+    current_locale = request.GET.get("current_locale")
+    target_locale = request.GET.get("target_locale")
+
+    if not all([current_slug, current_locale, target_locale]):
+        return JsonResponse({"error": "Missing required parameters", "found": False}, status=400)
+
+    try:
+        current_doc = Document.objects.get(slug=current_slug, locale=current_locale)
```
escattone, line 179: I would add `.select_related("parent")` to this, just to avoid the possible extra query for the parent below.

kitsune/sumo/views.py
```diff
@@ -156,3 +157,35 @@ def serve_cors(*args, **kwargs):
...
+        parent_doc = current_doc.parent or current_doc
+        translation = Document.objects.filter(parent=parent_doc, locale=target_locale).first()
```
escattone, line 181: Since the parent could be the "translation", I would change this to:
```python
translation = parent_doc if parent_doc.locale == target_locale else Document.objects.get(parent=parent_doc, locale=target_locale)
```
and then you can get rid of that `elif` clause below.

kitsune/sumo/views.py
```diff
@@ -156,3 +157,35 @@ def serve_cors(*args, **kwargs):
...
+
```
escattone, line 182: I would move all of the code below outside the `try-except` clause, because there's no way for it to trigger the exception.

escattone (changes requested):

kitsune/sumo/views.py
escattone: I just remembered another thing. You should used `Document.objects.visible(...` instead, because we shouldn't "leak" info about articles the user has no permission to see.

smithellis:
Closing for https://github.com/mozilla/kitsune/pull/6634

PR #6631, opened 2025-04-15
Allow advanced search to return archvied questions

smithellis:
We do a test to see if we are using simple or regular/advanced search. If we determine we are using simple search, we won't show archived questions. If we are using advanced search, we will return archived questions.

### DIFF
--- a/kitsune/search/base.py
+++ b/kitsune/search/base.py
@@ -284,7 +284,7 @@ class SumoSearchInterface(ABC):
         ...
 
     @abstractmethod
-    def get_filter(self):
+    def get_filter(self, is_simple_search: bool):
         """A query which filters for all documents to be searched over."""
         ...
 
@@ -344,6 +344,10 @@ class SumoSearch(SumoSearchInterface):
             return self.results[0]
         return self.results
 
+    def get_filter(self, is_simple_search: bool):
+        """Placeholder for subclasses, should be implemented there."""
+        raise NotImplementedError
+
     def build_query(self):
         """Build a query to search over a specific set of documents."""
         parsed = None
@@ -372,8 +376,9 @@ class SumoSearch(SumoSearchInterface):
             **settings.ES_SEARCH_PARAMS
         )
 
-        # add the search class' filter
-        search = search.query(self.get_filter())
+        # filter method needs simple_search status
+        # (True for keyword search, False for advanced syntax)
+        search = search.query(self.get_filter(is_simple_search=not self.parse_query))
         # add highlights for the search class' highlight_fields
         for highlight_field, options in self.get_highlight_fields_options():
             search = search.highlight(highlight_field, **options)
--- a/kitsune/search/search.py
+++ b/kitsune/search/search.py
@@ -111,7 +111,7 @@ class QuestionSearch(SumoSearch):
         ]
         return [(field, FVH_HIGHLIGHT_OPTIONS) for field in fields]
 
-    def get_filter(self):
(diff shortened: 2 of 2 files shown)


kitsune/search/base.py
```diff
@@ -284,7 +284,7 @@ def get_highlight_fields_options(self):
         ...
 
     @abstractmethod
-    def get_filter(self):
+    def get_filter(self, is_simple_search: bool):
```
escattone: I don't see the need to pass `is_simple_search` into this method, given that it's already available within all instances of `SumoSearch` as `not self.parse_query`. I would recommend the following instead:
- Keep the `get_filter` method signature the same.
- Within the `get_filter` implementations within the subclasses of `SumoSearch`, wherever you use `if is_simple_search:`, do the following instead:
    ```python
    is_simple_search = not self.parse_query
    if is_simple_search:
        ...
    ```
    OR
    ```python
    if not self.parse_query:
        ...
    ```

smithellis: Revised.

kitsune/search/base.py
```diff
@@ -344,6 +344,10 @@ def __getitem__(self, key):
             return self.results[0]
         return self.results
 
+    def get_filter(self):
+        """Placeholder for subclasses, should be implemented there."""
+        raise NotImplementedError
+
```
escattone: This can be removed because all of the sub-classes of `SumoSearch` already implement `get_filter`, and in general, it's impossible to instantiate a sub-class of `SumoSearch` without implementing all of its abstract methods.

kitsune/search/base.py
```diff
@@ -372,7 +376,8 @@ def run(self, key: Union[int, slice] = slice(0, settings.SEARCH_RESULTS_PER_PAGE
             **settings.ES_SEARCH_PARAMS
         )
 
-        # add the search class' filter
+        # filter method needs simple_search status
+        # (True for keyword search, False for advanced syntax)
```
escattone: These comments no longer apply.

kitsune/search/search.py
```diff
@@ -368,7 +376,12 @@ def get_highlight_fields_options(self):
 
     def get_filter(self):
         # `should` with `minimum_should_match=1` acts like an OR filter
-        return DSLQ("bool", should=self._from_children("get_filter"), minimum_should_match=1)
+        # Pass the flag down to children filters
+        return DSLQ(
+            "bool",
+            should=[child.get_filter() for child in self._children],
+            minimum_should_match=1,
+        )
```
escattone: These new lines should be removed, and the original line restored. The `self._from_children("get_filter")` is necessary because it flattens the list of filters.

kitsune/search/search.py
```diff
@@ -208,6 +210,11 @@ def get_filter(self):
             DSLQ("term", _index=self.get_index()),
             DSLQ("exists", field=f"title.{self.locale}"),
         ]
+
+        if not self.parse_query:
+            # Only exclude archived documents in simple search
+            filters.append(DSLQ("term", is_archived=False))
```
escattone, line 216: Does `is_archived` need to be added to the `WikiDocument` in order to filter on it?

escattone (changes requested):

smithellis:
Closing in favor or https://github.com/mozilla/kitsune/pull/6750

kitsune/search/search.py
escattone: Hm, I don't think so.

PR #6634, opened 2025-04-22
Support cross-locale slug lookup in language switcher

smithellis:
The language switcher previously failed when switching between locales where document slugs differ. For example, when switching from German "add-on-badges-abzeichen" back to English "add-on-badges", users would see a 404 error.

This commit enhances the get_visible_document_or_404 function to:
- Find correct documents across locales with different slugs
- Look up documents by finding related translations in any locale
- Maintain backward compatibility with existing behavior

Added comprehensive tests to verify the new functionality works with:
- Direct locale/slug matches
- Cross-locale lookups (any locale to any locale)
- Document fallbacks when translations don't exist

### DIFF
--- a/kitsune/wiki/tests/test_utils.py
+++ b/kitsune/wiki/tests/test_utils.py
@@ -25,11 +25,13 @@ from kitsune.wiki.utils import (
     generate_short_url,
     get_featured_articles,
     get_kb_visited,
+    get_visible_document_or_404,
     has_visited_kb,
     num_active_contributors,
     remove_expired_from_kb_visited,
     update_kb_visited,
 )
+from django.http import Http404
 
 
 class ActiveContributorsTestCase(TestCase):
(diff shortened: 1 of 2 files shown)


kitsune/wiki/utils.py
```diff
@@ -182,26 +182,54 @@ def get_visible_document_or_404(
...
 
-    # We couldn't find a visible translation in the requested non-default locale, so let's
-    # see if we can find a visible translation via its parent.
+    slug = kwargs.get("slug")
+
+    # For cross-locale lookup
+    if slug and look_for_translation_via_parent:
+        if locale == settings.WIKI_DEFAULT_LANGUAGE:
+            # Find documents with this slug in other locales
+            other_docs_qs = Document.objects.filter(slug=slug).exclude(locale=locale)
```
akatsoulas, line 194: You are repeating this code in the `else` clause too. This can be a bit more DRY by eliminating the second for loop and have everything in this one

akatsoulas: Adding here a `select_related` for the parent can also help with DB performance

akatsoulas (changes requested):

kitsune/wiki/utils.py
smithellis: Ah - got it. Thanks!

kitsune/wiki/tests/test_utils.py
```diff
@@ -700,3 +702,152 @@ def test_with_product_and_topic_2(self):
...
+            self.user,
+            locale="en-US",
+            slug="test-document-de",
+            look_for_translation_via_parent=True,
+        )
+        self.assertEqual(doc, self.en_doc)
+
+        # When switching from French to English, we should find the English document
+        # even though we're looking for the French slug in English
+        doc = get_visible_document_or_404(
```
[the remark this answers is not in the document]
smithellis: I'm not sure I follow you - all these tests have subtle differences, either slug or locale or other flags. Let me know if I've missed something.

smithellis:
I dropped most of the comments from the test file. I left many comments in the `utils.py` file - I personally find the way we do these fallbacks/failovers to be a bit difficult to follow, and if nothing else the comments there help me out.

kitsune/wiki/utils.py
```diff
@@ -182,26 +182,53 @@ def get_visible_document_or_404(
...
-        # make sense, because we're not making a locale-specific request or the locale we're
-        # requesting is already the default locale.
+    if not (locale := kwargs.get("locale")):
         raise Http404
 
-    # We couldn't find a visible translation in the requested non-default locale, so let's
-    # see if we can find a visible translation via its parent.
+    slug = kwargs.get("slug")
+
+    if slug and look_for_translation_via_parent:
```
akatsoulas, line 190: Again many redundant comments

kitsune/wiki/utils.py
```diff
@@ -182,26 +182,53 @@ def get_visible_document_or_404(
...
+                # Only consider documents that are translations
+                if not doc.parent:
+                    continue
+
+                # Check if this parent has a translation in our requested locale
+                translation = doc.parent.translated_to(locale, visible_for_user=user)
+                if translation:
+                    return translation
+
+    if not look_for_translation_via_parent or locale == settings.WIKI_DEFAULT_LANGUAGE:
```
akatsoulas, line 221: Is there a reason this check is so late in the code after all the expensive operations above? This should be combined with the earlier 404.

kitsune/wiki/utils.py
```diff
@@ -182,26 +182,53 @@ def get_visible_document_or_404(
...
         raise Http404
 
-    # We couldn't find a visible translation in the requested non-default locale, so let's
-    # see if we can find a visible translation via its parent.
+    slug = kwargs.get("slug")
+
+    if slug and look_for_translation_via_parent:
+        # Find documents with this slug in other locales
+        other_docs_qs = (
+            Document.objects.filter(slug=slug).exclude(locale=locale).select_related("parent")
```
akatsoulas, line 193: Shouldn't we use `visible` here too like above instead of `filter`?

kitsune/wiki/utils.py
```diff
@@ -182,26 +182,53 @@ def get_visible_document_or_404(
...
+        )
+
+        if locale == settings.WIKI_DEFAULT_LANGUAGE:
```
akatsoulas, line 196: This feels inefficient. We have basically 2 blocks of code (if/else). Each block has the same loop and a call to `translated_to`. This should have been the other way. Loop once over the docs and check what you need for each one

akatsoulas (changes requested):

kitsune/wiki/utils.py
smithellis: Removing.

kitsune/wiki/utils.py
smithellis: My thoughts were that since we are looking for a path to a document through other documents, the documents we look at to move to a valid document should all be collected. We check for visibility later before presenting a final document. What if a document is restricted in one language but not in another?

kitsune/wiki/utils.py
smithellis: Yes - the second check after the loop prevents the final fallback when we're in the default language. Checking it earlier breaks the case where a user requests a document in one locale but is using a slug from another. So - check to see if we should bother finding a translation.

kitsune/wiki/utils.py
akatsoulas: That's problematic and it shouldn't happen. The restricted visibility should be the same across all locales

kitsune/wiki/utils.py
```diff
@@ -174,64 +174,49 @@ def get_featured_articles(product=None, topics=None, locale=settings.WIKI_DEFAUL
 def get_visible_document_or_404(
     user, look_for_translation_via_parent=False, return_parent_if_no_translation=False, **kwargs
 ):
-    """
-    Get the document specified by the keyword arguments and visible to the given user, or 404.
```
akatsoulas, line 178: This is not supposed to be removed. Docstrings should always be part of a function/class etc

kitsune/wiki/utils.py
```diff
@@ -174,64 +174,49 @@ def get_featured_articles(product=None, topics=None, locale=settings.WIKI_DEFAUL
...
-    """
     try:
         return Document.objects.get_visible(user, **kwargs)
     except Document.DoesNotExist:
         pass
 
-    if not (locale := kwargs.get("locale")):
+    locale = kwargs.get("locale")
```
akatsoulas: What's the reason behind this change?

kitsune/wiki/utils.py
akatsoulas: most probably it's cleaner to work with the existence of the parent.doc instead:

```python
for doc in other_docs_qs:
    base_doc = doc.parent or doc
    if base_doc.locale == settings.WIKI_DEFAULT_LANGUAGE:
        if base_doc.locale == locale:
            return base_doc
        translation = base_doc.translated_to(locale, visible_for_user=user)
        if translation:
            return translation
    else:
        translation = base_doc.translated_to(locale, visible_for_user=user)
        if translation:
            return translation
```

akatsoulas (approved):
r+wc. Please squash the commits into a single one before merging

kitsune/wiki/utils.py
smithellis: I still believe this is the right approach - doing discovery of the redirect chain has nothing to do with visibility. If  there is a case where a document in the chain is not visible to the user, but the document the user should land on is visible - then they won't get there if we change this to use `.visible`. 
We could wind up with a case where locale leaders or content editors or others with unique permissions will get different redirects than other users. 
Unless you can think of a reason that this is harmful in some other way, I'd like to leave it.

kitsune/wiki/utils.py
smithellis: Overzealous compliance to comment reduction - automation error.

PR #6640, opened 2025-04-30
Upgrade Elasticsearch from 7.x to 8.18.0

smithellis:
This commit upgrades our Elasticsearch infrastructure from version 7.x to 8.18.0, 
including:

- Updated ES client library from elasticsearch 7.17.9 to 8.18.0
- Removed `elasticsearch-dsl` dependency in favor of `elasticsearch.dsl`
- Updated Docker configuration for Elasticsearch 8.18.0 with security settings
- Modified ES connection settings and client configuration options
- Updated ES module imports from `elasticsearch_dsl` to `elasticsearch.dsl`
- Renamed `test_es7_utils.py` to `test_es_utils.py` with updated test cases
- Added new ES client configuration options in `settings.py`
- Fixed ES URL prefix to use http:// protocol explicitly
- Made necessary code adjustments throughout search modules for compatibility

This upgrade provides better security, performance improvements, and access to newer 
Elasticsearch features while maintaining backward compatibility with our existing 
search functionality.


kitsune/search/base.py
```diff
@@ -95,17 +95,30 @@ def migrate_reads(cls):
...
-                {
-                    "actions": [
-                        {"remove": {"index": old_index, "alias": alias}},
-                        {"add": {"index": new_index, "alias": alias}},
-                    ]
-                }
-            )
+
+        # Handle special case where an index with same name as alias exists
+        try:
```
akatsoulas, line 100: Is there a reason for adding this check?

kitsune/search/es_utils.py
```diff
@@ -94,9 +92,36 @@ def es_client(**kwargs):
...
+        kwargs.update({"cloud_id": es_cloud_id, "basic_auth": settings.ES_HTTP_AUTH})
     else:
-        kwargs.update({"hosts": settings.ES_URLS})
+        # Elasticsearch 8.x settings
+        es_settings = {
+            "hosts": settings.ES_URLS,
+            "request_timeout": settings.ES_TIMEOUT,
+            "retry_on_timeout": settings.ES_RETRY_ON_TIMEOUT,
+            # SSL settings - these are needed for ES8 which requires SSL by default
+            "verify_certs": settings.ES_VERIFY_CERTS,
```
akatsoulas, line 103: what's the reason for introducing all these new settings?

kitsune/search/es_utils.py
```diff
@@ -94,9 +92,36 @@ def es_client(**kwargs):
...
+            "ssl_show_warn": settings.ES_SSL_SHOW_WARN,
+            # Disable auto-discovery which can cause connection issues
+            "sniff_on_start": settings.ES_SNIFF_ON_START,
+            "sniff_on_connection_fail": settings.ES_SNIFF_ON_CONNECTION_FAIL,
+        }
+
+        if settings.ES_HTTP_AUTH:
```
akatsoulas, line 110: Why is this needed?

kitsune/search/es_utils.py
```diff
@@ -134,10 +159,15 @@ def index_object(doc_type_name, obj_id):
         # just return
         return
 
+    kwargs = {}
+    # For ES8, use string "true" instead of boolean True for refresh parameter
+    if settings.TEST:
```
akatsoulas, line 164: You already have this check above

kitsune/search/es_utils.py
```diff
@@ -134,10 +159,15 @@ def index_object(doc_type_name, obj_id):
...
+        kwargs["refresh"] = "true"
```
akatsoulas, line 165: What's the purpose of this argument?

kitsune/search/tests/signals/test_forums.py
```diff
@@ -7,6 +7,7 @@
 
 class ForumDocumentSignalsTests(Elastic7TestCase):
     def setUp(self):
+        super().setUp()
```
akatsoulas, line 10: I am not sure that we need this

kitsune/search/tests/test_es_utils.py
```diff
@@ -46,9 +48,16 @@ def make_first_doc_throw_exception(self, *args, **kwargs):
             question = Question.objects.get(id=question_id)
             AnswerFactory(question=question, content=f"answer {question_id}")
 
+        # Force ES index refresh before testing
+        QuestionDocument._index.refresh()
```
akatsoulas, line 52: Why are you introducing this change? Did something change between 7 and 8?

kitsune/search/base.py
smithellis: I think I can drop this; I added this when troubleshooting test issues with index and alias collisions. I'll see if dropping it causes problems to recur.

kitsune/search/es_utils.py
smithellis: ES8 requires the SSL settings. `retry_on_timeout` was an attempt to improve reliability when troubleshooting test issues; the same with the `sniff_on_start` and `sniff_on_connection_fail`.  
We can drop the `sniff_*` settings, but should consider leaving the `retry_on_timeout`. May not be critical but it could help when/if there are real connection issues.

kitsune/search/es_utils.py
smithellis: There are three checks for `settings.TEST` in the file. The `es_client` has one to modify timeout and retry as part of local testing, `index_object()` and `delete_object()` both test to set the refresh to true. It didn't seem reasonable to pull that check into a helper function just for the second two tests. I don't think we want `refresh` set to true generally, as it forces an immediate refresh which would be expensive in prod.
I think we could remove the code I added for troubleshooting/resolving tests (so the es_client() timeout, retries and retry_on_timeout) and that would get rid of the test in `es_client()`. But we want the test in the other two places for sure.

kitsune/search/es_utils.py
smithellis: It forces ES to refresh, making changes immediately visible to search. So it helps for local testing more than anything else.

kitsune/search/tests/signals/test_forums.py
smithellis: Since I changed the `ElasticTestCase` it is needed. Now we set the test environment up cleanly, force a refresh, and get in front of any index issues.

smithellis:
Closing this in favor of https://github.com/mozilla/kitsune/pull/6728

PR #6642, opened 2025-05-02
remove unused `json_utils`

escattone:
I discovered that we don't use `kitsune.sumo.json_utils` anywhere in our code, so removing it and its tests.


akatsoulas (approved):

kitsune/sumo/tests/test_json_utils.py
zakir957, line 1: 22

zakir957 (approved):

PR #6649, opened 2025-05-08
Add migration to create and associate contributor topics and documents

smithellis:
- Ensures the "Contributors" product exists
- Creates "Templates" and "Canned responses" topics if missing
- Associates these topics with the Contributors product
- Links all KB article templates to the Templates topic and Contributors product
- Links the /en-US/kb/common-forum-responses document to the Canned responses topic and Contributors product

Handles the first section of https://github.com/mozilla/sumo/issues/1956


akatsoulas:
I am not sure that we are going to keep the `Contributor` product around. Even if we do though, there's a good chance that we will not use the migration. I am closing this for now and we can re-open if needed

PR #6651, opened 2025-05-09
Replace apscheduler with celery beat

smithellis:
Add `celery-beat` to docker-compose
Add schedules via `celery_schedules.py`
Add/update required settings
Add tasks to `tasks.py`
Add `django-celery-beat` to `pyproject.toml`


pyproject.toml
```diff
@@ -15,7 +15,7 @@ license = "BSD 3-Clause"
 
 [tool.poetry.dependencies]
 python = "^3.11"
-APScheduler = "^3.10.4"
+django-celery-beat = "^2.5.0"
```
akatsoulas, line 18: Why do we need this?

smithellis: Ah - this lets us schedule tasks in admin. We don't need this unless we want to do that; either way I'll drop it from this PR, as it should be its own PR if we decide to use it.

PR #6656, opened 2025-05-12
This attempts to make code ES version agnostic between

smithellis:
7 and 8.

We import both python libraries `elasticsearch7` and `elasticsearch8` We test the server version and set it in `settings.py` We use the correct library based on the server major version Server version check is now in `search/utils.py` and uses module level caching to prevent unneeded connections


smithellis:
Closing - new approach at https://github.com/mozilla/kitsune/pull/6665

PR #6658, opened 2025-05-12
initial AI-based question classification

escattone:
@akatsoulas I've created this is an initial take on question classification using only `langchain`. You may have had something else in mind, but at least this gives us something concrete to discuss when deciding how we'd like to move forward.


kitsune/ai/llms.py
```diff
@@ -0,0 +1,29 @@
...
+
+
+@cache
+def get_llm(
+    model_name: str, temperature: int = 0, max_tokens: int | None = None, max_retries: int = 2
+) -> BaseChatModel:
+    """
+    Returns a LangChain chat model instance based on the given LLM model name.
+    """
+    if is_openai_model(model_name):
```
akatsoulas, line 22: We are going to use vertex-ai. I would drop everything related to openai for now and add it if we need it instead of the other way around.

kitsune/ai/llms.py
```diff
@@ -0,0 +1,29 @@
...
+    Returns whether or not the given model name is an OpenAI model.
+    """
+    return any(
+        model_name.startswith(prefix) for prefix in ("gpt-", "chatgpt-", "o1-", "o3-", "o4-")
+    )
+
+
+@cache
+def get_llm(
+    model_name: str, temperature: int = 0, max_tokens: int | None = None, max_retries: int = 2
```
akatsoulas, line 17: I believe most models in vertex are having a temperature of 1 as default value

kitsune/ai/questions/tests/test_classify.py
```diff
@@ -0,0 +1,1791 @@
...
+from kitsune.sumo.tests import TestCase
+from kitsune.users.models import Profile
+
+
+@skipUnless(
+    os.getenv("GOOGLE_CLOUD_PROJECT") and os.getenv("GOOGLE_APPLICATION_CREDENTIALS"),
+    "Google application credentials have not been configured.",
+)
+class ClassifyQuestionsTests(TestCase):
+
```
akatsoulas, line 20: I believe we should avoid calling the actual llm and instead mock the call when we want to test that our code successfully calls the llm with the right params. The result of the llm should be part of the unittest. This will lead to slow tests, higher costs and flakiness because results will vary.

kitsune/ai/questions/utils.py
```diff
@@ -0,0 +1,62 @@
...
+    Flags the given question as spam with the given reason.
+    """
+    if question.is_spam:
+        return
+
+    sumo_bot = Profile.get_sumo_bot()
+
+    with transaction.atomic():
+        # Mark the question as spam, and assign it the "Undefined" topic.
+        question.topic = Topic.active.get(title="Undefined")
```
akatsoulas, line 23: Although I am not against this atm, at the same time I am not sure whether we should this or not. If it's spam, the topic should be irrelevant here. I am wondering if we are creating a bias towards `Undefined` if we are marking all spam as such

kitsune/ai/apps.py
```diff
@@ -0,0 +1,6 @@
+from django.apps import AppConfig
```
akatsoulas, line 1: shall we call the app `llm` instead of `ai` which is a bit more explicit? Open to another naming schemas as well

akatsoulas (changes requested):
This PR does many things in a specific way. My suggestion would be to use it for scaffolding by adding the new application, the prompt, parsers and the `get_llm` function. Once we have those we can start building the chains and the tooling.

kitsune/ai/questions/tests/test_classify.py
escattone: I forgot to mention with these tests, that I didn't intend them to ever be run as part of our test suite. I consider them the start of an evaluation tool for local development/testing of prompt/model changes.

kitsune/ai/questions/utils.py
escattone: I wasn't sure either. From the spam questions in the database currently, it appears that we don't also change the topic. Maybe we should just set the topic to `None`?

escattone:
Selecting pieces of this for the initial step. Closing in favor of https://github.com/mozilla/kitsune/pull/6659

PR #6659, opened 2025-05-13
initial pieces for llm-based question classification

escattone:


kitsune/llm/__init__.py
```diff
@@ -0,0 +1,15 @@
+from functools import cache
+
+from langchain_google_vertexai import ChatVertexAI
+
+
+@cache
+def get_llm(
+    model_name: str, temperature: int = 0, max_tokens: int | None = None, max_retries: int = 2
```
akatsoulas, line 8: Shall we set the temperature value to `1`? Default for most Google models

akatsoulas (approved):
r+wc

PR #6662, opened 2025-05-15
ES upgrade 

smithellis:
Removed many tests. Compatibility issues with tests was overwhelming; this is an attempt to check the the build and usage functionality without all the old ES7 tests. New tests can be developed to fill large gaps (signal testing, views, etc).

The goal of this PR is only to test if dual compatibility is otherwise achieved. 

This is for testing in DEV only


Nivekwaite:
I hope nobody's trying to break into my wallet

smithellis:
closing - new approach at https://github.com/mozilla/kitsune/pull/6665

PR #6663, opened 2025-05-16
Classify questions

akatsoulas:


kitsune/questions/classifiers.py
```diff
@@ -0,0 +1,43 @@
...
+                {
+                    "topic_result": (
+                        None
+                        if payload["spam_result"]["is_spam"]
+                        else topic_classification_chain.invoke(payload)
+                    )
+                }
+            ),
+        }
+    )
```
escattone, line 40: For me, this is difficult to read due to the `LangChain` abstractions. Since `LangChain` automatically wraps Python callables with `RunnableLambda`, what do you think of something like the following instead? I've tested it and it returns the same result. For me, it feels more "Pythonic" and readable. 

```python
    def conditional_topic_classification_chain(result):
        if result["is_spam"] or (result["confidence"] < 60):
            result.update(topic_result=None)
        else:
            result.update(topic_result=topic_classification_chain.invoke(payload))
        return result

    pipeline = spam_detection_chain | conditional_topic_classification_chain
```

kitsune/questions/classifiers.py
```diff
@@ -0,0 +1,43 @@
...
+    result = pipeline.invoke(payload)
+    # Draft - we need to process the result
+    return result["is_spam"], result["topic_result"]
```
escattone, line 43: I'm not sure if your plan is to process the result -- i.e., mark the question as spam or set its topic -- inside or outside of this function, but if outside, then I think we should also somehow include the spam reason and confidence level as well.

kitsune/questions/classifiers.py
```diff
@@ -0,0 +1,43 @@
...
+
+from kitsune.llm.questions.prompt import spam_parser, spam_prompt, topic_parser, topic_prompt
+from kitsune.llm.utils import get_llm
+from kitsune.products.utils import get_taxonomy
+from kitsune.questions.models import Question
+
+DEFAULT_LLM_MODEL = "gemini-2.5-flash-preview-04-17"
+
+
+def analyze_question(question: Question) -> tuple[bool, dict]:
```
escattone, line 11: If this function is going to also mark the question as spam or set its topic, I think we should change the name to something like `classify_question` instead.

escattone (changes requested):

kitsune/questions/classifiers.py
```diff
@@ -0,0 +1,43 @@
...
+    """
+    Analyze a question from spam and classifies the topic.
+    """
+    llm = get_llm(model_name=DEFAULT_LLM_MODEL)
```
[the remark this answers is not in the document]
akatsoulas: I would start from 1 which is the default and re-adjust based on the results

kitsune/questions/classifiers.py
akatsoulas: I was going for within the function. I just wanted a first pass before processing. I am updating the PR to include the processing logic.

akatsoulas: On second thought, I am going to process the result in https://github.com/mozilla/sumo/issues/2335

akatsoulas:
@escattone PR udpated

kitsune/questions/utils.py
```diff
@@ -138,3 +144,66 @@ def get_ga_submit_event_parameters_as_json(
...
+                question.mark_as_spam(sumo_bot)
+            case ModerationAction.FLAG_REVIEW:
+                _create_flag_object(
+                    question=question,
+                    by_user=sumo_bot,
+                    notes="LLM flagged for manual review",
+                    reason=FlaggedObject.REASON_SPAM,
+                )
+            case _:
+                if topic_title := result["topic_result"].get("topic"):
```
escattone, line 192: It seems important for us to know when a `topic_result` doesn't contain a `topic` key or its value isn't a valid topic title. That's probably something that we should add later somewhere in `kitsune.llm.questions`. For example, maybe we log every LLM request and response and also flag invalid cases.

escattone (changes requested):

kitsune/llm/questions/classifiers.py
```diff
@@ -0,0 +1,71 @@
...
+    SPAM = "spam", "Spam"
+    FLAG_REVIEW = "flag_review", "Flag for Review"
+
+
+def classify_question(question: "Question") -> dict[str, Any]:
+    """
+    Analyze a question for spam and, if not spam or low confidence, classify the topic.
+    Returns a dict with keys: action, spam_result, topic_result (optional).
+    """
+    llm = get_llm(model_name=DEFAULT_LLM_MODEL)
```
escattone, line 29: I don't think a non-zero `temperature`setting makes sense for this use case. We want as much consistency as possible.

kitsune/llm/questions/classifiers.py
```diff
@@ -0,0 +1,71 @@
...
+                ):
+                    result["action"] = ModerationAction.FLAG_REVIEW
+                case _:
+                    result["topic_result"] = topic_classification_chain.invoke(payload)
+
+        return result
+
+    pipeline = RunnablePassthrough.assign(spam_result=spam_detection_chain) | RunnableLambda(
+        decision_lambda
+    )
```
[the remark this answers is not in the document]
akatsoulas: Sounds good about the `spam_result`. Regarding the second suggestion, I would prefer to go with the RunnablePassthrough in this case. Apart from personal preference, reading through the documentation it is more evident why this is constructed this way. Using the two chains hides that imo

kitsune/questions/utils.py
```diff
@@ -138,3 +144,66 @@ def get_ga_submit_event_parameters_as_json(
...
+    if waffle.switch_is_active("auto-question-classifier"):
+        action = result.get("action")
+        match action:
+            case ModerationAction.SPAM:
+                question.mark_as_spam(sumo_bot)
+            case ModerationAction.FLAG_REVIEW:
+                _create_flag_object(
+                    question=question,
+                    by_user=sumo_bot,
+                    notes="LLM flagged for manual review",
```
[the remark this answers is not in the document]
akatsoulas: This falls under the manual processing (for now at least) so the moderator will need to flag it as such

kitsune/questions/utils.py
```diff
@@ -138,3 +144,66 @@ def get_ga_submit_event_parameters_as_json(
...
+    result: dict[str, Any],
+) -> None:
+    """
+    Process the classification result from the LLM and take moderation action.
+    Control flow:
+    - If 'auto-question-classifier' waffle is active, use LLM moderation logic.
+    - Else if 'flagit-spam-autoflag' waffle is active, always flag for manual review.
+    - Else, do nothing.
+    """
+    sumo_bot = Profile.get_sumo_bot()
```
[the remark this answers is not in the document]
akatsoulas: Currently we are deleting flagged objects upon user deletion. For now I would with SumoBot just b/c we can still differentiate the flags that are coming from automation. As long as we have discrete cases, I would keep what we have instead of introducing multiple accounts. Similarly, I don't mind if we update the `bio` but I would treat that as content and I wouldn't update the code for that.

kitsune/llm/questions/classifiers.py
```diff
@@ -0,0 +1,71 @@
...
+            match confidence:
+                case _ if confidence > HIGH_CONFIDENCE_THRESHOLD:
+                    result["action"] = ModerationAction.SPAM
+                case _ if (
+                    confidence > LOW_CONFIDENCE_THRESHOLD
+                    and confidence < HIGH_CONFIDENCE_THRESHOLD
+                ):
+                    result["action"] = ModerationAction.FLAG_REVIEW
+                case _:
+                    result["topic_result"] = topic_classification_chain.invoke(payload)
```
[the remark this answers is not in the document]
akatsoulas: thanks, updated

escattone (approved):

kitsune/llm/questions/classifiers.py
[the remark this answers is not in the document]
escattone: I saw that you had changed the name of `decision_lambda` to `spam_result`. I meant changing the name of the argument to `decision_lambda` from `payload` to `spam_result` (and removing the first line of that function, which woulds no longer be needed) in order to support the replacement of the `RunnablePassthrough.assign(...)` with `spam_detection_chain`. Since you wanted to keep `RunnablePassthrough.assign(...)`, I changed the name back to `decision_lambda`.

PR #6673, opened 2025-05-22
ensure consistent  topic and topic tag when classifying

escattone:
mozilla/sumo#2339

### DIFF
--- a/kitsune/questions/utils.py
+++ b/kitsune/questions/utils.py
@@ -7,11 +7,14 @@ from django.conf import settings
 from django.contrib.auth.models import User
 from django.contrib.contenttypes.models import ContentType
 from django.contrib.sessions.backends.base import SessionBase
+from django.db import transaction
 
 from kitsune.flagit.models import FlaggedObject
 from kitsune.llm.questions.classifiers import ModerationAction
 from kitsune.products.models import Product, Topic
 from kitsune.questions.models import Answer, Question
+
+# from kitsune.tags.models import SumoTag
 from kitsune.users.models import Profile
 from kitsune.wiki.utils import get_featured_articles as kb_get_featured_articles
 from kitsune.wiki.utils import has_visited_kb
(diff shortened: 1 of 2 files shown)


kitsune/questions/utils.py
```diff
@@ -186,20 +189,28 @@ def process_classification_result(
...
-                    flag_question(
-                        question,
-                        by_user=sumo_bot,
-                        notes=(
-                            "LLM classified as {topic.title}, for the following reason:\n"
-                            f"{result['topic_result']['reason']}"
-                        ),
-                        status=FlaggedObject.FLAG_ACCEPTED,
-                    )
+            if not (topic_title := result["topic_result"].get("topic")):
```
akatsoulas, line 192: imo the previous iteration was easier to follow. I am not sure why this needs changing.

kitsune/questions/utils.py
```diff
@@ -186,20 +189,28 @@ def process_classification_result(
...
+                return
+
+            try:
+                topic = Topic.active.get(title=topic_title, visible=True)
+            except (Topic.DoesNotExist, Topic.MultipleObjectsReturned):
+                return
+
+            with transaction.atomic():
```
akatsoulas, line 200: Why is this needed?

kitsune/questions/utils.py
```diff
@@ -186,20 +189,28 @@ def process_classification_result(
...
+                flag_question(
+                    question,
+                    by_user=sumo_bot,
+                    notes=(
+                        "LLM classified as {topic.title}, for the following reason:\n"
+                        f"{result['topic_result']['reason']}"
+                    ),
+                    status=FlaggedObject.FLAG_ACCEPTED,
+                )
+                if topic != question.topic:
```
akatsoulas, line 210: I would argue that the following 3 lines is all that we should include in this PR after the `else` clause without changing anything else. At least I cannot understand the reason for the rest of the changes

akatsoulas (approved):
r+wc

kitsune/questions/utils.py
escattone: I removed it. As longs as the `question_classifier` task is made idempotent, the task retry mechanism should take care of ensuring synchronization if any one of the steps fails.

PR #6674, opened 2025-05-22
ensure `question_classifier` task is idempotent

escattone:
mozilla/sumo#2343


kitsune/questions/utils.py
```diff
@@ -153,14 +153,21 @@ def flag_question(
...
-        status=status,
-        reason=reason,
-        notes=notes,
+        defaults=dict(
+            reason=reason,
+            status=status,
+            notes=notes,
+        ),
     )
+    if not created:
```
akatsoulas, line 166: This is not needed. You are already using `defaults={}` which will create the values if a CREATE event occurs.

akatsoulas (approved):
r+wc

PR #6675, opened 2025-05-23
support for image attachments when classifying questions

escattone:
mozilla/sumo#2344

### DIFF
--- a/kitsune/llm/questions/classifiers.py
+++ b/kitsune/llm/questions/classifiers.py
@@ -32,6 +32,7 @@ def classify_question(question: "Question") -> dict[str, Any]:
     payload: dict[str, Any] = {
         "subject": question.title,
         "question": question.content,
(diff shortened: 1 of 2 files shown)


akatsoulas:
I am not certain that we need to support images at the moment. Adding images will require a headless browser to do the capturing and it's going to raise the cost per query without knowing at this point if we need this. Preliminary testing shows that using just text works fine. Because of the above, I am hesitant to add code that we don't know if we are going to need

escattone:
> Adding images will require a headless browser to do the capturing

No, adding images like this does not require a headless browser. This code works as it is. I've already tested it. Our VertexAI backend grabs the images from the URL's we provide, and feeds them into the LLM.

akatsoulas:
Closing this for now to reduce clutter. As discussed in the platform meeting, we can revisit this after evaluating the performance of the classifier

PR #6676, opened 2025-05-23
migration for inactive user purge

smithellis:
1 - get all users who we consider inactive (3 years since login)
2 - divide into users having content and users without content
3 - hard deletes non-content users using the `_base_manager`
4 - pushes other users through the deletion pipeline
5 - implements batching to manage resources


kitsune/users/migrations/0035_batch_delete_inactive_users.py
```diff
@@ -52,61 +51,61 @@ def delete_inactive_users(apps, schema_editor):
...
-            f"Rate: {current_rate:.1f} users/min | "
-            f"Remaining: {timedelta(seconds=int(remaining_time))}"
-        )
+        # Progress reporting every 1000 users
+        if processed_count % 1000 == 0:
+            elapsed_time = time.time() - start_time
+            progress_pct = (processed_count/total_users*100) if total_users > 0 else 0
+            avg_time = elapsed_time / processed_count if processed_count > 0 else 0
+            remaining_time = (total_users - processed_count) * avg_time if processed_count > 0 else 0
+            current_rate = processed_count / elapsed_time * 60 if elapsed_time > 0 else 0
```
escattone, line 94: Nit. Same here. I think for all of these lines, the `processed_count` and `elapsed_time` values are guaranteed to be greater than zero, so I  don't think you need the `if x > 0 else 0` checks.

escattone (approved):
r+wc

kitsune/users/migrations/0035_batch_delete_inactive_users.py
```diff
@@ -0,0 +1,131 @@
...
+        | Q(wiki_thread_set__isnull=False)
+        | Q(locales_leader__isnull=False)
+        | Q(locales_reviewer__isnull=False)
+        | Q(locales_editor__isnull=False)
+        | Q(wiki_contributions__isnull=False)
+    )
+    
+    cutoff_date = timezone.now() - timedelta(days=3*365)
+    
+    query = User.objects.filter(last_login__lt=cutoff_date).annotate(has_content=has_content_criteria)
```
akatsoulas, line 45: Why are you using `annotate` here instead of `filter/exclude/Exists`?

kitsune/users/migrations/0035_batch_delete_inactive_users.py
```diff
@@ -0,0 +1,131 @@
+from datetime import timedelta
+import time
+import importlib
```
akatsoulas, line 3: is this used anywhere?

akatsoulas (changes requested):

kitsune/users/migrations/0035_batch_delete_inactive_users.py
smithellis: This use case absolutely works - I think it's just not specifically called out in the 4.2 docs but is in later docs.  I can run this query and see the output and it builds valid sql which executes properly. 

I'm annotating here so we can later divide our users into those with and those without content, so we can execute a quicker delete process on non-content users.

kitsune/users/migrations/0035_batch_delete_inactive_users.py
```diff
@@ -0,0 +1,131 @@
...
+from django.contrib.auth import get_user_model
+
+from kitsune.users.utils import delete_user_pipeline
+
+
+def delete_inactive_users(apps, schema_editor):
+    """
+    Delete users who haven't logged in for over three years.
+    """
+    User = get_user_model()
```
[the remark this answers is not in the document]
smithellis: This is necessary because the `delete_user_pipeline` function needs an actual User instance vs. a historical model.

kitsune/users/migrations/0035_batch_delete_inactive_users.py
smithellis: I changed this to two separate queries.

kitsune/users/migrations/0035_batch_delete_inactive_users.py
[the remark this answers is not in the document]
akatsoulas: If we cannot use the `apps.get_model()` here, probably we should refactor and change our approach. The migration needs to work the the model state and the time of its creation. If in the future the user model changes, the migration will most probably fail.

I would suggest to either refactor the migration in order to use methods compatible with the historical model (eg avoid properties and custom managers) or consider a different approach altogether.

smithellis:
Closing for https://github.com/mozilla/kitsune/pull/6747 

We decided to convert to a management command.

PR #6680, opened 2025-05-27
Introduce a product classifier

akatsoulas:


escattone (commented):
Overall, I like this approach. I'd suggest one refinement. I think we should only run the product re-classification when we know that the only reason the question was marked as spam was because of its irrelevance to its original product. In other words, it was a legitimate support request, but not relevant to its original product classification. The spam classification step can provide that extra information in its result without changing its current scope.

I created a [PR based on your work here](https://github.com/akatsoulas/kitsune/pull/9) that shows what I'm thinking. I tested it using two examples that Emil provided and it worked great.

kitsune/questions/utils.py
```diff
@@ -170,15 +170,78 @@ def flag_question(
...
+    topic_result = result.get("topic_result", {})
+    new_product_title = product_result.get("product")
+    new_topic_title = topic_result.get("topic")
+
+    update_fields = {}
+
+    if (
+        new_product_title
+        and hasattr(question, "product")
+        and getattr(question.product, "title", None) != new_product_title
```
escattone: Nit. The `hasattr` and `getattr` aren't necessary, and I think using:
```python
        and question.product
        and (question.product.title != new_product_title)
```
is more readable too.

kitsune/questions/utils.py
```diff
@@ -170,15 +170,78 @@ def flag_question(
...
+    ):
+        from kitsune.products.models import Product
```
escattone: This is not needed. `Product` is already imported above.

kitsune/questions/utils.py
```diff
@@ -170,15 +170,78 @@ def flag_question(
         flagged_object.save()
 
 
+def get_object_by_title(model, title, log_msg, manager=None):
+    try:
+        obj = (manager or model.objects).get(title=title)
+    except (model.DoesNotExist, getattr(model, "MultipleObjectsReturned", Exception)) as exc:
+        log.warning(f"{log_msg} Exception: {exc}", exc_info=True)
```
escattone: I think it would be nice to send this exception to Sentry, since it indicates that either the LLM has selected a bad product or topic title, or something's wrong with the products or topics that we've provided to the LLM in the prompt.

kitsune/questions/utils.py
```diff
@@ -170,15 +170,78 @@ def flag_question(
...
+        and hasattr(question, "product")
+        and getattr(question.product, "title", None) != new_product_title
+    ):
+        from kitsune.products.models import Product
+
+        if new_product := get_object_by_title(
+            Product,
+            new_product_title,
+            f"LLM suggested product {new_product_title} does not exist. Skipping product update.",
+        ):
```
escattone: I think this should also be restricted to active products, so I'd recommend adding `manager=Product.active`.

kitsune/questions/utils.py
```diff
@@ -170,15 +170,78 @@ def flag_question(
...
+
+    if update_fields:
+        for field, value in update_fields.items():
+            setattr(question, field, value)
+        question.save()
+
+        if topic := update_fields.get("topic"):
+            question.clear_cached_tags()
+            question.tags.clear()
+            question.auto_tag()
```
escattone: It seems to me that we'd also need to re-do the question's tags for the case when the question's product has been updated (since the product slug is a tag), but the topic wasn't -- which I guess means the question was moved to a different product, but was also considered spam under that product? Maybe impossible, but maybe not? I'm thinking that it would be better to do something like the following instead:

```python
        # Always update the tags, because we've updated product and/or topic fields.
        question.clear_cached_tags()
        question.tags.clear()
        question.auto_tag()
        # Only flag the question if we've updated the topic.
        if topic := update_fields.get("topic"):
            flag_question(
                ...
            )
```

escattone (approved):
I think this is a nice step forward.

r+wc

PR #6700, opened 2025-06-02
initial prompt and translate function for l10n

escattone:
mozilla/sumo#2355

# Notes
- The `gemini-2.5-pro-preview-05-06` model provides consistently excellent results given my "kitchen-sink" test case, while `gemini-2.5-flash-preview-04-17` frequently makes mistakes, which makes sense given the increased logical complexity of the task  relative to the spam and topic classification tasks.
- It might be worth limiting the use of the `gemini-2.5-pro-preview-05-06` model  to the KB article's `content` only, and use the cheaper/faster `gemini-2.5-flash-preview-04-17` model for the much simpler task of translating the `keywords`, `summary`, and `title`. I haven't done that in this PR.


kitsune/llm/l10n/wiki/prompt.py
```diff
@@ -0,0 +1,183 @@
...
+- The `prior-translation-wiki-map` is a Python `dict` built from the "prior translation", if it is provided.
+- A Python `dict` maps keys to their values.
+- Each `wiki-hook`, `wiki-article-link`, and `wiki-external-link` in the {{ source_language }} text of the "prior translation" becomes a key in the `prior-translation-wiki-map`, and each key's value is its corresponding translation found in the {{ target_language }} text of the "prior translation".
+- If no "prior translation" is provided, the `prior-translation-wiki-map` should be set to an empty `dict`.
+
+# Rules for translating special strings
+- Each of the following rules describes how to translate special strings that may be present within the {{ source_language }} text that you are translating.
+- **Strictly obey** all of these rules when freshly translating {{ source_language }} text.
+
+1. **Preserve unchanged** each of the following strings (each is wrapped with single backticks):
```
akatsoulas: I would extract the following list into a separate file and loop over it here. This makes the prompt easier to read. Eg translation_constants.jinja or something like that

kitsune/llm/l10n/wiki/translator.py
```diff
@@ -0,0 +1,64 @@
...
+    from kitsune.wiki.models import Document
+
+
+def translate(doc: "Document", target_locale: str) -> dict[str, dict[str, Any]]:
+    """
+    Translates the summary, keywords, content, and, conditionally, the title of the
+    given document into the target locale. The given document must be a parent
+    document.
+    """
+    llm = get_llm("gemini-2.5-pro-preview-05-06")
```
akatsoulas: I have a concern that it's going to add confusion to have multiple models defined. If we absolutely need this we can add something like into the utils and instantiate the `get_llm` with that here.
```python
L10N_MODEL="model_name"
```
A simpler solution would be to swap the model for everything. In any case, for this PR the model is irrelevant. I would just use what we have already defined and capture a new issue to switch to this model before we go live

kitsune/llm/l10n/wiki/translator.py
```diff
@@ -0,0 +1,64 @@
...
+
+    translation_chain = translation_prompt | llm | translation_parser
```
akatsoulas: Not for this PR, but I am wondering if there's a way to extract this logic which is similar with the classifier to a method to make it more DRY. Maybe it's worth capturing an issue about it

kitsune/llm/l10n/wiki/utils.py
```diff
@@ -0,0 +1,9 @@
+from django.conf import settings
+
+
+def get_language_in_english(locale: str) -> str:
```
akatsoulas: not sure if we need this or it's becoming to verbose

kitsune/llm/l10n/wiki/prompt.py
```diff
@@ -0,0 +1,183 @@
+from typing import Any
```
akatsoulas: I would remove (at least for now) the `wiki` directory from the l10n. If we ever need the nesting, we can easily update it in the future

kitsune/llm/l10n/wiki/utils.py
escattone: Yeah, it's probably too thin of a wrapper.

kitsune/llm/l10n/wiki/prompt.py
escattone: I like that idea.

PR #6708, opened 2025-06-05
Locale switch from parent non-default locale document should 404

smithellis:
This covers a case where the user is on a document that is not in the default locale (in our case en-US) and is not a translation.
If the user tries to use the locale switcher to go to English, since there is no parent as we are not on a translation, the user should get a 404.

### DIFF
--- a/kitsune/wiki/utils.py
+++ b/kitsune/wiki/utils.py
@@ -203,17 +203,17 @@ def get_visible_document_or_404(
             translation = base_doc.translated_to(locale, visible_for_user=user)
             if translation:
                 return translation
-        else:
-            translation = base_doc.translated_to(locale, visible_for_user=user)
-            if translation:
-                return translation
 
     # Don't try final fallback if not looking for translations or in default language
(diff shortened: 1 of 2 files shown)


kitsune/wiki/utils.py
```diff
@@ -203,17 +203,17 @@ def get_visible_document_or_404(
...
     if not look_for_translation_via_parent or locale == settings.WIKI_DEFAULT_LANGUAGE:
         raise Http404
 
     kwargs["locale"] = settings.WIKI_DEFAULT_LANGUAGE
-    parent = get_object_or_404(Document.objects.visible(user, **kwargs))
+    try:
+        parent = Document.objects.get_visible(user, **kwargs)
+    except Document.DoesNotExist:
+        raise Http404
+
```
escattone, line 216: This change does the exact same thing as what it replaces, so I guess you're making the case that the change is more readable, and I can buy that.

escattone (approved):

PR #6710, opened 2025-06-06
fix and simplify `get_visible_document_or_404`

escattone:
mozilla/sumo#2364
mozilla/sumo#2365

This PR also simplifies `get_visible_document_or_404()`.

### DIFF
--- a/kitsune/wiki/utils.py
+++ b/kitsune/wiki/utils.py
@@ -183,43 +183,31 @@ def get_visible_document_or_404(
     except Document.DoesNotExist:
         pass
 
-    if not (locale := kwargs.get("locale")):
+    if not (look_for_translation_via_parent and (locale := kwargs.get("locale"))):
         raise Http404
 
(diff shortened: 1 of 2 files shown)


kitsune/wiki/utils.py
```diff
@@ -183,44 +183,32 @@ def get_visible_document_or_404(
...
     kwargs["locale"] = settings.WIKI_DEFAULT_LANGUAGE
     try:
-        parent = Document.objects.get_visible(user, **kwargs)
+        return Document.objects.get_visible(user, **kwargs)
     except Document.DoesNotExist:
         raise Http404
 
-    translation = parent.translated_to(locale, visible_for_user=user)
-    if translation:
-        return translation
```
escattone, line 219: I removed these lines because they're redundant. This case is already handled by the `other_docs_qs`-related code above.

smithellis (approved):
Thanks @escattone !

PR #6714, opened 2025-06-10
consider locale when getting products with forums

escattone:
mozilla/sumo#2374

I tested this locally by creating a question exactly like https://support.allizom.org/it/questions/1480814 -- the question that @emilghittasv uses in mozilla/sumo#2374 -- with the `Thunderbird for Android Configuration` enabled for `it` and then disabled for `it`. When enabled, the question was re-classified to `Thunderbird for Android`, and when disabled, the LLM set the `maybe_misclassified` to `True` but since the `Thunderbird for Android` AAQ configuration wasn't enabled for `it`, the `product` returned was `None` -- so the question was marked as spam.

### DIFF
--- a/kitsune/llm/questions/classifiers.py
+++ b/kitsune/llm/questions/classifiers.py
@@ -62,7 +62,9 @@ def classify_question(question: "Question") -> dict[str, Any]:
         if not ((action == ModerationAction.SPAM) and spam_result.get("maybe_misclassified")):
             return {"action": action, "product_result": {}}
 
-        payload["products"] = get_products(only_with_forums=True, output_format="JSON")
+        payload["products"] = get_products(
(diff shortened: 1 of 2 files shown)


akatsoulas:
I would argue that we should ignore the locale completely. When I was implementing the filtering in the `get_products`, initially I was accounting for the locale too. I believe this is not the correct approach here. As far as the classifier is concerned the question is correct and it should go under the `Thunderbird for Android`. If we have an AAQ config enabled for any locale then the classifier needs to be aware of that product.

The solution, if we want to provide one, should lie outside of the classifier. An example of such a solution could be to update the locale in the example given in the issue to `en-us`. Personally, I am not 100% convinced that we should do anything here

escattone:
Good argument @akatsoulas. I'm closing this.

PR #6717, opened 2025-06-11
add metadata to products for LLM use

escattone:
mozilla/sumo#2375

Once this lands, we should consider mozilla/sumo#2377.


kitsune/products/models.py
```diff
@@ -50,6 +50,12 @@ class Product(BaseProductTopic):
     image_cachebuster = models.CharField(max_length=32, default=None, null=True, editable=False)
     sprite_height = models.IntegerField(default=None, null=True, editable=False)
 
+    metadata = models.JSONField(
```
akatsoulas: I would move this to the abstract `BaseProductTopic` and remove it from `Topic` to make it a bit more DRY.

PR #6732, opened 2025-06-23
Use css class instead of a data attribute

akatsoulas:


escattone:
My understanding is that Mozilla's location service (which the `geoip-suggestion` banner functionality relies on) is no longer supported. It was supposed to be shutdown in July of 2024, but it still seems to be running. Perhaps we should disable the `geoip-suggestion` announcement banner -- which is only used for Indonesia and Bangladesh -- until we can fix it?

akatsoulas:
The issue was a bit misleading. We don't have a flickering banner but an empty one. By looking closely, the problem is that we render an empty container when the `caller()` is empty. So we have `<p></p>` before JS kicks in to hide the container

PR #6737, opened 2025-06-25
Improve related document handling

smithellis:
- When adding related documents, you can only add documents that exist in the same locale as the document you're editing
- When displaying related documents, if you are viewing a translated document in a non-en-US locale, you should see related documents translated to that locale.


smithellis:
I happened to notice an issue with this PR - I'm setting it to draft pending a revision.

kitsune/wiki/jinja2/wiki/includes/document_macros.html
```diff
@@ -406,35 +406,32 @@ <h4 class="document-vote--heading">{{ header }}</h4>
   {% endif %}
 {%- endmacro %}
 
-{% macro related_documents(document) %}
-  {% if document.related_documents.count() > 0 %}
-    {% set docs = document.related_documents.all() %}
+{% macro related_documents(related_docs) %}
+  {% if related_docs|length > 0 %}
```
escattone: Nit. You can use `{% if related_docs %}` instead.

kitsune/wiki/views.py
```diff
@@ -595,10 +600,15 @@ def edit_document_metadata(request, document_slug, revision_id=None):
     can_edit_needs_change = doc.allows(user, "edit_needs_change")
     can_archive = doc.allows(user, "archive")
 
+    # Get related docs for both initial values and form choices
+    related_documents = _get_related_documents_for_locale(doc, request.LANGUAGE_CODE, request.user)
+
```
escattone: This is not needed (see details below).

kitsune/wiki/views.py
```diff
@@ -595,10 +600,15 @@ def edit_document_metadata(request, document_slug, revision_id=None):
...
     doc_form = DocumentForm(
-        initial=_document_form_initial(doc),
+        initial=_document_form_initial(doc, request.LANGUAGE_CODE, request.user),
         can_archive=can_archive,
         can_edit_needs_change=can_edit_needs_change,
+        locale=request.LANGUAGE_CODE,
+        related_documents=related_documents,
```
escattone: The `related_documents` is not needed here. The related documents are provided via the initial data when constructing the `DocumentForm` and then rendered in the view via the form.

kitsune/wiki/views.py
```diff
@@ -612,6 +622,8 @@ def edit_document_metadata(request, document_slug, revision_id=None):
             instance=doc,
             can_archive=can_archive,
             can_edit_needs_change=can_edit_needs_change,
+            locale=request.LANGUAGE_CODE,
+            related_documents=related_documents,
```
escattone: You don't need the `related_documents` here either. The edited related documents are provided via the `post_data`.

kitsune/wiki/views.py
```diff
@@ -1655,8 +1670,18 @@ def show_translations(request, document_slug):
     )
 
 
-def _document_form_initial(document):
+def _document_form_initial(document, locale=None, user=None):
```
escattone: We don't need the `locale` here.

kitsune/wiki/views.py
```diff
@@ -1655,8 +1670,18 @@ def show_translations(request, document_slug):
...
+    # Get related documents if locale and user are provided
+    if locale and user:
+        related_documents = _get_related_documents_for_locale(document, locale, user)
+        related_document_ids = [doc.id for doc in related_documents]
+    else:
+        # Fallback to original behavior
+        related_document_ids = Document.objects.filter(related_documents=document).values_list(
+            "id", flat=True
+        )
+
```
escattone: Since we're initializing the related documents of the form with the document's existing related documents, I think the only concern here should be restricting visibility. We don't need the locale, because this function is only called when editing metadata -- which can only be done on parent documents -- and also within the `translate` view, but the related documents are never shown for translations, since they should only be viewed/edited via parent documents.

I think this can be simplified to:

```python
    related_document_ids = Document.objects.visible(
        user, id__in=document.original.related_documents.all()
    ).values_list("id", flat=True)
```

The `document.original` is only to ensure that we're always using the parent document's related documents, but if the document was not a parent document, its related documents would never be shown anyway.

kitsune/wiki/views.py
```diff
@@ -1822,6 +1845,48 @@ def get_fallback_locale(doc, request):
     return (None, True)
 
 
+def _get_related_documents_for_locale(document, locale, user):
```
escattone, line 1879: I think this function is confusing, and also it introduces an N+1 query problem. As far as I can tell, it should only be called within the `document()` view to get the related documents to display to the user as cards, so it should take care of visibility, stripping-out related documents that have been archived, and handling translations with fallback to the default if necessary. It can be simplified to something like:

```python
def get_related_documents_for_display(document, locale, user):
    """Return visible related documents prioritizing the given locale."""

    related_docs = document.original.related_documents.filter(is_archived=False).all()

    return (
        Document.objects.visible(user, id__in=related_docs)
        .annotate(
            translation_id=Subquery(
                Document.objects.filter(parent=OuterRef("pk"), locale=locale).values("id")
            )
        )
        .annotate(desired_id=Coalesce("translation_id", "id"))
        .filter(id=F("desired_id"))
    )
```

kitsune/wiki/forms.py
```diff
@@ -228,6 +228,8 @@ def __init__(self, *args, **kwargs):
         can_archive = kwargs.pop("can_archive", False)
         can_edit_needs_change = kwargs.pop("can_edit_needs_change", False)
         initial_title = kwargs.pop("initial_title", "")
+        locale = kwargs.pop("locale", None)
+        related_documents = kwargs.pop("related_documents", None)
```
escattone: The `related_documents` is not needed.

kitsune/wiki/forms.py
```diff
@@ -244,7 +246,28 @@ def __init__(self, *args, **kwargs):
...
+
+            related_documents_field.choices = locale_choices
+
+            # Update the widget with related documents for proper rendering
+            if related_documents:
+                related_documents_field.widget = RelatedDocumentsWidget(
+                    related_documents=related_documents
+                )
+        else:
+            related_documents_field.choices = Document.objects.values_list("id", "title")
```
escattone: I think this entire section can be simplified to something like:

```python
        related_documents_choices = Document.objects.filter(is_template=False)
        if locale:
            related_documents_choices = related_documents_choices.filter(locale=locale)
        related_documents_field.choices = related_documents_choices.values_list("id", "title")
```

I think `locale` is only needed for the case of a non-English parent document.

kitsune/wiki/widgets.py
escattone, line 1: I don't think any of the changes in this file are needed.

kitsune/wiki/jinja2/wiki/includes/related_docs_widget.html
escattone, line 1: Even with my comment on `kitsune/wiki/widgets.py` -- that its changes are not needed -- I think these changes, although not needed, are fine. I like them a bit better because the `if` statements will cause  `.exists()` to be run on the Django queryset rather than `.count()`.

escattone (changes requested):

kitsune/wiki/views.py
smithellis: You're absolutely right about the N+1 issue, and your suggested query revealed an even bigger problem - it filters out documents incorrectly when translations exist. I've rewritten the function to fix both issues, using prefetch_related to avoid N+1 queries and implementing proper fallback logic for translations.

kitsune/wiki/views.py
smithellis: You're right about removing the redundant parameter passing. I've cleaned this up.

kitsune/wiki/jinja2/wiki/includes/document_macros.html
smithellis: You are right, this is cleaner. Updated.

kitsune/wiki/views.py
smithellis: You're correct - the form handles this through its initial data. I've removed the redundant parameter.

kitsune/wiki/views.py
smithellis: Agreed, removed the redundant parameter from the POST handling.

kitsune/wiki/views.py
smithellis: You're right that metadata editing is only done on parent documents, but we do need the locale parameter for a different reason: to determine which translations to show in the related documents field.
For example, when editing a Romanian parent document, we want to show Romanian translations of related documents (when they exist) rather than their English versions. This provides a better UX as users see titles in their own language. The locale parameter enables this translation lookup in _document_form_initial.

kitsune/wiki/views.py
smithellis: I tried your simplification, but it causes issues with non-English documents. The challenge is that related documents are stored as parent document references, but for UX we need to show translations in the TomSelect. For example, when editing a Romanian document, we want to show 'Flash 11.3 crashes Romanian' (the Romanian translation) rather than 'Flash 11.3 crashes' (the English parent).
The locale parameter is needed to find and display these translations. I've kept the visibility filtering as you suggested, but added translation lookup for better UX.

kitsune/wiki/forms.py
smithellis: Ah, got it.

kitsune/wiki/forms.py
smithellis: I initially simplified this as you suggested, but testing revealed that we need more complex handling here. The issue is that when a user selects a translated document 'Flash 11.3 crashes Romanian' with ID 50, we need to store its parent document ID (ID 1 for 'Flash 11.3 crashes').
The form needs to:
  - Show translated titles for better UX
  - Accept translation IDs as valid input
  - Map them to parent IDs for storage

So the additional complexity in the choices setup is actually necessary to handle this translation/parent mapping correctly. Without it, users get validation errors when selecting translated documents.

kitsune/wiki/widgets.py
smithellis: You're right, I've reverted all changes to widgets.py. The widget works fine with the existing implementation.

kitsune/wiki/jinja2/wiki/includes/related_docs_widget.html
smithellis: Thanks for the feedback! I've kept these template improvements since they're more efficient.

akatsoulas:
@escattone my understanding for this PR is that it's almost ready for merging. Since you are the one the reviewed it, can you take a look and decide on next steps? Thanks!

escattone:
Closing in favor of #6939

PR #6742, opened 2025-06-26
Delete kbforum and forum posts or threads created by user being deactivated

smithellis:


kitsune/questions/utils.py
```diff
@@ -57,6 +59,20 @@ def mark_content_as_spam(user, by_user):
     for answer in Answer.objects.filter(creator=user):
         answer.mark_as_spam(by_user)
 
+    # We will delete kbforums posts and threads
+    for thread in KBForumThread.objects.filter(creator=user):
```
akatsoulas: I wouldn't add this here but rather in the view that handles the deactivation and the marking of spam

akatsoulas: Most probably you can use just filter here - probably contributors and kb forums won't have the same problem with AAQ

smithellis: Thanks @akatsoulas ! I'll revise.

akatsoulas (approved):
Looks good! Let's merge after the release. Please squash before merging

PR #6743, opened 2025-06-26
6725 sorting forum posts issue

smithellis:
Nulls rise to the top of the sort order with Postgres - this corrects that issue for Questions by changing nulls to zero.

### DIFF
--- a/kitsune/questions/views.py
+++ b/kitsune/questions/views.py
@@ -13,8 +13,8 @@ from django.contrib.contenttypes.models import ContentType
 from django.contrib.sites.models import Site
 from django.core.exceptions import PermissionDenied
 from django.core.paginator import EmptyPage, PageNotAnInteger
-from django.db.models import Q
-from django.db.models.functions import Now
+from django.db.models import IntegerField, Q, Value
+from django.db.models.functions import Coalesce, Now
 from django.http import (
     Http404,
     HttpRequest,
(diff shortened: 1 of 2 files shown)


kitsune/questions/views.py
```diff
@@ -286,7 +286,19 @@ def question_list(request, product_slug=None, topic_slug=None):
...
+        question_qs = question_qs.annotate(
+            visits_nulls_as_zero=Coalesce(
+                "questionvisits__visits", Value(0), output_field=IntegerField()
+            )
+        )
+        question_qs = question_qs.order_by(
+            "visits_nulls_as_zero" if sort == "asc" else "-visits_nulls_as_zero"
+        )
+    else:
+        question_qs = question_qs.order_by(order_by if sort == "asc" else "-%s" % order_by)
```
escattone: [Django provides a nice way to deal with the NULL sorting issue](https://docs.djangoproject.com/en/5.2/ref/models/expressions/#using-f-to-sort-null-values) via the `nulls_first` and `nulls_last` keyword arguments to the `asc()` or `desc()` methods of `F()` instances, so you could do something like the following to cover all cases:

```python
if sort == "asc":
    order_by_expression = F(order_by).asc(nulls_first=True)
else:
    order_by_expression = F(order_by).desc(nulls_last=True)
question_qs = question_qs.order_by(order_by_expression)
```

There are several examples throughout the code that you can use as reference, for example https://github.com/mozilla/kitsune/blob/84d1badd7292e180074184259e991ac9b21917a9/kitsune/forums/views.py#L59.

smithellis: Thanks @escattone - awesome.

smithellis: Revised.

escattone (approved):

PR #6745, opened 2025-06-30
Expand assignee filter to the flagged queue.

akatsoulas:


akatsoulas:
The idea here is to quickly inspect the list of what Sumobot did. We are not assigning users to the flagged queue so it's going to be always unassigned. I am skipping for now until at least we have assignees in the flagged queue. Thanks for the review!

PR #6747, opened 2025-07-01

smithellis:


kitsune/users/management/commands/delete_inactive_users.py
```diff
@@ -0,0 +1,232 @@
...
+            help="How often to report progress (default: every 1000 users)",
+        )
+        parser.add_argument(
+            "--cutoff-date",
+            type=str,
+            help="Specific cutoff date in YYYY-MM-DD format (overrides --years)",
+        )
+
+    def handle(self, *args, **options):
+        User = get_user_model()
```
akatsoulas: It feels a bit redundant to call this method twice. You have a method to do the deletions but the queries are happening within the handle method. I would suggest one of the following

* Move everything in handle or
* add a method like `_find_and_delete_users` which will do the queries and deletions. This will also follow the paradigm that you set with the _bulk_delete_batch and _repor_progess and handle will only figure out the date span and call the _find_and_delete_users.

kitsune/users/management/commands/delete_inactive_users.py
```diff
@@ -0,0 +1,232 @@
...
+        start_time = time.time()
+        processed_count = 0
+        pipeline_count = 0
+        direct_count = 0
+        error_count = 0
+
+        self.stdout.write(
+            self.style.SUCCESS(f"Starting deletion of {total_users:,} inactive users")
+        )
+        if users_without_content.exists():
```
akatsoulas: I believe you can omit this query completely. Probably initializing an empty list like `users_no_content_batch` is cheaper than querying the database.

kitsune/users/management/commands/delete_inactive_users.py
```diff
@@ -0,0 +1,232 @@
...
+                        error_count,
+                        start_time,
+                    )
+            if users_no_content_batch:
+                batch_direct_count, batch_error_count = self._bulk_delete_batch(
+                    User, users_no_content_batch
+                )
+                direct_count += batch_direct_count
+                error_count += batch_error_count
+        if users_with_content.exists():
```
akatsoulas: Similar comment applies here too

kitsune/users/management/commands/delete_inactive_users.py
```diff
@@ -0,0 +1,232 @@
...
+
+    def _bulk_delete_batch(self, User, user_ids):
+        if not user_ids:
+            return 0, 0
+
+        try:
+            _, deleted_objects = User.objects.filter(id__in=user_ids).delete()
+            user_count = deleted_objects.get("auth.User", 0)
+            return user_count, 0
+        except Exception as e:
```
akatsoulas: This should be a specific exception

kitsune/users/management/commands/delete_inactive_users.py
```diff
@@ -0,0 +1,232 @@
...
+                    "Check error messages above."
+                )
+            )
+
+    def _bulk_delete_batch(self, User, user_ids):
+        if not user_ids:
+            return 0, 0
+
+        try:
+            _, deleted_objects = User.objects.filter(id__in=user_ids).delete()
```
akatsoulas: This probably can go outside of the try/except clause

kitsune/users/management/commands/delete_inactive_users.py
```diff
@@ -0,0 +1,232 @@
...
+        pipeline: int,
+        direct: int,
+        errors: int,
+        start_time: float,
+    ):
+        elapsed_time = time.time() - start_time
+        progress_pct = processed / total * 100
+
+        if elapsed_time > 0:
+            current_rate = processed / elapsed_time * 60
```
akatsoulas: The if could be removed and the current_rate could be condensed to:
`current_rate = (processed / elapsed_time * 60) if elapsed_time > 0 else 0`

kitsune/users/management/commands/delete_inactive_users.py
```diff
@@ -0,0 +1,240 @@
...
+
+        users_with_content = User.objects.filter(last_login__lt=cutoff_date).filter(
+            has_content_criteria
+        )
+        users_without_content = User.objects.filter(last_login__lt=cutoff_date).exclude(
+            has_content_criteria
+        )
+        total_users = users_with_content.count() + users_without_content.count()
+        content_count = users_with_content.count()
+        no_content_count = users_without_content.count()
```
akatsoulas, line 107: nit:  this could be rewritten as
```python
content_count = users_with_content.count()
no_content_count = users_without_content.count()
total_users = content_count + no_content_count()
```

kitsune/users/management/commands/delete_inactive_users.py
```diff
@@ -0,0 +1,240 @@
...
+                f"  Users with content: {content_count:,} (will use delete_user_pipeline)\n"
+                f"  Users without content: {no_content_count:,} (will use bulk delete)"
+            )
+        )
+
+        if dry_run:
+            self.stdout.write(self.style.SUCCESS("DRY RUN: No users were actually deleted"))
+            return
+
+        self._delete_users(users_with_content, users_without_content, total_users, options)
```
akatsoulas: this is not need. Just add the code of the _delete users in here to avoid code replication. `_delete_users` is not used anywhere else

akatsoulas (approved):
r+wc

kitsune/users/management/commands/delete_inactive_users.py
```diff
@@ -0,0 +1,235 @@
...
+                f"  Users with content: {content_count:,} (will use delete_user_pipeline)\n"
+                f"  Users without content: {no_content_count:,} (will use bulk delete)"
+            )
+        )
+
+        if dry_run:
+            self.stdout.write(self.style.SUCCESS("DRY RUN: No users were actually deleted"))
+            return
+
+        User = get_user_model()
```
akatsoulas: This is already defined above

kitsune/users/management/commands/delete_inactive_users.py
```diff
@@ -0,0 +1,235 @@
...
+        batch_size = options["batch_size"]
+        progress_interval = options["progress_interval"]
+
+        start_time = time.time()
+        processed_count = 0
+        pipeline_count = 0
+        direct_count = 0
+        error_count = 0
```
akatsoulas: I would put all the initialization code at the beginning of the function

akatsoulas (approved):
r+wc

PR #6748, opened 2025-07-01
Fix url warnings in document patterns urls

akatsoulas:


akatsoulas:
This is the first batch of changes in the wiki app which will help us minimize the chance of getting any regressions. Once it's merged I will continue with the next batch of urls.

kitsune/wiki/urls.py
```diff
@@ -8,52 +8,52 @@
 # These patterns inherit (?P<document_slug>[^\/]).
 document_patterns = [
     re_path(r"^$", views.document, name="wiki.document"),
```
escattone: You'll have to move this one out of the `document_patterns` and into the `urlpatterns` in order to avoid forcing all document URL's to include a trailing `/`.

akatsoulas:
@escattone pr updated

akatsoulas:
@escattone PR updated

kitsune/urls.py
```diff
@@ -21,7 +21,7 @@
 admin.autodiscover()
 
 urlpatterns = i18n_patterns(
-    path("kb", include("kitsune.wiki.urls")),
+    path("kb/", include("kitsune.wiki.urls")),
```
escattone, line 26: Wouldn't we have to move:
```python
    re_path(
        "^kb$", redirect_to, {"url": "products.product", "slug": "firefox"}, name="wiki.landing"
    ),
```
above this in order to keep the `wiki.landing` endpoint the same? Otherwise, it seems that it'll require a trailing forward-slash

PR #6750, opened 2025-07-02
Return archived question in search only if using advanced search operations

smithellis:


kitsune/search/search.py
```diff
@@ -104,6 +114,37 @@ def get_settings(self):
...
+            return True
+
+        try:
+            parsed = Parser(self.query)
+            return self._is_simple_parsed_token(parsed.parsed)
+        except ParseException:
+            # If parsing fails, it's definitely a simple search
+            return True
+
+    def _is_simple_parsed_token(self, token):
```
akatsoulas: I don't believe we need this method atm. Let's add it if we ever have to. For now, let's just move everything under is_simple_search

kitsune/search/search.py
```diff
@@ -124,10 +165,13 @@ def get_filter(self):
                     "gte": datetime.now(timezone.utc) - timedelta(days=QUESTION_DAYS_DELTA)
                 },
             ),
-            # exclude archived questions
-            DSLQ("term", question_is_archived=False),
         ]
 
+        # Only exclude archived questions for simple searches
+        # Advanced searches (with field:, AND, OR, NOT, etc.) should include archived questions
```
akatsoulas: This is already covered in the comments above.

kitsune/search/search.py
```diff
@@ -104,6 +114,37 @@ def get_settings(self):
...
+        try:
+            parsed = Parser(self.query)
+            return self._is_simple_parsed_token(parsed.parsed)
+        except ParseException:
+            # If parsing fails, it's definitely a simple search
+            return True
+
+    def _is_simple_parsed_token(self, token):
+        """Recursively check if a parsed token represents a simple search."""
+        # Advanced operators and tokens indicate an advanced search
```
akatsoulas, line 139: I would add the cases of which searches are advanced vs simple in the docstring

smithellis:
I refactored the token testing into `is_simple_search` and made it recursively test there. Much cleaner code - thanks for the suggestion.

akatsoulas (approved):
🚀

PR #6756, opened 2025-07-10
Move to uv

akatsoulas:


mcmcaxm (approved):

.pre-commit-config.yaml
```diff
@@ -8,17 +8,13 @@ repos:
       - id: check-case-conflict
       - id: check-merge-conflict
       - id: debug-statements
-  - repo: https://github.com/psf/black
-    rev: 25.1.0
+  - repo: https://github.com/astral-sh/ruff-pre-commit
+    rev: v0.4.0
```
escattone: I think we can use the latest instead (`v0.12.3` as of this comment).

Dockerfile
```diff
@@ -23,18 +22,16 @@ RUN set -xe \
...
-    # python
-    && python -m venv /venv \
-    && pip install --upgrade pip==${PIP_VERSION} \
-    && pip install --upgrade poetry==${POETRY_VERSION} \
-    && poetry config virtualenvs.create false \
     # clean up
     && rm -rf /var/lib/apt/lists/*
 
+# Copy uv from official image
+COPY --from=ghcr.io/astral-sh/uv:0.7.19 /uv /uvx /bin/
```
escattone: Could bump this to `uv:0.7.20`

escattone (approved):
r+wc

Nice work! I didn't check every file, but looked at most of the key ones. I think this is a nice step.

PR #6786, opened 2025-07-22
Update old VPN references

akatsoulas:


kitsune/customercare/forms.py
escattone, line 1: Do we also want to adjust the product slug used within the `ZendeskForm` to use `firefox-private-network-vpn` instead of `mozilla-vpn` so that the Zendesk support team doesn't have to change anything on their end?

kitsune/customercare/forms.py
```diff
@@ -66,8 +69,13 @@ def __init__(self, *args, product, user=None, **kwargs):
...
 
     def send(self, user, product):
         client = ZendeskClient()
-        return client.create_ticket(user, self.cleaned_data, product)
+        zendesk_product = (
+            ZENDESK_PRODUCT_SLUGS.get(product.slug, product.slug)
+            if product.slug == "mozilla-vpn"
+            else product.slug
+        )
+        return client.create_ticket(user, self.cleaned_data, zendesk_product)
```
escattone: I thought we'd also have to make this same adjustment to the product field within the cleaned data, but maybe not?

PR #6789, opened 2025-07-23
docs: update Elasticsearch docs with management commands and deployment info

smithellis:
Add comprehensive documentation for `es_init` and `es_reindex` management commands including usage examples, migration workflows, and performance tuning options. Integrate zero-downtime production deployment process with step-by-step instructions for Elasticsearch upgrades. Fix command syntax inconsistencies and update outdated version references.

### DIFF
--- a/docs/elastic_search.md
+++ b/docs/elastic_search.md
@@ -7,7 +7,7 @@
 
 Elastic supports adding new fields to an existing mapping,
 along with some other operations:
-https://www.elastic.co/guide/en/elasticsearch/reference/current/mapping.html#add-field-mapping
+https://www.elastic.co/guide/en/elasticsearch/reference/9.0/mapping.html#add-field-mapping
 
 To know whether a change you make to a Document will work in prod,
 try it locally having already set up the mapping:
@@ -36,7 +36,7 @@ So to ensure no data is missing from the index,
 run something like:
 
 ```
-./manage.py es_reindex --limit TestDocument --updated-after "2024-01-01 10:00:00" --updated-before "2024-01-01 12:00:00"
+./manage.py es_reindex --limit TestDocument --updated-after <datetime of deploy> --updated-before <datetime of mapping update>
 ```
 
 ### Indexing performance
@@ -61,7 +61,7 @@ to bring that number down.
 
 ## Search Management Commands
 
-Kitsune provides two key management commands for working with Elasticsearch indices: `es_init` and `es_reindex`. These commands handle index initialization, migration, and document reindexing.
+Kitsune provides two key management commands for working with Elasticsearch indices: `es_init` and `es_reindex`. These commands handle index initialization, migration via aliases, and document reindexing.
 
 ### es_init Command
 
@@ -80,20 +80,19 @@ The `es_init` command initializes Elasticsearch document types and manages index
 ./manage.py es_init --reload-search-analyzers
 ```
 
-#### Migration Options
+#### Migration Options Using Aliases
 
 The command supports two types of migrations:
 
-**Write Migration (`--migrate-writes`)**
+**Write Migration via Alias (`--migrate-writes`)**
 - Creates a new index with a timestamp suffix
 - Points the `_write` alias to the new index
-- Used when you need to create a completely new index with updated mappings
 
 ```bash
 ./manage.py es_init --migrate-writes
 ```
 
-**Read Migration (`--migrate-reads`)**
+**Read Migration via Alias (`--migrate-reads`)**
 - Updates the `_read` alias to point where the `_write` alias points
 - Used to switch read operations to the new index after write migration is complete
 
@@ -140,17 +139,17 @@ The `es_reindex` command populates Elasticsearch indices with data from the data
 ./manage.py es_reindex --count 1000
 ```
 
-#### Time-Based Filtering
+#### Time-Based Indexing
 
 ```bash
 # Reindex documents updated after a specific date
-./manage.py es_reindex --updated-after "2024-01-01"
+./manage.py es_reindex --updated-after "<updated date>"
 
 # Reindex documents updated before a specific date
-./manage.py es_reindex --updated-before "2024-01-01 12:00:00"
+./manage.py es_reindex --updated-before "<updated date>"
 
 # Reindex documents updated within a date range
-./manage.py es_reindex --updated-after "2024-01-01" --updated-before "2024-01-02"
+./manage.py es_reindex --updated-after "<updated date>" --updated-before "<updated date>"
 ```
 
 #### Performance Tuning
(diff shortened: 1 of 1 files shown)


docs/elastic_search.md
```diff
@@ -36,7 +36,7 @@ So to ensure no data is missing from the index,
 run something like:
 
 ```
-./manage.py es_reindex --limit TestDocument --updated-after <datetime of deploy> --updated-before <datetime of mapping update>
+./manage.py es_reindex --limit TestDocument --updated-after "2024-01-01 10:00:00" --updated-before "2024-01-01 12:00:00"
 ```
```
akatsoulas: let's revert this change, the previous example is more descriptive.

docs/elastic_search.md
```diff
@@ -7,7 +7,7 @@
 
 Elastic supports adding new fields to an existing mapping,
 along with some other operations:
-https://www.elastic.co/guide/en/elasticsearch/reference/7.9/mapping.html#add-field-mapping
+https://www.elastic.co/guide/en/elasticsearch/reference/current/mapping.html#add-field-mapping
```
akatsoulas: Let's update this to the version that we using and not `current`. If we get stuck with version 9 and ES is at version 12, current will point to 12

docs/elastic_search.md
```diff
@@ -59,6 +59,172 @@ Consider using some combination of
 or [annotations](https://docs.djangoproject.com/en/dev/ref/models/querysets/#annotate)
 to bring that number down.
 
+## Search Management Commands
+
+Kitsune provides two key management commands for working with Elasticsearch indices: `es_init` and `es_reindex`. These commands handle index initialization, migration, and document reindexing.
```
akatsoulas: `migration` might be adding confusion here, at least it does for me. What do you have in mind here?

docs/elastic_search.md
```diff
@@ -59,6 +59,172 @@ Consider using some combination of
...
+./manage.py es_init
+
+# Initialize specific document types only
+./manage.py es_init --limit WikiDocument QuestionDocument
+
+# Reload search analyzers (useful after updating synonyms)
+./manage.py es_init --reload-search-analyzers
+```
+
+#### Migration Options
```
akatsoulas: I would call this `Aliases` and not migration option
https://www.elastic.co/docs/manage-data/data-store/aliases

akatsoulas: This applies to the whole section

docs/elastic_search.md
```diff
@@ -59,6 +59,172 @@ Consider using some combination of
...
+
+The command supports two types of migrations:
+
+**Write Migration (`--migrate-writes`)**
+- Creates a new index with a timestamp suffix
+- Points the `_write` alias to the new index
+- Used when you need to create a completely new index with updated mappings
```
akatsoulas: I would remove this line, users may assume that they don't need to use this for an existing index which is not true

docs/elastic_search.md
```diff
@@ -59,6 +59,172 @@ Consider using some combination of
...
+./manage.py es_reindex --limit WikiDocument QuestionDocument
+
+# Reindex a percentage of documents (useful for testing)
+./manage.py es_reindex --percentage 10
+
+# Reindex a specific number of documents
+./manage.py es_reindex --count 1000
+```
+
+#### Time-Based Filtering
```
akatsoulas: s/Filtering/Indexing

docs/elastic_search.md
```diff
@@ -59,6 +59,172 @@ Consider using some combination of
...
+# Set custom timeout for bulk operations
+./manage.py es_reindex --timeout 60
+
+# Monitor SQL query performance
+./manage.py es_reindex --print-sql-count --count 100
+```
+
+#### Production Reindexing
+
+For production environments, use `CELERY_TASK_ALWAYS_EAGER=True` to process tasks synchronously:
```
akatsoulas: Why are you suggesting to skip Celery here?

docs/elastic_search.md
```diff
@@ -59,6 +59,172 @@ Consider using some combination of
...
+```bash
+# Reindex all document types in production
+CELERY_TASK_ALWAYS_EAGER=True ./manage.py es_reindex --limit WikiDocument
+CELERY_TASK_ALWAYS_EAGER=True ./manage.py es_reindex --limit QuestionDocument
+CELERY_TASK_ALWAYS_EAGER=True ./manage.py es_reindex --limit AnswerDocument
+CELERY_TASK_ALWAYS_EAGER=True ./manage.py es_reindex --limit ProfileDocument
+CELERY_TASK_ALWAYS_EAGER=True ./manage.py es_reindex --limit ForumDocument
+```
+
+#### How It Works
```
akatsoulas: I would skip this entirely. Although what you have is not inaccurate, it mixes different things and it gives the impression that all of these things are happening by the es_reindex command

docs/elastic_search.md
```diff
@@ -59,6 +59,172 @@ Consider using some combination of
...
+- Processes documents in configurable chunks to avoid memory issues
+- Sends bulk requests to Elasticsearch for efficient indexing
+- Uses Celery tasks for background processing (unless `CELERY_TASK_ALWAYS_EAGER=True`)
+- Supports both full reindexing and incremental updates based on modification dates
+
+### Common Workflows
+
+#### Creating New Indices
+
+When you need to create completely new indices (e.g., for mapping changes that require reindexing):
```
akatsoulas, line 184: I would remove the reference to new indices. Maybe we should phrase this in a way that indicates that we have mapping changes  that require reindexing. Using the word new is ambiguous here.

docs/elastic_search.md
```diff
@@ -59,6 +59,172 @@ Consider using some combination of
...
+
+```bash
```
akatsoulas, line 186: The same block is present in the Basic usage section

docs/elastic_search.md
```diff
@@ -59,6 +59,172 @@ Consider using some combination of
...
+./manage.py es_init --migrate-reads
+```
+
+#### Updating Existing Indices
+
+For compatible mapping changes that don't require full reindexing:
+
+```bash
+# Update mapping on existing index
+./manage.py es_init
```
akatsoulas, line 203: is this needed for reindexing data without mapping changes?

docs/elastic_search.md
```diff
@@ -59,6 +59,172 @@ Consider using some combination of
...
+
+# Optionally reindex recently updated documents
+./manage.py es_reindex --updated-after "2024-01-01"
```
akatsoulas: I would use something like `<<timeststamp before the reindex>>` or something to indicate the slice

docs/elastic_search.md
```diff
@@ -107,7 +273,7 @@ to `kitsune.search.views.simple_search`.
 
 The `kitsune/search/dictionaries/synonyms` path contains a text file for each of our search-enabled locales,
 where synonyms are in the
-[Solr format](https://www.elastic.co/guide/en/elasticsearch/reference/7.10/analysis-synonym-graph-tokenfilter.html#_solr_synonyms_2).
+[Solr format](https://www.elastic.co/guide/en/elasticsearch/reference/current/analysis-synonym-graph-tokenfilter.html#_solr_synonyms_2).
```
akatsoulas: Use specific version here

docs/elastic_search.md
```diff
@@ -303,3 +469,165 @@ however in practice plural words tend to be covered by the singular replacement
 (e.g. "add on" is a substring in "add ons",
 so "add ons" is replaced by "addons")
 and there is marginal benefit to defining _every single_ conjugation of a verb.
+
+## Production Deployment
```
akatsoulas: I would remove this completely. This is one possible way to do a major upgrade and is related to the current setup that we have. There are many others. It's not related to how search work and it should not dictate the policy on how to perform major version upgrades

akatsoulas:
Let's capture in a ticket for the current sprint

docs/elastic_search.md
smithellis: There is a migration flow between the two commands - setting up a new index, pointing writes to it, reindexing, then pointing reads.

docs/elastic_search.md
smithellis: I think it would be more confusing to label the section Aliases and then demonstrate three migration patterns. In all cases you are migrating to a new index. Maybe I call it "Using Aliases To Migrate Indexes".

docs/elastic_search.md
smithellis: No. If you aren't altering the index mapping, you only need `es_reindex`.

docs/elastic_search.md
smithellis: Like below, I think I'll add "via aliases" to be specific.

docs/elastic_search.md
smithellis: I removed this - this was an artifact of how we had to use the admin pods we created and isn't relevant to normal usage.

docs/elastic_search.md
smithellis: It creates a new index that you then re-index to - which you have to do when mappings change. I think it causes confusion if we don't differentiate between document types (`WikiDocument`, for instance) and an index (`sumo_wikidocument_<date>`). You will create new indexes that you then populate with your known document types - but that is what the command is doing, it is making a new index. This block in the docs is specific to a case where you need a new index. You do not need to create a new index if you are just reindexing without any mapping or settings changes.

PR #6796, opened 2025-07-28
improve contributor metrics

escattone:
mozilla/sumo#2428

## Notes
This work turned out to be much more involved than I anticipated, but it provides a solid "shim" to address deleted contributions until we have the will/time to create a more general solution.
- This PR addresses contributor metrics in the following areas:
    - The contribution metrics in the user profile (https://support.mozilla.org/en-US/user/{username}/)
    - https://support.mozilla.org/en-US/community/top-contributors/questions
    - https://support.mozilla.org/en-US/community/top-contributors/kb
    - https://support.mozilla.org/en-US/community/top-contributors/l10n
    - The awarding of revision and answer related badges.
- It does not yet address the following contribution details pages linked from the contributor's profile. That's left to another step, if desired.
    - https://support.mozilla.org/en-US/user/{username}/answers
    - https://support.mozilla.org/en-US/user/{username}/documents
- This PR fixes a bug in `top_contributors_l10n` where it was counting all `created_revisions`, when it should have been counting the `created_revisions` filtered by `revisions`.
- This PR also fixes a paging bug in `top_contributors_l10n` where it wasn't ensuring a predictable ordering **within** equal counts.


kitsune/community/apps.py
```diff
@@ -4,3 +4,6 @@
 class CommunityConfig(AppConfig):
     name = "kitsune.community"
     default_auto_field = "django.db.models.AutoField"
+
+    def ready(self):
+        from kitsune.community import signals  # noqa
```
akatsoulas, line 9: It doesn't even matter for the context of this PR so just ignore the comment. I am just passing along info: I believe `noqa` is not needed any more with ruff.

kitsune/community/models.py
```diff
@@ -0,0 +1,27 @@
...
+from kitsune.products.models import Product
+from kitsune.sumo.models import LocaleField
+
+
+class DeletedContribution(models.Model):
+    created = models.DateTimeField(auto_now_add=True)
+
+    # Contribution-related fields
+    content_type = models.ForeignKey(ContentType, on_delete=models.CASCADE)
+    # Use "metadata" to store any additional information needed for specific content types.
```
akatsoulas: the comments here feel a bit redundant

kitsune/community/models.py
```diff
@@ -0,0 +1,27 @@
...
+    metadata = models.JSONField(null=True, default=None)
```
akatsoulas, line 29: my main feedback is around json. I would propose some kind of validation like pydantic. This is just an example by using AI and not tested!

```python
from pydantic import BaseModel, Field, field_validator
from typing import Optional, List

class BaseMetadata(BaseModel):
    """Base metadata with common fields."""
    content_length: Optional[int] = Field(None, description="Length of content")
    tags: Optional[List[str]] = Field(default_factory=list, description="Content tags")

class AnswerMetadata(BaseMetadata):
    content: str = Field(..., min_length=1, description="The answer content")
    question_title: str = Field(..., min_length=1, description="Title of the question")
    is_solution: bool = Field(default=False, description="Whether this was the solution")
    
    @field_validator('content')
    def validate_content_not_empty(cls, v):
        if not v.strip():
            raise ValueError("Content cannot be empty")
        return v

class RevisionMetadata(BaseMetadata):
    is_approved: bool = Field(..., description="Whether revision was approved")
    document_title: Optional[str] = Field(None, description="Title of the document")

class DocumentMetadata(BaseMetadata):
    title: str = Field(..., min_length=1, description="Document title")
    slug: Optional[str] = Field(None, description="Document slug")

class DeletedContribution(models.Model):
    # ... existing fields ...
    
    def _validate_metadata(self):
        """Validate metadata using Pydantic models."""
        try:
            if self.content_type == self.ContentTypeChoices.ANSWER:
                AnswerMetadata(**self.metadata)
            elif self.content_type == self.ContentTypeChoices.REVISION:
                RevisionMetadata(**self.metadata)
            elif self.content_type == self.ContentTypeChoices.DOCUMENT:
                DocumentMetadata(**self.metadata)
        except ValidationError as e:
            raise ValidationError(f"Metadata validation failed: {e}")
    
    def save(self, *args, **kwargs):
        self._validate_metadata()
        super().save(*args, **kwargs)
```

kitsune/community/signals.py
```diff
@@ -0,0 +1,85 @@
...
+from django.db.models.signals import pre_delete
+from django.dispatch import receiver
+
+from kitsune.community.models import DeletedContribution
+from kitsune.questions.models import Answer
+from kitsune.wiki.models import Document, Revision
+
+
+@receiver(
+    pre_delete,
```
akatsoulas, line 12: I haven't tested this but I believe `pre_delete` signals may not work great with the m2m fields that we are using as relationships my be cleared already. If this is the case, we may have to switch to `post_delete`. In most cases in the project we are using `post_delete`. Is there a specific constraint that forced you to use `pre_delete` or just preference? Personally I am fine with either as long as m2m relationships work

kitsune/community/utils.py
```diff
@@ -18,17 +21,23 @@
 def top_contributors_questions(start=None, end=None, locale=None, product=None, count=10, page=1):
     """Get the top Support Forum contributors."""
 
+    if not start:
+        start = datetime.now() - timedelta(days=90)
```
akatsoulas: Since you are updating this here, let's use a constant like for 90 :)

akatsoulas: similar for the limit

kitsune/community/utils.py
```diff
@@ -59,10 +68,22 @@ def top_contributors_questions(start=None, end=None, locale=None, product=None,
...
+
+    if not (contribution_buckets or deletion_metrics_by_contributor):
         return [], 0
 
-    user_ids = [bucket.key for bucket in contribution_buckets]
+    user_ids = list(
+        {bucket.key for bucket in contribution_buckets}
+        | {str(contributor_id) for contributor_id in deletion_metrics_by_contributor.keys()}
+    )
     contributor_group_ids = list(
```
akatsoulas, line 90: This is used later in the `user_hits` query. I wonder if we need to do something like min(len(user_ids), 100) or any number for that matter to make sure that we are not fetching far too many profiles

kitsune/community/utils.py
```diff
@@ -18,17 +21,23 @@
 def top_contributors_questions(start=None, end=None, locale=None, product=None, count=10, page=1):
     """Get the top Support Forum contributors."""
 
+    if not start:
+        start = datetime.now() - timedelta(days=90)
+
+    limit = count * 10
```
akatsoulas: Shall we cap the count to number that makes sense. The code is smart and efficient but it can become a problem if the function is called with a count of 10K for example

akatsoulas (approved):
Great work! r+wc

kitsune/community/signals.py
escattone: Tasos and I discussed this via video. Key points:
- The m2m relationships do not work in `pre_save`. They work fine in `pre_delete`.
- I used `pre_delete` because I needed to make some queries and when I used `post_delete` I lost easy access to information that I needed.

kitsune/community/utils.py
escattone: Great idea.

kitsune/community/apps.py
escattone: I thought so too, but within `vscode`, `ruff` keeps suggesting to remove this line since it's unused, so the `#  noqa` silences `ruff`.

kitsune/community/models.py
escattone: Good idea.

PR #6799, opened 2025-07-29
Point to the form submission if no search results

akatsoulas:


akatsoulas:
Resolves https://github.com/mozilla/sumo/issues/1227

PR #6801, opened 2025-07-29
Strategy pattern scaffolding for l10n

akatsoulas:


kitsune/llm/l10n/service.py
```diff
@@ -0,0 +1,81 @@
...
+
+from kitsune.llm.l10n.strategies import (
+    TranslationMethod,
+    TranslationRequest,
+    TranslationStrategyFactory,
+)
+from kitsune.wiki.models import Revision
+
+
+@dataclass
```
escattone: I don't think `@dataclass` makes sense here does it?

kitsune/wiki/views.py
```diff
@@ -814,9 +814,13 @@ def review_revision(request, document_slug, revision_id):
...
                 doc.save()
 
-            # Send notifications of approvedness and readiness:
-            if rev.is_ready_for_localization or rev.is_approved:
+            # Send notifications of approvedness
+            if rev.is_approved:
                 ApprovedOrReadyUnion(rev).fire(exclude=[rev.creator, request.user])
+            if rev.is_ready_for_localization:
+                # Trigger automatic translation with strategy pattern
+                from kitsune.llm.tasks import trigger_automatic_translation_with_strategy
```
escattone: Did you mean to define this task later, or did you mean it to be in this PR?

escattone (approved):
r+wc Let's try this. I like it's high-level interface.

kitsune/wiki/views.py
```diff
@@ -814,9 +814,13 @@ def review_revision(request, document_slug, revision_id):
 
                 doc.save()
 
-            # Send notifications of approvedness and readiness:
-            if rev.is_ready_for_localization or rev.is_approved:
```
akatsoulas, line 818: I made this change here b/c for a revision to be ready for localization it needs first to be approved. So just checking for is_approved is sufficient here

PR #6802, opened 2025-07-29

smithellis:
Reviving - previously reverted


smithellis:
Git snaffu - easier to close this and push a new one.

PR #6803, opened 2025-07-29

smithellis:


kitsune/users/management/commands/delete_inactive_users.py
```diff
@@ -0,0 +1,234 @@
...
+                )
+            )
+
+    def _bulk_delete_batch(self, User, user_ids):
+        _, deleted_objects = User.objects.filter(id__in=user_ids).delete()
+
+        try:
+            user_count = deleted_objects.get("auth.User", 0)
+            return user_count, 0
+        except (DatabaseError, IntegrityError) as e:
```
akatsoulas: Why are we using these 2 exceptions here?

akatsoulas (commented):
How is this management command different from `cleanup_expired_users`? Is this going to be a one off deletion or are we going to reuse it?

smithellis:
This was initially dealing with [Migration to bulk delete inactive users regardless of content](https://github.com/mozilla/sumo/issues/2345) which we decided to then approach as a [management command ](https://github.com/mozilla/kitsune/pull/6747) which then got reverted due to a polluted commit. 
So this is the one time bulk clean-up of *any* expired user regardless of content, and it handles all cases - users with/without content.

`cleanup_expired_users` is what should handle ongoing maintenance, as it only uses the deletion pipeline. This code was to handle a first, large purge where some users could just be bulk deleted.

kitsune/users/management/commands/delete_inactive_users.py
smithellis: These should be covering the actual `.delete()` vs. the retrieval of the user count. I'll move the deletion activity into the try. Good catch - thanks!

kitsune/users/management/commands/delete_inactive_users.py
```diff
@@ -0,0 +1,233 @@
...
+                self.style.WARNING(
+                    f"Warning: {error_count} users could not be deleted. "
+                    "Check error messages above."
+                )
+            )
+
+    def _bulk_delete_batch(self, User, user_ids):
+        try:
+            _, deleted_objects = User.objects.filter(id__in=user_ids).delete()
+            user_count = deleted_objects.get("auth.User", 0)
```
akatsoulas: this line and the next should be better in the `else` clause of a try/except/else

kitsune/users/management/commands/delete_inactive_users.py
akatsoulas: My point was mostly about why these two specific exceptions.  Why not for example `OperationalError` (which we have in all environments). Shall we just use `Exception` here?

kitsune/users/management/commands/delete_inactive_users.py
smithellis: You are correct! Thanks!

kitsune/users/management/commands/delete_inactive_users.py
smithellis: They seemed more specific to what could go wrong, but maybe being more general makes sense. I'll alter to use `Exception`. 
Thanks!

PR #6806, opened 2025-08-01
Adding trigger_delete API for test accounts

emilghittasv:
Should fix https://github.com/mozilla/sumo/issues/2420

### DIFF
--- a/kitsune/users/api.py
+++ b/kitsune/users/api.py
@@ -375,3 +375,30 @@ def create_test_user(request):
             ],
         }
     )
+
+@group_required("Staff")
+@require_POST
+@csrf_exempt
+def trigger_delete(request):
+    """
+    Deletes a given test user.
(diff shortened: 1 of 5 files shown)


kitsune/users/api.py
```diff
@@ -375,3 +375,30 @@ def create_test_user(request):
             ],
         }
     )
+
+@group_required("Staff")
+@require_POST
+@csrf_exempt
+def trigger_delete(request):
```
akatsoulas, line 382: let's move this back under the views.py. This is a regular django view

akatsoulas (approved):
r+wc

kitsune/users/api.py
akatsoulas: Feel free to merge this PR as is if you prefer the `api.py` convention

PR #6810, opened 2025-08-04
Use a dedicated node container.

akatsoulas:
* Use sharp instead of optipng


Dockerfile
```diff
@@ -15,69 +32,59 @@ ENV LANG=C.UTF-8 \
 RUN useradd -d /app -M --uid 1000 --shell /bin/bash kitsune
 
 RUN set -xe \
-    && apt-get update && apt-get install apt-transport-https \
```
escattone, line 18: Nice housekeeping!

Dockerfile
```diff
@@ -1,7 +1,24 @@
...
+FROM node:22-bookworm AS frontend-builder
+
+WORKDIR /app
+
+COPY package*.json ./
+RUN npm install
+
+COPY . .
+RUN cp .env-build .env && \
+    mkdir -p jsi18n/jsi18n && \
```
escattone, line 13: What do you think about adding `npm install -g npm@latest`?

Makefile
```diff
@@ -47,7 +43,7 @@ build:
 rebuild: clean build
 
 start: .docker-build
-	${DC} up web
+	${DC} up web node
 run: start
 
 init: .docker-build
```
escattone, line 48: The `run-bootstrap.sh` script run within this `init` target doesn't work any longer, because its `npm` portion needs to run on `node` rather than `web`. I suppose we could split that into two separate scripts, one for Python and one for `node`:

```make
init: .docker-build
    ${DC} run web bin/run-web-bootstrap.sh $(INIT_ARGS_WEB)
    ${DC} run node bin/run-node-bootstrap.sh $(INIT_ARGS_NODE)
```

where we do the following in the Python script:
- `scripts/l10n-fetch-lint-compile.sh`
- `python manage.py compilejsi18n`
- `python manage.py migrate`

and the `npm run development && npm run webpack:build && npm run webpack:build:pre-render` in the `node` script.

Makefile
escattone, line 1: Maybe also change the `format` target to use `ruff` instead of `black`?
```make
format: .docker-build
	${DC} run web ruff format kitsune
```

Makefile
escattone: The `run-bootstrap.sh` script is also used within `devcontainer.json`, so we'll need to adjust that as well.

.devcontainer/devcontainer.json
```diff
@@ -46,7 +46,7 @@
         // Uncomment the next line if you want to keep your containers running after VS Code shuts down.
         "shutdownAction": "none",
         // Uncomment the next line to run commands after the container is created - for example installing git.
-        "postCreateCommand": "./bin/run-bootstrap.sh",
+        "postCreateCommand": "./bin/run-devcontainer-bootstrap.sh",
```
escattone: Could we eliminate the need for `bin/run-devcontainer-bootstrap.sh` by using the following instead?

```suggestion
        "postCreateCommand": ["./bin/run-web-bootstrap.sh", "./bin/run-node-bootstrap.sh"],
```

Makefile
```diff
@@ -47,11 +43,12 @@ build:
...
-	${DC} run web bin/run-bootstrap.sh $(INIT_ARGS)
+	${DC} run web bin/run-web-bootstrap.sh $(INIT_ARGS)
```
escattone: Should we just get rid of `INIT_ARGS`?

escattone (approved):
r+wc

PR #6818, opened 2025-08-04
manage pending machine translations

escattone:
mozilla/sumo#2454

## Notes
This PR assumes the following for now:
- Simple settings rather than admin-controlled settings.
- No recording yet of important statistics, like machine translations automatically rejected or approved.
- We'll continue to use APScheduler rather than Celery beat for periodic tasks.


escattone:
Closing this so I can instead implement it within the new content manager and translation strategy framework.

PR #6823, opened 2025-08-05
Move draft management into strategy

akatsoulas:


kitsune/wiki/drafts.py
escattone, line 1: I'm thinking we should rename this to `content_managers.py` or something like that.

kitsune/wiki/drafts.py
```diff
@@ -0,0 +1,186 @@
...
+        Returns:
+            bool: True if draft was successfully deleted
+        """
+        try:
+            draft = DraftRevision.objects.get(id=draft_id, creator=user)
+        except DraftRevision.DoesNotExist:
+            return False
+        else:
+            draft.delete()
+            return True
```
escattone: Nit. Maybe do it all in one shot?

```python
        try:
            DraftRevision.objects.get(id=draft_id, creator=user).delete()
        except DraftRevision.DoesNotExist:
            return False
        else:
            return True
```

kitsune/wiki/drafts.py
```diff
@@ -0,0 +1,186 @@
...
+            user: The user publishing the draft
+            form_data: Form data for revision creation
+            based_on_id: Optional ID of the revision this is based on
+            base_rev: Optional base revision object
+        Returns:
+            Revision: The created revision
+        """
+        try:
+            draft = DraftRevision.objects.get(id=draft_id, creator=user)
+            document = draft.document
```
escattone: To get rid of an extra query, maybe something like the following instead?

```python
draft = DraftRevision.objects.select_related("document").get(id=draft_id, creator=user)
```

and then use `draft.document` below?

kitsune/wiki/drafts.py
```diff
@@ -0,0 +1,186 @@
...
+                    EditDocumentEvent(revision).fire(exclude=exclude_users)
+                case NotificationType.TRANSLATION_WORKFLOW:
+                    ApprovedOrReadyUnion(revision).fire(exclude=exclude_users)
+                case NotificationType.READY_FOR_L10N:
+                    ReadyRevisionEvent(revision).fire(exclude=exclude_users)
+
+    def mark_ready_for_localization(self, revision: Revision, user=None, send_notifications=True) -> None:
+        """Mark a revision as ready for localization and optionally send notifications."""
+        revision.is_ready_for_localization = True
+        revision.readied_for_localization = revision.reviewed
```
escattone: This doesn't work for the `mark_ready_for_l10n` case, because that happens some period of time after the revision has been reviewed. In that case, you'd need to use `datetime.now()` instead.

kitsune/wiki/forms.py
```diff
@@ -340,22 +342,16 @@ def save(self, creator, document, based_on_id=None, base_rev=None, **kwargs):
...
-
-        # If the document doesn't allow the revision creator to edit the
-        # keywords, keep the old value.
-        if base_rev and not document.allows(creator, "edit_keywords"):
-            new_rev.keywords = base_rev.keywords
-
-        new_rev.save()
-        return new_rev
+        form_data = self.cleaned_data.copy()
+        strategy = ManualTranslationStrategy()
```
escattone: Why  `ManualTranslationStrategy` instead of `ManualContentManager`? It doesn't make sense to me to use `ManualTranslationStrategy`.

kitsune/wiki/strategies.py
```diff
@@ -93,9 +97,15 @@ def translate(self, l10n_request: TranslationRequest) -> TranslationResult:
         return TranslationResult(success=True, method=TranslationMethod.MANUAL, metadata={})
 
 
+@dataclass
 class AITranslationStrategy(TranslationStrategy):
     """AI-based translation strategy without human review."""
 
+    draft_manager: "AIContentManager" = field(init=False)
```
escattone: Why `draft_manager`? It handles much more than drafts, so it seems this should be `content_manager` instead.

kitsune/wiki/strategies.py
```diff
@@ -110,9 +120,15 @@ def translate(self, l10n_request: TranslationRequest) -> TranslationResult:
         )
 
 
+@dataclass
 class HybridTranslationStrategy(TranslationStrategy):
     """Hybrid translation strategy (AI + Human review)."""
 
+    draft_manager: "HybridContentManager" = field(init=False)
```
escattone: Same here.

kitsune/wiki/strategies.py
```diff
@@ -127,9 +143,15 @@ def translate(self, l10n_request: TranslationRequest) -> TranslationResult:
         )
 
 
+@dataclass
 class ManualTranslationStrategy(TranslationStrategy):
     """Manual translation workflow coordinator."""
 
+    draft_manager: "ManualContentManager" = field(init=False)
```
escattone: And here.

kitsune/wiki/strategies.py
```diff
@@ -158,18 +180,14 @@ def translate(self, l10n_request: TranslationRequest) -> TranslationResult:
...
-        revision.is_ready_for_localization = True
-        revision.readied_for_localization = datetime.now()
-        if l10n_request.user:
-            revision.readied_for_localization_by = l10n_request.user
-        revision.save()
 
     def _handle_mark_ready_for_l10n(self, l10n_request: TranslationRequest) -> TranslationResult:
         """Handle mark_ready_for_l10n trigger."""
-        self._update_revision_for_localization(l10n_request)
+        self.draft_manager.mark_ready_for_localization(
```
escattone: As I noted above, in this case, the `readied_for_localization` field of the revision should be set to `datetime.now()` instead of `revision.reviewed`.

escattone (commented):
Within the `ManualTranslationStrategy.translate` method, the following chunk of code seems off to me:

```python
match l10n_request.trigger:
    case "mark_ready_for_l10n":
        if revision.is_ready_for_localization:
            message = "Revision is already ready for localization"
        return self._handle_mark_ready_for_l10n(l10n_request)
    case "review_revision":
        if not revision.is_ready_for_localization:
            message = "Revision is not ready for localization"
        return self._handle_review_revision(l10n_request)
```

It seems that it instead be something like:

```python
match l10n_request.trigger:
    case "mark_ready_for_l10n":
        if not revision.is_ready_for_localization:
            return self._handle_mark_ready_for_l10n(l10n_request)
        message = "Revision is already ready for localization"
    case "review_revision":
        return self._handle_review_revision(l10n_request)
```

because in the `mark_ready_for_l10n` case, if we're marking is again, we don't want to notify users again if it's already been done, and in the `review_revision` case, by the time we've reached that case, we've already ensured that `revision.is_ready_for_localization` is True.

kitsune/wiki/drafts.py
escattone: We'll create a ticket for this one, and do it after this PR merges.

PR #6827, opened 2025-08-06
Fix rendering issues

akatsoulas:


escattone (commented):
For consistency, do we also want to remove the trailing `:` from the `title` and `slug` field labels?

PR #6828, opened 2025-08-06
Replace simple search with semantic

smithellis:
* Set trial license in `docker-compose.yml` to avoid CI issue
* Increase `mapping.nested_fields.limit` and `mapping.total_fields.limit` to accomodate `semantic_text` needs
* Create `SumoLocaleAwareSemanticTextField` and `SemanticTextField` types
* Add semantic fields to document types
* Add `prepare_` methods for semantic fields in document types
* `es_utils.py` update to circumvent a CI/build issue - this needs a better solution but does work
* Add new search classes to `search.py`: `SemanticWikiSearch` and `SemanticQuestionSearch`
* Modify `views.py` to make semantic search the priority, and fall back to traditional search if there is an issue
* Add `USE_SEMANTIC_SEARCH` to `settings.py` in case we want to disable in some use cases

Prepare for stage

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -40,6 +40,7 @@ services:
       - discovery.type=single-node
       - LOG4J_FORMAT_MSG_NO_LOOKUPS=true
       - xpack.security.enabled=false
+      - xpack.license.self_generated.type=trial
       - action.destructive_requires_name=false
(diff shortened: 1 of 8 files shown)


smithellis:
Closing for https://github.com/mozilla/kitsune/pull/6857

PR #6829, opened 2025-08-06
fixes for the manual l10n strategy

escattone:
This should resolve the test failures in `main`.


akatsoulas:
This won't be applicable after #6823. Thanks @escattone for taking the time!

PR #6831, opened 2025-08-07
create translations when revision is ready for localization

escattone:
mozilla/sumo#2453

This is a really rough. Still lots to think through and do.

While I was doing this it struck me that maybe `HybridTranslationStrategy` should inherit from `AITranslationStrategy` which should inherit from `ManualTranslationStrategy`?


escattone:
Closing this in favor of a better organization of this work by @akatsoulas in https://github.com/mozilla/kitsune/pull/6832.

PR #6835, opened 2025-08-11
Display announcements per OS

akatsoulas:


kitsune/announcements/models.py
```diff
@@ -60,22 +61,23 @@ def content_parsed(self):
         return wiki_to_html(self.content.strip())
 
     @classmethod
-    def get_site_wide(cls):
-        return cls._visible_query(groups=None, locale=None)
+    def get_site_wide(cls, platform_slugs: Iterable[str] | None = None):
+        """Returns announcements that are visible to everyone (no group/locale restrictions)."""
+        return cls._visible_query(platforms=platform_slugs).filter(groups__isnull=True, locale__isnull=True)
```
escattone, line 66: In the Windows case, it's possible for `platform_slugs` to be `["win10", "win11"]` and for an announcement to be for both Windows 10 and 11. For that case, you'd get a duplicate for that announcement, so I think we need to add `distinct()` to this. I would also add this case to the tests.

escattone: I added a fix as well as a test for the fix.

escattone (approved):
I fixed two tests, as well as added a fix + tests for a case when duplicate announcements could occur.

PR #6854, opened 2025-08-13
handle hybrid auto approvals/rejections

escattone:
mozilla/sumo#2454


kitsune/wiki/content_managers.py
```diff
@@ -156,25 +166,50 @@ def create_revision(
...
+        self, revision: Revision, user, comment=None, send_notifications=True
+    ) -> Revision:
+        """Publish (approve) a revision."""
+        return self.review_revision(revision, user, True, comment, send_notifications)
+
+    def reject_revision(self, revision: Revision, user, comment=None) -> Revision:
+        """Reject (disapprove) a revision."""
+        return self.review_revision(revision, user, False, comment, send_notifications=False)
+
+    def review_revision(
```
akatsoulas: should this be an "internal" method?

kitsune/wiki/content_managers.py
```diff
@@ -261,18 +299,77 @@ def create_revision(
...
-            send_notifications: Whether to send content creation notifications
-        """
+    def publish_revision(
+        self, revision: Revision, user=None, comment=None, send_notifications=True
+    ) -> Revision:
+        """Publish (approve) a revision using the sumo bot."""
         user = user or Profile.get_sumo_bot()
-        return super().publish_revision(revision, user, send_notifications)
+        rev = super().publish_revision(revision, user, comment, send_notifications)
+        self.fire_notifications(rev, [NotificationType.REVIEW_WORKFLOW])
```
akatsoulas: Can't we incorporate this into the publish method? If it's a problem that we just pass a boolean flag, maybe we need to change the signature and pass a list of notifications instead?

kitsune/wiki/strategies.py
```diff
@@ -263,6 +266,7 @@ def translate(
         rev = self.content_manager.create_revision(data, doc, send_notifications=True)
         if publish:
             rev = self.content_manager.publish_revision(rev)
+            self.content_manager.fire_notifications(rev, [NotificationType.REVIEW_WORKFLOW])
```
akatsoulas: Similar comment to include the notification in the publish method

kitsune/wiki/content_managers.py
```diff
@@ -156,25 +166,50 @@ def create_revision(
...
+            comment: An optional comment to assign to the revision
             send_notifications: Whether to send content creation notifications
         """
-        revision.is_approved = True
+        revision.is_approved = approve
         revision.reviewed = datetime.now()
-        revision.reviewer = user
+        revision.reviewer = reviewer
+        if comment:
+            revision.comment = comment
```
akatsoulas: any chance to overwrite existing comment here?

kitsune/wiki/content_managers.py
```diff
@@ -261,18 +299,77 @@ def create_revision(
...
+        render_document_cascade.delay(rev.document.id)
+        return rev
 
 
 class HybridContentManager(AIContentManager):
     """Content manager for hybrid translation workflow."""
 
-    pass
+    def __init__(self):
+        unreviewed_translations = Revision.objects.filter(
```
akatsoulas: Up until now, the manager was responsible for the operations on a single revision and strategy was doing the orchestration. This change breaks this architecture and couples document discovery for translation approval with document management.  The manager could provide methods like `is_stale`, `can_be_approved` (names are just examples) which would be responsible for introspecting the single revision.
My suggestion would be to keep the separation of concerns and move the discovery of documents to the strategy and let the management of the document to the manager

kitsune/wiki/strategies.py
```diff
@@ -296,6 +300,49 @@ def translate(self, l10n_request: TranslationRequest) -> TranslationResult:
...
+                    f" {settings.HYBRID_REVIEW_GRACE_PERIOD} day(s)."
+                ),
+            )
+            results.append(rev)
+        return results
+
+    def handle_pending_reviews(self, asynchronous=True):
+        if asynchronous:
+            handle_pending_reviews_task.delay()
+        self.reject_stale_translations()
```
akatsoulas: Do we need to do this in bulk through Celery?
I was thinking that rejection could be part of the revision creation. When a new revision is saved we could check there if we need to invalidate anything and just call the manager methods and helpers. This should greatly simplify the implementation and offloads the revision management to the manager.
Eg
- New parent revision -> invalidate pending translations 
- New revision on the translation -> invalidate pending translation

kitsune/wiki/strategies.py
```diff
@@ -296,6 +300,49 @@ def translate(self, l10n_request: TranslationRequest) -> TranslationResult:
...
+
+        results = []
+        for rev in self.content_manager.get_stale_translations(
+            locales=settings.HYBRID_ENABLED_LOCALES
+        ):
+            self.content_manager.reject_revision(rev, comment="No longer relevant.")
+            results.append(rev)
+        return results
+
+    def publish_unreviewed_within_grace_period(self) -> list[Revision]:
```
akatsoulas: how about `approve_translations` | `auto_approve_translations` | `approve_pending_translations`? It's closer to the naming schema that we have so far and the details are described in the docstring

kitsune/wiki/strategies.py
```diff
@@ -296,6 +300,49 @@ def translate(self, l10n_request: TranslationRequest) -> TranslationResult:
...
+        """
+        Publish fresh machine translations that have not been reviewed within the grace period.
+        """
+        if not (settings.HYBRID_REVIEW_GRACE_PERIOD and settings.HYBRID_ENABLED_LOCALES):
+            return []
+
+        results = []
+        for rev in self.content_manager.get_fresh_translations(
```
akatsoulas: The query should be part of this method - similar comment was made for the manager to handle single revisions

kitsune/wiki/tasks.py
```diff
@@ -39,15 +40,16 @@
...
     try:
-        revision = Revision.objects.get(id=revision_id)
-        document = Document.objects.get(id=document_id)
-    except (Revision.DoesNotExist, Document.DoesNotExist) as err:
+        revision = Revision.objects.select_related("document").get(id=revision_id)
+    except Revision.DoesNotExist as err:
         capture_exception(err)
         return
+    else:
+        document = revision.document
```
akatsoulas, line 53: Is there a case where this is not true and we need to pass a the document id? (revision.document.id != document_id from the function signature)

akatsoulas (changes requested):

kitsune/wiki/content_managers.py
escattone: I was thinking the same thing.

kitsune/wiki/content_managers.py
escattone: I'll add a check that avoids this if there is already a comment.

kitsune/wiki/strategies.py
escattone: That's a good idea.

kitsune/wiki/tasks.py
escattone: No, I don't think so. The document is always the revision's document, but I'll double check.

kitsune/wiki/content_managers.py
akatsoulas: I am not sure whether we need to avoid or append

kitsune/wiki/strategies.py
```diff
@@ -290,18 +276,92 @@ class HybridTranslationStrategy(TranslationStrategy):
     def __post_init__(self):
         self.content_manager = HybridContentManager()
 
+        sumo_bot = Profile.get_sumo_bot()
```
akatsoulas: I am not sure whether this should be part of the strategy. The strategy up until now and similarly to the content manager operates per document. This mixes bulk operations and document operations.  On top of that, doing the queries in the init means that every time we initialize the strategy we will perform the queries. Even if we decide to keep this pattern, I would suggest to move all the queries in methods that we will call explicitly. 

PR #6856 introduces services. Maybe we can adjust that service to have methods to update stale translations and do the approvals. Or even introduce a new service for the approvals - totally up to you. Services basically do the bulk operations by calling the right method from strategies and/or content managers.

kitsune/wiki/strategies.py
```diff
@@ -290,18 +276,92 @@ class HybridTranslationStrategy(TranslationStrategy):
...
+            created__lt=Now() - timedelta(days=settings.HYBRID_REVIEW_GRACE_PERIOD),
+        ).exclude(another_already_approved | translations_discontinued)
+
     def translate(self, l10n_request: TranslationRequest) -> TranslationResult:
         """Perform hybrid translation."""
         result = AITranslationStrategy().translate(l10n_request, publish=False)
         result.method = TranslationMethod.HYBRID
         return result
 
+    def reject_obsolete_translations(self, document: Document) -> None:
```
akatsoulas: In my first review, I suggested that we don't need to reject documents in bulk and this could be part of the revision creation.

When a new revision is created we simply invalidate the previous one under certain conditions:
* A new revision with adequate changes to warrant a translation on the parent
* A new revision on the localized version

This could be handled by the HybridContentManager when creating a new revision?

kitsune/wiki/strategies.py
```diff
@@ -290,18 +276,92 @@ class HybridTranslationStrategy(TranslationStrategy):
...
+                    f" reviewed within {grace_period}."
+                )
+
 
 class TranslationStrategyFactory:
     """Factory for creating and selecting translation strategies."""
 
-    def __init__(self):
-        self._strategies = {
+    @cached_property
```
akatsoulas: The performance benefits here will be negligible. Using `@cached_property` on the internal structure can be confusing with regards to the interface of the class. This should be accessible outside of the class. I would revert to `__init__` here

kitsune/wiki/strategies.py
escattone: This just defines the queries once, it doesn't perform them, but I can move this into a service.

kitsune/wiki/strategies.py
escattone: I didn't do this for performance. I did it to decouple the running of all of the strategy constructors whenever the `TranslationStrategyFactory` is constructed -- especially since it's [constructed at import time in `wiki/views.py`](https://github.com/mozilla/kitsune/blob/e180bd0a6b86d205bb75c93e82ad331c38a60a74/kitsune/wiki/views.py#L94). I was calling `Profile.get_sumo_bot()` in the `__post_init__` of my old version of `HybridTranslationStrategy`, and that failed because the `UserConfig.ready` method had not yet been called. I'm going to move that code out of `HybridTranslationStrategy` into a service, so I'll revert this. However, we should keep in mind that whatever is done within the `__post_init__` methods of the translation strategies will run at import time.

kitsune/wiki/strategies.py
escattone: I'm not rejecting documents in bulk, but one at a time via the `post_save` signal whenever a document is updated. However, that one document update can precipitate multiple obsolete machine translations. The only events that can precipitate obsolete translations are when a document's `current_revision` and/or `latest_localizable_revision` is updated, and that can happen only when one of its revisions is approved or its current revision is deleted (so within the `Revision.save` and `Revision.delete` methods with their logic for when to update the document's current and latest-localizable revision).

It's not clear how that could be done within a content manager unless we made extensive changes.

PR #6860, opened 2025-08-18
display warning for auto-published translations

escattone:
mozilla/sumo#2455


kitsune/wiki/content_managers.py
```diff
@@ -349,6 +350,22 @@ def reject_revision(
         user = user or Profile.get_sumo_bot()
         return super().reject_revision(revision, user, comment, send_notifications)
 
+    def is_auto_published_translation(self, revision: Revision | None) -> bool:
```
akatsoulas, line 357: This could be part of the HybridContentManager

PR #6878, opened 2025-08-27
2488 - Logged out users should see announcments

smithellis:
- Allow logged out users to see announcements that that are locale restricted

### DIFF
--- a/kitsune/announcements/models.py
+++ b/kitsune/announcements/models.py
@@ -61,9 +61,18 @@ class Announcement(ModelBase):
         return wiki_to_html(self.content.strip())
 
     @classmethod
-    def get_site_wide(cls, platform_slugs: Iterable[str] | None = None):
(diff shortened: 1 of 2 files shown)


escattone (changes requested):
This fixes the case for when the user is logged out, but doesn't fix the case for when the user is logged-in. When the user is logged-in, the locale of the announcement is not respected. For example, when I create an announcement for the Spanish locale, it's always shown to logged-in users, even for non-Spanish locales.

emilghittasv:
Just a note that the case in which the locale rule is not respected is tracked in https://github.com/mozilla/sumo/issues/1197

escattone:
@emilghittasv (cc @smithellis) Thanks for that Emil. Since this satisfies mozilla/sumo#2488, I'll approve this PR with the assumption that mozilla/sumo#1197 is also fixed before we release.

escattone (approved):

PR #6904, opened 2025-09-09
set default AI-based L10N locales

escattone:


escattone:
Closing in favor of setting these values via a webservices-infra PR.

PR #6915, opened 2025-09-18
better translation response handling

escattone:
mozilla/sumo#2528
mozilla/sumo#2530

## Notes
This PR improves the translation response handling in several ways:
- Uses the `text()` method -- which assembles the text response from the `AIMessage` instance and handles both the case when the message `content` is a string, and when it's a list.
- Simplifies the LLM's response formatting.
- Strictly parses the LLM response, and if the parsing fails, raises an `OutputParserException` exception which will trigger a retry if called within `wiki.tasks.translate`.


kitsune/llm/l10n/prompt.py
```diff
@@ -1,10 +1,18 @@
...
 
 from langchain.prompts import ChatPromptTemplate
 from langchain.schema import AIMessage
+from langchain_core.exceptions import OutputParserException
 
 from kitsune.llm.l10n.config import L10N_PROTECTED_TERMS
 
+DELIMITER = "<<<<translation-and-explanation-delimiter>>>>"
+TRANSLATION_PARSER_REGEX = re.compile(
+    rf"^(```wiki\s+)?\s*(?P<translation>.+?)(?(1)\s*```)\s*{DELIMITER}\s*(?P<explanation>.+)$",
```
akatsoulas, line 12: Wouldn't this regex fail if there are nested delimiters?

escattone: No, it won't fail. If there are nested delimiters, the first one will satisfy the regex, and the second one will be included in the explanation. For example,
```python
>>> text = f"This is the translation. {DELIMITER} This is the explanation {DELIMITER}."
>>> mo = TRANSLATION_PARSER_REGEX.match(text)
>>> mo.group("translation")
'This is the translation.'
>>> mo.group("explanation")
'This is the explanation <<<<translation-and-explanation-delimiter>>>>.'
```

PR #6916, opened 2025-09-18
do not translate redirects

escattone:
mozilla/sumo#2534


escattone:
Closing this. @emilghittasv pointed out that we probably should create translations of redirects so the translations will also redirect.

PR #6928, opened 2025-09-25
Do not send notifications for system accounts

akatsoulas:


escattone (commented):
It seems to me that we'd still want to send notifications if the revision has been created by the `sumo_bot`, so that the other contributors are notified. I don't see why we wouldn't want to do that.

Is https://github.com/mozilla/sumo/issues/2542 saying that we should never send the `Your revision has been...` email to the reviewer in all cases or just when the revision's creator is `sum_bot`?

kitsune/wiki/content_managers.py
```diff
@@ -252,8 +252,10 @@ def fire_notifications(
                         exclude_users.append(revision.readied_for_localization_by)
                     ReadyRevisionEvent(revision).fire(exclude=exclude_users)
                 case NotificationType.REVIEW_WORKFLOW:
-                    send_reviewed_notification.delay(revision.id, comment)
                     send_contributor_notification.delay(revision.id, comment)
+                    profile = getattr(revision.creator, "profile", None)
+                    if not (profile and profile.is_system_account):
+                        send_reviewed_notification.delay(revision.id, comment)
```
escattone: Within `send_reviewed_notification()` we do already skip emails to system accounts (and we skip the whole thing when the reviewer and the creator are the same), so it seems that with this change we're skipping an email to the reviewer when the reviewer is not a system account and the revision was created by a system account?

I still don't see why we'd need to do that. It seems to me that we should either never send emails to the reviewer (because they're already aware that they performed the action) or continue to send emails to the reviewer no matter who created the revision (but change the language of the email so it doesn't say `Your revision...`).

akatsoulas:
@escattone PR updated

kitsune/wiki/content_managers.py
```diff
@@ -226,7 +226,7 @@ def _review_revision(
         revision.save()
 
         if notification_types:
-            self.fire_notifications(revision, notification_types)
+            self.fire_notifications(revision, notification_types, comment)
```
escattone, line 229: Good catch.

kitsune/wiki/views.py
```diff
@@ -825,7 +825,7 @@ def review_revision(request, document_slug, revision_id):
                     _execute_l10n_strategy(rev, TranslationTrigger.REVIEW_REVISION, request.user)
                 case (True, False):
                     ManualContentManager().fire_notifications(
-                        rev, [NotificationType.TRANSLATION_WORKFLOW], [rev.creator, request.user]
+                        rev, [NotificationType.TRANSLATION_WORKFLOW]
```
escattone, line 828: Another good catch.

PR #6946, opened 2025-10-07
Create missing translations

akatsoulas:


kitsune/wiki/services.py
```diff
@@ -80,42 +93,110 @@ def find_stale_translation_candidates(
...
+                document__locale=locale,
+                creator=self.sumo_bot,
+                is_approved=False,
+                reviewed__isnull=True,
+            )
+
+            docs = (
+                english_docs_without_translation.annotate(has_pending=Exists(pending_revisions))
+                .filter(has_pending=False)
+            )
```
escattone: I don't think this code is needed.  When there is a pending revision in a locale, there will be a translation document in that locale, so the `.exclude(translations__locale=locale)` above already takes care of this.

escattone (changes requested):
This looks good, but I think there's a chunk of code that could be removed.

kitsune/wiki/services.py
akatsoulas: good point!

akatsoulas:
Closing if favor of #6949

PR #6947, opened 2025-10-07
expand pinned articles

escattone:
mozilla/sumo#2562


kitsune/wiki/utils.py
```diff
@@ -112,14 +112,44 @@ def _active_contributors_id(from_date, to_date, locale, product):
...
     """Returns up to 4 random articles from the most visited.
 
     Args:
         product: Optional product to filter by
         topics: Optional iterable of topics to filter by
         locale: Locale to get articles for, defaults to WIKI_DEFAULT_LANGUAGE
+        include_pinned_articles: Whether or not to include the pinned articles (defaults to True)
     """
+    if include_pinned_articles:
+        pinned_articles = list(get_pinned_articles(product=product, locale=locale))
```
akatsoulas: Shouldn't this pass the `user` as well?

kitsune/wiki/utils.py
```diff
@@ -229,7 +260,9 @@ def build_topics_data(request: HttpRequest, product: Product, topics: list[Topic
     """
     topics_data: list[dict] = []
 
-    featured_articles = get_featured_articles(product, locale=request.LANGUAGE_CODE, topics=topics)
+    featured_articles = get_featured_articles(
+        product, locale=request.LANGUAGE_CODE, topics=topics, include_pinned_articles=False
```
akatsoulas: Similar comment about the `user` object

kitsune/wiki/utils.py
```diff
@@ -112,14 +112,44 @@ def _active_contributors_id(from_date, to_date, locale, product):
...
+def get_featured_articles(
+    product=None, topics=None, locale=settings.WIKI_DEFAULT_LANGUAGE, include_pinned_articles=True
+):
     """Returns up to 4 random articles from the most visited.
 
     Args:
         product: Optional product to filter by
         topics: Optional iterable of topics to filter by
         locale: Locale to get articles for, defaults to WIKI_DEFAULT_LANGUAGE
+        include_pinned_articles: Whether or not to include the pinned articles (defaults to True)
```
akatsoulas: let's also add what this functions returns

kitsune/wiki/models.py
```diff
@@ -1147,6 +1147,44 @@ def __str__(self):
...
+        constraints = [
+            # Ensure each product can only have one config.
+            models.UniqueConstraint(
+                fields=["product"],
+                condition=Q(product__isnull=False),
+                name="unique_pinned_article_config_per_product",
+            ),
+            # Ensure there is only one config where the product field is NULL.
+            models.UniqueConstraint(
+                Value(1, output_field=models.IntegerField()),
```
akatsoulas, line 1170: nice!

kitsune/wiki/utils.py
escattone: We don't currently take into account the user for `get_featured_articles()`, so I don't pass a user to this `get_pinned_articles()` call. If you don't pass a `user` to `get_pinned_articles()`, it effectively limits the pinned articles to those that have a current revision and are not restricted.

However, I wrote `get_pinned_articles()` with the future in mind -- so it does properly handle user visibility.

Given that we'll probably be using restricted articles more (perhaps for Firefox Enterprise?), I can adjust `get_featured_articles()` to take into account the visibility of the user, and then I'll also pass the user to `get_pinned_articles()`.

escattone:
> How about unifying the two systems?

Great! I would prefer that as well! I'll do that.

escattone:
> My concern with this approach is that we are forcing the same pinned articles for the product landing page and the AAQ. Adding the pinned article config as a FK to the AAQConfig will allow for greater flexibility by having a different configuration file per page. Thoughts?

I don't see when we'd ever want different pinned articles on a product landing page versus its AAQ product page, so it feels like we're adding something that we might not ever use? I can add it though.

akatsoulas:
> > My concern with this approach is that we are forcing the same pinned articles for the product landing page and the AAQ. Adding the pinned article config as a FK to the AAQConfig will allow for greater flexibility by having a different configuration file per page. Thoughts?
> 
> I don't see when we'd ever want different pinned articles on a product landing page versus its AAQ product page, so it feels like we're adding something that we might not ever use? I can add it though.

The AAQ has a subset of topics compared to the product so I would assume that in most cases the AAQ will need to differentiate when we have the need to divert from the default behavior and have pinned articles

kitsune/questions/models.py
```diff
@@ -869,7 +868,6 @@ def __str__(self) -> str:
 class AAQConfig(ModelBase):
     title = models.CharField(max_length=255, default="")
     product = models.ForeignKey(Product, on_delete=models.CASCADE, related_name="aaq_configs")
-    pinned_articles = models.ManyToManyField(Document, null=True, blank=True)
```
akatsoulas, line 872: Don't we need to migrate this to the new configuration since we are deleting the field and we already have entries in prod?

kitsune/questions/admin.py
```diff
@@ -46,18 +48,31 @@ class QuestionLocaleAdmin(admin.TabularInline):
...
         "is_active",
-        "pinned_articles",
         "associated_tags",
         "extra_fields",
+        "pinned_articles",
     )
+    readonly_fields = ("pinned_articles",)
+
+    @admin.display(description="Pinned Articles")
+    def pinned_articles(self, obj):
```
akatsoulas: nit: shall we make it a bit more obvious to the function name that it's returning a link and not articles?

kitsune/wiki/utils.py
```diff
@@ -112,61 +113,135 @@ def _active_contributors_id(from_date, to_date, locale, product):
...
+    )
 
-    filter_kwargs = {
-        f"{parent_prefix}restrict_to_groups__isnull": True,
-    }
+    if len(pinned_articles) == limit:
+        return pinned_articles
+
+    qs = (
+        Document.objects.visible(user, locale=locale, is_template=False, is_archived=False)
```
akatsoulas, line 201: Let's add a comment about what this query is doing for future reference

kitsune/wiki/models.py
```diff
@@ -1147,6 +1147,71 @@ def __str__(self):
...
+        Product,
+        null=True,
+        blank=True,
+        on_delete=models.CASCADE,
+        related_name="pinned_article_config",
+        help_text=_lazy(
+            "Leave empty to configure for the Home Page or when linking to an AAQ configuration."
+        ),
+    )
+    aaq_config = models.ForeignKey(
```
akatsoulas: Instead of using this inverse FK, I would suggest to move this relationship to the AAQConfig by having a FK pointing to a PinnedArticleConfig.

Doing so will have some benefits like:
* It's more intuitive from a UX perspective to either select or create on the fly a configuration for the pinned articles when you are creating/editing a configuration for the AAQ. All of the configuration for the AAQ will be centralized.
* Eliminates the reverse lookup and the use of `first()`
* Simpler constraints and database queries
* Loosely coupling of the two configuration entries
* Multiple AAQs can share the same configuration file
* Separation of concerns. AAQConfig configures forums and the pinned config is responsible for the pinned articles

kitsune/questions/models.py
escattone: Good catch, thanks!

kitsune/questions/migrations/0022_remove_aaqconfig_pinned_articles_and_more.py
```diff
@@ -0,0 +1,86 @@
...
+
+        aaq_config.pinned_article_config = already_created[signature]
+        aaq_config.save()
+        print(
+            f'Associated AAQConfig "{aaq_config.title}" with PinnedArticleConfig '
+            f'"{aaq_config.pinned_article_config.title}".'
+        )
+
+
+def reverse_migrate_pinned_articles(apps, schema_editor):
```
akatsoulas: Let's remove the reverse. If this is run with the `pinned_articles` field is dropped from the schema it's going to fail and will block all the backwards operations under the `questions` app.

akatsoulas (approved):
looks good 🚀 r+wc

PR #6949, opened 2025-10-08
Create missing tranlsations for enabled locales

akatsoulas:
* Automatically approve AI based revisions for archived documents
* Introduce a query builder for l10n
* Refactor services
* Update strategy routing

### DIFF
--- a/docs/l10n-system.md
+++ b/docs/l10n-system.md
@@ -66,10 +66,33 @@ Different content managers handle the publishing workflow:
 
 #### 5. Services (`kitsune/wiki/services.py`)
 
+**TranslationQueryBuilder**
+- Centralized repository for complex translation discovery queries
+- Provides methods to find stale and missing translations
+- Implements document-aware routing logic for AI vs HYBRID flows
+- Key methods:
+  - `get_stale_docs_hybrid()`: Finds non-archived stale translations in HYBRID locales
+  - `get_stale_docs_ai()`: Finds all stale translations for AI flow (all AI docs + archived HYBRID docs)
+  - `get_missing_docs_hybrid()`: Finds non-archived documents without translations in HYBRID locales
+  - `get_missing_docs_ai()`: Finds documents without translations for AI flow
+  - `get_pending_translations()`: Finds unreviewed translations exceeding grace period
+  - `get_obsolete_translations()`: Finds translations that are no longer useful
+
+**BaseTranslationService**
+- Shared processing logic for translation services
+- Handles metadata building and strategy execution
+- Uses TranslationQueryBuilder for document discovery
+
 **StaleTranslationService**
 - Identifies outdated translations that need updates
 - Processes batch updates using appropriate strategies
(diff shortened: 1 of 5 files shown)


kitsune/wiki/services.py
```diff
@@ -12,130 +12,269 @@
...
+        """Find non-archived stale translations in HYBRID locales.
+
+        Args:
+            limit: Maximum number of candidates to return
 
-        candidates = [
+        Returns:
+            List of tuples: (english_document, translation_document, target_locale)
+        """
+        target_locales = settings.HYBRID_ENABLED_LOCALES
```
escattone: Nit. Seems like it's only used once, so maybe better to just use `settings.HYBRID_ENABLED_LOCALES` below?

kitsune/wiki/services.py
```diff
@@ -12,130 +12,269 @@
...
+    def get_missing_docs_hybrid(self, limit: int | None = None) -> list[tuple[Document, None, str]]:
+        """Find non-archived English docs without translations in HYBRID locales.
+
+        Args:
+            limit: Maximum number of candidates to return
+
+        Returns:
+            List of tuples: (english_document, None, target_locale)
+        """
+        target_locales = settings.HYBRID_ENABLED_LOCALES
```
escattone: Nit. Same comment as above.

kitsune/wiki/services.py
```diff
@@ -12,130 +12,269 @@
...
+            if remaining_limit:
+                docs = docs[:remaining_limit]
 
-class HybridTranslationService:
-    def __init__(self):
-        self.content_manager = HybridContentManager()
+            for doc in docs:
+                missing_translations.append((doc, None, locale))
+                if limit and len(missing_translations) >= limit:
+                    break
```
escattone: This seems unnecessary to me, since if we have a `limit`, we're already restricting the `docs` above using `docs[:remaining_limit]`. If so, seems like we could simplify the addition of tuples to `missing_translations` to something like:

```python
missing_translations.extend((doc, None, locale) for doc in docs)
```

kitsune/wiki/services.py
```diff
@@ -12,130 +12,269 @@
...
+            )
+
+            remaining_limit = limit - len(missing_translations) if limit else None
+            if remaining_limit:
+                docs = docs[:remaining_limit]
+
+            for doc in docs:
+                missing_translations.append((doc, None, locale))
+                if limit and len(missing_translations) >= limit:
+                    break
```
escattone: Same comment as above.

docs/l10n-system.md
```diff
@@ -163,7 +204,18 @@ graph TD
...
+Updates outdated translations across all AI/Hybrid enabled locales. The service automatically routes documents based on their archive status and locale configuration.
+
+### Create Missing Translations
+
+```bash
+python manage.py create_missing_translations --limit 50
+```
+
+Creates initial translations for English documents that don't have translations yet. Uses document-aware routing:
+- AI locales: translates all documents (archived and non-archived)
```
escattone, line 216: What do you think about adding `and auto-publishes`?

```
- AI locales: translates and auto-publishes all documents (archived and non-archived)
```

PR #6956, opened 2025-10-14
Automatically sync product versions

akatsoulas:


kitsune/products/management/commands/sync_product_versions.py
```diff
@@ -0,0 +1,228 @@
...
+            help="Sync only this product (slug). Options: firefox, mobile, thunderbird",
+        )
+        parser.add_argument(
+            "--dry-run",
+            action="store_true",
+            help="Show what would be changed without modifying the database",
+        )
+
+    def handle(self, *args, **options):
+        dry_run = options["dry_run"]
```
escattone, line 26: Based on the code below and some testing, the `dry-run` option doesn't show what would have changed.

kitsune/products/management/commands/sync_product_versions.py
```diff
@@ -0,0 +1,228 @@
...
+        verbosity = options["verbosity"]
```
escattone, line 27: The `verbosity` is never used anywhere as far as I can tell.

PR #6958, opened 2025-10-15
Fix ESR versioning

akatsoulas:


kitsune/products/management/commands/sync_product_versions.py
```diff
@@ -63,7 +64,18 @@ def _get_product_configs(self):
                 "version_data": product_details.firefox_versions,
                 "version_key": "LATEST_FIREFOX_VERSION",
                 "history_data": product_details.firefox_history_major_releases,
+                "esr_keys": [],
+                "esr_only": False,
+            },
+            "firefox-enterprise": {
+                "name": "Firefox for Enterprise",
+                "slug_prefix": "fx",
```
escattone, line 72: Shouldn't it have its own unique `slug_prefix`?

kitsune/products/management/commands/sync_product_versions.py
```diff
@@ -96,20 +119,19 @@ def _sync_product(self, product: Product, config: dict, dry_run: bool, verbosity
         if verbosity >= 1:
             self.stdout.write(f"\nSyncing {product.title} (latest: {latest_version})...")
 
-        history_data = config["history_data"]
-        available_versions = self._get_available_versions(history_data, latest_major)
+        esr_only = config.get("esr_only", False)
+
+        # Use transaction only if not dry-run
+        context = transaction.atomic() if not dry_run else nullcontext()
```
escattone, line 125: Nice.

kitsune/products/management/commands/sync_product_versions.py
akatsoulas: I believe `fx` is fine here. The distinction is in the versions and the release cycle. We are avoiding colisions b/c of the use of `-esr` and semantically makes sense. I am concerned that if we differentiate here, we are going to make the manual management through the admin more prone to errors. Merging as is. We can update if needed.

PR #6970, opened 2025-10-21
randomly cycle pinned articles if more than limit

escattone:
mozilla/sumo#2591

### DIFF
--- a/kitsune/wiki/utils.py
+++ b/kitsune/wiki/utils.py
@@ -174,11 +174,14 @@ def get_featured_articles(
             product=product,
             locale=locale,
             fetch_for_aaq=fetch_for_aaq,
-        )[:limit]
+        )
     )
 
-    if len(pinned_articles) == limit:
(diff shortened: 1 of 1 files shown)


kitsune/wiki/utils.py
```diff
@@ -174,12 +174,15 @@ def get_featured_articles(
...
             fetch_for_aaq=fetch_for_aaq,
-        )[:limit]
+        )
     )
 
-    if len(pinned_articles) == limit:
+    if (num_pinned_articles := len(pinned_articles)) == limit:
         return pinned_articles
 
+    elif num_pinned_articles > limit:
```
akatsoulas, line 183: nit - this could be reduced to:
```python
if len(pinned_articles) <= limit:
    return pinned_articles
return random.sample(pinned_articles, limit)
```

akatsoulas:
merging this to allow for QA

PR #6983, opened 2025-10-23
better featured articles but with the same performance

escattone:
## Notes
- Expands the functionality of `get_featured_articles()`:
    - **Visibility**. Restricted articles can now be included in the featured articles depending on the permissions of the viewer.
    - **No longer excludes featured articles from the `Firefox Focus`, `Thunderbird`, and `Thunderbird for Android` product landing pages**.
    - **Excludes non-default category articles**.
    - **Non-English parent articles**. Non-English parent articles, although rare, can now be included in the featured articles for their specific locale.
- **The performance is the same as the original**, even with the extra pinned-article handling. Tested in production during full load:
```
>>> import random
>>> from typing import Any
>>> from django.conf import settings
>>> from django.db.models import Q
>>> from django.db.models.query import QuerySet
>>> from kitsune.dashboards import LAST_7_DAYS
>>> from kitsune.wiki.config import REDIRECT_HTML
>>> from kitsune.wiki.models import Document, PinnedArticleConfig
>>> 
l>>> PINNED_ARTICLE_LIMIT_Q = Q(
...     parent__isnull=True,
...     is_template=False,
...     is_archived=False,
...     category__in=settings.IA_DEFAULT_CATEGORIES,
... ) & ~Q(html__startswith=REDIRECT_HTML)
>>> 
>>> def get_q_object_for_parent(locale: str | None = None, **kwargs: dict[str, Any]) -> Q:
...     """
...     Returns a Django Q object that ensures that each of the provided keyword...     arguments is applied only on the parent. Provide the "locale" argument only
...     if already filtering for that locale.
...     """
...     parent_q = Q(**kwargs)
...     if locale == settings.WIKI_DEFAULT_LANGUAGE:
...         return parent_q
...     # Non-English documents can be parents too.
...     child_q = Q(**{f"parent__{k}": v for k, v in kwargs.items()})
...     return (Q(parent__isnull=True) & parent_q) | (Q(parent__isnull=False) & child_q)
... 
>>> def get_pinned_articles(
...     user=None, product=None, locale=settings.WIKI_DEFAULT_LANGUAGE, fetch_for_aaq=False
... ) -> QuerySet[Document]:
...     """
...     Given the product, locale, and whether or not we're getting pinned articles for the
...     AAQ, returns a queryset of the pinned articles that are visible to the given user.
...     """
...     qs = PinnedArticleConfig.objects
...     if fetch_for_aaq:
...         qs = qs.filter(aaq_configs__product=product, aaq_configs__is_active=True)
...     elif product:
...         qs = qs.filter(products=product)
...     else:
...         qs = qs.filter(use_for_home_page=True)
...     if not (config := qs.first()):
...         return Document.objects.none()
...     pinned_articles = config.pinned_articles.filter(PINNED_ARTICLE_LIMIT_Q)
...     condition = Q(id__in=pinned_articles)
...     if locale != settings.WIKI_DEFAULT_LANGUAGE:
...         condition = condition | Q(parent__in=pinned_articles)
...     return (
...         Document.objects.visible(user, locale=locale)
...         .filter(condition)
...         .select_related("current_revision")
...     )
... 
>>> def gf2(
...     user=None,
...     product=None,
...     topics=None,
...     locale=settings.WIKI_DEFAULT_LANGUAGE,
...     fetch_for_aaq=False,
...     limit=4,
... ):
...     """Returns up to 4 random articles from the most visited.
...     Args:
...         user: Optional user for visibility
...         product: Optional product to filter by
...         topics: Optional iterable of topics to filter by
...         locale: Locale to get articles for, defaults to WIKI_DEFAULT_LANGUAGE
...         fetch_for_aaq: Optional boolean (currently ignored)
...         limit: Optional integer to limit the number of articles
...     """
...     pinned_articles = list(
...         get_pinned_articles(
...             user=user,
...             product=product,
...             locale=locale,
...             fetch_for_aaq=fetch_for_aaq,
...         )
...     )
...     if (num_pinned_articles := len(pinned_articles)) >= limit:
...         if num_pinned_articles > limit:
...             pinned_articles = random.sample(pinned_articles, limit)
...         return pinned_articles
...     filter_kwargs = {"category__in": settings.IA_DEFAULT_CATEGORIES}
...     if product:
...         filter_kwargs.update(products=product)
...     if topics:
...         filter_kwargs.update(topics__in=topics)
...     qs = Document.objects.visible(
...         user,
...         locale=locale,
...         is_template=False,
...         is_archived=False,
...         current_revision__isnull=False,
...     ).filter(get_q_object_for_parent(locale=locale, **filter_kwargs))
...     if (not product) and (excluded_slugs := settings.EXCLUDE_PRODUCT_SLUGS_FEATURED_ARTICLES):
...         # If we're not filtering by a specific product, exclude the products...         # that have been configured for exclusion from featured articles.
...         qs = qs.exclude(
...             get_q_object_for_parent(
...                 locale=locale,
...                 products__slug__in=excluded_slugs,
...             )
...         )
...     qs = (
...         qs.filter(visits__period=LAST_7_DAYS)
...         .exclude(html__startswith=REDIRECT_HTML)
...         .select_related("current_revision")
...         .order_by("-visits__visits")
...     )
...     if topics:
...         # Documents that match multiple topics will be repeated,
...         # so remove any duplicates when we're matching by topics.
...         qs = qs.distinct()
...     # Only include the "limit * 2" most visited articles for sampling.
...     docs = list(qs[: limit * 2])
...     remaining_limit = limit - len(pinned_articles)
...     if len(docs) > remaining_limit:
...         docs = random.sample(docs, remaining_limit)
...     return pinned_articles + docs
... 
>>> timeit("gf2()", globals=globals(), number=100)
5.793605691025732
>>> product = Product.objects.get(slug="firefox")
>>> product
<Product: Firefox>
>>> timeit("gf2(product=product)", globals=globals(), number=100)
3.727559871011181
>>> timeit(setup="from kitsune.wiki.utils import get_featured_articles", stmt="get_featured_articles()", number=100)
6.661416374990949
>>> timeit(setup="from kitsune.wiki.utils import get_featured_articles", stmt="get_featured_articles(product=product)", globals=globals(), number=100)
3.410288790008053
```

### DIFF
--- a/kitsune/wiki/utils.py
+++ b/kitsune/wiki/utils.py
@@ -1,7 +1,6 @@
 import random
 import time
 from itertools import chain, islice
-from typing import Any
 
 import requests
 from django.conf import settings
@@ -173,10 +172,8 @@ def get_featured_articles(
         )
     )
 
-    if (num_pinned_articles := len(pinned_articles)) >= limit:
-        if num_pinned_articles > limit:
-            pinned_articles = random.sample(pinned_articles, limit)
-        return pinned_articles
+    if len(pinned_articles) >= limit:
+        return random.sample(pinned_articles, limit)
 
     filter_kwargs = {"category__in": settings.IA_DEFAULT_CATEGORIES}
 
@@ -186,23 +183,30 @@ def get_featured_articles(
     if topics:
         filter_kwargs.update(topics__in=topics)
 
+    def q_for_parent(**kwargs):
+        """
+        Return a Q object that ensures that each of the
+        provided conditions is applied only on the parent.
+        """
+        parent_q = Q(**kwargs)
+        if locale == settings.WIKI_DEFAULT_LANGUAGE:
+            return parent_q
+        # Non-English documents can be parents too.
+        child_q = Q(**{f"parent__{k}": v for k, v in kwargs.items()})
+        return (Q(parent__isnull=True) & parent_q) | (Q(parent__isnull=False) & child_q)
+
     qs = Document.objects.visible(
         user,
         locale=locale,
         is_template=False,
         is_archived=False,
         current_revision__isnull=False,
-    ).filter(get_q_object_for_parent(locale=locale, **filter_kwargs))
+    ).filter(q_for_parent(**filter_kwargs))
 
     if (not product) and (excluded_slugs := settings.EXCLUDE_PRODUCT_SLUGS_FEATURED_ARTICLES):
         # If we're not filtering by a specific product, exclude the products
         # that have been configured for exclusion from featured articles.
-        qs = qs.exclude(
-            get_q_object_for_parent(
-                locale=locale,
-                products__slug__in=excluded_slugs,
-            )
-        )
+        qs = qs.exclude(q_for_parent(products__slug__in=excluded_slugs))
 
     qs = (
         qs.filter(visits__period=LAST_7_DAYS)
@@ -455,17 +459,3 @@ def get_kb_visited(
             urls.extend(visits.keys())
 
     return urls
-
-
-def get_q_object_for_parent(locale: str | None = None, **kwargs: dict[str, Any]) -> Q:
-    """
-    Returns a Django Q object that ensures that each of the provided keyword
-    arguments is applied only on the parent. Provide the "locale" argument only
-    if already filtering for that locale.
-    """
-    parent_q = Q(**kwargs)
-    if locale == settings.WIKI_DEFAULT_LANGUAGE:
-        return parent_q
-    # Non-English documents can be parents too.
-    child_q = Q(**{f"parent__{k}": v for k, v in kwargs.items()})
-    return (Q(parent__isnull=True) & parent_q) | (Q(parent__isnull=False) & child_q)
--- a/kitsune/wiki/tests/test_utils.py
+++ b/kitsune/wiki/tests/test_utils.py
@@ -3,9 +3,7 @@ import time
 from datetime import date, timedelta
 from unittest import mock
 
-from django.conf import settings
 from django.contrib.sessions.backends.base import SessionBase
-from django.db.models import Q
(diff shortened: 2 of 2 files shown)


kitsune/wiki/utils.py
```diff
@@ -156,64 +156,75 @@ def get_featured_articles(
...
+            locale=locale,
+            fetch_for_aaq=fetch_for_aaq,
+        )
+    )
 
-    filter_kwargs = {
-        f"{parent_prefix}restrict_to_groups__isnull": True,
-    }
+    if (num_pinned_articles := len(pinned_articles)) >= limit:
+        if num_pinned_articles > limit:
```
akatsoulas: This is redundant. You can directly use:
```python
if len(pinned_articles) >= limit:
    return random.sample(pinned_articles, limit)
```

kitsune/wiki/utils.py
```diff
@@ -444,3 +455,17 @@ def get_kb_visited(
             urls.extend(visits.keys())
 
     return urls
+
+
+def get_q_object_for_parent(locale: str | None = None, **kwargs: dict[str, Any]) -> Q:
```
akatsoulas: I would move this as a closure within the get_featured_articles to scope it locally and help readability. This is highly specific to that query

kitsune/wiki/utils.py
escattone: I tried that, but the side effect is that if the `limit` number of articles are pinned, their order is randomly shuffled. If that's OK, then I like this better too.

kitsune/wiki/utils.py
escattone: 👍 I originally had this as a closure, but this is a pattern that occurs frequently, so I thought it would be useful outside of this function. I'll move it inside and we can later consider a more general function if needed.

PR #6993, opened 2025-10-31
replace APScheduler with Celery beat

escattone:
mozilla/sumo#1897
mozilla/sumo#2279
mozilla/sumo#2586

## Notes
- This PR replaces `APScheduler` with Celery `beat`, **and also configures the Celery `beat` auto-discovery for Sentry Crons**.
- This PR depends on mozilla/webservices-infra#8216, **which should be merged immediately prior to merging this PR**.
- Much of this PR is moving code out of Django management commands and into Celery tasks that can be used directly by Celery `beat`.
- Periodic tasks have been moved from `APScheduler` to Celery `beat`.
- The Celery `beat` singleton asynchronously submits periodic tasks directly to the Celery queue, and since all tasks run within Celery workers are already covered by Sentry, we get full Sentry coverage of any errors within our periodic tasks.
- The managements commands that were created to be run by `APScheduler` but converted into tasks with this PR have been removed.
- The Celery `beat` singleton needs to reliably persist its status across terminations and restarts, so the `django-celery-beat` package has been included to provide that via the database, which is the standard, production-worthy approach for Django applications.
- Note that for read-only environments (our `stage` and `prod` European deployments), we do not run Celery (neither `beat` or workers) at all.
- The `beat` service within `docker-compose` is not required for local development, so it's not included by default. However, it can be included if desired by using the `--profile beat` option, so for example, `docker-compose --profile beat up --build -d` will include the Celery `beat` service for local development/testing. I've also included the `flower` service (for Celery task monitoring) under the `beat` profile.

### DIFF
--- a/kitsune/settings.py
+++ b/kitsune/settings.py
@@ -8,9 +8,9 @@ import re
 
 import dj_database_url
 import django_cache_url
-from celery.schedules import crontab
 from decouple import Csv, config
 
+from kitsune.celery_beat import PERIODIC_TASKS_ALL, PERIODIC_TASKS_PRODUCTION_ONLY
 from kitsune.lib.sumo_locales import LOCALES
 
 DEBUG = config("DEBUG", default=False, cast=bool)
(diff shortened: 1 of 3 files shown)


docker-compose.yml
```diff
@@ -37,6 +37,22 @@ services:
     depends_on:
       - web
 
+  beat:
```
akatsoulas, line 40: I wonder if we want to also introduce flower locally

akatsoulas (commented):
Merely a review. I will wait for all tests to pass in order to review it. I was wonder though, now with celery beat, why are we keeping the management commands around?

escattone:
> Merely a review. I will wait for all tests to pass in order to review it. I was wonder though, now with celery beat, why are we keeping the management commands around?

Great question. I was simply maintaining the status quo so to speak. There's really no need for them with this change, and I would actually prefer removing them. I'll do that!

docker-compose.yml
escattone: That's a good idea.

kitsune/settings.py
```diff
@@ -1387,3 +1354,166 @@ def setup_logger(name, formatter_str, level=logging.DEBUG):
 )
 STALE_TRANSLATION_BATCH_SIZE = config("STALE_TRANSLATION_BATCH_SIZE", default=10, cast=int)
 HYBRID_REVIEW_GRACE_PERIOD = config("HYBRID_REVIEW_GRACE_PERIOD", default=72, cast=int)  # Hours
+
+# Celery beat configuration
+DJANGO_CELERY_BEAT_TZ_AWARE = False  # This should match USE_TZ.
+CELERY_BEAT_SCHEDULER = "django_celery_beat.schedulers:DatabaseScheduler"
+CELERY_BEAT_SCHEDULE: dict[str, dict] = {}
+if not READ_ONLY:
```
akatsoulas: Can't we move this in a separate file eg `celery.py` to reduce clutter?

kitsune/settings.py
```diff
@@ -1387,3 +1354,166 @@ def setup_logger(name, formatter_str, level=logging.DEBUG):
...
+            "schedule": crontab(hour="1", minute="11"),
+        },
+        # Daily at 01:40.
+        "update_weekly_votes": {
+            "task": "kitsune.questions.tasks.update_weekly_votes",
+            "schedule": crontab(hour="1", minute="40"),
+        },
+        # Daily at 04:00.
+        "auto_archive_old_questions": {
+            "task": "kitsune.quetions.tasks.auto_archive_old_questions",
```
akatsoulas: typo

kitsune/wiki/tasks.py
```diff
@@ -198,6 +208,23 @@ def schedule_rebuild_kb():
     rebuild_kb.delay()
 
 
+@shared_task
+@skip_if_read_only_mode
+def run_rebuild_kb() -> None:
+    """Try to run a KB rebuild, if we're allowed to."""
+    if waffle.switch_is_active("wiki-rebuild-on-demand"):
```
akatsoulas, line 215: This is missing the`settings.CELERY_TASK_ALWAYS_EAGER` check

kitsune/wiki/tasks.py
```diff
@@ -220,9 +247,30 @@ def add_short_links(doc_ids):
         pass
 
 
+@shared_task
+@skip_if_read_only_mode
+def generate_missing_share_links() -> None:
```
akatsoulas, line 251: Here we have a task that calls another one. We should move the query withing the add_short_links task

kitsune/settings.py
```diff
@@ -1387,3 +1354,166 @@ def setup_logger(name, formatter_str, level=logging.DEBUG):
 )
 STALE_TRANSLATION_BATCH_SIZE = config("STALE_TRANSLATION_BATCH_SIZE", default=10, cast=int)
 HYBRID_REVIEW_GRACE_PERIOD = config("HYBRID_REVIEW_GRACE_PERIOD", default=72, cast=int)  # Hours
+
+# Celery beat configuration
+DJANGO_CELERY_BEAT_TZ_AWARE = False  # This should match USE_TZ.
+CELERY_BEAT_SCHEDULER = "django_celery_beat.schedulers:DatabaseScheduler"
+CELERY_BEAT_SCHEDULE: dict[str, dict] = {}
+if not READ_ONLY:
+    CELERY_BEAT_SCHEDULE |= {
```
akatsoulas: nit: It could be more readable if we organize the imports per app

akatsoulas: Are we missing the reindex task that we had in cron?

akatsoulas (approved):
r+wc

kitsune/settings.py
escattone: > Are we missing the reindex task that we had in cron?

Yes, we're missing the reindex task that reindexed all items updated within the last 90 minutes. However, that reindex task is no longer necessary now that the main and read-only clusters share the same ES instance. Also, if we switch over to the read-only cluster, the main cluster's beat process and related worker tasks will still be running, so any unfinished indexing tasks will still complete.

kitsune/wiki/tasks.py
escattone: As far as I can tell, we don't check that currently at any point for the `rebuild_kb` management command that is called by `cron` (which this task replaces), but I suppose we should add it?

kitsune/wiki/tasks.py
escattone: The `add_short_links` task is used separately by the `DocumentForm` class, so I think I'll make the `add_short_links` task call within `generate_missing_share_links` synchronous (so it runs within that same task instead of kicking-off another one).

PR #6998, opened 2025-11-03
chore: remove NewRelic agent

Basma1912:
as part of the effort of ending the newrelic contract, SUMO is using NewRelic APM Agent and this PR removes newrelic from the code.

### DIFF
--- a/bin/run-cron.sh
+++ b/bin/run-cron.sh
@@ -1,3 +1,3 @@
 #!/bin/bash
 
-exec run-program python manage.py runscript cron
+exec python manage.py runscript cron


bin/run-cron.sh
```diff
@@ -1,3 +1,3 @@
 #!/bin/bash
 
-exec newrelic-admin run-program python manage.py runscript cron
+exec run-program python manage.py runscript cron
```
escattone:
```suggestion
exec python manage.py runscript cron
```

escattone (changes requested):
Just a small tweak, but otherwise looks good!

Basma1912:
> Just a small tweak, but otherwise looks good!

Thank you @escattone

akatsoulas (approved):
We also need to drop the newrelic package from the deps. With that said though, I am happy to follow up with another PR for this

PR #7000, opened 2025-11-03
Introduce spam classifier for Zendesk

akatsoulas:
* Refactor existing classifier.

### DIFF
--- a/kitsune/llm/categorization/classifiers.py
+++ b/kitsune/llm/categorization/classifiers.py
@@ -3,16 +3,16 @@ from typing import Any
 from kitsune.llm.categorization.prompt import (
     DEFAULT_PRODUCT_RESULT,
     DEFAULT_TOPIC_RESULT,
+    build_product_prompt,
     build_topic_prompt,
     product_parser,
-    product_prompt,
     topic_parser,
 )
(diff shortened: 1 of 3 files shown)


kitsune/customercare/models.py
```diff
@@ -0,0 +1,45 @@
...
+    product = models.ForeignKey(Product, on_delete=models.CASCADE, related_name="support_tickets")
+    user = models.ForeignKey(
+        User, on_delete=models.SET_NULL, null=True, blank=True, related_name="support_tickets"
+    )
+    zendesk_tags = models.JSONField(default=list, blank=True)
+    status = models.CharField(
+        max_length=20, choices=STATUS_CHOICES, default=STATUS_PENDING, db_index=True
+    )
+    zendesk_ticket_id = models.CharField(max_length=255, null=True, blank=True)
+    created = models.DateTimeField(auto_now_add=True, db_index=True)
```
escattone, line 39: Super nit. Consider using `Meta.indexes` instead of `db_index=True`.

escattone (approved):
r+wc

Regarding the minor prompt changes, I'm a bit concerned that the LLM will handle the term `content` differently than the more specific term `question`, especially when asking it to determine `its primary intent or main concern`.

PR #7009, opened 2025-11-05
Update styleguide build & deployment

### LINKED ISSUE #6696: Styleguide (KSS) usage & deployment
Wondering how the styleguide content and its entrypoints fit within the project, how are supposed to be used, and if/where are generally available, the impression I got was they are being somewhat forgotten over time, which is a shame given the effort put into it — nonetheless I wasn't able to find any mention in the docs or any workflows so it very well could just have been some tribal knowledge only…

Surprisingly I wasn't even able to get `kss` to build locally with current configs, so wasn't sure how — if ever — it is getting used, or supposed to be consumed or updated. After a couple of tweaks to get it to build it looked like that's something worth having as a reference online, in the published docs, for anyone to be able to look into things without having to either envision it just from sources, or having to build own copies of the styleguide.

Would it make sense to have it as part of the kitsune docs deployment?

I made a quick PoC: https://janbrasna.github.io/kitsune/_kss/ just getting the content updated with CI.

(I chose the _kss subdir to make sure it won't interfere with built mkdocs outputs in the same workdir… does that make sense or it should be renamed/moved elsewhere?)

If that looks like something worth adding, I'll also add a link to it to the main docs, and open a PR with it.

janbrasna:
Resolves #6696

After #6785 is published, there can now be links added to the KSS site.

This also moves the directories used in the runner a bit, as the built artifacts were being included twice in the deployment.

Bumps a few versions while at it.

### DIFF
--- a/docs/frontend.md
+++ b/docs/frontend.md
@@ -86,6 +86,11 @@ Its configuration lives in `postcss.config.js`.
 When building a production bundle,
 [cssnano](https://cssnano.co/) minifies our CSS.
 
+### KSS styleguide
+
+Components and layouts are documented in [SUMO Styleguide](https://mozilla.github.io/kitsune/_kss/) using `kss-node`
+automation to generate the documentation from SCSS sources.
+
 ## Images
 
 Images directly imported in JS files,


janbrasna:
FWIW the build run `npm ci` on _main_ failed for npm woes https://github.com/mozilla/kitsune/pull/7137#issuecomment-3699946620 that should turn green once the next changeset touching the docs triggers another deploy.

PR #7015, opened 2025-11-06
fix N+1 query issue in `question.details`

escattone:
mozilla/sumo#2623

### DIFF
--- a/kitsune/questions/jinja2/questions/question_details.html
+++ b/kitsune/questions/jinja2/questions/question_details.html
@@ -396,7 +396,6 @@
       <li class="sidebar-subheading sidebar-nav--heading-item">{{ _('Tags') }}</li>
       <li id="tags-list">
         <div class="folder">
-          {% set tags = question.my_tags %}
           {% if tags or can_tag %}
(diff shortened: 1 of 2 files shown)


kitsune/questions/jinja2/questions/question_details.html
```diff
@@ -413,7 +412,7 @@ <h3 class="sumo-card-heading">{{ _('Post a Reply') }}</h3>
                         {% for tag in all_tags %}
                           <option value="{{ tag.id }}"
                                   data-url="{{ url('questions.list', product.slug if product else 'all')|urlparams(tagged=tag.slug) }}"
-                                  {% if tag in question.tags.all() %}selected{% endif %}>
+                                  {% if tag.id in tag_ids %}selected{% endif %}>
```
escattone, line 415: Just FYI, checking for `tag.id in tag_ids`, where `tag_ids` is a set of integers, is over 10x faster than checking for `tag in tags`, where `tags` is a list of `SumoTag` objects (even when the list is small). In the larger scheme of things, it's probably not important, but it's easy to do, so why not.

PR #7021, opened 2025-11-10
Correct the wording about marking the reply as a solution in new_answer_to_asker templates

denyshon:
Resolves [#2617](https://github.com/mozilla/sumo/issues/2617)


denyshon:
@kelimuttu what do you think of this?

kelimuttu:
I'll probably slightly change the button text to "This solves my problem". But I'm also not a UX expert here.

denyshon:
Hmm that makes sense

PR #7028, opened 2025-11-12
Add context to differentiate the Updated strings

denyshon:
Add context to differentiate the "Updated" strings from the l10n dashboard and the Forum sorting dropdown.
Replace "order" with "sort" in l10n comments for better clarity.

Fully resolves [#2630](https://github.com/mozilla/sumo/issues/2630)

### DIFF
--- a/kitsune/dashboards/readouts.py
+++ b/kitsune/dashboards/readouts.py
@@ -19,6 +19,7 @@ from django.utils.translation import gettext as _
 from django.utils.translation import gettext_lazy as _lazy
 from django.utils.translation import pgettext_lazy
 from markupsafe import Markup
+from tower import ugettext_lazy
 
 from kitsune.dashboards import LAST_30_DAYS, PERIODS
 from kitsune.dashboards.models import WikiDocumentVisits
@@ -45,7 +46,7 @@ MOST_VIEWED = 1
 MOST_RECENT = 2
 REVIEW_STATUSES = {
     1: (_lazy("Review Needed"), "wiki.document_revisions", "review"),
-    0: (_lazy("Updated"), "", "ok"),
+    0: (ugettext_lazy("Updated", context="article l10n status"), "", "ok"),
 }
 SIGNIFICANCE_STATUSES = {
     MEDIUM_SIGNIFICANCE: (_lazy("Update Needed"), "wiki.edit_document", "update"),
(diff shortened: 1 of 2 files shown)


denyshon:
@akatsoulas excuse me, is this the wrong way to add context?
image
I've taken it [from the docs](https://mozilla.github.io/kitsune/localization/#adding-context-with-msgctxt)... Is that the place where `_lazy()` differs from `_()`?

kitsune/dashboards/readouts.py
```diff
@@ -45,7 +45,7 @@
 MOST_RECENT = 2
 REVIEW_STATUSES = {
     1: (_lazy("Review Needed"), "wiki.document_revisions", "review"),
-    0: (_lazy("Updated"), "", "ok"),
+    0: (_lazy("Updated", context="article l10n status"), "", "ok"),
```
janbrasna: Ah now I get why it maybe wasn't wrapped before, for the msgid collision here;D 

My cheap fix here would be: change the status text. To something like "Up to date" — which would then live as a distinct message, not having to disambiguate from another one.

denyshon: @janbrasna I was thinking about the exact same thing... Though I'd really want to find out what's wrong with the context attribute (and probably find a way to make it work)

janbrasna: TL;DR might be the docs are outdated, and in previous versions used the gettext directly, or imported a different function and no longer do; now just rely on Django utils `gettext` & `gettext_lazy` — that don't take message context — for that it would need migrating to some `pgettext` https://docs.djangoproject.com/en/5.2/topics/i18n/translation/#contextual-markers but that means rewriting the whole view I assume?

denyshon: I guess I've figured out what's happening. The docs suggest using Tower's `_()` and `_lazy()`, but we're using Django's ones (which do not have context). Let's see if the Tower's ones will work.

denyshon: (without rewriting the whole code)

denyshon: Aha, so Tower is not available 😁

denyshon: I'll proceed with changing the string then

denyshon:
FYI I've filed https://github.com/mozilla/sumo/issues/2637

kitsune/dashboards/readouts.py
denyshon: Huh, looks like I had to use pgettext_lazy. There's no point in changing it here already, but I'll note it for the future

PR #7038, opened 2025-11-14
Optimize questions_list queries

akatsoulas:


kitsune/questions/views.py
```diff
@@ -258,7 +273,14 @@ def question_list(request, product_slug=None, topic_slug=None):
...
     )
 
-    question_qs = question_qs.prefetch_related("topic", "product")
+    question_qs = question_qs.prefetch_related("tags")
+
+    # Annotate with visit counts to avoid N+1 queries
+    question_qs = question_qs.annotate(
+        visits_count=Subquery(
+            QuestionVisits.objects.filter(question=OuterRef("pk")).values("visits")[:1]
+        )
```
escattone: You could simplify the annotation by using the related name if you'd like:

`visits_count=F("questionvisits__visits")`

kitsune/questions/models.py
```diff
@@ -429,6 +431,9 @@ def kb_visits_prior_to_creation(self) -> list[str]:
     @property
     def my_tags(self):
         """A caching wrapper around self.tags.all()."""
+        if hasattr(self, '_prefetched_objects_cache') and 'tags' in self._prefetched_objects_cache:
```
escattone: This would be perfect if `_prefetched_objects_cache` was a public interface, but I don't feel totally comfortable using it given that it isn't. If you like it though, let's use it.

We could instead use something like `sorted(question.tags.all(), key=lambda t: t.name)` in the `question_list.html` template -- hm, I suppose we'd have to do that in a helper function -- instead of `question.my_tags`.

escattone (approved):
Nice, it feels good to address this!

r+wc

PR #7045, opened 2025-11-16
Remove unnecessary if statement in macros.html

denyshon:
The code in the `if` and `else` blocks is identical, except for the "most-visited-row" `<tr>` class. However, this class is not defined or used anywhere outside of this statement.

The PR removes the conditional statement, leaving the `else` block content in the code.


akatsoulas (approved):
nice catch!

PR #7057, opened 2025-11-18
Send tags to ZD

akatsoulas:


kitsune/llm/support/classifiers.py
```diff
@@ -105,25 +108,27 @@ def handle_spam(payload: dict[str, Any], spam_result: dict[str, Any]) -> dict[st
         action = determine_action_from_spam_result(spam_result)
 
         if not ((action == ModerationAction.SPAM) and spam_result.get("maybe_misclassified")):
-            return {"action": action, "product_result": {}}
+            return {"action": action, "product_result": {}, "topic_result": {}}
```
escattone, line 111: Nit. Adding the `topic_result` is not necessary here, since it's included via the `base_result` used within the `decision_lambda` below.

escattone (approved):
r+wc

escattone:
I verified that the test pass locally. The test failure was due to a flaky test.

PR #7063, opened 2025-11-20
Display announcements per product

akatsoulas:


escattone:
## Notes
@akatsoulas Nice work! I made some significant changes though (made in the second commit):
- Placement. Product announcements are always included at the top of the page along with any other announcements.
- Ordering. If there is more than one announcement, they are shown in the reverse order in which they were created, so most recent on top.
- Simplification:
    - The Jinja helpers have been collapsed into a single `get_announcements` function with an optional `product` keyword argument.
    - All of the announcements are created in the `base.html` template. Child templates that want to include product-specific announcements can simply set the `announcement_product` Jinja variable.
- Product-specific announcements have been added to the AAQ flow pages.

## Question
- **Should we also include product-specific announcements on all of the question pages that include their product?** For example, a Firefox announcement on the `/questions/firefox` page.

escattone (approved):

romanrozputniak-eng:
пт, 21 лист. 2025 р., 00:45 користувач Ryan Johnson <
***@***.***> пише:

PR #7087, opened 2025-11-27
Rework Share data fallback message formatting

### LINKED ISSUE #6881: AAQ "Share Data" Fallback Needs Target Attribute
* Example problem page: https://support.mozilla.org/en-US/questions/new/firefox/form
* Code location: https://github.com/mozilla/kitsune/blob/main/kitsune/questions/jinja2/questions/includes/aaq_macros.html#L201

When the Share Data button is not able to retrieve Troubleshooting Information automatically, the page displays this:

`<p id="troubleshooting-manual" class="hide-until-expanded" style="display: block;">`
`We can't automatically get your browser's troubleshooting data, please <a href="https://support.mozilla.org/kb/use-troubleshooting-information-page-fix-firefox">try these manual steps</a>.`
`</p>`

This link navigates away from the user's question, destroying the effort put into drafting it. That link should have `target="_blank"` to avoid this problem.

Over the years, numerous users have complained about this, but I hadn't seen Share Data fail before so this is the first time I could demonstrate it for myself.

(One might also take the view that going "Back" to the AAQ form page AND to /questions/ pages should restore data from cache, but that's a battle for another day. Let's fix this one ASAP.)

denyshon:
- Move the tags from the Share data fallback message to the trans block
- Add target="_blank" to the link in the Share data fallback message

Resolves #6881.

The old string is kept as deprecated to preserve existing translations and give localizers time to transfer them. We'll remove that string later.

### DIFF
--- a/kitsune/questions/jinja2/questions/includes/aaq_macros.html
+++ b/kitsune/questions/jinja2/questions/includes/aaq_macros.html
@@ -197,9 +197,16 @@
         </button>
       </div>
       <p id="troubleshooting-manual" class="hide-until-expanded">
-        {% trans %}
-          We can't automatically get your browser's troubleshooting data, please
-          <a href="https://support.mozilla.org/kb/use-troubleshooting-information-page-fix-firefox">try these manual steps</a>.
+        {# L10n: Deprecated #}
+        {% set deprecated_share_data_fallback %}
+          {% trans %}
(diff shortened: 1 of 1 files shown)


denyshon:
Converting to draft until I clarify how Pontoon handles the formatting

denyshon:
Well, it should look fine on Pontoon

PR #7089, opened 2025-11-28
Switch from naturaltime to timesince

denyshon:
- Switch from `naturaltime` to `timesince` in `datetimeformat`, as `naturaltime` breaks l10n
- Rename `datetimeformat`'s `use_naturaltime` argument to `use_timesince`

Hopefully resolves [#1978](https://github.com/mozilla/sumo/issues/1978)


denyshon:
To clarify: I have tested it locally, and it works. My only concern is that it may not fully resolve [#1978](https://github.com/mozilla/sumo/issues/1978), since there may be several issues in one.

PR #7091, opened 2025-12-01
Improve pagination for group profiles

denyshon:
Switch to the paginator template for group member lists


denyshon:
The test failure seems to be unrelated to my changes. I would appreciate rerunning the tests

PR #7102, opened 2025-12-02
Separate spam queue for Zendesk

akatsoulas:


akatsoulas:
This PR also introduces a celery task which will auto-reject the tickets after 14 days in the queue.

kitsune/flagit/views.py
```diff
@@ -330,6 +342,34 @@ def moderate_content(request):
...
+    """Display Zendesk spam tickets for review by Staff members."""
+    ct_support_ticket = ContentType.objects.get_for_model(SupportTicket)
+
+    objects = (
+        get_flagged_objects(
+            reason=FlaggedObject.REASON_SPAM,
+            content_types=[ct_support_ticket],
+        )
+        .select_related("content_type", "creator")
+        .prefetch_related("content_object", "content_object__product", "content_object__user")
```
escattone: Recommend adding `"content_object__topic"` to the `prefetch_related` call.

kitsune/flagit/views.py
```diff
@@ -330,6 +342,34 @@ def moderate_content(request):
     )
 
 
+@group_required(settings.STAFF_GROUP)
```
escattone, line 339: Since this view is only for members of the staff group, should we also only show support tickets within the `moderate_content` view to members of the staff group as well? Currently in that view we check for the `customercare.view_supportticket` perm.

akatsoulas: Good call! After talking through slack another PR will follow to address this.

PR #7104, opened 2025-12-02
change access to moderate queue

escattone:
mozilla/sumo#2688


akatsoulas:
@denyshon that's true but this is the `/moderate` queue, not the `/flagged` queue. To my knowledge the `/moderate` queue has been deprecated so the idea here is to re-use it, at least for now, to remove the customercare type tickets from all the other queues

denyshon:
Oh, you're right, sorry for being inattentive

Issue #4349, opened 2020-04-28
L10n/L12y: Expose "Firefox Account" in full sentence for full localizers


petercpg:
We have a sandwich string in
https://github.com/mozilla/kitsune/blob/930dd6310ae12c052e43b60b8db5601dedb05604/kitsune/sumo/jinja2/handlers/403.html#L13

In Chinese (Taiwan) locale, the translation of the full line shall be:
> 請先試著使用 Firefox Accounts 登入 / 註冊
> Please try using <a>Firefox Accounts</a> to login /register

Note that the first slice of sandwich is broken again into two parts in valid translation and the HTML in the center is out of localizer's control.

Usually we could move "login /register" into the last slice as a workaround in translation, but this is not possible as in this case, "Firefox Accounts" is reused in `kitsune/sumo/jinja2/base.html:288` which may break other occurrences.

Changes like https://github.com/mozilla/kitsune/commit/32dbed0fb9a2a5a4f9487a7e32d8a82efde09eb2#diff-83a46d54cc7cba6ca9b1d31f9d539614 should solve such issue.

petercpg:
Similar issue happens within this file: https://github.com/mozilla/kitsune/blob/master/kitsune/users/jinja2/users/includes/login-fxa.html , too. We should avoid these cases.

PR #7107, opened 2025-12-03
Fix a split string in kitsune\sumo\jinja2\handlers\403.html & Add l10n comments

denyshon:
- Fix a split string in kitsune\sumo\jinja2\handlers\403.html to ensure proper l10n
- Add l10n comments for strings in kitsune\sumo\jinja2\handlers\403.html

Resolves [#4349](https://github.com/mozilla/kitsune/issues/4349).

### DIFF
--- a/kitsune/sumo/jinja2/handlers/403.html
+++ b/kitsune/sumo/jinja2/handlers/403.html
@@ -1,15 +1,22 @@
 {% extends "base.html" %}
+{# L10n: The 403 Access denied page title. #}
 {% set title = _('Access denied') %}
 
 {% block content %}
   <article id="error-page" class="main sumo-page-section">
     <h1 class="sumo-page-heading">{{ title }}</h1>
     <div class="center-on-mobile">
+      {# L10n: A message displayed on the 403 Access denied page. #}
       <p>{{ _('You do not have permission to access this page.') }}</p>
 
       {% if not user.is_authenticated %}
         <p>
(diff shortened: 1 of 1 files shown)


kitsune/sumo/jinja2/handlers/403.html
```diff
@@ -1,15 +1,23 @@
...
       <p>{{ _('You do not have permission to access this page.') }}</p>
 
       {% if not user.is_authenticated %}
         <p>
-          {{ _('Please try to log in / sign up first with') }} <a rel="nofollow" href="{{ url('users.fxa_authentication_init') }}">{{ _('Mozilla accounts') }}</a>.
+          {# L10n: Deprecated. A message displayed for non-authenticated users on the 403 Access denied page. #}
+          {% set deprecated_403 = _('Please try to log in / sign up first with') %}
+          {% set fxa_url = url('users.fxa_authentication_init') %}
+          {# L10n: A message displayed for non-authenticated users on the 403 Access denied page. #}
+          {% trans a_open='<a rel="nofollow" href="{{ fxa_url }}">'|safe, a_close='</a>'|safe %}
```
akatsoulas: I believe the `fxa_url` here will not work. Probably we need something similar to the [profile](https://github.com/mozilla/kitsune/blob/main/kitsune/users/jinja2/users/profile.html#L49)

denyshon: @akatsoulas Oops, thank you for the catch! I've fixed that (the test failure is unrelated)

PR #7118, opened 2025-12-12
Correct the message when deleting an only revision

denyshon:
Resolves [#2706](https://github.com/mozilla/sumo/issues/2706).


escattone (approved):
Thanks @denyshon!

PR #7123, opened 2025-12-16
Hybrid flow models

akatsoulas:


kitsune/customercare/forms.py
```diff
@@ -51,23 +49,42 @@ def __init__(self, *args, product, user=None, **kwargs):
...
-        for category in self.product_categories:
-            category_choices.append((category["slug"], category["topic"]))
+            topics = zendesk_config.topics.filter(loginless_only=is_loginless).order_by(
+                "display_order"
+            )
+
+            if not topics.exists() and not is_loginless:
+                topics = zendesk_config.topics.filter(loginless_only=False).order_by(
+                    "display_order"
+                )
```
escattone, line 76: I think this block should be deleted? If the `topics.exists()` is `False` and `is_loginless` is `False`, it recreates the same `topics` query as before, the one for which `topics.exists()` is `False`.

escattone (approved):
r+wc

PR #7124, opened 2025-12-17
Remove fallback topics for loginless

akatsoulas:


akatsoulas:
This is an omission from the review in #7123

PR #7136, opened 2025-12-29
remove unused docs dependencies

escattone:
While considering #7097, I realized that we're no longer using Sphinx for our docs, so this PR performs two clean-up tasks:
- Removes the unused Sphinx-related dependencies.
- Fixes the `docs` target in the `Makefile`


pyproject.toml
```diff
@@ -113,10 +113,6 @@ dev = [
     "debugpy>=1.8.0,<2",
     "urlwait~=1.0",
     "testfixtures>=7.2.2,<8",
-    "sphinx-rtd-theme>=1.3.0,<2",
-    "sphinx>=5.3.0,<6",
-    "recommonmark>=0.7.1,<0.8",
-    "sphinxcontrib-httpdomain>=1.8.1,<2",
```
janbrasna, line 119: For some reason while `sphinx` was in dev dependencies, current `mkdocs` is not, and is in the default group instead — should it also be moved to dev extras? Or it's in prod dependencies to be included in the `web` image preinstalled for local/containerized work on the docs?

(For GitHub Pages/Actions deploy, the deps, make targets or containers are not being used — the CI runs its own version of it.)

escattone: Hm, good point. It seems to me that it should be in dev extras.

PR #7137, opened 2025-12-30
update package-lock.json

escattone:
For some reason, the `package-lock.json` file is out of sync. I think an older version of `npm` was used.


janbrasna:
Some weeks ago dependabot started using npm 11 instead of 10 — right in the middle of npm 11.5/11.6 lockfile wrinkles, so that might have introduced these dirty lockfile states, see e.g.: https://github.com/mozilla/bedrock/pull/16906#discussion_r2586619356

I've written up some notes earlier: https://github.com/mozilla/bedrock/pull/16906#pullrequestreview-3530062502

PR #7139, opened 2025-12-30
update stylelint and related packages

escattone:
This updates `stylelint` and its related packages, and also removes the indentation rule, which is no longer supported. If we'd prefer to keep the indentation rule, we can add the `@stylistic/stylelint-plugin`, which provides it.

The `npm run stylelint` command runs without error, but there are plenty of warnings, mostly about empty comments and ordering. This PR doesn't address the warnings, which we can add to this PR or take care of in another PR if desired.

### DIFF
--- a/.stylelintrc
+++ b/.stylelintrc
@@ -1,10 +1,12 @@
 {
   "extends": "stylelint-config-recommended-scss",
   "plugins": [
-    "stylelint-order"
+    "stylelint-order",
+    "@stylistic/stylelint-plugin"
   ],
   "defaultSeverity": "warning",
   "rules": {
+    "@stylistic/indentation": 2,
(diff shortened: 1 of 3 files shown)


.stylelintrc
```diff
@@ -5,7 +5,6 @@
   ],
   "defaultSeverity": "warning",
   "rules": {
-    "indentation": 2,
```
akatsoulas, line 8: the default for web projects used to be 2. I am not sure how strict is this rule these days but I would keep it unless there is a good reason not too

escattone: I added the `@stylistic/stylelint-plugin` package, so we can use its `@stylistic/indentation` rule.

PR #7149, opened 2026-01-05
adjustments for hybrid flow

escattone:
mozilla/sumo#2693

### DIFF
--- a/kitsune/products/managers.py
+++ b/kitsune/products/managers.py
@@ -13,7 +13,7 @@ class ProductManager(NonArchivedManager):
     def with_question_forums(self, language_code: str = ""):
         q_kwargs: dict[str, Any] = {
             "support_configs__is_active": True,
-            "support_configs__forum_config__isnull": False,
+            "support_configs__forum_config__is_active": True,
         }
 
         if language_code:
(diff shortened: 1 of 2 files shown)


kitsune/products/managers.py
```diff
@@ -12,13 +12,14 @@ def get_queryset(self):
 class ProductManager(NonArchivedManager):
     def with_question_forums(self, language_code: str = ""):
         q_kwargs: dict[str, Any] = {
-            "aaq_configs__is_active": True,
+            "support_configs__is_active": True,
+            "support_configs__forum_config__isnull": False,
```
escattone: If the support config is active, I'm assuming that its forum config is considered active even if the forum config's `is_active` is `False`, or, in other words, that a forum config's `is_active` field is overridden by its containing support config's `is_active` field.

akatsoulas: This could be risky behavior. We should not assume any overrides. If a support config is active it's assumed that an option at least is active for the product. If the support config contains a forum config that is not active, then the forum config is not active. Maybe the solution here is to make sure that if the support config is active then it contains at least an active configuration. For this line though probably we need to make sure that both queries are active. 

A different approach would be to reconsider the placement of the active flags. Both approaches though are beyond the scope of this PR. Feel free to open a new issue to track this problem

escattone: https://github.com/mozilla/sumo/issues/2721

PR #7154, opened 2026-01-09
make the support switcher a banner

escattone:
mozilla/sumo#2720

I don't know if this is what we want, but it's an option.

image

image

image

image


escattone:
Closing in favor of #7156

PR #7171, opened 2026-01-13
Allow product exemption from spam classifier

akatsoulas:


kitsune/llm/support/classifiers.py
```diff
@@ -122,12 +122,18 @@ def classify_zendesk_submission(submission: "SupportTicket") -> dict[str, Any]:
...
-
-    base_result = {
-        "spam_result": spam_result,
+    # Check if spam moderation should be skipped for this product
+    skip_spam = False
+    try:
+        config = product.support_configs.get(is_active=True)
+        if config.zendesk_config and config.zendesk_config.skip_spam_moderation:
+            skip_spam = True
+    except Exception:
```
escattone: What do you think about replacing `Exception` with `product.support_configs.model.DoesNotExist` or, if we also import `ProductSupportConfig`,  `ProductSupportConfig.DoesNotExist`? We shouldn't see any other kinds of exceptions, and if we do, we should probably let them be raised?

escattone (approved):
r+wc

PR #7175, opened 2026-01-13
improve performance of two product support methods

escattone:
While working on #7173, I noticed that we could reduce the query count of the `Product.has_ticketing_support` property and the `Product.questions_enabled()` method.

### DIFF
--- a/kitsune/products/models.py
+++ b/kitsune/products/models.py
@@ -97,25 +97,18 @@ class Product(BaseProductTopic):
     @property
     def has_ticketing_support(self):
         """Return boolean if a product has Zendesk ticketing support"""
-        try:
-            config = self.support_configs.get(is_active=True)
-        except ProductSupportConfig.DoesNotExist:
-            return False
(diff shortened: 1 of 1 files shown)


kitsune/products/models.py
```diff
@@ -97,25 +97,18 @@ def image_alternate_url(self):
     @property
     def has_ticketing_support(self):
         """Return boolean if a product has Zendesk ticketing support"""
-        try:
-            config = self.support_configs.get(is_active=True)
-        except ProductSupportConfig.DoesNotExist:
-            return False
-        else:
-            return config.enable_zendesk_support
+        return self.support_configs.filter(
```
akatsoulas, line 100: nitpicking: this query and the original one perform both a single query. The performance gain is negligible here and comes from the fetching and object initialization.

escattone: It won't perform another query for the `zendesk_config` object?

PR #7177, opened 2026-01-14
Nested groups

akatsoulas:


kitsune/groups/models.py
```diff
@@ -42,3 +75,36 @@ def save(self, *args, **kwargs):
         self.information_html = wiki_to_html(self.information)
 
         super().save(*args, **kwargs)
+
+    def can_user_edit(self, user):
+        """Check if user can edit this group.
+
+        If no leaders exist, delegates to parent leaders.
+        """
+        if user.is_superuser or user in self.leaders.all():
```
escattone: It would be more efficient to use the following for the leaders check:
```python
self.leaders.filter(pk=user.pk).exists()
```

kitsune/groups/models.py
```diff
@@ -42,3 +75,36 @@ def save(self, *args, **kwargs):
...
+                return parent.can_user_edit(user)
+        return False
+
+    def can_user_manage_members(self, user):
+        """Check if user can add/remove members."""
+        return self.can_user_edit(user)
+
+    def can_user_manage_leaders(self, user):
+        """Check if user can add/remove leaders."""
+        return user.is_superuser or user in self.leaders.all()
```
escattone: Same here:
```python
self.leaders.filter(pk=user.pk).exists()
```

kitsune/groups/models.py
```diff
@@ -42,3 +75,36 @@ def save(self, *args, **kwargs):
...
+
+    def can_user_see(self, user):
+        """Check if user can see this group based on visibility setting."""
+        if self.visibility == self.Visibility.PRIVATE:
+            return user in self.group.user_set.all()
```
escattone: More efficient to do:
```python
return self.group.user_set.filter(pk=user.pk).exists()
```

kitsune/groups/models.py
```diff
@@ -42,3 +75,36 @@ def save(self, *args, **kwargs):
         self.information_html = wiki_to_html(self.information)
 
         super().save(*args, **kwargs)
+
+    def can_user_edit(self, user):
```
escattone: Nit. What do you think of removing the `user_` part of each of these `can_user_*` methods? So, for example,`can_edit(self, user)` instead of `can_user_edit(self, user)`, and so on?

kitsune/groups/models.py
```diff
@@ -42,3 +75,36 @@ def save(self, *args, **kwargs):
...
+        """Check if user can see this group based on visibility setting."""
+        if self.visibility == self.Visibility.PRIVATE:
+            return user in self.group.user_set.all()
+        if self.visibility == self.Visibility.PUBLIC:
+            return True
+        return self.visible_to_groups.filter(pk__in=user.groups.all()).exists()
+
+    def get_visible_children(self, user):
+        """Get child groups visible to this user."""
+        return [child for child in self.get_children() if child.can_user_see(user)]
```
escattone: I think this is an N+1 query situation. I think `get_children()` returns a queryset? If so, could we somehow build on that queryset to get the visible children in a single query? I haven't tried the following, but maybe something like it?

```python
if user.is_superuser:
    return list(self.get_children())

public = Q(visibility=self.Visibility.PUBLIC)
private = Q(visibility=self.Visibility.PRIVATE) & Q(group__user_set__pk=user.pk)
moderated = Q(visibility=self.Visibility.MODERATED) & Q(visible_to_groups__in=user.groups.all())
return list(self.get_children().filter(public | private | moderated))
```

Also, do we always want to return a list or should we return a queryset instead just in case we'd like to further filter the queryset?

kitsune/groups/tests/__init__.py
```diff
@@ -9,3 +9,7 @@ class Meta:
         model = GroupProfile
 
     group = factory.SubFactory(GroupFactory)
+
+    @classmethod
+    def _create(cls, model_class, *args, **kwargs):
```
escattone, line 14: Do we need this class method?

kitsune/groups/models.py
```diff
@@ -42,3 +75,36 @@ def save(self, *args, **kwargs):
...
+
+    def can_user_manage_members(self, user):
+        """Check if user can add/remove members."""
+        return self.can_user_edit(user)
+
+    def can_user_manage_leaders(self, user):
+        """Check if user can add/remove leaders."""
+        return user.is_superuser or user in self.leaders.all()
+
+    def can_user_see(self, user):
```
escattone: I just realized that this method doesn't handle the case when the user is a superuser. I guess that should be checked right away, and return `True` if so.

escattone (commented):
Nice work on the migrations. Those were tricky.

kitsune/groups/tests/__init__.py
akatsoulas: yes, without it we will get an integrity error from the db because the internal fields of treabeard that are added to the class will be null

PR #7179, opened 2026-01-15
move switcher to sidebar

escattone:
mozilla/sumo#2720

image

image

image

image

image

image


akatsoulas (approved):
This looks good! Let's just center the button in the widget and I think this is good to go. We can iterate later if needed. r+wc

PR #7189, opened 2026-01-20
Introduce group isolation.

akatsoulas:
* Optmize manager and model methods
* Add tests
* Add group documentation


kitsune/groups/models.py
```diff
@@ -72,81 +76,130 @@ def __str__(self):
     def get_absolute_url(self):
         return reverse("groups.profile", args=[self.slug])
 
+    def is_root_node(self):
+        """Check if this is a root node using path length."""
+        return len(self.path) == self.steplen
```
escattone: The equivalent of this method is already provided by the `MP_Node` sub-class as the `is_root` method.

kitsune/groups/models.py
```diff
@@ -72,81 +76,130 @@ def __str__(self):
...
-            )
-            .values_list("is_leader", flat=True)
-            .first()
-        )
+        if self.leaders.filter(pk=user.pk).exists():
+            return True
+
+        if not self.is_root_node():
+            root = self.get_root()
+            if root and root.leaders.filter(pk=user.pk).exists():
```
escattone: I think the `get_root()` method will either return the root node or raise an exception, so I don't think it makes sense to check if root is truthy.

docs/groups.md
```diff
@@ -0,0 +1,222 @@
...
+**With Isolation Enabled (default):**
+```
+Root (member: Alice, Bob; moderator: Mike)
+├── SubA (member: Alice, Charlie)
+└── SubB (member: Bob)
+
+Mike (root moderator):     Sees Root, SubA, SubB (entire tree)
+Alice (root member):       Sees Root, SubA, SubB (entire tree, view-only)
+Bob (root member):         Sees Root, SubA, SubB (entire tree, view-only)
+Charlie (node member):     Sees Root, SubA ONLY (sibling isolation - no SubB)
```
escattone, line 78: Does Charlie see `Root`? That doesn't seem right to me.

I think we need a `**With Isolation Disabled:**` example. If isolation is disabled, does that mean Charlie can see `SubB`?

Also, it's not clear to me how isolation affects `MODERATED` groups.

docs/groups.md
```diff
@@ -0,0 +1,222 @@
...
+
+### PUBLIC
+- **Who can see**: Everyone (anonymous and authenticated users)
+- **Isolation**: None - all PUBLIC groups are globally visible
+- **Behavior**: No restrictions on visibility
+
+### PRIVATE
+- **Who can see**: Depends on `isolation_enabled` setting (default: True)
+- **With isolation enabled**: Members/moderators see only their hierarchy (sibling isolation)
+- **With isolation disabled**: Members see entire tree
```
escattone, line 55: With isolation disabled, what does `Members see entire tree` mean? For example, does it mean that members now also see their sibling groups and the groups below each of their sibling groups?

kitsune/groups/models.py
```diff
@@ -72,81 +76,130 @@ def __str__(self):
...
-                # Moderators (including delegated) can see private groups
-                return self.can_moderate_group(user)
-            case _:
-                return False
+        if self.visibility == self.Visibility.MODERATED:
+            return self.visible_to_groups.filter(user=user).exists()
+
+        if self.visibility == self.Visibility.PRIVATE:
+            return (self.group.user_set.filter(pk=user.pk).exists() or
+                    self.can_moderate_group(user))
```
escattone: Shouldn't the `isolation_enabled` setting affect the return value for these two cases?

Also, it seems like this view isn't handling the cases where the user is a member/leader of an ancestor group of the current group.

It'd be nice to make this more DRY and not repeat the logic of the `GroupProfileManager.visible` method.

kitsune/groups/models.py
```diff
@@ -72,81 +76,130 @@ def __str__(self):
...
+        Only root moderators can delete subgroups.
+        """
+        if not (user and user.is_authenticated):
+            return False
+
+        if user.is_superuser:
+            return True
+
+        root = self.get_root()
+        return root.leaders.filter(pk=user.pk).exists() if root else False
```
escattone: Likewise here, I don't think it makes sense to check the truthiness of `root`.

kitsune/groups/managers.py
escattone, line 1: I think we could rewrite the filtering within the `visible` method without running extra queries, but in the meantime, I have some questions below.

kitsune/groups/managers.py
```diff
@@ -1,37 +1,82 @@
...
+        root_nodes = self.filter(path__in=root_paths).prefetch_related(
+            "group__user_set", "leaders"
         )
 
-        return queryset.filter(public | private | moderated).distinct()
+        full_access_paths = set()
+        for root in root_nodes:
+            is_root_member = root.group.user_set.filter(pk=user.pk).exists()
+            is_root_moderator = root.leaders.filter(pk=user.pk).exists()
+            if is_root_member or is_root_moderator or not root.isolation_enabled:
```
escattone, line 58: If the isolation of the root group is not enabled, it seems like this means that members and moderators of lower-level groups -- even if they're not members or moderators of the root group -- are allowed to see all of their ancestor groups, which doesn't seem right to me.

kitsune/groups/managers.py
```diff
@@ -1,37 +1,82 @@
...
+
+        for profile in all_user_profiles:
+            root_path = profile.path[:steplen]
+            if root_path not in full_access_paths:
+                ancestor_paths = [
+                    profile.path[:i] for i in range(steplen, len(profile.path), steplen)
+                ]
+
+                visibility_q |= Q(path__in=ancestor_paths)
+                visibility_q |= Q(path__startswith=profile.path)
```
escattone, line 74: This chunk of code doesn't seem right to me either. It seems to be saying that members and moderators of groups should be able to see not only their own groups, but all of the groups above their own groups (all of the ancestor groups of their own groups). Shouldn't it be the opposite, where members and moderators of groups can see the child groups below their own groups?

escattone (commented):
I'm still confused about the following in terms of visibility:
- What should happen if isolation is disabled for a private hierarchy?
- What should happen if isolation is disabled for a moderated hierarchy?
- Do members/leaders of non-root groups that have child groups see all of those child groups?

docs/groups.md
akatsoulas: Yes. The whole tree. From top to bottom. The depth of the tree is an organizational structure. If a group needs more that 2 levels, it should be just for organizing members.
* With isolation enabled, members of a group can see: group + ancestors + descendants 
* With isolation disabled, members of a group can see the whole tree: group + siblings + ancestors + descendants

kitsune/groups/managers.py
akatsoulas: Yes, there's visibility for the whole tree. Why do you believe this is not right? For context and navigation it makes sense to allow members of a group to see where they are in the hierarchy.
Permissions and access is what should flow one way, in this case only to descendants.

docs/groups.md
akatsoulas: The same way it affects `PRIVATE` groups. It's just allows visibility only on the branch of the tree.

kitsune/groups/managers.py
akatsoulas: No, see also my previous answers. This is not about what permissions each group allows and how those permissions are inherited from descendants. It's just a simple hierarchy.

docs/groups.md
```diff
@@ -0,0 +1,222 @@
...
+Groups have three visibility levels that control discovery and access.
+
+### PUBLIC
+- **Who can see**: Everyone (anonymous and authenticated users)
+- **Isolation**: None - all PUBLIC groups are globally visible
+- **Behavior**: No restrictions on visibility
+
+### PRIVATE
+- **Who can see**: Depends on `isolation_enabled` setting (default: True)
+- **With isolation enabled**: Members/moderators see only their hierarchy (sibling isolation)
```
escattone, line 54: For me, it would be more clear to say "Members/moderators see only their ancestors and their own subtree (sibling isolation)". What do you think of that?

kitsune/groups/managers.py
```diff
@@ -1,37 +1,82 @@
...
+        all_user_profiles = self.filter(Q(group__user=user) | Q(leaders=user)).distinct()
 
-        moderated = Q(visibility=GroupProfile.Visibility.MODERATED) & Q(
-            visible_to_groups__user=user
+        if not all_user_profiles:
+            return queryset.filter(visibility=self.model.Visibility.PUBLIC)
+
+        root_paths = {p.path[:steplen] for p in all_user_profiles}
+
+        root_nodes = self.filter(path__in=root_paths).prefetch_related(
```
escattone, line 50: I don't think the `prefetch_related` here will help with the `is_root_member` and `is_root_moderator` expressions below, because I think the `.filter()` on each of those querysets will initiate a fresh query?

escattone (approved):
r+wc

docs/groups.md
akatsoulas: The see both their ancestors and their descendants. Basically their whole branch/hierarchy so I think this is more accurate

PR #7201, opened 2026-01-22
Group management

akatsoulas:


kitsune/groups/admin.py
```diff
@@ -1,10 +1,14 @@
 from django.contrib import admin
+from treebeard.admin import TreeAdmin
 
 from kitsune.groups.models import GroupProfile
 
 
-class GroupProfileAdmin(admin.ModelAdmin):
+class GroupProfileAdmin(TreeAdmin):
```
escattone, line 8: Several things here:
- We need to add `form = movenodeform_factory(GroupProfile)` to this `GroupProfileAdmin` class in order for `django-treebeard` to create the proper form needed for the admin interface. The import for that is `from treebeard.forms import movenodeform_factory`.
- Currently, the `leaders` field is required, but should be optional. I think we're going to have to set `blank=True` on the `GroupProfile.leaders` field and create a new DB migration along with that.
- Do we also want to allow the `information` field to be blank?
- I'd suggest using `"depth"` instead of `"get_depth"`. I think it looks better as a column name in the list display, and it's already directly available as a field.

kitsune/groups/jinja2/groups/list.html
```diff
@@ -13,7 +13,12 @@ <h1 class="sumo-page-heading">{{ title }}</h1>
 
     <ul class="mzp-u-list-styled">
       {% for profile in groups %}
-        <li><a href="{{ profile.get_absolute_url() }}">{{ profile.group.name }}</a></li>
+        <li style="margin-left: {{ profile.get_depth() * 20 }}px;">
```
escattone: - The inline style will be blocked by our CSP in production.
- Also, it seems misleading to "indent" using the dynamic `margin-left` based on depth, since it might indicate to users that the profile is a sub-group of the group immediately above, when there's no guarantee that the group immediately above is its parent.
- Would it make sense to somehow show the groups as a tree?

kitsune/groups/jinja2/groups/profile.html
escattone, line 1: Nice! This looks so much better than what we currently have. 🎉

PR #7208, opened 2026-01-27
Upgrade to Django 5.2 LTS

akatsoulas:
* Enable native db pooling


kitsune/search/views.py
```diff
@@ -26,7 +26,7 @@
 def cache_control(resp, cache_period):
     """Inserts cache/expires headers"""
     resp["Cache-Control"] = "max-age=%s" % (cache_period * 60)
-    resp["Expires"] = (datetime.utcnow() + timedelta(minutes=cache_period)).strftime(
+    resp["Expires"] = (datetime.now(UTC) + timedelta(minutes=cache_period)).strftime(
```
escattone: What do you think about standardizing on `django.utils.timezone.now()` instead?

kitsune/users/middleware.py
```diff
@@ -30,11 +30,19 @@ def process_request(self, request):
...
+                    first_seen = None
+
+                if first_seen:
+                    change_time = user.profile.fxa_password_change
+                    if change_time and change_time > first_seen:
+                        logout(request)
+                        return HttpResponseRedirect(reverse("home"))
             else:
-                request.session["first_seen"] = datetime.utcnow()
+                request.session["first_seen"] = datetime.now(UTC).isoformat()
```
escattone: Same here, perhaps standardize on `django.utils.timezone.now()` instead?

kitsune/settings.py
```diff
@@ -55,22 +55,33 @@ def path(*parts):
...
-    DATABASES["default"]["CONN_MAX_AGE"] = DB_CONN_MAX_AGE
+    # Django 5.1+ Native Connection Pooling with psycopg3
+    # CRITICAL: CONN_MAX_AGE must be 0 when using pool
+    DATABASES["default"]["CONN_MAX_AGE"] = 0
+    DATABASES["default"]["CONN_HEALTH_CHECKS"] = config(
+        "CONN_HEALTH_CHECKS", default=True, cast=bool
+    )
+
+    # Configure connection pooling to prevent exhaustion
+    # Max connections capacity estimated with the formula: 75 pods x 3 workers x 3 max = 675 connections
```
escattone: Just thinking out loud: Our current HPA maximum is 100 web replicas and our current DB connection limit is 800. We'll have to leave room for DB connections within our Celery workers, which can scale up to a total of 16 workers (15 plus the one email worker). Maybe we need to lower the HPA maximum web replicas?

wsgi/config.py
```diff
@@ -9,13 +9,12 @@
 accesslog = "-"
 errorlog = "-"
 loglevel = getenv("WSGI_LOG_LEVEL", "info")
-worker_class = getenv("GUNICORN_WORKER_CLASS", "gevent")
+worker_class = getenv("GUNICORN_WORKER_CLASS", "sync")
```
escattone, line 12: I was really worried about this, that it would explode the number of web pods that we'd need to scale up to, but, after looking into the `gevent` monkey-patching, wow, you're right, our Postgres DB connections aren't using the `gevent` monkey-patched sockets at all (because psycopg doesn't use the Python sockets)! 

I'm still a bit concerned, but let's give it a try.

kitsune/settings.py
akatsoulas: In order to be on the safe side, we should bring down the number of replicas to 85

escattone: I think our cache requests use Python sockets, so they're currently using the `gevent` monkey-patched sockets and are non-blocking. Since we're using the cache within some key endpoints, like our KB articles (and I think within the questions views?), I suspect that moving to `sync` workers will cause an increase in the number of web pods needed to serve our traffic.

kitsune/settings.py
```diff
@@ -1378,11 +1393,12 @@ def setup_logger(name, formatter_str, level=logging.DEBUG):
...
-DJANGO_CELERY_BEAT_TZ_AWARE = False  # This should match USE_TZ.
+DJANGO_CELERY_BEAT_TZ_AWARE = config("USE_TZ", default=True, cast=bool)
 CELERY_BEAT_SCHEDULER = "django_celery_beat.schedulers:DatabaseScheduler"
 if READ_ONLY:
     CELERY_BEAT_SCHEDULE = {}
 else:
     CELERY_BEAT_SCHEDULE = PERIODIC_TASKS_ALL
     if not STAGE:
         CELERY_BEAT_SCHEDULE |= PERIODIC_TASKS_PRODUCTION_ONLY
+        CELERY_BEAT_SCHEDULE |= PERIODIC_TASKS_PRODUCTION_ONLY
```
escattone: Typo. Looks like this line was duplicated.

kitsune/search/base.py
```diff
@@ -163,14 +163,14 @@ def prepare(cls, instance, parent_id=None, **kwargs):
...
-                        value = timezone.make_aware(value).astimezone(timezone.utc)
+                        value = timezone.make_aware(value).astimezone(UTC)
 
                     if not parent_id:
                         setattr(obj, f, value)
         else:
             obj.es_discard_doc = "unindex_me"
 
-        obj.indexed_on = datetime.now(timezone.utc)
+        obj.indexed_on = datetime.now(UTC)
```
escattone: Nit. Should we use `timezone.now()` here as well?

escattone (commented):
- Don't we also need to change all of the other models with `DateTimeField` fields with `default=datetime.now` -- like `Announcement.created`, `Announcement.show_after`, `FlaggedObject.created`, `FlaggedObject.handled`, etc. -- to use `default=timezone.now` and then create migrations for those changes?
- I assume we also need to upgrade our ES servers to 9.2.4, or will our current 9.0.3 version still run fine with these changes?

escattone (commented):
### The following fields still need `default=timezone.now` and migrations:
- In `kitsune/forums/models.py`:
    - [ ] `Thread.created`
    - [ ] `Post.created`
    - [ ] `Post.updated`
- In `kitsune/kbforums/models.py`:
    - [ ] `Thread.created`
    - [ ] `Post.created`
    - [ ] `Post.updated`

### The following files need `datetime.now` replaced with `timezone.now`
- [ ] `kitsune/community/utils.py`
- [ ] `kitsune/community/tests/test_cron.py`
- [ ] `kitsune/community/tests/test_utils.py`
- [ ] `kitsune/dashboards/tests/test_readouts.py`
- [ ] `kitsune/forums/tests/__init__.py`
- [ ] `kitsune/forums/tests/test_feeds.py`
- [ ] `kitsune/forums/tests/test_models.py`
- [ ] `kitsune/forums/tests/test_posts.py`
- [ ] `kitsune/kbforums/tests/test_models.py`
- [ ] `kitsune/kpi/api.py`
- [ ] `kitsune/kpi/tests/test_api.py`
- [ ] `kitsune/questions/api.py`
- [ ] `kitsune/questions/handlers.py`
- [ ] `kitsune/questions/managers.py`
- [ ] `kitsune/questions/tasks.py`
- [ ] `kitsune/questions/views.py`
- [ ] `kitsune/kbforums/views.py`
- [ ] `kitsune/forums/views.py`
- [ ] `kitsune/questions/tests/__init__.py`
- [ ] `kitsune/questions/tests/test_api.py
- [ ] `kitsune/questions/tests/test_feeds.py`
- [ ] `kitsune/questions/tests/test_handlers.py`
- [ ] `kitsune/questions/tests/test_tasks.py`
- [ ] `kitsune/questions/tests/test_templates.py`
- [ ] `kitsune/search/base.py`
- [ ] `kitsune/search/models.py`
- [ ] `kitsune/search/search.py`
- [ ] `kitsune/search/management/commands/es_init.py`
- [ ] `kitsune/sumo/context_processors.py`
- [ ] `kitsune/sumo/tasks.py`
- [ ] `kitsune/sumo/utils.py`
- [ ] `kitsune/sumo/management/commands/enqueue_lag_monitor_task.py`
- [ ] `kitsune/sumo/management/commands/extract.py`
- [ ] `kitsune/sumo/templatetags/jinja_helpers.py`
- [ ] `kitsune/users/api.py`
- [ ] `kitsune/users/tasks.py`
- [ ] `kitsune/users/utils.py`
- [ ] `kitsune/users/tests/test_api.py`
- [ ] `kitsune/wiki/content_managers.py`
- [ ] `kitsune/wiki/models.py`
- [ ] `kitsune/wiki/sampledata.py`
- [ ] `kitsune/wiki/views.py`
- [ ] `kitsune/wiki/tests/__init__.py`
- [ ] `kitsune/wiki/tests/test_models.py`
- [ ] `kitsune/wiki/tests/test_services.py`
- [ ] `kitsune/wiki/tests/test_tasks.py`
- [ ] `kitsune/wiki/tests/test_templates.py`
- [ ] `scripts/contributor_counts.py`

escattone (approved):
🎉

PR #7211, opened 2026-01-27
save images in view similarly to admin

escattone:
mozilla/sumo#2748

### DIFF
--- a/kitsune/groups/views.py
+++ b/kitsune/groups/views.py
@@ -13,7 +13,7 @@ from kitsune.groups.forms import AddUserForm, GroupAvatarForm, GroupProfileForm
 from kitsune.groups.models import GroupProfile
 from kitsune.sumo.urlresolvers import reverse
 from kitsune.sumo.utils import get_next_url, paginate
-from kitsune.upload.tasks import _create_image_thumbnail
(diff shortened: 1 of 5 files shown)


kitsune/groups/views.py
```diff
@@ -103,12 +101,8 @@ def edit_avatar(request, group_slug):
         if old_avatar_path:
             default_storage.delete(old_avatar_path)
 
-        content = _create_image_thumbnail(
-            form.instance.avatar.file, settings.AVATAR_SIZE, pad=True
```
akatsoulas, line 107: Shouldn't we apply this in the admin interface instead? There is a good chance that users can upload massive images here.

escattone: We already limit the size of the avatar upload to `MAX_AVATAR_FILE_SIZE` (1MB), so this won't change that.

akatsoulas: why are we dropping the thumbnails? Reversing this, 1MB without a thumbnail is too small

escattone: Applying this approach in the admin instead but also removing the `pad=True` so we avoid the problem of avatars not filling their circle on the profile page.

PR #7217, opened 2026-01-29
Ensure root node always has a moderator

akatsoulas:


kitsune/groups/models.py
```diff
@@ -177,3 +177,18 @@ def get_visible_children(self, user):
         """Return child groups visible to this user."""
         children = self.get_children()
         return self.__class__.objects.visible(user).filter(pk__in=children)
+
+    def can_remove_leader(self, user):
```
escattone: The `user` argument isn't used within this method. Is it there for consistency with the other `can_*` methods -- which would mean the `user` is the user trying to remove a leader not the user that is going to be removed -- or is it supposed to be the user that is going to be removed as a leader?

kitsune/groups/models.py
```diff
@@ -177,3 +177,18 @@ def get_visible_children(self, user):
...
+        """
+        Check if a leader can be removed from this group.
+
+        Rules:
+        - Root groups must always have at least one leader
+        - Subgroups can have no leaders (they inherit from root)
+
+        Returns True if the user can be removed as a leader, False otherwise.
```
escattone: This comment suggests that the `user` argument is the user that is going to be removed as a leader, not the user trying to remove a leader, but that's not consistent with the other `can_*` methods, so if that's the case, maybe we should rename this method and/or drop the `user` argument altogether since it's not used?

kitsune/groups/views.py
```diff
@@ -16,6 +16,32 @@
...
+    Remove a member/leader from a group with validation.
+
+    Handles both regular members and leaders, with appropriate validation
+    and error messages for the last leader scenario.
+
+    Returns True if removed successfully, False if validation failed.
+    """
+    is_leader = profile.leaders.filter(pk=user.pk).exists()
+
+    if is_leader and not profile.can_remove_leader(user):
```
escattone: See comments above regarding the `user` passed to `profile.can_remove_leader(user)`.

PR #7218, opened 2026-01-29
Update visibilty for descendants

akatsoulas:


kitsune/groups/admin.py
```diff
@@ -39,6 +41,23 @@ def get_form(self, request, obj=None, **kwargs):
                     "Changing this will update all descendants in the tree."
                 )
 
+        if "visible_to_groups" in form.base_fields:
+            if obj and not obj.is_root():
+                parent = obj.get_parent()
```
escattone, line 46: Since `visibility` and `visible_to_groups` are determined by the root, I'm thinking maybe for both we should, instead of the parent, use `root = obj.get_root()` and use the root's groups? Effectively it ends up the same, but just thinking it might be more consistent.

escattone (approved):
r+wc

PR #7159, opened 2026-01-11
Add SUMO rules mentions & Improve AAQ step 3

denyshon:
Rebased [#7114](https://github.com/mozilla/kitsune/pull/7114).

### Changes
- Add SUMO rules mentions when creating/editing a question, an answer, a forum thread, a forum post, a KB forum thread or a KB forum post ([#2196](https://github.com/mozilla/sumo/issues/2196)).
- In AAQ step 3, replace "Subject:" with "Summarize your question" and add help text for this field.
- In AAQ step 3, display the Troubleshooting Info help text above the Share Data button, right after the field name.
- In AAQ step 3, add the "tb_version" extra field ([#2536](https://github.com/mozilla/sumo/issues/2536)).
- In AAQ step 3, autofill the OS version field for Thunderbird.
- Fix the issue with the Share Data button not working when editing a question ([#1642](https://github.com/mozilla/sumo/issues/1642)).
- Do not display the Common responses button when editing a question.
- Improve accessibility.
- Use "Windows 10/11" when there is not enough info to identify the exact Windows NT 10.0 version ([#2700](https://github.com/mozilla/sumo/issues/2700)).
- Add l10n comments.
- Improve i18n, internationalize some alt strings
- Move repeated Helpful Tip code to a macro
- Remove an unused variable from forms.py

Resolves [#2196](https://github.com/mozilla/sumo/issues/2196), [#2536](https://github.com/mozilla/sumo/issues/2536), [#1642](https://github.com/mozilla/sumo/issues/1642), [#2700](https://github.com/mozilla/sumo/issues/2700).

### Screenshots

<details>
<summary>AAQ step 3 (Firefox): [click to expand]</summary>
image_2025-12-08_17-48-27
</details>

<details>
<summary>AAQ step 3 (Firefox) after clicking the Share Data button, in case we failed to pull the data: [click to expand]</summary>
image_2025-12-08_17-50-11
</details>

<details>
<summary>AAQ step 3 (Thunderbird): [click to expand]</summary>
image_2025-12-08_17-52-38
</details>

<details>
<summary>An example of how the editor looks on the updated surfaces (except for questions): [click to expand]</summary>
image_2025-12-08_17-58-57
</details>

_Note: Some icons in the screenshots may be missing because they were taken on a test setup._

### TODO [admin]
Once this PR is merged, the following changes have to be made via the admin console:
[...]

PR #7229, opened 2026-02-04
Encourage question posters to provide additional information

herbmarshcentral:
We don't need to keep asking what page they are on, and screenshots are preferable to them describing what is happening on their screen.


denyshon (changes requested):
I believe it's not an appropriate place for such kind of advice. We have a dedicated field for screenshots, and all other specific advices should go to the "How can we help?" help text. Please take a look at #7159 (which IMO should be merged first anyway).

Also, we shouldn't remove the old string from the localization file, creating a new version instead.

herbmarshcentral:
You're right, it would better placed after the "How can we help?" help text. I'll edit this after your mentioned PR has been implemented on the site.

akatsoulas:
Closing due to inactivity, please re-open if there's an update

PR #7234, opened 2026-02-05
establish consistent AAQ routing

escattone:
mozilla/sumo#2785

### DIFF
--- a/kitsune/products/jinja2/products/includes/topic_macros.html
+++ b/kitsune/products/jinja2/products/includes/topic_macros.html
@@ -41,7 +41,8 @@
 
 
 
-{% macro topic_metadata(topics, product=None) %}
+{% macro support_callout(product=None, aaq_context=None) %}
+{% set is_ticketed = (aaq_context.current_support_type == 'zendesk') if aaq_context else False %}
 {% if product and has_support_config and not settings.READ_ONLY %}
   <section class="support-callouts mzp-l-content sumo-page-section--inner">
     <div class="card card--ribbon is-inverse heading-is-one-line">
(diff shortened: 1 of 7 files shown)


kitsune/products/jinja2/products/includes/topic_macros.html
```diff
@@ -56,34 +57,34 @@ <h3 class="card--title">
...
-            {{ _('We\'re here for you! If you haven\'t found a solution after exploring our help articles, you can get in touch with our support team.') }}
+          {% if is_ticketed %}
+            {% if product.slug == "firefox-enterprise" %}
+              {{ _("If you're a Firefox Support for Organizations admin, contact us for technical assistance, deployment guidance, and priority help.") }}
+            {% else %}
+              {{ _("We're here for you! If you haven't found a solution after exploring our help articles, you can get in touch with our support team.") }}
+            {% endif %}
           {% else %}
-            {{ _('We\'re here for you. Post a question to our support forums and get answers from our community of experts.') }}
+            {{ _("We're here for you. Post a question to our support forums and get answers from our community of experts.") }}
```
janbrasna, line 74: Would it be possible that just changing quote escaping somehow breaks the gettext logic?
- https://github.com/mozilla/sumo/issues/2828

(Like it's seen as a different string by gettext, but not by Pontoon so it's not exposed as a new string for localization, while at the same time not applied by gettext from the existing one?)

escattone: Hi @janbrasna! Hm, it looks like Pontoon has an apostrophe character within `We're...` rather than a single quote character, but I don't think this would cause that.

janbrasna: So I checked the string history before posting, seeing the quote character there — but you're right @escattone there ought to be a proper apostrophe in the first place, and according to translation memory it actually was the case:

> quoting

The string https://pontoon.mozilla.org/cs/sumo/LC_MESSAGES/django.po/?search=found+a+solution+after+exploring&string=330811 appears to have a time stamp of "CREATED: February 24, 2026 11:40:09 AM" so it indeed must have been a recent change.

Looking further. Thanks.

janbrasna: Oh, right, wrong PR — it appears to have regressed here:
- https://github.com/mozilla/kitsune/pull/7227

What would be the advice… revert back to the proper apostrophe, in hopes to get the original translations from translation memory back?

PR #7245, opened 2026-02-10
Introduce search reindex api for playwright usage

emilghittasv:
Search related playwright tests are slow due to the need of waiting for the default reindex to kick in. 

This PR tries to solve that issue by creating a secured endpoint which should trigger a reindex when called (only against a given obj_id)

### DIFF
--- a/kitsune/search/api.py
+++ b/kitsune/search/api.py
@@ -1,0 +1,75 @@
+import json
+
+from django.conf import settings
+from django.http import Http404, JsonResponse
+from django.views.decorators.csrf import csrf_exempt
+from django.views.decorators.http import require_POST
+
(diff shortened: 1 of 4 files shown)


kitsune/search/api.py
```diff
@@ -0,0 +1,75 @@
...
+
+    try:
+        data = json.loads(request.body)
+    except json.JSONDecodeError:
+        return JsonResponse({"message": "Invalid JSON payload."}, status=400)
+
+    doc_type_name = data.get("doc_type")
+    obj_id = data.get("obj_id")
+
+    if not doc_type_name or obj_id is None:
```
escattone:
```suggestion
    if not (doc_type_name and obj_id):
```

PR #7246, opened 2026-02-10
gracefully handle legacy user sessions

escattone:
mozilla/sumo#2799

### DIFF
--- a/kitsune/users/middleware.py
+++ b/kitsune/users/middleware.py
@@ -28,22 +28,26 @@ class LogoutInvalidatedSessionsMiddleware(MiddlewareMixin):
     """
 
     def process_request(self, request):
-        user = request.user
+        if not request.user.is_authenticated:
+            return
 
-        if user.is_authenticated:
-            first_seen = request.session.get("first_seen")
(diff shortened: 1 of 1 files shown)


kitsune/users/middleware.py
```diff
@@ -31,14 +31,14 @@ def process_request(self, request):
...
-            if first_seen_str:
-                # Convert ISO string back to datetime for comparison
-                try:
-                    first_seen = datetime.fromisoformat(first_seen_str)
-                except (ValueError, AttributeError):
-                    # Handle legacy pickle-serialized datetime objects or invalid strings
-                    first_seen = None
+            first_seen = request.session.get("first_seen")
+            # Gracefully handle legacy pickle-serialized datetime objects.
+            if first_seen and isinstance(first_seen, str | datetime):
```
akatsoulas: to make this more readable and less nested and shorter you could use a `match` statement 

```python
match first_seen:
    case str():
        first_seen = datetime.isoformat(first_seen)
    case _:
        first_seen = None
```

escattone: I didn't realize until now that the class-pattern matching feature existed. Nice!

PR #7248, opened 2026-02-11
add data migration for updated topic descriptions

escattone:
mozilla/sumo#2797


akatsoulas:
Let's use the admin interface for this. Adding a data migration to manage content just adds noise to the commit history.

PR #7255, opened 2026-02-13
improve support configuration

escattone:
mozilla/sumo#2721

## Highlights
- The `ProductSupportConfig` is now the sole source of **product** support information. It's the only support-related config that can be associated with a product and activated/deactivated.
- The `AAQConfig` is no longer activated/deactivated and no longer directly associated with a product. It's associated with a product only by associating it with a `ProductSupportConfig` instance. A single `AAQConfig` may be used within any number of `ProductSupportConfig` instances.
- A `ProductSupportConfig` instance cannot be activated unless it has at least one support channel (Zendesk or forum), and also, if it has a forum support channel, that forum config must have at least one enabled locale.
- Assumes that `ProductSupportConfig` instances are created only via the admin console (not programmatically), so enforces validation only during `clean`, not `save`.

### DIFF
--- a/kitsune/flagit/views.py
+++ b/kitsune/flagit/views.py
@@ -320,7 +320,11 @@ def moderate_content(request):
             "objects": objects,
             "locale": request.LANGUAGE_CODE,
             "products": [
-                (p.slug, p.title) for p in Product.active.filter(aaq_configs__is_active=True)
+                (p.slug, p.title)
+                for p in Product.active.filter(
+                    support_configs__is_active=True,
+                    support_configs__forum_config__enabled_locales__isnull=False,
+                ).distinct()
             ],
             "selected_product": product_slug,
             "assignees": sorted(
--- a/kitsune/questions/forms.py
+++ b/kitsune/questions/forms.py
@@ -7,9 +7,9 @@ from django.utils.html import strip_tags
 from django.utils.translation import gettext as _
 from django.utils.translation import gettext_lazy as _lazy
 
-from kitsune.products.models import Topic
+from kitsune.products.models import ProductSupportConfig, Topic
 from kitsune.questions.events import QuestionReplyEvent
-from kitsune.questions.models import AAQConfig, Answer, Question
+from kitsune.questions.models import Answer, Question
 from kitsune.questions.utils import remove_pii
 from kitsune.sumo.forms import KitsuneBaseForumForm
 from kitsune.upload.models import ImageAttachment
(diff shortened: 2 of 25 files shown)


kitsune/flagit/views.py
```diff
@@ -320,7 +320,11 @@ def moderate_content(request):
             "objects": objects,
             "locale": request.LANGUAGE_CODE,
             "products": [
-                (p.slug, p.title) for p in Product.active.filter(aaq_configs__is_active=True)
+                (p.slug, p.title)
+                for p in Product.active.filter(
+                    support_configs__is_active=True,
+                    support_configs__forum_config__enabled_locales__isnull=False,
+                ).distinct()
```
akatsoulas, line 327: Why do we need `distinct()` here? We should have only on active support config per product.

kitsune/questions/forms.py
```diff
@@ -101,8 +109,15 @@ def __init__(self, product=None, *args, **kwargs):
...
 
         if product:
-            aaq_config = AAQConfig.objects.get(product=product, is_active=True)
-            extra_fields = aaq_config.extra_fields
+            psc = (
+                ProductSupportConfig.objects.filter(
+                    product=product, is_active=True, forum_config__isnull=False
+                )
+                .select_related("forum_config")
+                .first()
```
akatsoulas: Similarly to the `distinct()` comment above, why do we need `first()` here?

kitsune/flagit/views.py
escattone: The `distinct()` here is necessary due to the join through `forum_config` to the `enabled_locales`, which could introduce duplicate rows without it.

kitsune/questions/forms.py
escattone: Since the result could be an object or None, I used `first()` to avoid doing a `get(...)` within a try-except that handles the `DoesNotExist` case. Would you prefer the latter to avoid the suggestion that there could be more than one product-support config?

escattone: Would you prefer something like the following instead?
```python
try:
    psc = product.support_configs.select_related("forum_config").get(
        is_active=True, forum_config__isnull=False
    )
except ProductSupportConfig.DoesNotExist:
    pass
else:
    extra_fields = psc.forum_config.extra_fields
```

escattone: Hm, I like that better than what I have, so I'll switch to that.

denyshon:
Hi @escattone! In order for l10n comments to be preserved, they must be one line above the corresponding strings. Any new lines will vanish a comment. Please do not add newlines to localizable strings.
I've submitted #7351 to revert the breaking changes.

PR #7256, opened 2026-02-16
Bump playwright to 1.57.0 and update playwright tests to use the reindex api

emilghittasv:


akatsoulas (approved):
For future reference, let's use separate PRs for upgrades and code changes.

PR #7258, opened 2026-02-17
Add tb_version to handle_metadata_tags & Make the version input parser more permissive

### LINKED ISSUE #2810: Autoasssign version tags for Thunderbird
We need to assign version tags to Thunderbird questions based on the `tb_version` field value, just like with Firefox.

### LINKED ISSUE #2811: When autoassigning version tags, make the input parser more permissive
Currently, the only formats that will be handled are:
- `<major_version>.0`
- `<major_version>.<minor_version_maybe_with_another_dot>`
- the above with `a*` and `b*` suffixes
If the input doesn't match the patterns above, no tags will be assigned.

I believe the initial reason for such restrictive behavior is that we never aimed to parse actual user input, because for Firefox, the version field is almost always autofilled. However, version autofill is impossible for Thunderbird, and with #2810 we need to make the parser more permissive. I suggest the following:
- We can safely remove leading and trailing whitespaces.
- A version number is a sequence of digits and dots, starting and ending with a digit.
- Any text following the version number can just be removed (since we don't rely on any suffixes).
- For integers, we need to add ".0" to match them to major versions.
- After that, we can search for the version in our archives just like we do now.

denyshon:
Resolves [#2810](https://github.com/mozilla/sumo/issues/2810), resolves [#2811](https://github.com/mozilla/sumo/issues/2811).

### DIFF
--- a/kitsune/questions/models.py
+++ b/kitsune/questions/models.py
@@ -286,25 +286,34 @@ class Question(AAQBase):
             for tag in product_config.associated_tags.all():
                 tags.append(tag)
 
-        version = self.metadata.get("ff_version", "")
+        version = self.metadata.get("ff_version", "") or self.metadata.get("tb_version", "")
 
-        # Remove the beta (b*), aurora (a2) or nightly (a1) suffix.
-        version = re.split("[a-b]", version)[0]
+        # Remove from the version string anything following the version number.
+        # A version number is a sequence of digits and dots, starting and ending with a digit.
+        versionMatch = re.match(r'\d(?:[\d.]*\d)?', version)
+        version = versionMatch.group(0) if versionMatch else ""
+        # If there are no dots, version is an integer, and we need to add ".0" to find matches among major versions.
+        if '.' not in version:
+            version += ".0"
 
-        dev_releases = product_details.firefox_history_development_releases
+        development_releases = product_details.firefox_history_development_releases or product_details.thunderbird_history_development_releases
+        stability_releases = product_details.firefox_history_stability_releases or product_details.thunderbird_history_stability_releases
+        major_releases = product_details.firefox_history_major_releases or product_details.thunderbird_history_major_releases
 
-        if (
-            version in dev_releases
-            or version in product_details.firefox_history_stability_releases
-            or version in product_details.firefox_history_major_releases
-        ):
-            tags.append("Firefox {}".format(version))
-            tenths = _tenths_version(version)
-            if tenths:
-                tags.append("Firefox {}".format(tenths))
-        elif _has_beta(version, dev_releases):
-            tags.append("Firefox {}".format(version))
-            tags.append("beta")
+        if version:
+            product_name = "Firefox" if self.metadata.get("ff_version", "") else "Thunderbird"
+            if (
+                version in development_releases
+                or version in stability_releases
+                or version in major_releases
+            ):
+                tags.append(product_name + " {}".format(version))
+                tenths = _tenths_version(version)
+                if tenths:
+                    tags.append(product_name + " {}".format(tenths))
+            elif _has_beta(version, development_releases):
+                tags.append(product_name + " {}".format(version))
+                tags.append("beta")
 
         # Add a tag for the OS but only if it already exists as a non-segmentation tag.
         if os := self.metadata.get("os"):
--- a/kitsune/questions/models.py
+++ b/kitsune/questions/models.py
@@ -289,16 +289,23 @@ class Question(AAQBase):
         version = self.metadata.get("ff_version", "") or self.metadata.get("tb_version", "")
 
         # Remove from the version string anything following the version number.
-        # A version number is a sequence of digits and dots, starting and ending with a digit.
-        versionMatch = re.match(r'\d(?:[\d.]*\d)?', version)
-        version = versionMatch.group(0) if versionMatch else ""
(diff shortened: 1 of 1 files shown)


kitsune/questions/models.py
```diff
@@ -286,25 +286,34 @@ def handle_metadata_tags(self, action):
...
-            if tenths:
-                tags.append("Firefox {}".format(tenths))
-        elif _has_beta(version, dev_releases):
-            tags.append("Firefox {}".format(version))
-            tags.append("beta")
+        version = self.metadata.get("ff_version", "") or self.metadata.get("tb_version", "")
+
+        # Remove from the version string anything following the version number.
+        # A version number is a sequence of digits and dots, starting and ending with a digit.
+        versionMatch = re.match(r'\d(?:[\d.]*\d)?', version)
```
escattone: Several comments here:
- Use snake case instead of camel case, so `version_match`.
- Your regular expression will match multiple periods in a row like (`"148..1.01"`). You could avoid that with something like `r"\d+(?:\.\d+)*"` instead.
- In our current code, although `version` removes suffixes like `a1` and `b2`, it keeps suffixes like `rc1` and `rc2`. Your `version` strips the `rc*` suffixes.
- As far as I can tell, your `version` will never match anything in the development releases (`firefox_history_development_releases` and `thunderbird_history_development_releases`), since those only contain versions ending in `a*`, `b*`, and `rc*`. The current `version` can match versions ending in `rc*`.
- I think with your version, release candidate versions will fall through to the `_has_beta` handling, and be labelled as `beta`.
- So, I think your version regular expression will have to include `rc*` suffixes, or maybe just continue using `version = re.split("[a-b]", version)[0]`?

kitsune/questions/models.py
```diff
@@ -286,25 +286,34 @@ def handle_metadata_tags(self, action):
...
+        version = versionMatch.group(0) if versionMatch else ""
+        # If there are no dots, version is an integer, and we need to add ".0" to find matches among major versions.
+        if '.' not in version:
+            version += ".0"
+
+        development_releases = product_details.firefox_history_development_releases or product_details.thunderbird_history_development_releases
+        stability_releases = product_details.firefox_history_stability_releases or product_details.thunderbird_history_stability_releases
+        major_releases = product_details.firefox_history_major_releases or product_details.thunderbird_history_major_releases
```
escattone: These variables will always contain the Firefox releases, because they're always "truthy".

kitsune/questions/models.py
```diff
@@ -286,25 +286,34 @@ def handle_metadata_tags(self, action):
...
+
+        if version:
+            product_name = "Firefox" if self.metadata.get("ff_version", "") else "Thunderbird"
+            if (
+                version in development_releases
+                or version in stability_releases
+                or version in major_releases
+            ):
+                tags.append(product_name + " {}".format(version))
```
escattone: I think it's more readable to use f-strings (`f"{product_name} {version}"`) instead (here and the lines below).

kitsune/questions/models.py
```diff
@@ -286,25 +286,34 @@ def handle_metadata_tags(self, action):
...
-            tags.append("Firefox {}".format(version))
-            tags.append("beta")
+        version = self.metadata.get("ff_version", "") or self.metadata.get("tb_version", "")
+
+        # Remove from the version string anything following the version number.
+        # A version number is a sequence of digits and dots, starting and ending with a digit.
+        versionMatch = re.match(r'\d(?:[\d.]*\d)?', version)
+        version = versionMatch.group(0) if versionMatch else ""
+        # If there are no dots, version is an integer, and we need to add ".0" to find matches among major versions.
+        if '.' not in version:
```
escattone: If the version is empty, this will make it `.0`.

escattone (changes requested):

denyshon:
Hi @escattone, thank you for the review! I've addressed your comments in a follow-up commit (keeping it separate for readability).

My only concern is related to "rc" versions. You are right that the current code respects them, and I've updated my regex string to do the same, but... is there any practical reason to do so? From what I can see, every "rc" entry in the development releases has a corresponding normal entry (without "rc") in the stability/major releases, and modern versions don't have "rc" entries at all. If that's no longer needed, we could remove "rc" from the regex and check `development_releases` only in helper functions (like `_has_beta()`). That won't make the existing "rc" versions be labeled as beta, since we'll just find them (the part preceding "rc) in `stability_releases`/`major_releases`.

escattone (commented):
@denyshon Your question about no longer needing to hande `rc*` led to some research and I realized that I'd like to change some more things. I ended up creating #7443. Could you review that one, and if you like it, we can close this in favor of  #7443?

denyshon:
Thank you @escattone! Your patch is definitely better, so let's stick to it and close this one

wsmwk:
@denyshon does this mean if a user submits version as 150.0b1 that it will be stored as 150.0?

denyshon:
@wsmwk we don't modify user input at all – in Question Details (or, if you click Edit Question, in the version field) you'll see the exact same text that the user has entered
we only apply tags based on the user input (and that's what this PR is all about)
if a user submits version as 150.0b1, tags "Thunderbird 150.0" and "beta" will be applied

denyshon:
generally, I was going to wait until the QA passes before letting you know, but since you are here... you should already be able to test this on stage :D

PR #7265, opened 2026-02-19
Switch ZendeskForm to ModelForm

### LINKED ISSUE #2806: 500 error message when trying to submit a ticket with a long subject for a premium product
**Preconditions**
Sign in to SuMo.

**Steps to reproduce**
1. Go to any premium product contact form (zendesk form)
2. Add a long string inside the Subject field (example: `Lorem ipsum dolor sit amet, consectetur adipiscing elit. Integer vitae justo nec urna facilisis placerat. Curabitur non turpis ac odio varius fermentum. Sed dignissim, nisl at convallis tincidunt, massa lorem lacinia augue, vitae elit nulla sit amet erat.3`)
3. Add something inside the `Tell us more` field.
4. Submit the question.

**Expected behavior**
The Subject field has a character limit, the ticket is not submitted and the user is prompted that the data introduced inside the "Subject" field is too long.

**Actual behavior**
500 error message.

**Desktop:**
 - OS: Windows 11
 - Browser: Firefox

**Additional context**
This issue is reproducible in both `stage` and `prod`

denyshon:
- Switch ZendeskForm from Form to ModelForm to benefit from model validation.
- Specify SupportTicket.description `max_length` based on the Zendesk API.
- Set `blank=True` for SupportTicket.category to prevent a model validation error in case we decide to make this field optional.
- Adjust kitsune.customercare.tests.test_forms.ZendeskFormTests.test_send_with_no_category_selected to ignore a form validation error.

Resolves [#2806](https://github.com/mozilla/sumo/issues/2806).


kitsune/customercare/tests/test_forms.py
```diff
@@ -301,6 +301,9 @@ def test_send_with_no_category_selected(self, mock_task):
                 "description": "Test description",
                 "category": "",  # Empty category
             }
+        # An empty category field will result in a form validation error in send()->save().
```
akatsoulas, line 356: Given the change in the model, the empty category shouldn't raise a validation error

akatsoulas (commented):
Thanks @denyshon for the PR! I added a couple of comment

kitsune/customercare/tests/test_forms.py
denyshon: @akatsoulas It raises a form (not model) validation error, because the field is still required (and from the form perspective cannot be empty)

denyshon:
@akatsoulas Thank you for the review! I've replied to your comment and rebased the patch onto the latest main.

kitsune/customercare/models.py
```diff
@@ -150,8 +143,8 @@ class SupportTicket(ModelBase):
     }
 
     subject = models.CharField(max_length=255)
-    description = models.TextField()
-    category = models.CharField(max_length=255)
+    description = models.TextField(max_length=65563)
+    category = models.CharField(max_length=255, blank=True, default="")
```
akatsoulas, line 154: Given that the form does not allow an empty value, I would argue that the model shouldn't allow this as well.

kitsune/customercare/models.py
```diff
@@ -150,8 +143,8 @@ class SupportTicket(ModelBase):
     }
 
     subject = models.CharField(max_length=255)
-    description = models.TextField()
-    category = models.CharField(max_length=255)
+    description = models.TextField(max_length=65563)
```
akatsoulas: Shouldn't the value be `65535`?

kitsune/customercare/tests/test_forms.py
```diff
@@ -352,6 +352,9 @@ def test_send_with_no_category_selected(self, mock_task):
...
+        # Since we don't care about the form, we'll just clear the errors to run send().
+        form._errors = ""
```
akatsoulas, line 358: I am not sure why we are bypassing the form errors here. We shouldn't be able to send() given the invalid form

kitsune/customercare/models.py
akatsoulas: I suspect this change will also need a schema migration

denyshon: form does allow an empty value: `category=self.cleaned_data.get("category", "")`
in case we'll unmark the field as required

denyshon: I have a handwritten migration, but from what I understand, it should ideally be generated by django?

kitsune/customercare/models.py
denyshon: I believe I took that value from the docs, but now they [say](https://developer.zendesk.com/api-reference/ticketing/tickets/ticket_comments/#body-size-limits) just "64 kB", so let me make change this to 65535

kitsune/customercare/models.py
denyshon: (updated the comment above, the current behavior regarding an empty value is not consistent, so you'll be the best person to decide on this)

kitsune/customercare/tests/test_forms.py
denyshon: we can pass a random topic if you want

kitsune/customercare/models.py
akatsoulas: you can run `./manage.py makemigrations` to create the migration and `./manage migrate` to apply it

denyshon:
@akatsoulas migration added. Please let me know if any further changes are needed

kitsune/customercare/models.py
```diff
@@ -150,8 +150,8 @@ class SupportTicket(ModelBase):
     }
 
     subject = models.CharField(max_length=255)
-    description = models.TextField()
-    category = models.CharField(max_length=255)
+    description = models.TextField(max_length=65535)
```
akatsoulas: This will not be enforced. From Django docs:
```
If you specify a max_length attribute, it will be reflected in the Textarea widget of the auto-generated form field. However it is not enforced at the model or database level. Use a CharField for that.
```

kitsune/customercare/forms.py
```diff
@@ -277,22 +282,13 @@ def send(self, user, product):
...
-            zendesk_tags=zendesk_tags,
-            submission_status=SupportTicket.STATUS_PENDING,
-        )
+        self.instance.product = product
+        self.instance.user = user if (user and user.is_authenticated) else None
+        self.instance.org_group = resolve_org_group(user, product)
+        self.instance.zendesk_tags = zendesk_tags
+        self.instance.submission_status = SupportTicket.STATUS_PENDING
+
+        submission = super().save()
```
akatsoulas: this should be `self.save()`

kitsune/customercare/tests/test_forms.py
```diff
@@ -353,6 +353,9 @@ def test_send_with_no_category_selected(self, mock_task):
...
         submission = form.send(self.user, self.vpn_product)
```
akatsoulas, line 359: Submission is possible only if the form is valid. I am not even sure we need this test

kitsune/customercare/models.py
denyshon: Oh, indeed. Let's switch to CharField then

kitsune/customercare/tests/test_forms.py
denyshon: my understanding is that we're keeping it in case we decide to make the category field optional or remove it, but I can delete the test if you want

kitsune/customercare/forms.py
denyshon: fixed

denyshon:
@akatsoulas thank you, I've updated the PR

akatsoulas:
I believe the update from TextField to CharField will force a validation in every row and it's going to fail if an entry is more that 65535 chars. Maybe an easier path is to keep the TextField and introduce a validator instead. 

Otherwise we need to first make sure that we don't have any entries that have more than 65535 chars in all environments. Let me know which path you prefer

denyshon:
@akatsoulas yeah, I thought about the same thing, but it's up to you to decide :D

Any entry longer than 65535 chars will most likely be spam (and, therefore, safe to delete), but I don't know how easy is to actually handle that.

If you want, I'll be happy to write a custom validator instead

akatsoulas:
A custom validator is the easier approach. Emil for example has entries with more than 65K chars to test staging. This means that the migration will fail. I assume it will fail in prod as well either b/c of spam or b/c of testing

PR #7272, opened 2026-02-24
improve AAQ voting rate limits and messaging

escattone:
mozilla/sumo#1216

### Question Voting
image

image

### Answer Voting
image

### DIFF
--- a/kitsune/questions/views.py
+++ b/kitsune/questions/views.py
@@ -1074,6 +1074,13 @@ def question_vote(request, question_id):
         raise PermissionDenied
 
     if not question.has_voted(request):
+        if request.limited:
+            msg = _("You've hit your voting limit for now. Please try again later.")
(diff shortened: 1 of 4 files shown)


akatsoulas (changes requested):
My main concern with this PR is that we are introducing a modal for an operation that it's informative. If we are going to interrupt the user flow, it should be for something that demands a user action or it's about critical information. The message about hitting the rate-limiting is neither. On top of that, in one of the 2 modals the only 2 actions are the OK and the dismissal button. 

An idea to display the information could be to disable/gray-out the voting button when the ratelimit is reached and add a label to inform the user about it. This requires no user interaction and the information is tied to the voting action

escattone:
Updated, and ready for another review (see second image in introduction for an image of the update).

Just FYI, I didn't change anything about the `Thanks!` message other than adding the box shadow to emphasize the box/form that's overlaid on the page. It's not a modal.

PR #7277, opened 2026-03-02
Update number of childs in groups

akatsoulas:


kitsune/groups/views.py
```diff
@@ -54,12 +56,21 @@ def _remove_group_member(profile, user, request, remove_from_group=True):
...
-    groups = GroupProfile.objects.visible(request.user).select_related("group")
+    groups = [*GroupProfile.objects.visible(request.user).select_related("group")]
+
+    path_to_id = {g.path: g.id for g in groups}
+
+    visible_child_counts = Counter(
+        path_to_id[g.path[:-GroupProfile.steplen]]
+        for g in groups
+        if g.depth > 1 and g.path[:-GroupProfile.steplen] in path_to_id
+    )
```
escattone, line 67: Just a comment that I think something like the following would make it more readable.
```python
    visible_child_counts = Counter(
        path_to_id[parent_path]
        for g in groups
        if g.depth > 1 and ((parent_path := g.path[:-GroupProfile.steplen]) in path_to_id)
    )
```

PR #7284, opened 2026-03-03
Add segmentation tags for Enterprise

akatsoulas:


kitsune/tags/migrations/0002_auto_20241115_0241.py
escattone, line 1: Why are these new segmentation tags being added to a migration that has already been applied? I would expect them to be added via a new migration.

akatsoulas: It's there just for reference. The segmentation tags will be added manually through the admin since it's just content.

PR #7286, opened 2026-03-03
Update routing options for ZD

akatsoulas:


kitsune/products/managers.py
```diff
@@ -57,6 +61,18 @@ def route_support_request(self, request, product):
             return (None, False)
 
         user = request.user
+
+        # Subscription gate: check before any other routing logic
+        if support_config.subscription_only:
+            is_subscribed = (
+                user.is_authenticated
+                and user.profile.products.filter(id=product.id).exists()
```
escattone, line 69: Should we add a profile check?

```python
... and user.profile and user.profile.products.filter(id=product.id).exists()
```

escattone (approved):
r+wc

kitsune/products/managers.py
akatsoulas: This should be redundant given that the manager is used in places where a user should have a profile

PR #7289, opened 2026-03-04
Fix 500 on default aaq topic selection

emilghittasv:


kitsune/questions/views.py
```diff
@@ -831,6 +831,8 @@ def aaq_step3(request, product_slug):
         request.headers.get("hx-trigger-name") == "category"
     ):
         topic_id = request.GET.get("category")
+        if not topic_id:
+            return HttpResponse(status=204)
```
akatsoulas, line 835: We are getting a validation error that instead of the topic `id` which we need to retrieve it form the db (and it's an integer) we are getting an empty string. The proposed response will silence the error but it's not fixing the underlying problem

denyshon: Hey @akatsoulas, could you please elaborate on what underlying problem you meant above? Did you mean that we need to validate that topic_id is an integer here, or that we shouldn't send a hx-request at all when the category is empty?

PR #7291, opened 2026-03-04
Strip NUL characters from tagged

emilghittasv:


kitsune/questions/views.py
```diff
@@ -172,7 +172,7 @@ def question_list(request, product_slug=None, topic_slug=None):
     if show not in FILTER_GROUPS:
         show = "needs-attention"
 
-    tagged = request.GET.get("tagged")
+    tagged = request.GET.get("tagged", "").replace("\x00", "") or None
```
akatsoulas: Why are you adding the `or None` part when you are defaulting to an empty string?

kitsune/questions/views.py
```diff
@@ -173,9 +175,9 @@ def question_list(request, product_slug=None, topic_slug=None):
     if show not in FILTER_GROUPS:
         show = "needs-attention"
 
-    tagged = request.GET.get("tagged")
+    tagged = strip_nul_bytes(request.GET.get("tagged", ""))
```
akatsoulas: Why are you defaulting to an empty sring?

kitsune/questions/views.py
```diff
@@ -173,9 +175,9 @@ def question_list(request, product_slug=None, topic_slug=None):
...
     tags = None
-    topic_slug = request.GET.get("topic", "") or topic_slug
+    topic_slug = strip_nul_bytes(request.GET.get("topic", "")) or topic_slug
```
akatsoulas: similar here. Before the default value was `None`, now it's an empty string

kitsune/flagit/views.py
```diff
@@ -132,6 +132,8 @@ def flag(request, content_type=None, model=None, object_id=None, **kwargs):
 def flagged_queue(request):
     """Display the flagged queue with optimized queries."""
     reason = request.GET.get("reason")
+    if reason and reason not in dict(FlaggedObject.REASONS):
```
akatsoulas: We could use a frozenset to avoid rebuilding this in every request.

akatsoulas (approved):
r+wc

kitsune/questions/views.py
emilghittasv: Having another look at this, yeah there is no real benefit of defaulting this to an empty string.

Changed.

kitsune/questions/views.py
emilghittasv: Having another look at this, yeah there is no real benefit of defaulting this to an empty string.

Changed.

kitsune/flagit/views.py
emilghittasv: Done!

kitsune/questions/views.py
```diff
@@ -1283,7 +1010,7 @@ def solve(request, question_id, answer_id):
     """Accept an answer as the solution to the question."""
 
     if not (
-        ((request.method == "GET") and (watch_secret := request.GET.get("watch", None)))
+        ((request.method == "GET") and (watch_secret := strip_nul_bytes(request.GET.get("watch", None))))
```
akatsoulas, line 1013: `None` is not needed in the `get()` method. It's the default value

kitsune/postcrash/views.py
```diff
@@ -3,11 +3,12 @@
 from django.views.decorators.cache import cache_page
 
 from kitsune.postcrash.models import Signature
+from kitsune.sumo.utils import strip_nul_bytes
 
 
 @cache_page(60 * 60 * 24)  # One day.
 def api(request):
-    s = request.GET.get("s", None)
+    s = strip_nul_bytes(request.GET.get("s", None))
```
akatsoulas, line 11: `None` is not needed in the `get()` method. It's the default value

kitsune/users/api.py
```diff
@@ -57,8 +55,8 @@ def to_internal_value(self, data):
 @json_view
 def usernames(request):
     """An API to provide auto-complete data for user names."""
-    term = request.GET.get("term", "")
-    query = request.GET.get("query", "")
+    term = strip_nul_bytes(request.GET.get("term", ""))
```
akatsoulas, line 58: Some calls like this one, pass an empty string as a default value and others rely on None. We should be consistent

akatsoulas: A good way to enforce this is by adding a `str -> str` contract in the signature of the function

kitsune/sumo/utils.py
```diff
@@ -142,6 +160,13 @@ class TruncationException(Exception):
     pass
 
 
+def strip_nul_bytes(value):
```
akatsoulas, line 163: My main concern with this PR is that we need to remember to sanitize which leaves a lot of room for errors.  For example this PR already missed a call I believe in wiki/views. I suspect this will happen often. We could use a middleware for `GET` and `POST` requests only.

akatsoulas: @escattone thoughts on this?

akatsoulas:
@escattone can you take a look on this ?

escattone:
Sorry for the delayed review. I agree with @akatsoulas that this approach feels fragile in that we'd have to remember to strip NULL bytes every time we obtain data from `request.GET`. I think it would be worthwhile to explore solving this problem with a small custom middleware.

emilghittasv:
Thank you! I am going to close this PR in favor of the middleware approach

PR #7295, opened 2026-03-04
add protection for image uploads

escattone:
mozilla/sumo#2678

## Notes
- We already enforce a file-size limit of 10MB, but this PR adds another check, one that enforces a pixel limit (20 MP) on images in order to protect against excessive memory demands during decompression of the image.
- It also adds graceful handling of `Image.DecompressionBombError`.
- These checks have been centralized in a new `open_as_pil_image` utility function.
- It also adds these checks into the `GroupAvatarForm.clean_avatar` method to protect the group avatar uploads.
- It adds tests as needed.
- It re-organizes some files as needed:
    - The test cases in `kitsune/upload/tests/__init__.py` have been moved into a new `kitsune/upload/tests/test_utils.py` file.
    - Moves two non-Celery-task functions out of `kitsune/upload/tasks.py` and into `kitsune/upload/utils.py`.

### DIFF
--- a/kitsune/groups/admin.py
+++ b/kitsune/groups/admin.py
@@ -5,27 +5,14 @@ from treebeard.admin import TreeAdmin
 from treebeard.forms import movenodeform_factory
 
 from kitsune.groups.models import GroupProfile
-from kitsune.upload.utils import (
-    FileTooLargeError,
-    check_file_size,
-    create_image_thumbnail,
-    open_as_pil_image,
-)
+from kitsune.upload.utils import create_image_thumbnail, validate_avatar_size
 
 
 class GroupProfileAdminForm(movenodeform_factory(GroupProfile)):  # type: ignore[misc]
     def clean_avatar(self):
         avatar = self.cleaned_data.get("avatar")
         if avatar and hasattr(avatar, "size"):
-            try:
-                check_file_size(avatar, settings.MAX_AVATAR_FILE_SIZE)
-            except FileTooLargeError as e:
-                raise forms.ValidationError(e.args[0])
-            try:
-                with open_as_pil_image(avatar):
-                    pass
-            except FileTooLargeError as e:
-                raise forms.ValidationError(e.args[0])
+            validate_avatar_size(avatar)
         return avatar
 
 
--- a/kitsune/groups/forms.py
+++ b/kitsune/groups/forms.py
@@ -7,7 +7,7 @@ from kitsune.groups.models import GroupProfile
 from kitsune.sumo.form_fields import MultiUsernameField
 from kitsune.sumo.widgets import ImageWidget
 from kitsune.upload.forms import LimitedImageField
-from kitsune.upload.utils import FileTooLargeError, check_file_size, open_as_pil_image
+from kitsune.upload.utils import validate_avatar_size
 
 
 class GroupProfileForm(forms.ModelForm):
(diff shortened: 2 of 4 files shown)


akatsoulas:
### Code review

Found 2 issues:

1. `save_model` in the admin path doesn't catch `FileTooLargeError` — before this PR, `create_image_thumbnail` called `Image.open()` directly and never raised `FileTooLargeError`. After this PR, `create_image_thumbnail` calls `open_as_pil_image()` internally, which raises `FileTooLargeError` when pixel limits are exceeded. The normal web form path is protected by `GroupAvatarForm.clean_avatar()`, but `GroupProfileAdmin.save_model()` doesn't use `GroupAvatarForm` — it uses the raw treebeard model admin form. An oversized image uploaded via the Django admin will now cause an unhandled `FileTooLargeError` and a 500 error instead of a clean validation message.

https://github.com/mozilla/kitsune/blob/036cebf596a25b82486ca9c9169b9576216ac1e8/kitsune/groups/admin.py#L66-L72

2. Redundant comment `# This should raise` inside an `assertRaises` context manager (CLAUDE.md says "do not add redundant comments")

https://github.com/mozilla/kitsune/blob/036cebf596a25b82486ca9c9169b9576216ac1e8/kitsune/upload/tests/test_utils.py#L53-L57

🤖 Generated with [Claude Code](https://claude.ai/code)

<sub>- If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

escattone:
Trying to upload an over-sized avatar for a group profile via the admin console is now checked as well. ✅ 

@akatsoulas Ready for another review.

kitsune/groups/admin.py
```diff
@@ -5,11 +5,32 @@
...
+class GroupProfileAdminForm(movenodeform_factory(GroupProfile)):  # type: ignore[misc]
+    def clean_avatar(self):
+        avatar = self.cleaned_data.get("avatar")
+        if avatar and hasattr(avatar, "size"):
+            try:
+                check_file_size(avatar, settings.MAX_AVATAR_FILE_SIZE)
+            except FileTooLargeError as e:
+                raise forms.ValidationError(e.args[0])
+            try:
+                with open_as_pil_image(avatar):
```
akatsoulas: The `with` statement probably is not needed. The `open_as_pil_image` is used here just to validate the size. The actual image is never used

kitsune/groups/forms.py
```diff
@@ -46,6 +46,11 @@ def clean_avatar(self):
                 check_file_size(avatar, settings.MAX_AVATAR_FILE_SIZE)
             except FileTooLargeError as e:
                 raise forms.ValidationError(e.args[0])
+            try:
+                with open_as_pil_image(avatar):
```
akatsoulas: similar comment

akatsoulas (approved):
r+wc

kitsune/groups/admin.py
akatsoulas: Given that `open_as_pil_image` seems to only validate the image, we could use something more purposeful like `validate_image_size` or something similar

escattone: > The `with` statement probably is not needed. The `open_as_pil_image` is used here just to validate the size. The actual image is never used

The actual image is used, because the `open_as_pil_image` function does call `Image.open()` to read the file and validate the sizes, so the `with` does ensure it's immediately closed after the check.

escattone: > Given that `open_as_pil_image` seems to only validate the image, we could use something more purposeful like `validate_image_size` or something similar

We use `open_as_pil_image` for more than just validating the image, but it would be more DRY to encapsulate the validation (which is used in two places) in a separate utility function.

PR #7298, opened 2026-03-05
handle nullified profile on account events

escattone:
mozilla/sumo#2816

I was able to confirm the root cause of these errors. It's due to duplicate Mozilla Account events. For example, I can see two `delete-user` account events for a specific `fxa_uid` created within a second of each other. They both initially have profiles, but when the first is processed, the user deletion cascades such that the profile of the second is set to `None`, so when the second is processed, it fails when trying to get the user via its profile.

This PR doesn't attempt to de-duplicate account events. It only takes the defensive approach, detecting account events that have had their profiles nullified, and changing their status to `IGNORED`. We could, I think, de-duplicate events as well, if desired, with something like the following in the loop within `WebhookView.process_events`:

```python
for long_id, event in events.items():
    short_id = ...
    # Ignore duplicate events.
    if AccountEvent.objects.filter(fxa_uid=fxa_uid, event_type=short_id, status=AccountEvent.UNPROCESSED).exists():
        continue
    ...
```

By **not** doing that, we do gain a record of **all** incoming events, so I've opted for that approach (at least for now).

### DIFF
--- a/kitsune/users/tasks.py
+++ b/kitsune/users/tasks.py
@@ -22,6 +22,15 @@ shared_task_with_retry = shared_task(
 log = logging.getLogger("k.task")
 
 
+def ignore_if_missing_profile(event: AccountEvent) -> bool:
+    """Checks if an event has an associated profile and, if not, marks it as ignored."""
+    if event.profile is None:
+        event.status = AccountEvent.IGNORED
+        event.save(update_fields=["status"])
+        return True
+    return False
+
+
 @shared_task_with_retry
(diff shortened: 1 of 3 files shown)


escattone:
I modified the code to avoid duplicate unprocessed delete-user events for the same profile. I didn't add de-duplication for other events since that would involve checking the `body` as well, which would probably mean that we should also index the `body`, and I didn't think that was worth it. I kept the defensive approach as well, as a safeguard.

PR #7350, opened 2026-03-25
add zendesk webhook for push updates

escattone:
mozilla/sumo#2856

### DIFF
--- a/kitsune/customercare/tasks.py
+++ b/kitsune/customercare/tasks.py
@@ -5,6 +5,7 @@ from celery import shared_task
 from django.contrib.auth.models import User
 from django.contrib.contenttypes.models import ContentType
 from django.utils import timezone
+from django.utils.dateparse import parse_datetime
 
 from kitsune.customercare.models import SupportTicket
 from kitsune.customercare.utils import process_zendesk_classification_result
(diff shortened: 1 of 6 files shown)


kitsune/customercare/tasks.py
```diff
@@ -119,3 +123,52 @@ def process_failed_zendesk_tickets() -> None:
...
+    Updates the matching SupportTicket with status, comments, and tags
+    from the payload. Handles partial payloads gracefully — only fields
+    present in the payload are updated.
+    """
+    if not (ticket_id := payload.get("ticket_id")):
+        log.warning("Zendesk webhook payload missing ticket_id.")
+        return
+
+    try:
+        ticket = SupportTicket.objects.get(zendesk_ticket_id=str(ticket_id))
```
akatsoulas: There's a really good chance that we are going to have race conditions here. Maybe we can modify the existing code by using `select_for_udpate` method and wrap everything in a `transaction.atomic`

kitsune/customercare/views.py
```diff
@@ -41,3 +51,59 @@ def update_topic(request, ticket_id):
...
+        signature_header = request.headers.get("x-zendesk-webhook-signature")
+        timestamp = request.headers.get("x-zendesk-webhook-signature-timestamp")
+
+        if not (signature_header and timestamp):
+            raise SuspiciousOperation("Missing signature or timestamp header.")
+
+        secret = settings.ZENDESK_WEBHOOK_SIGNING_SECRET.encode("utf-8")
+        message = timestamp.encode("utf-8") + request.body
+        computed = hmac.new(secret, message, hashlib.sha256).digest()
+        expected = base64.b64decode(signature_header)
```
akatsoulas: shouldn't we catch errors here to avoid having 500s instead of 403? This could be potentially a ddos endpoint

akatsoulas (approved):
r+wc

kitsune/customercare/tasks.py
escattone: Fantastic catch, thanks!

kitsune/customercare/views.py
escattone: Another great catch, thanks!

PR #7369, opened 2026-03-31
add ZD ticket status/comment write methods

escattone:
mozilla/sumo#2921

### DIFF
--- a/kitsune/customercare/zendesk.py
+++ b/kitsune/customercare/zendesk.py
@@ -1,5 +1,6 @@
 from django.conf import settings
 from zenpy import Zenpy
+from zenpy.lib.api_objects import Comment as ZendeskComment
 from zenpy.lib.api_objects import Identity as ZendeskIdentity
 from zenpy.lib.api_objects import Ticket
 from zenpy.lib.api_objects import User as ZendeskUser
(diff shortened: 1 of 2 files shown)


kitsune/customercare/zendesk.py
```diff
@@ -181,3 +184,32 @@ def create_ticket(self, user, ticket_fields):
...
+
+        If a Django user is provided, the comment will be attributed to their
+        corresponding Zendesk user. The user must have a zendesk_id set on their profile.
+        If no user is provided, the comment defaults to the ticket's requester.
+        """
+        ticket = self.client.tickets(id=ticket_id)
+        comment = ZendeskComment(body=comment_body, public=public)
+        if user:
+            if not (user.is_authenticated and (zendesk_id := user.profile.zendesk_id)):
+                raise ValueError(f'User "{user}" does not have a "zendesk_id".')
```
akatsoulas: If an authenticated user doesn't have a zendesk_id, shouldn't we create one? How can we end up with a ticket_id in our system attributed to a user that doesn't have a zendesk_id?

akatsoulas: Also probably we shouldn't mix the existence of the zendesk_id with the login status of the user.

kitsune/customercare/zendesk.py
```diff
@@ -181,3 +184,32 @@ def create_ticket(self, user, ticket_fields):
...
+            comment.author_id = int(zendesk_id)
+        else:
+            comment.author_id = ticket.requester_id
```
akatsoulas: Under which conditions can we end up here in the flow? This is basically similar to the 2nd question of the first comment.

kitsune/customercare/zendesk.py
escattone: > If an authenticated user doesn't have a zendesk_id, shouldn't we create one? 

👍 

> How can we end up with a ticket_id in our system attributed to a user that doesn't have a zendesk_id?

Should the user associated with the comment be allowed to be a user other than the requester of the ticket? I'm assuming yes, but maybe that's not the case?

> Also probably we shouldn't mix the existence of the zendesk_id with the login status of the user.

That's what we already do above in the `create_ticket` method. How else would we exclude anonymous users?

akatsoulas (approved):

PR #7370, opened 2026-04-01
Update question detailed view

akatsoulas:
* Introduce a detailed view for tickets
* Streamline the UI where possible

### DIFF
--- a/kitsune/customercare/jinja2/customercare/includes/ticket_replies.html
+++ b/kitsune/customercare/jinja2/customercare/includes/ticket_replies.html
@@ -1,0 +1,46 @@
+{% from "sumo/includes/thread_post.html" import post_body %}
+
+{% if sync_error %}
+  <div class="thread-detail--notice thread-detail--notice--info">
+    {{ _('We could not fetch the latest replies. Please try again later.') }}
+  </div>
+{% endif %}
+
+{% if ticket.public_comments %}
+  <h2 class="sumo-card-heading">
+    {% trans count=ticket.public_comments | length %}
+      {{ count }} Reply
+    {% pluralize %}
+      {{ count }} Replies
+    {% endtrans %}
+  </h2>
+  {% for comment in ticket.public_comments %}
+    {% set _author = comment.get("author", {}) %}
+    {% set _created_str = comment.get("created_at", "") %}
+    {% set _created_dt = _created_str | parse_datetime %}
+    <div class="thread-post" id="comment-{{ comment.get('id', loop.index) }}">
+      <div class="thread-post--header">
+        <div class="thread-post--avatar">
+          <div class="question-entry--icon" aria-hidden="true">
+            <svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 24 24" fill="none"
+                 stroke="currentColor" stroke-width="2" stroke-linecap="round" stroke-linejoin="round">
+              <path d="M20 21v-2a4 4 0 0 0-4-4H8a4 4 0 0 0-4 4v2"/>
+              <circle cx="12" cy="7" r="4"/>
+            </svg>
+          </div>
+        </div>
+        <div class="thread-post--author">
+          <span class="thread-post--author-name">{{ _author.get("name", _("Support Agent")) }}</span>
+          {% if _created_dt %}
+            <time class="thread-post--time" datetime="{{ _created_str }}">
+              {{ datetimeformat(_created_dt, use_timesince=True) }}
+            </time>
+          {% endif %}
+        </div>
+      </div>
+      {{ post_body(comment.get("body", "") | e | replace('\n', '<br>') | safe) }}
+    </div>
+  {% endfor %}
+{% else %}
+  <p class="thread-comments--empty">{{ _("No replies yet.") }}</p>
+{% endif %}
--- a/kitsune/customercare/models.py
+++ b/kitsune/customercare/models.py
@@ -1,5 +1,6 @@
 from django.contrib.auth.models import User
 from django.db import models
+from django.urls import reverse
 from django.utils.translation import gettext_lazy as _lazy
 
 from kitsune.products.models import Product, Topic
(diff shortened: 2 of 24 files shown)


akatsoulas:
This also pulls comments from ZD if the last sync was more than an hour ago

kitsune/customercare/models.py
```diff
@@ -92,7 +91,7 @@ def channel(self):
         return "direct_support"
 
     def get_absolute_url(self):
-        return "#"  # no detail page yet
+        return reverse("customercare.ticket_detail", args=[self.user.username, self.id])
```
escattone, line 94: You're probably already aware, but just noting that the `customercare.ticket_detail` endpoint doesn't exist yet.

kitsune/urls.py
```diff
@@ -27,7 +27,7 @@
     path("search/", include("kitsune.search.urls")),
     path("forums/", include("kitsune.forums.urls")),
     path("questions/", include("kitsune.questions.urls")),
-    path("customercare/", include("kitsune.customercare.urls")),
+    path("", include("kitsune.customercare.urls")),
```
escattone, line 30: `flagit.js` has a hard-coded URL that still uses the `customercare/` path. Would it make sense to move that hard-coded URL into a `data-*` attribute instead?

kitsune/customercare/views.py
```diff
@@ -1,14 +1,50 @@
...
+        user__username=username,
+    )
+    if not (ticket.user_id == request.user.id or request.user.has_perm("customercare.change_supportticket")):
+        raise Http404
+
+    if request.headers.get("HX-Request"):
+        sync_error = False
+        try:
+            sync_ticket_from_zendesk(ticket)
+        except Exception:
```
escattone, line 33: Probably too general. Perhaps catch just `zenpy.lib.exception.APIException` and `requests.exceptions.RequestException` instead?

kitsune/customercare/views.py
```diff
@@ -1,14 +1,50 @@
...
+def ticket_detail(request, username, ticket_id):
+    ticket = get_object_or_404(
+        SupportTicket.objects.select_related("product", "topic", "user"),
+        id=ticket_id,
+        user__username=username,
+    )
+    if not (ticket.user_id == request.user.id or request.user.has_perm("customercare.change_supportticket")):
+        raise Http404
+
+    if request.headers.get("HX-Request"):
```
escattone, line 29: Would it make sense to double check here if we really need to sync before we do it, as another layer of protection against wasting our Zendesk API quota/rate-limit? Probably an edge case.

escattone (commented):
General comments:

- There's too much white-space in the detailed view above the original question as well as each of the replies.

Image


- I think the question details CSS class hierarchy changed from `.question > .main-content` to `.question > .content`, but the `initCrashIdLinking` function within `questions.js` is still using `.question > .main-content`.

kitsune/customercare/jinja2/customercare/includes/ticket_replies.html
```diff
@@ -0,0 +1,48 @@
...
+{% if ticket.comments %}
+  <h2 class="sumo-card-heading">
+    {% trans count=ticket.comments | length %}
+      {{ count }} Reply
+    {% pluralize %}
+      {{ count }} Replies
+    {% endtrans %}
+  </h2>
+  {% for comment in ticket.comments %}
+    {% if comment.get("public", True) %}
```
escattone, line 18: This seems risky. If the comment has no `public` attribute, I'm thinking we should probably assume it's private just to be on the safe side.

escattone: Maybe we should create a read-only `public_comments` property on the `SupportTicket` class and use that here?

escattone:
In general, I like this unified approach, and the asynchronous fetching of support ticket comments that haven't been synchronized recently.

kitsune/customercare/utils.py
```diff
@@ -15,6 +16,28 @@
...
 
+def sync_ticket_from_zendesk(ticket: SupportTicket) -> None:
+    """Fetch fresh ticket status and comments from Zendesk and save to DB."""
+    client = ZendeskClient()
+    zd_ticket = client.get_ticket(ticket.zendesk_ticket_id)
+    zd_comments = client.get_ticket_comments(ticket.zendesk_ticket_id)
+
+    ticket.zd_status = zd_ticket.status
+    ticket.zd_updated_at = zd_ticket.updated_at
+    ticket.comments = [
```
escattone, line 27: My only concern here is the case when the local ticket's comments are actually more up-to-date than the comments fetched from ZD. For example, what if a comment was added via SUMO, and either hasn't yet been updated in ZD, or the ZD update failed, so when we get the comments from ZD, that newly-added comment would be wiped-out by this sync.

kitsune/customercare/utils.py
```diff
@@ -15,6 +16,28 @@
...
+        {
+            "id": c.id,
+            "body": c.body,
+            "created_at": c.created_at,
```
escattone, line 31: I think Zenpy returns a `datetime` instance for `created_at`? If so, we'd have to convert it to a string for JSON.

kitsune/customercare/models.py
```diff
@@ -92,11 +91,15 @@ def channel(self):
...
```
escattone, line 96: Do we have to consider the case of a ticket without a `user`, since `user` can be `None`?

escattone (commented):
I think the change from `.question .main-content` to `.question .content` still breaks `initCrashIdLinking` in `kitsune/sumo/static/sumo/js/questions.js`?

kitsune/customercare/utils.py
escattone: I put this challenge to `Claude`, and it recommended an "Outbox table":

>   Outbox table (recommended). A SupportTicketCommentOutbox model holds SUMO-authored replies. A small worker (or per-write Celery task with retry) drains it to Zendesk; on success the row is deleted, and the next sync — or the response from the create-comment call itself — populates `comments` with ZD's authoritative version. The UI displays `comments` + outbox.pending_for(ticket) so users see their reply immediately. Net effect: `comments` is always exactly what ZD says, and retries/failures live in one obvious place.

escattone: Assuming we go with some kind of "outbox" approach, I would say your current code here is fine for now, and perhaps we can address the `outbox` approach in a separate PR?

PR #7396, opened 2026-04-07
Skip classification on local env

akatsoulas:


escattone (changes requested):
Both the `dev` and `stage` environments have `settings.DEV=True`, so this would affect them as well.

akatsoulas:
@escattone PR udpated

escattone (approved):

PR #7408, opened 2026-04-08
Display spam questions for moderators in the user view

### LINKED ISSUE #2961: Spam questions are not displayed for moderators in user profiles
When viewing questions authored by a user from their profiles, spam questions are not visible even for moderators (despite the counter including them). That's a critical functionality that must be restored ASAP. For more details, see https://support.mozilla.org/forums/forum-moderators/717594.

denyshon:
Resolves [#2961](https://github.com/mozilla/sumo/issues/2961).

Restore displaying of spam questions to moderators on the Questions Asked pages, and introduce a moderator-specific view consisting of 3 tabs:
- **All** (default), listing all the questions, including spam ones
- **Valid**, listing only questions not marked as spam
- **Spam**, listing only questions marked as spam

The tabs will be hidden when there are no spam questions (so that it's easy to see if there are any) and when viewing your own profile. The tabs will _not_ be hidden if you have legitimately opened one, even if there are no spam questions at the moment (so that you can always navigate back).

image

image

image

### DIFF
--- a/kitsune/users/jinja2/users/questions_contributed.html
+++ b/kitsune/users/jinja2/users/questions_contributed.html
@@ -10,6 +10,7 @@
       {% if is_owner %}
         {{ _("My Questions") }}
       {% else %}
+        {# L10n: This is a title for user questions pages (like https://support.mozilla.org/user/kelimuttu/questions), displayed when a page is viewed by someone other than the profile owner. The page lists questions authored by the user. #}
         {{ _("Questions Asked") }}
       {% endif %}
     </h2>
@@ -21,10 +22,29 @@
           <a href="?" {% if not channel %}class="selected"{% endif %}><span>{{ _("All") }}</span></a>
         </li>
         <li class="tabs--item">
+          {# L10n: Displayed for users viewing their own questions. This is a label for a tab that lists community forum questions authored by the user. #}
           <a href="?channel=forum" {% if channel == "forum" %}class="selected"{% endif %}><span>{{ _("Forum") }}</span></a>
         </li>
         <li class="tabs--item">
+          {# L10n: A label/tab for questions that are Premium support tickets handled by paid support agents. #}
           <a href="?channel=direct_support" {% if channel == "direct_support" %}class="selected"{% endif %}><span>{{ _("Direct Support") }}</span></a>
+        </li>
+      </ul>
+    </nav>
+    {% endif %}
+
+    {% if is_moderator and (user_has_spam_questions or show) and not is_owner %}
+    <nav class="tabs">
+      <ul class="tabs--list">
+        <li class="tabs--item">
+          <a href="?" {% if not show %}class="selected"{% endif %}><span>{{ _("All") }}</span></a>
+        </li>
+        <li class="tabs--item">
+          {# L10n: Displayed for moderators only on user questions pages (like https://support.mozilla.org/user/kelimuttu/questions). This is a label for a tab that lists all non-spam forum questions authored by the user. #}
+          <a href="?show=valid" {% if show == "valid" %}class="selected"{% endif %}><span>{{ _("Valid") }}</span></a>
+        </li>
+        <li class="tabs--item">
+          <a href="?show=spam" {% if show == "spam" %}class="selected"{% endif %}><span>{{ _("Spam") }}</span></a>
         </li>
       </ul>
     </nav>
--- a/kitsune/users/views.py
+++ b/kitsune/users/views.py
@@ -223,6 +223,7 @@ def questions_contributed(request, username):
     profile = get_object_or_404(Profile, user__username=username, user__is_active=True)
 
     channel = request.GET.get("channel")
+    show = request.GET.get("show")
     product_slug = request.GET.get("product")
     topic_slug = request.GET.get("topic")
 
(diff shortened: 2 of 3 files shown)


kelimuttu:
Hi Denys, 

Would it be possible to update the "Non-spam" copy into "Valid questions" instead? Let's avoid using "spam" to anchor our thinking or copy. 

Also, am I correct to assume that the idea is to make this view only available for forum moderators?

denyshon:
@kelimuttu sure, let me do this. Yes, this view will be visible to moderators only, since only they can see spam questions

denyshon:
Converting to draft while we're deciding which name to choose for the "Non-spam" tab. Note for myself: also add l10n comments for the new strings.

denyshon:
We've decided to stick with "Valid". The PR has been updated and is now ready for review again.

denyshon:
@akatsoulas PR updated according to our discussion

kitsune/users/views.py
```diff
@@ -223,25 +223,37 @@ def questions_contributed(request, username):
...
+    show = request.GET.get("show")
     product_slug = request.GET.get("product")
     topic_slug = request.GET.get("topic")
 
     is_owner = request.user == profile.user
     if not is_owner and channel == "direct_support":
         raise Http404()
 
+    is_moderator = request.user.has_perm("flagit.can_moderate")
+    print (is_moderator)
```
akatsoulas: Let's remove this statement

kitsune/users/jinja2/users/questions_contributed.html
```diff
@@ -21,15 +22,34 @@ <h2 class="sumo-page-subheading">
...
         </li>
         <li class="tabs--item">
+          {# L10n: A label/tab for questions that are Premium support tickets handled by paid support agents. #}
           <a href="?channel=direct_support" {% if channel == "direct_support" %}class="selected"{% endif %}><span>{{ _("Direct Support") }}</span></a>
         </li>
       </ul>
     </nav>
     {% endif %}
 
+    {% if is_moderator and (spam_forum_questions or show) and not is_owner %}
```
akatsoulas: This will force the query `spam_forum_questions` to evaluate. Maybe we should use a boolean flag here from the view instead and evaluate the query if needed.
eg 
```python
user_has_spam_questions = profile.user.questions(is_spam=True).exists()
```

kitsune/users/views.py
```diff
@@ -223,25 +223,37 @@ def questions_contributed(request, username):
...
+        raise Http404()
+
     if channel != "direct_support":
         forum_questions = (
-            profile.user.questions.filter(is_spam=False)
+            profile.user.questions
             .select_related("solution", "product", "topic", "last_answer", "last_answer__creator")
             .order_by("-created")
         )
+        spam_forum_questions = forum_questions.filter(is_spam=True)
```
akatsoulas: Here spam_forum_questions is assigned with forum_questions that have spam and a couple of lines below the forum_questions is swapped with spam_forum_questions which is a bit confusing

kitsune/users/views.py
```diff
@@ -223,25 +223,37 @@ def questions_contributed(request, username):
...
+        if is_moderator and show == "spam":
```
akatsoulas: If I pass in the url `?show=spam` this will reveal the tabs even in the case the user hasn't any. Be using the boolean flag suggested in the template we could probably mitigate that

akatsoulas (commented):
nice work. Some small improvements and I believe this is good to go

kitsune/users/views.py
denyshon: oops, I'm so sorry...

kitsune/users/views.py
denyshon: That's intended. Otherwise, if I navigate to the spam tab, unmark all spam questions (let's say there's only one), and reload the tab to make sure there are no spam questions left, I'll end up in a situation with a blank page when I can't navigate back. To prevent such kind of issues, we display the tabs when `show` is specified (and the user is allowed to see the tabs at all).

kitsune/users/jinja2/users/questions_contributed.html
denyshon: addressed

kitsune/users/views.py
denyshon: addressed

denyshon:
@akatsoulas thank you for the review! I've updated the PR to address your feedback

akatsoulas (approved):

kitsune/users/views.py
akatsoulas: When I first read your comment, I misunderstood the intention. Apologies about that. If we want the described behavior, then the tabs should remain visible - with an appropriate comment - at all times. Otherwise it's a weird UX where you can make the tabs appear by adding a query in the url. That's what drove me to open a [PR](https://github.com/mozilla/kitsune/pull/7457) to ignore the filter if there's no spam. 

My suggestion would be to go with the PR that I opened in order to streamline the uses cases of when the tabs are visible. If the tabs are needed for the case that you described, let's follow up with another one to make them permanently visible.

denyshon: Ok, let's stick to your approach for now, and if the mods complain, then make the tabs visible permanently

Issue #5326, opened 2022-12-07
Dark theme


rakleed:
Could you please add a dark theme to the site? Because using your site at night or in dark rooms is very unpleasant because of the bright white background.

dannycolin:
Duplicate of #4662

dannycolin:
This is blocked by [mozilla/protocol#647](https://github.com/mozilla/protocol/issues/647)

anewuser:
@dannycolin That issue has been closed by a designer. Please explain to the design team that this is a security issue. People who use extensions like Dark Reader are told to clear `extensions.webextensions.restrictedDomains` in `about:config` to be able to make the background of AMO and SUMO dark.

denyshon:
Hi, does anyone happen to know the current state of the issue? @emilghittasv (sorry if it's a wrong mention, but I've seen you dealing with the issues...)

emilghitta:
Hi @denyshon 👋 

The dark theme implementation is not currently under the engineering team's radar/not currently being worked on due to some more urgent priorities.

kelimuttu:
See also: https://bugzilla.mozilla.org/show_bug.cgi?id=1761983

PR #7417, opened 2026-04-10
Add dark mode

emilghittasv:
- Added CSS custom property tokens for dark mode in `_project-theme.scss`: page background, card backgrounds, text, borders, shadows, form fields, and interactive states
- All dark overrides follow a consistent three-rule pattern:
@media (prefers-color-scheme: dark) { ... }  ← OS preference
[data-theme="dark"] { ... }                  ← manual toggle
[data-theme="light"] { ... }                 ← override OS dark → light
-Added a theme toggle button to the top navbar via Alpine.js, with a cookie-backed preference that is pre-read server-side in base.html to eliminate the flash-of-wrong-theme on page load

### DIFF
--- a/kitsune/sumo/jinja2/base.html
+++ b/kitsune/sumo/jinja2/base.html
@@ -3,8 +3,14 @@
 {% if request.LANGUAGE_CODE == 'xx' %}
   {% set meta = [('robots', 'noindex')] %}
 {% endif %}
+{# Read the theme cookie set by theme-toggle.js so we can pre-render the correct     #}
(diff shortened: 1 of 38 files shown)


denyshon:
We might also want to add dark versions of images in the gallery. What I was thinking of was storing two images under the same entity (light/dark), but there may be different implementations... Anyway, treat this like a note for the future, since this definitely shouldn't be a blocker

kelimuttu:
Related GH issue https://github.com/mozilla/kitsune/issues/5326

PR #7419, opened 2026-04-10
show tags async per question in list

escattone:
mozilla/sumo#2975

- Asynchronously loads each question's tags in the question list view, and uses ES to get each question's tags.
- The tags are only shown to contributors.

### DIFF
--- a/kitsune/questions/views.py
+++ b/kitsune/questions/views.py
@@ -80,6 +80,7 @@ from kitsune.tidings.events import ActivationRequestFailed
 from kitsune.tidings.models import Watch
 from kitsune.upload.models import ImageAttachment
 from kitsune.users.models import Setting
+from kitsune.users.utils import user_is_contributor
 from kitsune.wiki.facets import topics_for
 from kitsune.wiki.utils import build_topics_data, get_featured_articles, get_kb_visited
 
@@ -423,6 +424,7 @@ def question_list(request, product_slug=None, topic_slug=None):
         "product_slug": product_slug,
         "topic_navigation": topic_navigation,
         "has_support_config": product_with_aaq,
+        "show_inline_tags": False,
     }
 
     if products:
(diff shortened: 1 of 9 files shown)


kitsune/questions/views.py
```diff
@@ -439,9 +441,101 @@ def question_list(request, product_slug=None, topic_slug=None):
...
+    try:
+        search = QuestionDocument.search()
+        search = search.filter("ids", values=question_ids)
+        search = search.source(includes=["question_tag_slugs"])
+        search = search.extra(size=len(question_ids))
+        for hit in search.execute():
+            slugs = list(hit.question_tag_slugs or [])
+            if slugs:
+                question_tags[hit.meta.id] = slugs
+    except Exception:
```
akatsoulas: let's make this a bit more explicit

kitsune/questions/views.py
```diff
@@ -439,9 +441,101 @@ def question_list(request, product_slug=None, topic_slug=None):
...
+    if not question_tags:
+        return HttpResponse("")
+
+    tagged_set = set(tagged.split(",")) if tagged else set()
+
+    return render(
+        request,
+        "questions/includes/inline_tags.html",
+        {
+            "question_tags": question_tags,
```
akatsoulas, line 543: this will render tag slugs instead of names

kitsune/questions/views.py
```diff
@@ -439,9 +441,101 @@ def question_list(request, product_slug=None, topic_slug=None):
...
+                "questions.list_by_topic" if topic_navigation else "questions.list",
+                kwargs=(
+                    {"topic_slug": topic_slug}
+                    if topic_navigation
+                    else {"product_slug": product_slug}
+                ),
+            ),
+            **{k: v for k, v in request.GET.items() if k not in ("tagged", "page")},
+        )
+        data["tag_base_url"] = tag_base_url
```
akatsoulas: why is this needed?

akatsoulas (changes requested):

kitsune/questions/views.py
escattone: Good catch, thanks! That was leftover from a previous iteration.

kitsune/questions/views.py
escattone: Good catch.

akatsoulas:
Closing for now. We can revisit this if there's a need for it in the future

PR #7423, opened 2026-04-10
add simple periodic task watchdog

escattone:
mozilla/sumo#2784

This PR provides an option for a simple but effective "watchdog" of all Celery beat tasks -- except for any that are explicitly excluded (`WATCHDOG_EXCLUDED_TASKS`) -- that alerts a list of email recipients (`WATCHDOG_EMAIL_RECIPIENTS`) of all tasks that are currently missing more than a configurable number of scheduled runs (`WATCHDOG_ALLOWED_MISSED_RUNS`). By default, a task can be included in the list of alerts only once per day (`WATCHDOG_ALERT_COOLDOWN_SECONDS`).

Any new Celery beat tasks added will automatically be "watched" unless explicitly excluded.


kitsune/sumo/watchdog.py
```diff
@@ -0,0 +1,115 @@
...
+    beat_schedule = settings.CELERY_BEAT_SCHEDULE
+    if not beat_schedule:
+        return []
+
+    now = timezone.now()
+    excluded = set(settings.WATCHDOG_EXCLUDED_TASKS)
+    overdue = []
+
+    for task_name, task_config in beat_schedule.items():
+        if task_name == "watchdog":
```
akatsoulas: Why do we need both here? Can't we add `watchdog` in excluded?

kitsune/sumo/watchdog.py
```diff
@@ -0,0 +1,115 @@
...
+
+        if not isinstance(schedule, crontab):
+            continue
+
+        try:
+            periodic_task = PeriodicTask.objects.get(name=task_name)
+        except PeriodicTask.DoesNotExist:
+            continue
+
+        last_run_at = periodic_task.last_run_at
```
akatsoulas: This will be updated when the task is dispatched. Probably what it won't tell us is if the task has crashed right after dispatching. This way we can end up with a monitoring solution that reports that everything is working fine but the tasks are crashing. Should we add the task results into the solution?

kitsune/sumo/watchdog.py
```diff
@@ -0,0 +1,115 @@
...
+log = logging.getLogger("k.watchdog")
+
+
+class OverdueTask(NamedTuple):
+    name: str
+    last_run_at: datetime
+    expected_next_run: datetime
+
+
+def compute_period(schedule):
```
akatsoulas: What happens if a task has a non uniform schedule like the `cleanup_old_spam`?

kitsune/sumo/watchdog.py
```diff
@@ -0,0 +1,115 @@
...
+        if task_name in excluded:
+            continue
+
+        schedule = task_config["schedule"]
+
+        if not isinstance(schedule, crontab):
+            continue
+
+        try:
+            periodic_task = PeriodicTask.objects.get(name=task_name)
```
akatsoulas: Although the number of periodic tasks is small, this introduces a N+1 query. Could we use here `in_bulk` queryset/

kitsune/sumo/watchdog.py
```diff
@@ -0,0 +1,115 @@
...
+        overdue_by = timesince(task.expected_next_run)
+        lines.append(
+            f"- {task.name}\n"
+            f"  Last run: {task.last_run_at:%Y-%m-%d %H:%M:%S %Z}\n"
+            f"  Expected next run: {task.expected_next_run:%Y-%m-%d %H:%M:%S %Z}\n"
+            f"  Overdue by: {overdue_by}"
+        )
+
+    message = "\n\n".join(lines)
+    send_mail(subject, message, settings.DEFAULT_FROM_EMAIL, recipients)
```
akatsoulas: Any reason why we are skipping here post office?

kitsune/sumo/tasks.py
```diff
@@ -36,3 +39,33 @@ def process_queued_mail():
...
+    if not settings.WATCHDOG_EMAIL_RECIPIENTS:
+        return
+
+    overdue_tasks = get_overdue_tasks()
+    if not overdue_tasks:
+        return
+
+    try:
+        redis_conn = redis_client("default")
+    except RedisError:
```
akatsoulas: do we need a test for this?

kitsune/sumo/tests/test_watchdog.py
```diff
@@ -0,0 +1,540 @@
...
+        }
+    )
+    @mock.patch("kitsune.sumo.tasks.try_alert", return_value=True)
+    @mock.patch("kitsune.sumo.tasks.send_email_alert")
+    @mock.patch("kitsune.sumo.tasks.redis_client")
+    @mock.patch("kitsune.sumo.watchdog.PeriodicTask")
+    def test_sends_alert_for_overdue_task(
+        self, MockPeriodicTask, mock_redis_client, mock_send_email, mock_try_alert
+    ):
+        from kitsune.sumo.tasks import watchdog
```
akatsoulas: Let's move all imports top level unless there's a circular dependency

akatsoulas (changes requested):

kitsune/sumo/watchdog.py
escattone: Great catch. That's a huge hole in this current iteration. I'll look into resolving this.

kitsune/sumo/watchdog.py
escattone: We're still using `django-post-office` because Django's `send_mail` gets it connection through our configured email backend of `post_office.EmailBackend`, which in turn uses `django_ses.SESBackend`.

kitsune/sumo/watchdog.py
escattone: I missed that we had a task with a non-uniform schedule. In the case of `cleanup_old_spam`, it'll return either 2 days or 3 days depending on when it's run. Given that our default value for `WATCHDOG_ALLOWED_MISSED_RUNS` allows a grace period of 1 skipped run, it should still work. This function was one of the trickiest to work out, but I'll take another look.

kitsune/sumo/watchdog.py
escattone: This has been resolved with the new approach based on Celery's `task_success` signal and the new `TaskHealth` model.

kitsune/sumo/watchdog.py
escattone: The new `next_run_after` and `nth_run_after` functions are simpler and work with non-uniform `crontab` schedules.

kitsune/sumo/watchdog.py
escattone: This has been resolved in the new code as well.

kitsune/sumo/tasks.py
escattone: Added.

kitsune/sumo/tests/test_watchdog.py
escattone: Done.

escattone:
Thanks for the review @akatsoulas! This is ready for another pass. It's so much better now:

- Uses Celery's `task_success` signal to record periodic task completions **ONLY** if they succeed.
- Can handle non-uniform `crontab` schedules.
- Thorough tests.
- Provides a read-only admin interface.
- Re-organized as its own app.

kitsune/sumo/watchdog.py
escattone: Done.

kitsune/watchdog/core.py
```diff
@@ -0,0 +1,153 @@
...
+    ]
+    if not watched:
+        return []
+
+    # Ensure a TaskHealth row exists for every watched task with a single
+    # bulk insert (ignore_conflicts skips already-existing rows), then fetch
+    # them all in one query. Race-safe against the task_success signal
+    # handler concurrently inserting rows for the same names.
+    watched_names = [name for name, _ in watched]
+    TaskHealth.objects.bulk_create(
```
akatsoulas, line 73: The method is called `get_overdue_tasks` but it does more than that - there's a mutating state by creating objects in the db. We should either rename the function or move the object creation out of the function.

kitsune/watchdog/core.py
```diff
@@ -0,0 +1,153 @@
...
+    (no completions yet) this same logic gives a per-task grace period
+    proportional to the schedule's interval.
+    """
+    if not settings.CELERY_BEAT_SCHEDULE:
+        return []
+
+    excluded = set(settings.WATCHDOG_EXCLUDED_TASKS)
+    watched = [
+        (name, config["schedule"])
+        for name, config in settings.CELERY_BEAT_SCHEDULE.items()
```
akatsoulas: Why are we excluding everything else? We don't have for example `timedelta` for now but we are silently ignoring them. At least we should log something

kitsune/watchdog/apps.py
```diff
@@ -0,0 +1,10 @@
+from django.apps import AppConfig
+
+
+class WatchdogConfig(AppConfig):
+    name = "kitsune.watchdog"
+    default_auto_field = "django.db.models.AutoField"
```
akatsoulas, line 6: why is this needed?

kitsune/watchdog/tasks.py
```diff
@@ -0,0 +1,46 @@
...
+
+    prune_stale_health_records()
+
+    overdue_tasks = get_overdue_tasks()
+    if not overdue_tasks:
+        return
+
+    try:
+        redis_conn = redis_client("default")
+    except RedisError:
```
akatsoulas: If we have an error, we are going to send a few emails probably that don't make sense.

kitsune/settings.py
```diff
@@ -1420,3 +1421,15 @@ def setup_logger(name, formatter_str, level=logging.DEBUG):
     CELERY_BEAT_SCHEDULE = PERIODIC_TASKS_ALL
     if not STAGE:
         CELERY_BEAT_SCHEDULE |= PERIODIC_TASKS_PRODUCTION_ONLY
+
+# Reverse map for the watchdog's task_success signal handler, which receives
+# the dotted-path Celery task name and needs the schedule's display name.
+CELERY_BEAT_NAME_BY_TASK = {config["task"]: name for name, config in CELERY_BEAT_SCHEDULE.items()}
```
akatsoulas: let's rename `config` to something like `cfg` or similar to help readability b/c it classes with the `config` from decouple. Obviously different scope so there's no problem with that, just to make the code more obvious

akatsoulas (approved):
r+wc

kitsune/watchdog/core.py
escattone: Good point. Since the `remaining_estimate` method is available on all schedules, we don't need to limit to just `crontab`'s.

kitsune/watchdog/apps.py
escattone: You're right, it's not necessary. It's there only to maintain consistency with all of our other `AppConfig` declarations.

kitsune/watchdog/tasks.py
escattone: The emails will still make sense, in the sense of being accurate, but we'll get more than one alert for a specific task per day (the default value of `WATCHDOG_ALERT_COOLDOWN_SECONDS`).

escattone: I removed the need for Redis entirely.

PR #7433, opened 2026-04-20
Search ES if no match is found client-side.

akatsoulas:


kitsune/questions/views.py
escattone, line 468: I wonder if we should rate limit `question_tags` to protect against anyone hammering this endpoint?

escattone (approved):
r+wc

kitsune/questions/views.py
akatsoulas: Good catch! I added a ratelimit of 30/m. Hopefully it's not too aggressive

PR #7439, opened 2026-04-21
Sanitize input and upgrade justHTML.

akatsoulas:


escattone (approved):
r+wc

I was hoping there was some kind of workaround for the import of the "private" attribute `_SERIALIZABLE_ATTR_NAME_RE`, but I couldn't find anything either.

Maybe add a test for the linkify case that failed prior to this change?

PR #7443, opened 2026-04-22
add tb_version handling and improve handling

escattone:
mozilla/sumo#2810
mozilla/sumo#2811

Thanks to @denyshon for getting this started with #7258

### DIFF
--- a/kitsune/questions/models.py
+++ b/kitsune/questions/models.py
@@ -293,25 +293,46 @@ class Question(AAQBase):
             for tag in product_config.associated_tags.all():
                 tags.append(tag)
 
-        version = self.metadata.get("ff_version", "")
+        ff_version = self.metadata.get("ff_version", "").strip()
+        tb_version = self.metadata.get("tb_version", "").strip()
+        if ff_version or tb_version:
+            if ff_version:
+                product_name = "Firefox"
+                major_releases = product_details.firefox_history_major_releases
+                stability_releases = product_details.firefox_history_stability_releases
+                development_releases = product_details.firefox_history_development_releases
+                raw_version = ff_version
+            else:
+                product_name = "Thunderbird"
+                major_releases = product_details.thunderbird_history_major_releases
+                stability_releases = product_details.thunderbird_history_stability_releases
+                development_releases = product_details.thunderbird_history_development_releases
+                raw_version = tb_version
 
-        # Remove the beta (b*), aurora (a2) or nightly (a1) suffix.
-        version = re.split("[a-b]", version)[0]
+            # Parse "X", "X.Y", or "X.Y.Z" followed by an optional pre-release
+            # suffix ("aN" for aurora/nightly, "bN" for beta). A suffix drives
+            # the "beta" tag iff this version had a beta cycle in
+            # development_releases — product_details tracks betas but not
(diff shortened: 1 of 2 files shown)


kitsune/questions/models.py
```diff
@@ -293,25 +293,46 @@ def handle_metadata_tags(self, action):
...
-            or version in product_details.firefox_history_major_releases
-        ):
-            tags.append("Firefox {}".format(version))
-            tenths = _tenths_version(version)
-            if tenths:
-                tags.append("Firefox {}".format(tenths))
-        elif _has_beta(version, dev_releases):
-            tags.append("Firefox {}".format(version))
-            tags.append("beta")
+        ff_version = self.metadata.get("ff_version", "").strip()
```
denyshon: We don't need strip() here and below since it's done by Django (by default, but I've also added an explicit parameter passing to ensure that). Unless you want to be even more sure, of course :D Just a nit, it doesn't make any harm anyway

escattone: Ah, I see that now, thanks @denyshon!

denyshon (commented):
Thank you very much @escattone for working on this! This code is definitely cleaner than the previously existing one. The only thing I wonder about is if we could determine beta/alpha with product_details.firefox_versions / product_details.thunderbird_versions, like in https://github.com/mozilla/kitsune/blob/main/kitsune/products/management/commands/sync_product_versions.py
What do you think?

escattone:
> The only thing I wonder about is if we could determine beta/alpha with product_details.firefox_versions / product_details.thunderbird_versions, like in https://github.com/mozilla/kitsune/blob/main/kitsune/products/management/commands/sync_product_versions.py
What do you think?

@denyshon I think that approach would be limited to just the latest versions, but we need to handle historic versions as well, as least within the last few years. So, I would stay with what we have, unless I've misunderstood your question?

denyshon:
Oh right, it's me misunderstanding the "beta" tag. It's fine to keep this version then.
TBH I don't think the tag is very useful now, but just labelling versions according to the release schedule will probably also not make much sense... so let's leave it as it is.

akatsoulas (approved):

denyshon (approved):

kitsune/questions/tests/test_models.py
```diff
@@ -264,21 +263,96 @@ def test_auto_tagging_restraint(self):
...
+        mock_pd.thunderbird_history_stability_releases = {}
+        mock_pd.thunderbird_history_development_releases = {"129.0b2": "2024-08-12"}
+        q = self.question
+        q.add_metadata(tb_version="129.0b2")
+        q.save()
+        q.auto_tag()
+        tags_eq(q, ["Thunderbird 129.0", "beta"])
+
+    @mock.patch("kitsune.questions.models.product_details")
+    def test_auto_tagging_strips_whitespace_and_trailing_noise(self, mock_pd):
```
denyshon, line 291: Django's stripping happens in the form fields, not in the model. That's probably the reason this test doesn't work

denyshon: For reference, see https://github.com/mozilla/kitsune/blob/f074f5c730031a9f3d29b58c523896819dd5464b/kitsune/questions/forms.py#L121-L127 and below

PR #7449, opened 2026-04-24
fix hover highlight on tags

escattone:
mozilla/sumo#2986

image


ceacloud-hue:
Status: Out of office. We are monitoring the situation.

akatsoulas (approved):

Issue #3012, opened 2018-01-15
Database Architecture


glogiotatidis:
- Kitsune uses MySQL (See #2937) on SCL3.
 - On AWS Kitsune will running in multiple regions
 - Since all visitors can potentially write on the database (from the simple task to subscribe to a question to the more complex editing and managing of KB articles), we can't have an admin only domain as we do with other apps. We want visitors on `support.mozilla.org` to be able to write to database and thus we need a database solution which supports writes from multiple locations. 
 - Recently AWS announced Aurora Multi Master support. Aurora is MySQL compatible so we could "just switch" with the downside of getting more locked-in into AWS.  Either way Aurora only supports multi-AZ Multi Master at the moment (2018-01-15) with a plan to support multi-Region in the future. We can re-eval when multi-Region becomes available but for the time being we need an alternative plan.

## Suggested Database Architecture (2nd iteration)

 - One Multi-AZ RDS Database in Oregon running MySQL
 - One Multi-AZ RDS Database in Frankfurt, replicating Oregon's DB
 - Replication from SCL3 to AWS
   - RDS will be promoted to master and the replication will stop when we flip the switch to AWS
 - SUMO will run on two different k8s clusters, in two different AZs but in the same AWS region
   - This will allow having the same RDS instance, while protecting as from cluster level and AZ level catastrophic failures
   - In the case of a AWS Region failure, an automatic failover (using Route 53 Traffic Policies) will route traffic to an App instance running on Frankfurt based k8s cluster. This cluster will read from the Frankfurt based RDS database replica.
   - We can manually upgrade the replica to master and make Frankfurt a r/w instance.

 

![image](https://user-images.githubusercontent.com/584352/35628242-f4289d20-06a3-11e8-86de-e6b6659cd314.png)

glogiotatidis:
In yesterday's DB mtg the following happened:

 - We agreed continue fwd with the plan to establish replication between SCL3 and AWS
 - We agreed to develop and deploy the infra needed to replicate RDS from Oregon to other regions
 - We agreed to develop and deploy Kitsune backing services (ElasticSearch, ElasticCache) in more than one regions
 - We agreed to deploy Kitsune in more than one regions

 - There were concerns expressed about the stability and performance of the of the link to the Master database from other regions. It's suggested the we initially we keep other regions as hot spares and focus on developing a automatic fail-over to a hot spare in case Oregon dies.
[...]

PR #7452, opened 2026-04-27
Allow saving drafts when editing documents

### LINKED ISSUE #3010: Allow saving drafts when editing KB articles
Currently, the draft functionality is only available when translating a document. We should also allow saving drafts when editing a document in the default locale (en-US).

_Created from [bug 1761997](https://bugzilla.mozilla.org/show_bug.cgi?id=1761997)._

denyshon:
- Extend the draft functionality to the edit_document case. Update the related names/comments.
- Remove an unused import in edit_metadata.html

Resolves [#3010](https://github.com/mozilla/sumo/issues/3010).

Limitations:
- The "expiry date" (corresponding to the `expires` field) is not included in drafts. See also [#3012](https://github.com/mozilla/sumo/issues/3012)
- This PR doesn't enable drafts when creating a new document.

### DIFF
--- a/kitsune/wiki/jinja2/wiki/edit.html
+++ b/kitsune/wiki/jinja2/wiki/edit.html
@@ -23,6 +23,19 @@
         {{ _('This document is restricted.') }}
       </div>
     {% endif %}
+    {% if draft_revision %}
+      <li class="mzp-c-notification-bar mzp-t-info info">
+        <button class="mzp-c-notification-bar-button" type="button"></button>
+        <div>
+        <p><strong>{{ _('You have a draft revision for this article saved on {date_time}')|f(date_time=draft_revision.created) }}</strong></p>
+        <form action="{{ url('wiki.edit_document', document.slug) }}" method="post">
+          {% csrf_token %}
+          <input class="btn" name="restore" value={{ _('Restore') }} type="submit">
+          <input class="btn" name="discard" value={{ _('Discard') }} type="submit">
+        </form>
+        </div>
+      </li>
+    {% endif %}
     {{ edit_messages(document, show_revision_warning) }}
     {{ document_lock_warning() }}
   </div>
@@ -42,7 +55,7 @@
           <a href="{{ url }}">Support Document Guide</a>.{% endtrans %}
       </p>
       {{ errorlist(revision_form) }}
-      <form action="" method="post">
+      <form id="edit_form" action="" method="post">
         {% csrf_token %}
           {% for field in revision_form.visible_fields() %}
             <div class="field {% if field.name == 'content' %}has-large-textarea{% endif %}">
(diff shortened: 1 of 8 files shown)


kitsune/wiki/views.py
```diff
@@ -574,18 +574,45 @@ def edit_document(request, document_slug, revision_id=None):
...
+                        "summary": draft_data.get("summary", ""),
+                        "keywords": draft_data.get("keywords", ""),
+                        "based_on": draft_data.get("based_on"),
+                    }
+                )
+                rev_form = RevisionForm(instance=rev, initial=rev_initial)
+        else:
+            rev_form = RevisionForm(request.POST)
+            rev_form.instance.document = doc  # for rev_form.clean()
+            if rev_form.is_valid():
```
akatsoulas, line 610: Shouldn't also delete the draft in this case?

kitsune/wiki/views.py
```diff
@@ -574,18 +574,45 @@ def edit_document(request, document_slug, revision_id=None):
         return init_check
     user, doc, rev = init_check
 
-    rev_form = RevisionForm(instance=rev, initial={"based_on": rev.id, "comment": ""})
+    rev_initial = {"based_on": rev.id, "comment": ""}
+    rev_form = RevisionForm(instance=rev, initial=rev_initial)
+
+    # Check if the user has draft revision saved for the document (with the default locale)
+    content_manager = ManualContentManager()
+    draft = content_manager.get_draft(user, doc, settings.WIKI_DEFAULT_LANGUAGE)
```
akatsoulas: Shouldn't we use here the `request.LANGUAGE_CODE` to retrieve the draft? I believe that's what we are using to save the draft

kitsune/wiki/jinja2/wiki/edit.html
```diff
@@ -63,15 +76,18 @@ <h1 class="sumo-page-heading">{{ _('<em>Editing Content For:</em><br>{title}')|f
...
+          <div id="preview"></div>
+          <div id="preview-diff">
+            <div class="from">{{ revision_form.content.value() }}</div>
+            <div class="to"></div>
+            <div class="output"></div>
+          </div>
+          <div class="submit" id="preview-bottom">
+            {{ submit_revision(revision_form, buttons_only=True, include_diff=True, allow_draft=True) }}
+          </div>
+          <div id="draft-message" class="alert" hidden>
```
akatsoulas: I believe a `</div> is missing for the `button-and-preview` class

denyshon: Oh, indeed

kitsune/wiki/views.py
denyshon: De facto, we only use this view for en-US, so there will be no difference. But you are right, `request.LANGUAGE_CODE` would be more correct here

kitsune/wiki/views.py
denyshon: Indeed, thanks for spotting this! I've added the piece of code deleting the draft.
One nit: I think it should be placed after saving the form, so I've moved it in translate() too.

denyshon:
@akatsoulas thank you for the review, PR updated! Apart from the changes mentioned in the comments, I've also changed `user-messages` from `<div>` to `<ul>` in edit.html (similarly to translate.html)

akatsoulas (approved):

PR #7454, opened 2026-04-27
select default show filter based on contributor status

escattone:
mozilla/sumo#3013

This PR doesn't change the default order of the questions. That remains at `updated`. If we'd prefer to change that to `views`, that's left for a separate PR.


akatsoulas (approved):
r+ with tests passing

PR #7456, opened 2026-04-27
disallow links in profile bios

escattone:
mozilla/sumo#2994


escattone:
Closing, since we're probably going to allow links.

PR #7457, opened 2026-04-28
Do not display tabs on url queries

akatsoulas:


denyshon (commented):
This will redirect to the default tab if the user doesn't have spam questions anymore, but I still think that's less intuitive than just keeping the tab view

akatsoulas:
Thanks @denyshon. I replied in the other PR. Linking here for clarity
https://github.com/mozilla/kitsune/pull/7408#discussion_r3153370347

PR #7458, opened 2026-05-01
add periodic task for active support ticket sync

escattone:
mozilla/sumo#2857

### Assumptions
- Rate-limiting is handled by `zenpy`. We're using `zenpy`'s default reactive handling that occurs when we receive a 429 from Zendesk (sleep for second, check if budget allows, repeat). This is probably not an issue as long as we only have 10's/100's of active tickets at any moment in time rather than thousands.

### DIFF
--- a/kitsune/customercare/tasks.py
+++ b/kitsune/customercare/tasks.py
@@ -1,7 +1,7 @@
 import logging
 from datetime import timedelta
 
-from celery import shared_task
+from celery import group, shared_task
 from django.contrib.auth.models import User
 from django.contrib.contenttypes.models import ContentType
 from django.db import transaction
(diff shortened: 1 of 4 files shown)


kitsune/customercare/tasks.py
```diff
@@ -189,3 +192,52 @@ def process_zendesk_update(payload: dict) -> None:
...
+            .get(id=ticket_id)
+        )
+    except SupportTicket.DoesNotExist:
+        return
+
+    sync_ticket_from_zendesk(ticket)
+
+
+@shared_task
+def sync_active_support_tickets() -> None:
```
akatsoulas, line 218: shall we skip when readonly?

kitsune/customercare/tasks.py
```diff
@@ -189,3 +192,52 @@ def process_zendesk_update(payload: dict) -> None:
...
+            zd_status__in=active_statuses,
+            zendesk_ticket_id__isnull=False,
+        )
+        .exclude(zendesk_ticket_id="")
+        .values_list("id", flat=True)
+    )
+
+    dispatched = 0
+    for ticket_id in ticket_ids:
+        sync_support_ticket.delay(ticket_id)
```
akatsoulas: we can use `group` from celery here

akatsoulas (approved):
r+wc

kitsune/customercare/tasks.py
escattone: Good idea!

PR #7461, opened 2026-05-04
fix favicon on Firefox for Android

escattone:
mozilla/sumo#2889

This is another attempt to fix the favicon for SUMO on Firefox for Android. It also ensures that all of our favicons are the same and satisfy the brand guidelines.

## Rationale

The previous markup gave Fenix two broken candidates:

1. An ICO file (`type="image/x-icon"`) — Fenix's icon parser de-prioritizes ICO on Android, so it tended to skip it.
2. A PNG declared as 192×192 that was actually 196×196 — Fenix uses the declared sizes to rank candidates, and the size mismatch likely caused it to drop the candidate.

With both candidates failing, Fenix had nothing to display.

The fix:
- added an SVG favicon (Fenix prefers SVG since v90, no size or decoder issues)
- replaced the mismatched PNG with properly-sized 16×16 and 32×32 PNGs
- demoted the ICO to a shortcut icon legacy fallback
- added a 192×192 manifest entry for speed dial.

Now Fenix has a clean primary candidate and consistent fallbacks.

Caveat: Fenix caches favicons aggressively, so post-deploy verification should use a fresh profile or cleared site data.


kitsune/sumo/views.py
```diff
@@ -58,7 +58,7 @@ def geoip_suggestion(request):
         with translation.override(locale):
             # This is using our JS-style string formatting.
             response[locale] = {
-                "suggestion": _("Would you like to view this page in " "%(language)s instead?"),
+                "suggestion": _("Would you like to view this page in %(language)s instead?"),
```
akatsoulas, line 61: This introduces a bit of scope creep. The change is subtle so probably is OK this time but we should  avoid this

kitsune/sumo/views.py
```diff
@@ -151,7 +161,7 @@ def test_memcached(host, port):
 def serve_cors(*args, **kwargs):
     """A wrapper around django.views.static.serve that adds CORS headers."""
     if not settings.DEBUG:
-        raise RuntimeError("Don't use kitsune.sumo.views.serve_cors " "in production.")
+        raise RuntimeError("Don't use kitsune.sumo.views.serve_cors in production.")
```
akatsoulas, line 164: Similar comment with the previous

akatsoulas (approved):

PR #7465, opened 2026-05-04
account for owner filter with tag counters

escattone:
mozilla/sumo#3020

Requires a re-index after merging/deploying.


kitsune/questions/views.py
```diff
@@ -358,7 +358,7 @@ def question_list(request, product_slug=None, topic_slug=None):
 
     try:
         questions_page = simple_paginate(request, question_qs, per_page=config.QUESTIONS_PER_PAGE)
-    except (PageNotAnInteger, EmptyPage):
```
akatsoulas, line 361: This is not part of the scope. Although 3.14 allows for it, the `pyproject.toml` file declares python `>=3.11` which will break with this

akatsoulas (approved):
r+wc

PR #7484, opened 2026-05-12
allow replies to support tickets (synchronous)

escattone:
mozilla/sumo#2923

### DIFF
--- a/kitsune/customercare/forms.py
+++ b/kitsune/customercare/forms.py
@@ -248,3 +248,21 @@ class ZendeskForm(forms.Form):
         zendesk_submission_classifier.delay(submission.id)
 
         return submission
+
+
+class SupportTicketReplyForm(forms.Form):
+    """Form for a ticket owner to author a reply that will be posted to Zendesk."""
+
+    # Matches Zendesk's documented per-comment limit:
+    # https://support.zendesk.com/hc/en-us/articles/4408820879258
+    BODY_MAX_LENGTH = 65_535
+
+    body = forms.CharField(
+        label=_lazy("Reply"),
+        widget=forms.Textarea(attrs={"maxlength": BODY_MAX_LENGTH}),
+        max_length=BODY_MAX_LENGTH,
+        error_messages={
+            "required": _lazy("Please enter a reply."),
+            "max_length": _lazy("Replies are limited to %(limit_value)d characters."),
+        },
+    )
--- a/kitsune/customercare/jinja2/customercare/includes/ticket_replies.html
+++ b/kitsune/customercare/jinja2/customercare/includes/ticket_replies.html
@@ -1,20 +1,22 @@
 {% from "sumo/includes/thread_post.html" import post_body %}
 
+<div id="thread-replies">
 {% if sync_error %}
   <div class="thread-detail--notice thread-detail--notice--info">
     {{ _('We could not fetch the latest replies. Please try again later.') }}
   </div>
 {% endif %}
 
-{% if ticket.public_comments %}
+{% set sorted_comments = ticket.public_comments | sort(attribute='id') %}
+{% if sorted_comments %}
   <h2 class="sumo-card-heading">
-    {% trans count=ticket.public_comments | length %}
+    {% trans count=sorted_comments | length %}
       {{ count }} Reply
     {% pluralize %}
       {{ count }} Replies
     {% endtrans %}
   </h2>
-  {% for comment in ticket.public_comments %}
+  {% for comment in sorted_comments %}
     {% set _author = comment.get("author", {}) %}
     {% set _created_str = comment.get("created_at", "") %}
     {% set _created_dt = _created_str | parse_datetime %}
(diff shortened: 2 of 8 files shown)


kitsune/customercare/views.py
```diff
@@ -33,6 +60,135 @@ def _ticket_needs_sync(ticket):
     return ticket.last_synced_at < timezone.now() - threshold
 
 
+def _user_can_reply(user, ticket):
+    if not user.is_authenticated:
+        return False
+    if ticket.zd_status not in ACTIVE_ZD_STATUSES:
+        return False
+    return (user.id == ticket.user_id) or user.has_perm("customercare.change_supportticket")
```
akatsoulas, line 68: Who should have the permission to change the supporticket?

kitsune/customercare/views.py
```diff
@@ -33,6 +60,135 @@ def _ticket_needs_sync(ticket):
     return ticket.last_synced_at < timezone.now() - threshold
 
 
+def _user_can_reply(user, ticket):
+    if not user.is_authenticated:
+        return False
+    if ticket.zd_status not in ACTIVE_ZD_STATUSES:
```
akatsoulas, line 66: Why is this needed? If a ticket is closed, we can automatically open it when a reply is submitted. Through emails, a user can reply to an old thread without any limitation. We should do the same

kitsune/customercare/views.py
```diff
@@ -33,6 +60,135 @@ def _ticket_needs_sync(ticket):
...
+    try:
+        ticket_audit = ZendeskClient(timeout=ZENDESK_REPLY_TIMEOUT).add_ticket_comment(
+            user=request.user,
+            ticket_id=ticket.zendesk_ticket_id,
+            comment_body=body,
+            public=True,
+        )
+    except (APIException, requests.exceptions.RequestException, ValueError) as exc:
+        log.exception(f"Failed to post reply to Zendesk ticket {ticket.zendesk_ticket_id}")
+        if _is_permanent_zendesk_error(exc):
```
akatsoulas, line 163: why the need for this complexity? Can't we just catch the exception and log it along with the message? This feels redundant at least at first sight

kitsune/customercare/views.py
```diff
@@ -33,6 +60,135 @@ def _ticket_needs_sync(ticket):
...
+    return {
+        "ticket": ticket,
+        "sync_error": sync_error,
+        "post_error": post_error,
+        "reply_form": reply_form or SupportTicketReplyForm(),
+        "can_reply": _user_can_reply(viewer, ticket),
+    }
+
+
+def _extract_new_comment(ticket_audit):
```
akatsoulas, line 81: Why all of these helper internal methods? Most if not all are specific to the view and the view should hold that logic. Fragmenting the functionality in many methods hurts readability imo and adds an extra layer of complexity

kitsune/customercare/views.py
```diff
@@ -9,20 +9,47 @@
...
 
+ACTIVE_ZD_STATUSES = {
+    SupportTicket.ZD_STATUS_OPEN,
+    SupportTicket.ZD_STATUS_PENDING,
+    SupportTicket.ZD_STATUS_WAITING,
+}
+
+# Per-HTTP-request timeout for the user-facing reply POST. Bounds worst-case wall time
+# at ~2x this (first-time replier triggers an extra create_user call).
+ZENDESK_REPLY_TIMEOUT = 10
```
akatsoulas, line 38: This is probably the only variable that we need here. I would move it to settings if that's the case

kitsune/customercare/forms.py
```diff
@@ -248,3 +248,21 @@ def send(self, user, product):
...
+class SupportTicketReplyForm(forms.Form):
+    """Form for a ticket owner to author a reply that will be posted to Zendesk."""
+
+    # Matches Zendesk's documented per-comment limit:
+    # https://support.zendesk.com/hc/en-us/articles/4408820879258
+    BODY_MAX_LENGTH = 65_535
+
+    body = forms.CharField(
+        label=_lazy("Reply"),
+        widget=forms.Textarea(attrs={"maxlength": BODY_MAX_LENGTH}),
```
akatsoulas, line 262: We could have a countdown counter instead or completely omit this. It doesn't add much value to inform a user that has so many characters

kitsune/customercare/forms.py
```diff
@@ -248,3 +248,21 @@ def send(self, user, product):
...
 
         return submission
+
+
+class SupportTicketReplyForm(forms.Form):
+    """Form for a ticket owner to author a reply that will be posted to Zendesk."""
+
+    # Matches Zendesk's documented per-comment limit:
+    # https://support.zendesk.com/hc/en-us/articles/4408820879258
+    BODY_MAX_LENGTH = 65_535
```
akatsoulas, line 258: this feels redundant

kitsune/customercare/jinja2/customercare/includes/ticket_replies.html
```diff
@@ -44,3 +46,44 @@ <h2 class="sumo-card-heading">
 {% else %}
   <p class="thread-comments--empty">{{ _("No replies yet.") }}</p>
 {% endif %}
+
+{% if can_reply %}
```
akatsoulas, line 50: Per view comment, this is not needed

akatsoulas (changes requested):

kitsune/customercare/views.py
```diff
@@ -33,6 +60,135 @@ def _ticket_needs_sync(ticket):
...
+        log.exception(f"Failed to post reply to Zendesk ticket {ticket.zendesk_ticket_id}")
+        if _is_permanent_zendesk_error(exc):
+            post_error = _("Sorry, your reply couldn't be posted. Please contact support.")
+        else:
+            post_error = _("Something went wrong. Please try again.")
+        return _render_replies(request, ticket, is_htmx, reply_form=form, post_error=post_error)
+
+    new_comment = _extract_new_comment(ticket_audit)
+    if new_comment is None:
+        log.warning(f"Zendesk audit had no comment event for ticket {ticket.zendesk_ticket_id}")
```
akatsoulas, line 171: let's log to sentry.

kitsune/customercare/forms.py
escattone: I'll omit it.

escattone:
Closing in favor of #7492

PR #7485, opened 2026-05-12
return early on "falsy" input to wiki_to_safe_html

escattone:
mozilla/sumo#2962


kitsune/sumo/templatetags/jinja_helpers.py
```diff
@@ -273,7 +275,7 @@ def datetimeformat(context, value, format="shortdatetime", use_timesince=False):
                     convert_tzinfo = (
                         Profile.objects.get(user=request.user).timezone or default_tzinfo
                     )
-                except (Profile.DoesNotExist, AttributeError):
```
akatsoulas, line 276: Not part of the scope. Let's remove these changes for now.

akatsoulas: before accepting any of these, we need to change the python versions that the app support - probably in a different PR

akatsoulas (commented):
r+wc

kitsune/sumo/templatetags/jinja_helpers.py
escattone: Yeah, let's change the Python versions. It's getting annoying given that `ruff format` automatically removes the parentheses.

PR #7492, opened 2026-05-14
allow replies to support tickets (simpler async)

escattone:
mozilla/sumo#2923

I've tested this locally while connected to the stage Zendesk service, and it works great.

NOTE: This PR does not address the fact that Zendesk tickets support HTML in replies as well as attachments. Those are left for future PR's.


kitsune/customercare/utils.py
```diff
@@ -15,13 +20,37 @@
...
+# when raised without a response attached (e.g. client-side validation in
+# zenpy.lib.request raises TooManyValuesException with no response).
+PERMANENT_ZENPY_EXCEPTIONS = (
+    RecordNotFoundException,
+    TooManyValuesException,
+    SearchResponseLimitExceeded,
+)
+
+
+def is_permanent_zendesk_error(exc) -> bool:
```
akatsoulas: This introduces some complexity that I am not sure it's justified. It basically allows to re-raise an exception to either send something to Sentry or adjust the UX based on errors. Moreover it captures specific errors but ignores others (eg how about 500, 429). I would propose to completely drop all this code and just use multiple exceptions where this method is called

```python
except (ConnectionError, Timeout):
    do_stuff
except APIException:
    do_stuff
except another_exception:
```

kitsune/customercare/views.py
```diff
@@ -33,6 +35,31 @@ def _ticket_needs_sync(ticket):
     return ticket.last_synced_at < timezone.now() - threshold
 
 
+def _replies_context(ticket, reply_form=None, sync_error=False):
```
akatsoulas: I would argue that we should completely drop this method and the one that follows. They add a layer of indirection when reading the code with little benefit. The repetition is extremely small especially if we only use a single helper method like

```python
def _initial_reply(ticket):
    pending = ticket.effective_pending("comment")
    if pending and pending.get("status") == "failed":
         return {"body": pending["body"]
    return None
```

kitsune/customercare/views.py
```diff
@@ -41,24 +68,91 @@ def ticket_detail(request, username, ticket_id):
         id=ticket_id,
         user__username=username,
     )
-    if not (ticket.user_id == request.user.id or request.user.has_perm("customercare.change_supportticket")):
+
+    if not (
+        ticket.user_id == request.user.id
+        or request.user.has_perm("customercare.change_supportticket")
```
akatsoulas: why do we need this permission check? Only the submitter should have access to the ticket

kitsune/customercare/views.py
```diff
@@ -41,24 +68,91 @@ def ticket_detail(request, username, ticket_id):
...
+    ):
         raise Http404
 
-    if request.headers.get("HX-Request"):
+    is_htmx = bool(request.headers.get("HX-Request"))
+
+    if request.method == "POST":
```
akatsoulas: I would rewrite this in terms of user actions. Instead of checking HTTP vocabulary, we can focus on what the user does. 

Eg and assuming the helper I mentioned before
```python
form = SupportTicketReplyForm(request.POST or None, initial=_initial_reply(ticket)

if form.is_vali():
   ...
   # handle post request and clean data
if is_html and _ticket_needs_Sync(ticket):
   ... 
   # reply handling
return render(request,...)
```

kitsune/customercare/tasks.py
```diff
@@ -137,6 +141,131 @@ def process_failed_zendesk_tickets() -> None:
...
+            exc, APIException | requests.exceptions.RequestException
+        ) and not is_permanent_zendesk_error(exc)
+        _mark_pending_failed(ticket_id, allow_retries=allow_retries)
+        if allow_retries:
+            return
+        raise
+
+
+@contextmanager
+def _locked_ticket(ticket_id, select_related=None):
```
akatsoulas: I am not sure there's big benefit for the selecte_related here and most importantly is very specific to ZD functionality. Can't we use the select_related outside of the locking mechanism, within the scope is needed? Right now, there's a single call to it and 2 other calls pass nothing

kitsune/customercare/tasks.py
```diff
@@ -137,6 +141,131 @@ def process_failed_zendesk_tickets() -> None:
...
+        if select_related:
+            qs = SupportTicket.objects.select_for_update(of=("self",)).select_related(
+                *select_related
+            )
+        else:
+            qs = SupportTicket.objects.select_for_update()
+        yield qs.get(id=ticket_id)
+
+
+def _mark_pending_failed(ticket_id, allow_retries):
```
akatsoulas: I don't believe this method is needed, we can just add these lines to the caller

kitsune/customercare/models.py
```diff
@@ -69,6 +74,7 @@ class SupportTicket(ModelBase):
     zd_updated_at = models.DateTimeField(null=True, blank=True)
     last_synced_at = models.DateTimeField(null=True, blank=True)
     comments = models.JSONField(default=list, blank=True)
+    pending_changes = models.JSONField(default=dict, blank=True)
```
akatsoulas: Introducing this for pending changes has a cascade effect to the whole architecture. This requires to read, modify and write under a row lock for every change. I was thinking that we could introduce a table like

```python
class SupportTicketPendingChange(models.Model):
    ticket = models.ForeignKey(...) 
    kind = models.CharField(...) # comment, attachment, etc
    status = ...
    payload = 
    other_fields_as_needed
    
    class Meta:
         ... # unique constraint on ticket and kind

# then the task can query the new table with skip_locked
def post_reply_to_zendesk(ticket_id):
    with transaction.atomic():
        SupportTicketPendingChanges.objects
                                                           .select_for_udpate(skip_locked=True)
                                                           .filter(ticket_id=ticket_id, ...)
       # do_work_based_on_pending_status
```

This approach will significantly reduce the complexity and will also allow to easily introduce other types of replies apart from comments. Thoughts?

akatsoulas (changes requested):

kitsune/customercare/models.py
escattone: That sounds better to me as well.

kitsune/customercare/tasks.py
escattone: Yeah, I had done some refactoring, and missed that this was no longer useful as a separate helper.

kitsune/customercare/views.py
escattone: I think that's the way it was, but it should be removed.

akatsoulas (approved):
Nice update! r+

PR #7495, opened 2026-05-15
Update question details UI

akatsoulas:


kitsune/questions/jinja2/questions/question_details.html
```diff
@@ -217,6 +180,15 @@
       {% endif %}
     </div>
   {% endif %}
+  {% if pinned_solution %}
+    <div class="solution-tree">
+      {% set answer = pinned_solution %}
+      {% set is_pinned_copy = True %}
+      {% include "questions/includes/answer.html" %}
+      {% set is_pinned_copy = False %}
+      {% set answer = None %}
```
escattone, line 189: I don't think these are needed/useful, because they go out of scope after the `{% endblock %}`?

kitsune/questions/jinja2/questions/question_details.html
```diff
@@ -161,47 +161,10 @@
               datetime=datetimeformat(question.updated, format='longdatetime')) }}
       </em></p>
     {% endif %}
-    {% if question.solution %}
-      {% set solver = question.solver %}
```
escattone, line 165: I don't think we've added this back anywhere, so in the new code the solver is never set?

escattone (commented):
When we pin a solution, do we want to show the "controls" (voting buttons, undo solution button) for both the pinned solution and its copy?

PR #7499, opened 2026-05-18
Switch GeoIP to Fastly

akatsoulas:


kitsune/sumo/jinja2/base.html
```diff
@@ -67,6 +67,9 @@
   {% if canonical_url %}
     <link rel="canonical" href="{{ canonical_url|safe }}" />
   {% endif %}
+
+  <meta name="geo-country-code" content="{{ GEO_COUNTRY_CODE }}" />
+  <meta name="geo-country-name" content="{{ GEO_COUNTRY_NAME }}" />
```
escattone: Our document caching uses the locale as part of its key, but I suspect this might require us to add the country code/name to `DOC_HTML_CACHE_KEY` and update `doc_html_cache_key` and `doc_page_cache`.

Also, are there cases when `GEO_COUNTRY_CODE` and `GEO_COUNTRY_NAME` will be falsy? Maybe check that one or both are truthy before including these?

akatsoulas:
@escattone can you have another pass?

kitsune/sumo/static/sumo/js/geoip-locale.js
```diff
@@ -1,38 +1,13 @@
...
-    .fail(function(error) {
-      console.warn('Error retrieving geoip data');
-      $('#announce-geoip-suggestion').remove();
-    })
-    .always(function() {
-      handleLocale(countryData.country_name);
-    });
-  } else {
-    handleLocale(countryData.country_name);
+  var match = document.cookie.match(/(?:^|; )geoip_country_name=([^;]*)/);
```
escattone, line 4: Since we're only using `geoip_country_name`, do we still need to set the `geoip_country_code` cookie?

kitsune/sumo/static/sumo/js/geoip-locale.js
```diff
@@ -1,38 +1,13 @@
...
+  var value = match ? match[1] : '';
+  // SimpleCookie wraps values with special chars (e.g. spaces in country names) in double quotes
+  if (value.length >= 2 && value.charAt(0) === '"' && value.charAt(value.length - 1) === '"') {
+    value = value.slice(1, -1);
   }
+  handleLocale(decodeURIComponent(value));
```
escattone: Is `decodeURIComponent` necessary?

escattone (approved):
r+wc

kitsune/sumo/jinja2/base.html
akatsoulas: Addressed by the current design: country data is no longer rendered into the (locale-cached) HTML. The middleware sets it as a cookie on the response after the cache layer, and JS reads it client-side, so there's no need to add it to DOC_HTML_CACHE_KEY. The cookie is also only set when the Fastly header is truthy.

kitsune/sumo/static/sumo/js/geoip-locale.js
akatsoulas: No. Nothing reads geoip_country_code (the locale-suggestion map and the Zendesk country field both key off the name), so I dropped it from the middleware and tests.

kitsune/sumo/static/sumo/js/geoip-locale.js
akatsoulas: No, removed it. Django's SimpleCookie doesn't percent-encode values (it double-quotes/backslash-escapes), so the quote-stripping above is the only decoding needed.

PR #7500, opened 2026-05-18
allow status changes to support tickets

escattone:
mozilla/sumo#2922

This PR builds on top of #7492. The status change is performed asynchronously. I've tested this locally while connected to the stage Zendesk service, and it works great.

image

image


escattone:
Closing in favor of #7532.

PR #7516, opened 2026-05-21
Sync tickets upon webhooks

akatsoulas:


kitsune/customercare/tasks.py
```diff
@@ -174,24 +159,7 @@ def process_zendesk_update(payload: dict) -> None:
...
-            comment["created_at"] = updated_at
-            comment["public"] = comment.pop("is_public", False)
-            ticket.comments.append(comment)
-            update_fields.append("comments")
-
-        if update_fields:
-            ticket.zd_updated_at = parse_datetime(updated_at)
-            update_fields.append("zd_updated_at")
-            ticket.save(update_fields=update_fields)
+        sync_ticket_from_zendesk(ticket)
```
escattone: This is making two calls to the Zendesk API while the support ticket's row is locked. The calls to the API could be slow or rate-limited -- the `zenpy` client will wait for the rate-limit to pass -- so this could hold the row for seconds. Would it work to make the Zendesk API calls first, and then lock the ticket for the update?

kitsune/customercare/migrations/0008_normalize_zd_status_values.py
```diff
@@ -0,0 +1,33 @@
...
+
+    uppercase_updated = SupportTicket.objects.exclude(zd_status__isnull=True).update(
+        zd_status=Lower(F("zd_status"))
+    )
+
+    legacy_waiting_updated = SupportTicket.objects.filter(zd_status="waiting").update(
+        zd_status="hold"
+    )
+
+    print(
```
escattone, line 19: Nit. Perhaps use `schema_editor.stdout.write` instead?

kitsune/customercare/utils.py
```diff
@@ -24,6 +24,8 @@ def sync_ticket_from_zendesk(ticket: SupportTicket) -> None:
 
     ticket.zd_status = zd_ticket.status
```
escattone: Should we also explicitly lowercase the `zd_ticket.status` here?

escattone (approved):
r+wc

PR #7519, opened 2026-05-22
Enable ticket history for groups

akatsoulas:


kitsune/customercare/utils.py
```diff
@@ -10,12 +10,37 @@
...
 from kitsune.flagit.models import FlaggedObject
+from kitsune.groups.models import GroupProfile
 from kitsune.llm.spam.classifier import ModerationAction
-from kitsune.products.models import ProductSupportConfig, Topic
+from kitsune.products.models import Product, ProductSupportConfig, Topic
 from kitsune.questions.utils import flag_object
 from kitsune.users.models import Profile
 
 
+def _first_ancestor_org(user, candidates) -> GroupProfile | None:
```
escattone: Is this supposed to return the nearest ancestor org or just the first one it finds? The model says `org_group` should be the nearest ancestor, so I was thinking this would have to sort the candidates by their path length.

kitsune/customercare/views.py
```diff
@@ -33,32 +33,38 @@ def _ticket_needs_sync(ticket):
...
 @login_required
 def ticket_detail(request, username, ticket_id):
     ticket = get_object_or_404(
-        SupportTicket.objects.select_related("product", "topic", "user"),
+        SupportTicket.objects.select_related("product", "topic", "user", "org_group"),
         id=ticket_id,
         user__username=username,
     )
-    if not (ticket.user_id == request.user.id or request.user.has_perm("customercare.change_supportticket")):
+    if not SupportTicket.objects.accessible_to(request.user).filter(id=ticket.id).exists():
```
escattone: I think we can get rid of this gate altogether by moving the `accessible_to(request.user)` up into the queryset that the `get_object_or_404` acts on?

escattone (approved):
r+wc

The two comments above, and also one more. Why doesn't the visibility (`GroupProfileManager.visible`) factor into this?

akatsoulas:
`accessible_to` is a subset of `visible()` by inspecting only group membership regarding of group visibility

PR #7520, opened 2026-05-22
D3 chartjs migration

akatsoulas:


kitsune/sumo/static/sumo/js/charts/index.js
```diff
@@ -0,0 +1,39 @@
...
+  TimeScale,
+  Legend,
+  Tooltip,
+  Title,
+  Filler,
+  CategoryScale,
+} from "chart.js";
+// Side-effect import: registers the date-fns adapter with Chart.js
+import "chartjs-adapter-date-fns";
+import { MatrixController, MatrixElement } from "chartjs-chart-matrix";
```
escattone: These are imported but never used. Maybe remove them and the `chartjs-chart-matrix"` dependency until a use case arises?

kitsune/sumo/static/sumo/js/charts/index.js
```diff
@@ -0,0 +1,39 @@
...
+  CategoryScale,
+  MatrixController,
+  MatrixElement
+);
+
+export function renderLineChart(el, config) {
+  return new Chart(el, config);
+}
+
+export function renderMatrixChart(el, config) {
```
escattone: This is never used either. Maybe remove until a use case arises?

escattone (approved):
r+wc

escattone:
Oh, I forgot one thing. If they don't already comply, I think we should configure the chart colors to comply with the [`Tertiary palette` of the Mozilla brand guidelines](https://brand.mozilla.com/document/231#/-/color), which is intended for data visualizations.

PR #7517, opened 2026-05-21
Add ESR version to Firefox

akatsoulas:

PR #7530, opened 2026-05-26
Sync *_ESR_NEXT & Make only supported ESRs visible

### LINKED ISSUE #3054: Improve ESR version automatic syncing
Currently, we have a couple of issues with how sync_product_versions works for ESR:
- It will only add new ESRs in 3-4 months (since they first appear in the `FIREFOX_ESR_NEXT`/`THUNDERBIRD_ESR_NEXT` fields that we don't handle).
- It does not update visibility for old ESR versions.

Practically, this means that once ESR 153 is released, it will not automatically appear in the version picker, and once ER 140 reaches EOL, it will not disappear.

denyshon:
This PR consists of two patches:
- The first one cleans up some leftovers after #7517, particularly:
  - Remove the hardcoded `esr_major_versions` array. Old ESR versions are never deleted, so we don't need to recreate them, and new ones are synced and created automatically.
  - Remove the `esr_only` property. This was something exclusive for the Enterprise product, and with it removed, it's no longer needed.
- The second patch adds `"FIREFOX_ESR_NEXT"`/`"THUNDERBIRD_ESR_NEXT"` to `esr_keys` and updates ESR visibility, so that only supported ESRs (`*_ESR` and `*_ESR_NEXT`) are visible. Resolves [#3054](https://github.com/mozilla/sumo/issues/3054).

### DIFF
--- a/kitsune/products/management/commands/sync_product_versions.py
+++ b/kitsune/products/management/commands/sync_product_versions.py
@@ -67,8 +67,6 @@ class Command(BaseCommand):
                 "alpha_version_key": "FIREFOX_NIGHTLY",
                 "history_data": product_details.firefox_history_major_releases,
                 "esr_keys": ["FIREFOX_ESR"],
-                "esr_major_versions": [52, 60, 68, 78, 91, 102, 115, 128, 140],
-                "esr_only": False,
             },
             "mobile": {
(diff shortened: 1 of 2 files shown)


akatsoulas (approved):
thanks @denyshon

PR #7532, opened 2026-05-26
allow status changes to support tickets

escattone:
mozilla/sumo#2922

image

### DIFF
--- a/kitsune/customercare/jinja2/customercare/includes/ticket_status.html
+++ b/kitsune/customercare/jinja2/customercare/includes/ticket_status.html
@@ -1,0 +1,61 @@
+{% from "sumo/includes/question_entry.html" import state_badges %}
+
+{% set pending = ticket.pending_change('zd_status') %}
+{% set is_sending = pending and pending.effective_status == 'sending' %}
+{% set is_failed = pending and pending.effective_status == 'failed' %}
+{% set permitted = ticket.permitted_zd_status_targets() %}
+
+{% set _status_url = url('customercare.ticket_status', username=ticket.user.username, ticket_id=ticket.id) %}
+
+{% if permitted or is_sending %}
+  <div id="thread-status" class="thread-detail--status"
+       {% if is_sending %}
+       hx-get="{{ _status_url }}"
+       hx-trigger="load delay:1s, every 2s"
+       hx-swap="outerHTML"
+       {% endif %}>
+
+    {% if is_sending %}
+      <span class="thread-detail--status-busy">
+        <span class="loading-spinner" aria-hidden="true"></span>
+        {{ _('Updating status…') }}
+      </span>
+    {% elif permitted %}
+      <form method="post"
+            action="{{ _status_url }}"
+            hx-post="{{ _status_url }}"
+            hx-target="#thread-status"
+            hx-swap="outerHTML"
+            hx-disabled-elt="find button[type=submit]"
+            class="thread-detail--status-actions">
+        {% csrf_token %}
+        {% if ticket.ZD_STATUS_SOLVED in permitted %}
+          <button type="submit" name="target_status" value="{{ ticket.ZD_STATUS_SOLVED }}"
+                  class="sumo-button secondary-button button-sm">
+            {{ _('Mark as solved') }}
+          </button>
+        {% endif %}
+        {% if ticket.ZD_STATUS_OPEN in permitted %}
+          <button type="submit" name="target_status" value="{{ ticket.ZD_STATUS_OPEN }}"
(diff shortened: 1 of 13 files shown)


kitsune/customercare/views.py
```diff
@@ -128,6 +132,54 @@ def ticket_detail(request, username, ticket_id):
     )
 
 
+@login_required
+@require_http_methods(["GET", "POST"])
+def ticket_status(request, username, ticket_id):
+    ticket = get_object_or_404(SupportTicket, id=ticket_id, user__username=username)
+
+    if request.user.id != ticket.user_id:
```
akatsoulas, line 139: I don't think this is needed. The `get_object_or_404` should cover this

kitsune/users/jinja2/users/questions_contributed.html
```diff
@@ -62,7 +62,8 @@ <h2 class="sumo-page-subheading">
       <ul class="question-entry-list">
         {% set _user_is_contributor = is_contributor(request.user) %}
         {% for question in questions %}
-          <li class="question-entry">
+          {% set _is_inactive_ticket = question.channel == "direct_support" and question.zd_status in ("solved", "closed") %}
```
akatsoulas, line 70: Ideally we should avoid hardcoding strings here

kitsune/customercare/jinja2/customercare/includes/ticket_status.html
```diff
@@ -0,0 +1,61 @@
...
+        {{ _('Updating status…') }}
+      </span>
+    {% elif permitted %}
+      <form method="post"
+            action="{{ _status_url }}"
+            hx-post="{{ _status_url }}"
+            hx-target="#thread-status"
+            hx-swap="outerHTML"
+            hx-disabled-elt="find button[type=submit]"
+            class="thread-detail--status-actions">
```
akatsoulas, line 30: I would move the buttons to the right side for consistency with the rest of the site. I believe we are using the `align-end` class

kitsune/sumo/static/sumo/scss/layout/_thread-detail.scss
```diff
@@ -131,6 +129,28 @@
   }
 }
 
+.thread-detail--status {
+  margin-top: p.$spacing-lg;
+  padding-top: p.$spacing-lg;
+  border-top: 1px solid var(--color-border);
+}
+
+.thread-detail--status-actions {
```
akatsoulas, line 140: by using `align-end` I believe we can drop this

kitsune/customercare/jinja2/customercare/includes/ticket_status.html
```diff
@@ -0,0 +1,61 @@
...
+        {% if ticket.ZD_STATUS_SOLVED in permitted %}
+          <button type="submit" name="target_status" value="{{ ticket.ZD_STATUS_SOLVED }}"
+                  class="sumo-button secondary-button button-sm">
+            {{ _('Mark as solved') }}
+          </button>
+        {% endif %}
+        {% if ticket.ZD_STATUS_OPEN in permitted %}
+          <button type="submit" name="target_status" value="{{ ticket.ZD_STATUS_OPEN }}"
+                  class="sumo-button secondary-button button-sm">
+            {{ _('Reopen') }}
```
akatsoulas, line 41: I wonder if we actually need this. The only case that can be useful is if a user accidentally marks the ticket as solved. This can be easily remedied by having some help text either below the answer box or within the answer box explaining that replying will reopen the ticket.

kitsune/customercare/jinja2/customercare/includes/ticket_status.html
```diff
@@ -0,0 +1,61 @@
...
+            hx-post="{{ _status_url }}"
+            hx-target="#thread-status"
+            hx-swap="outerHTML"
+            hx-disabled-elt="find button[type=submit]"
+            class="thread-detail--status-actions">
+        {% csrf_token %}
+        {% if ticket.ZD_STATUS_SOLVED in permitted %}
+          <button type="submit" name="target_status" value="{{ ticket.ZD_STATUS_SOLVED }}"
+                  class="sumo-button secondary-button button-sm">
+            {{ _('Mark as solved') }}
```
akatsoulas, line 35: If we have a long thread, a user will have to scroll on top to click that button. That small friction could leave tickets open. I believe a better placement could be right below the last question. Contrary to the AAQ, we just need a single button to solve the question here that trails the last question

akatsoulas (approved):
I am approving this to unblock work but we should iterate over the design a bit

PR #7542, opened 2026-05-29
allow support tickets to be marked as solved

escattone:
mozilla/sumo#2922

Replaces #7532.

image

image

After marking as solved:

image

### DIFF
--- a/kitsune/customercare/views.py
+++ b/kitsune/customercare/views.py
@@ -42,27 +42,33 @@ def _ticket_needs_sync(ticket):
     return ticket.last_synced_at < timezone.now() - threshold
 
 
-@login_required
-@require_http_methods(["GET", "POST"])
-def ticket_detail(request, username, ticket_id):
-    ticket = get_object_or_404(
-        SupportTicket.objects.accessible_to(request.user).select_related(
-            "product", "topic", "user", "org_group"
-        ),
-        id=ticket_id,
-        user__username=username,
-    )
-
+def _accessible_ticket(request, username, ticket_id, select_related=None):
+    qs = SupportTicket.objects.accessible_to(request.user)
+    if select_related:
+        qs = qs.select_related(*select_related)
+    ticket = get_object_or_404(qs, id=ticket_id, user__username=username)
     can_reply = ticket.can_reply(request.user)
     if request.method == "POST" and not can_reply:
         raise PermissionDenied
+    return ticket, can_reply
 
-    placeholder = (
+
+def _reply_placeholder(ticket):
+    return (
         _("Reply here to reopen this ticket.")
         if ticket.zd_status == SupportTicket.ZD_STATUS_SOLVED
         else None
     )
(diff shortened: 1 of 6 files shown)


kitsune/customercare/views.py
```diff
@@ -165,6 +165,7 @@ def ticket_detail(request, username, ticket_id):
         "reply_form": form,
         "sync_error": sync_error,
         "reply_error": reply_error,
+        "status_error": False,
```
akatsoulas, line 176: Won't this silently disregard any errors?

kitsune/customercare/views.py
```diff
@@ -177,6 +178,72 @@ def ticket_detail(request, username, ticket_id):
     )
 
 
+@login_required
+@require_POST
+def ticket_mark_solved(request, username, ticket_id):
```
akatsoulas, line 191: the main differences with the ticket_detail view are which ZD method is called. A shared method to render and maybe a SupportTicket one will collapse both views. If we week it as-is we will need to remember to make changes to both places which is susceptible to bugs.

kitsune/customercare/views.py
```diff
@@ -177,6 +178,72 @@ def ticket_detail(request, username, ticket_id):
...
+    ticket = get_object_or_404(
+        SupportTicket.objects.accessible_to(request.user).select_related(
```
akatsoulas: Why are we using `select_related` here?

kitsune/customercare/views.py
```diff
@@ -177,6 +178,72 @@ def ticket_detail(request, username, ticket_id):
...
+    form = SupportTicketReplyForm(placeholder=placeholder)
+
+    context = {
+        "ticket": ticket,
+        "reply_form": form,
+        "sync_error": False,
+        "reply_error": False,
+        "status_error": status_error,
+        "status_changed": status_changed,
+        "can_reply": True,
```
akatsoulas: Let's not hardcode the value here. If the checks ever change on who has access to reply, this will grant owner-level access to non-owners (eg staff)

akatsoulas (approved):
r+wc

kitsune/customercare/views.py
escattone: We don't check for status errors in the `ticket_detail` view. It seems nearly impossible that the reply could succeed but an associated status flip fail. I think it's best to keep this the way it is.

kitsune/customercare/views.py
escattone: The render helper doesn't really make sense given that `ticket_mark_solved` redirects rather than renders in the non-htmx case, but I added a support-ticket helper and a simple reply-placeholder helper.

kitsune/customercare/views.py
escattone: Good question. We only need  `user` here, not the others, and I also confirmed that for the `ticket_detail` view and the rendering of `ticket_detail.html` we don't ever use the `org_group`. It's only used by `accessible_to` and that only uses the `org_group_id` when the query is made.

PR #7544, opened 2026-05-29
Ma topic

akatsoulas:


kitsune/products/migrations/0040_backfill_topic_metadata.py
```diff
@@ -0,0 +1,121 @@
...
+        "t1-browse",
+        "t1-download-and-save",
+        "t1-email-and-messaging",
+        "t1-installation-and-updates",
+        "t1-performance-and-connectivity",
+        "t1-privacy-and-security",
+        "t1-search-tag-and-share",
+        "t1-settings",
+    },
+    "payment": {"t1-billing-and-subscriptions"},
```
escattone: I think this should this be `"payments"` instead of `"payment"`?

kitsune/customercare/tests/test_forms.py
escattone, line 1: Do we need a test or tests to exercise when the `ZendeskTopic.topic` is defined?

kitsune/customercare/tests/test_utils.py
```diff
@@ -138,48 +115,40 @@ def test_three_tier_topic(self):
...
+        )
 
-        tier1 = TopicFactory(title="Billing and subscriptions", parent=None, is_archived=False)
+    def test_no_automation_tag_when_no_zendesk_topic_match(self):
+        """No automation tag is appended when no ZendeskTopic links to the classified Topic."""
+        tier1 = TopicFactory(
+            title="Billing and subscriptions",
+            parent=None,
+            is_archived=False,
+            legacy_tag="payment",
```
escattone: Should this be `"payments"` instead?

escattone (approved):
r+wc

It appears that the legacy tag sent to Zendesk for the `backup-recovery-and-sync` topic changes from `accounts` to `technical`? Just wanted to call it out in case it wasn't intentional.

Additional comments below.

PR #7548, opened 2026-06-01
improve breadcrumbs for ticket detail

escattone:
mozilla/sumo#3072


kitsune/customercare/views.py
```diff
@@ -64,10 +55,18 @@ def _reply_placeholder(ticket):
...
+    ticket = get_object_or_404(
+        SupportTicket.objects.accessible_to(request.user).select_related(
+            "product", "topic", "user", "org_group"
+        ),
+        id=ticket_id,
+        user__username=username,
     )
 
+    can_reply = ticket.can_reply(request.user)
+    if request.method == "POST" and not can_reply:
```
akatsoulas: should we change this to `if not request.GET`?

akatsoulas (approved):

kitsune/customercare/views.py
escattone: Good point, safer.

PR #7553, opened 2026-06-02
Add product version to question cards for contributors

### LINKED ISSUE #2958: Display product version on question cards
Before, product version was displayed among the tags on question cards, making it easy to understand whether the question is related to ESR or release, for example. Now you have to open the question to see the version number. That has to be changed.

denyshon:
Resolves [#2958](https://github.com/mozilla/sumo/issues/2958).

To avoid fetching tags, we can use question metadata to get product versions (and display them). However, if we use raw data from the corresponding fields, this would expose us to all kinds of spam and UI breakages related to user input. In order to prevent that, this patch saves our validated version number (used to apply tags) into a separate question metadata field, and then displays it on question cards. For older questions, we fall back to the raw data from the version fields.

The SCSS changes do not represent any visual change, and are purely for clarity/future.

### DIFF
--- a/kitsune/questions/models.py
+++ b/kitsune/questions/models.py
@@ -200,6 +200,9 @@ class Question(AAQBase):
             if update:
                 self.updated = timezone.now()
 
+        # Clear cached product_slug
+        self._product_slug = None
+
         super().save(*args, **kwargs)
 
         # Ensure that the metadata doesn't contain a "solver_id" if there's no solution.
@@ -214,7 +217,7 @@ class Question(AAQBase):
             if settings.QUESTION_CLASSIFIER_ENABLED:
                 question_classifier.delay(self.id)
 
-    def add_metadata(self, **kwargs):
+    def add_metadata(self, update=False, **kwargs):
         """Add (save to db) the passed in metadata.
 
         Usage:
@@ -223,7 +226,10 @@ class Question(AAQBase):
 
         """
         for key, value in list(kwargs.items()):
-            QuestionMetaData.objects.create(question=self, name=key, value=value)
+            if update:
+                QuestionMetaData.objects.update_or_create(question=self, name=key, defaults={"value": value})
+            else:
+                QuestionMetaData.objects.create(question=self, name=key, value=value)
         self._metadata = None
 
     def clear_mutable_metadata(self):
@@ -254,6 +260,18 @@ class Question(AAQBase):
         return self._metadata
 
     @property
+    def product_version(self):
+        metadata = self.metadata
+        if "sanitized_product_version" in metadata:
+            return metadata.get("sanitized_product_version")
+        elif self.product_slug in ("firefox", "mobile", "ios"):
+            return metadata.get("ff_version")
+        elif self.product_slug in ("thunderbird", "thunderbird-android"):
+            return metadata.get("tb_version")
+        else:
+            return None
+
+    @property
     def solver(self):
         """Get the user that solved the question."""
         solver_id = self.metadata.get("solver_id")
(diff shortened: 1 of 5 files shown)


kitsune/questions/models.py
```diff
@@ -327,14 +328,17 @@ def handle_metadata_tags(self, action):
...
                         tags.append(f"{product_name} {version}")
                         if tenths and tenths != version:
                             tags.append(f"{product_name} {tenths}")
                         tags.append("beta")
                 elif version in major_releases or version in stability_releases:
+                    sanitized_product_version = version
                     tags.append(f"{product_name} {version}")
                     if tenths and tenths != version:
                         tags.append(f"{product_name} {tenths}")
+            self.add_metadata(sanitized_product_version=sanitized_product_version)
```
akatsoulas: This will be called on every update of the question. The `add_metadata` will call basically the `QuestionMetaData.objects.create(...)` but he model has a unique constraint applied. I suspect this is is going to lead to an `IntegrityError` from the DB.

kitsune/sumo/jinja2/sumo/includes/question_entry.html
```diff
@@ -199,6 +200,18 @@ <h3 class="question-entry--title">
           </svg>
           {{ question.product.title }}
         </a>
+        {% if user_is_contributor %}
```
akatsoulas: I wonder if we can push the logic with the nested `if` statements into python in a `@property` like `display_version` to simplify the template to something like

```jinja
{% if user_is_contributor and question.display_version %}
  {{ question.display_version }}
{% endif %}

**Comment by akatsoulas** [comment]
@denyshon let me know when this is good for another pass

File: kitsune/questions/models.py (hunk 1 above)
Comment (akatsoulas) [reply kitsune/questions/models.py]: Another idea here could be to do something like
```python
 if action == "add":
    self.metadata_set.update_or_create(
         name="sanitized_product_version",
          defaults={"value": sanitized_product_version},
      )
      self._metadata = None
```

kitsune/sumo/jinja2/sumo/includes/question_entry.html
denyshon: done

kitsune/questions/models.py
denyshon: addressed

denyshon:
Hi @akatsoulas, I've updated the patch to address your feedback. When you have time, could you please take another look?

kitsune/questions/models.py
```diff
@@ -223,7 +226,10 @@ def add_metadata(self, **kwargs):
 
         """
         for key, value in list(kwargs.items()):
-            QuestionMetaData.objects.create(question=self, name=key, value=value)
+            if update:
+                QuestionMetaData.objects.update_or_create(question=self, name=key, value=value)
```
akatsoulas: I believe we should have something like
```python
QuestionMetaData.objects.update_or_create(question=self, name=key, defaults={"value": value})
```
Otherwise the query will always come back empty because the value won't be in the database and the unique constraint will fire the Integrity error.

kitsune/sumo/jinja2/sumo/includes/question_entry.html
```diff
@@ -193,6 +200,10 @@ <h3 class="question-entry--title">
           </svg>
           {{ question.product.title }}
         </a>
+        {% if user_is_contributor and question.product_version %}
+         {{ question.product_version }}
```
akatsoulas, line 204: This will force a query for each metadata entry, creating a N+1 query problem. We should add `metadata_set` to prefetch_related

kitsune/questions/models.py
denyshon: thank you, done

kitsune/sumo/jinja2/sumo/includes/question_entry.html
denyshon: I've added `prefetch_related("metadata_set")` in two places where this macro is used, and I've also added a check here to make sure that `question.channel != "direct_support"`

denyshon:
Hey @akatsoulas, thank you for reviewing this once again! I've updated the patch to address your comments

akatsoulas (approved):

PR #7569, opened 2026-06-04
handle Zendesk ticket deletions

escattone:
mozilla/sumo#3085

image

image

image

image

image


kitsune/customercare/tests/test_models.py
```diff
@@ -67,6 +67,14 @@ def test_str_method(self):
         self.assertIn("Test Subject", str_repr)
         self.assertIn("processing_failed", str_repr)
 
+    def test_is_zendesk_deleted_reflects_zd_deleted_at(self):
```
akatsoulas: This test is not needed - it's testing the language and not the behavior

kitsune/customercare/tasks.py
```diff
@@ -154,17 +166,31 @@ def process_zendesk_update(payload: dict) -> None:
...
+                return
+
+            ticket.zd_deleted_at = zd_deleted_at
+            ticket.save(update_fields=["zd_deleted_at"])
+
         return
 
-    zd_ticket, zd_comments = fetch_zendesk_ticket_data(str(ticket_id))
+    try:
+        ticket = qs.filter(zd_deleted_at__isnull=True).get()
```
akatsoulas: Given that we introduce more cases on when it's appropriate to sync the tickets, I wonder if we need a manager to handle that. We could very well override the default manager in the model. This will allow us to make changes on what's updated in a single place

kitsune/customercare/models.py
```diff
@@ -128,6 +133,11 @@ def content(self):
     def channel(self):
         return "direct_support"
 
+    @property
+    def is_zendesk_deleted(self):
```
akatsoulas: Although having a timestamp makes a lot of sense and adds metadata that we can potentially need, this `property` also introduces another way of querying about the status of the ticket. Status should only be reflected by the status choices.

With this new addition, the status of a ticket is only derived by looking at two places and it adds complexity to templates and queries. A potential solution could be here either to collapse this into the statuses or use managers in a single place to get the right tickets

kitsune/customercare/tests/test_tasks.py
```diff
@@ -248,59 +251,91 @@ def _payload(self, event_type, *, ticket_id="12345"):
...
+    @patch("kitsune.customercare.tasks.sync_ticket_from_zendesk")
+    def test_soft_deleted_marks_ticket_deleted(self, mock_sync):
+        process_zendesk_update(self._payload("zen:event-type:ticket.soft_deleted"))
+
+        mock_sync.assert_not_called()
+        self.ticket.refresh_from_db()
+        self.assertIsNotNone(self.ticket.zd_deleted_at)
+
+    @patch("kitsune.customercare.tasks.sync_ticket_from_zendesk")
+    def test_permanently_deleted_marks_ticket_deleted(self, mock_sync):
```
akatsoulas: I would also drop this

akatsoulas (changes requested):

kitsune/customercare/models.py
escattone: I agree. Since Zendesk doesn't have a "deleted" status, I'd rather not fold that into `zd_status`, but I'll look into a clean solution.

escattone:
@akatsoulas Ready for another review.

kitsune/sumo/jinja2/sumo/includes/question_entry.html
```diff
@@ -90,41 +90,50 @@
...
 
+{# Status pill for a Direct Support (Zendesk) ticket. Shared by the My Questions
+   list, the ticket detail hero, and the group tickets table so all three render
+   the same .status-label--* colors. The status -> color mapping lives in SCSS
+   (the class is just status-label--<zd_status>), so re-coloring a status is a
+   stylesheet change, not a template one. "size" controls only padding/scale
+   ("sm" default; "lg" for the roomier group table). Tips use a neutral "This
+   ticket" voice so they read correctly on both the owner's pages and the staff
+   table. #}
+{% macro zd_status_badge(ticket, size="sm") %}
```
akatsoulas, line 95: Can we collapse this into a single per status registry?
```python
ZD_STATUS_META = {
    ZDStatus.NEW:     StatusMeta(group="active", css="open",    tip=_("waiting review")),
    ZDStatus.OPEN:    StatusMeta(group="active", css="open",    tip=_("under review")),
    ZDStatus.PENDING: StatusMeta(group="active", css="pending", tip=_("awaiting reply")),
    ZDStatus.HOLD:    StatusMeta(group="active", css="hold",    tip=_("third party")),
    ZDStatus.SOLVED:  StatusMeta(group="solved", css="solved",  tip=_("resolved")),
    ZDStatus.CLOSED:  StatusMeta(group="solved", css="closed",  tip=_("closed")),
}

@dataclass(frozen=True)
class StatusMeta:
   group...
   tip....
```

akatsoulas (approved):
r+wc

PR #7573, opened 2026-06-08
Add the Reviewer badge

### LINKED ISSUE #2847: Badge for KB reviewers with 25 or more reviews in a year
We want to have a badge to recognize KB reviewers with 25 or more reviews in a year. 

Badge name: 2026 Reviewer Badge
Badge slug: 2026-reviewer-badge
Badge description: This badge is awarded to KB reviewers with 25 or more reviews in 2026.
Badge design: [Gdrive](https://drive.google.com/file/d/13WFiauzD3s2r2kmD_JYIhcaGkdWEAzsl/view?usp=drive_link)

denyshon:
Resolves [#2847](https://github.com/mozilla/sumo/issues/2847).

### DIFF
--- a/kitsune/wiki/badges.py
+++ b/kitsune/wiki/badges.py
@@ -20,6 +20,12 @@ WIKI_BADGES = {
         "description": "This badge is awarded to contributors with 10 "
         "approved translations edits during {year}.",
     },
+    "reviewer-badge": {
+        "slug": "{year}-reviewer-badge",
+        "title": "{year} Reviewer Badge",
+        "description": "This badge is awarded to KB reviewers with 25 "
+        "or more reviews in {year}.",
+    },
 }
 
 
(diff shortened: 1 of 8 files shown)


kitsune/wiki/badges.py
```diff
@@ -28,24 +34,32 @@ def on_revision_save(sender, instance, **kwargs):
...
+    * We award the Reviewer badge on 25 en-US reviews.
     """
     rev = instance
     year = rev.created.year
     creator = rev.creator
+    reviewer = rev.reviewer
+
+    # If the revision is reviewed, maybe award the reviewer.
+    if reviewer:
+        from kitsune.wiki.tasks import maybe_award_badge
```
akatsoulas: I would suggest to just import the method instead of using conditionals. This check also forces the duplication of the `import` a few lines later.

kitsune/wiki/badges.py
```diff
@@ -28,24 +34,32 @@ def on_revision_save(sender, instance, **kwargs):
...
     * We award the L10n badge on 10 approved translation edits.
+    * We award the Reviewer badge on 25 en-US reviews.
     """
     rev = instance
     year = rev.created.year
     creator = rev.creator
+    reviewer = rev.reviewer
+
+    # If the revision is reviewed, maybe award the reviewer.
+    if reviewer:
```
akatsoulas: How about adding a check here about the locale? If the locale is different that `en-US` we don't even need to spawn the task

kitsune/wiki/badges.py
```diff
@@ -28,24 +34,32 @@ def on_revision_save(sender, instance, **kwargs):
...
+        from kitsune.wiki.tasks import maybe_award_badge
 
-    # We only care about approved revisions.
-    if not rev.is_approved:
-        return
+        maybe_award_badge.delay(
+            WIKI_BADGES["reviewer-badge"], year, reviewer.id
```
akatsoulas: `year` here is the creation date of the revision and not the review date. Is this intended?

kitsune/wiki/badges.py
denyshon: Fixing this...

kitsune/wiki/badges.py
denyshon: Indeed, let me add this

kitsune/wiki/badges.py
denyshon: Oops, that's not intended. And the same issue is in tasks.py. Let me fix those

denyshon:
@akatsoulas addressed your feedback, thank you so much for the review

akatsoulas (approved):
This is great, thanks @denyshon! r+ 🚀

PR #7582, opened 2026-06-09
Strengthen l10n prompt

akatsoulas:


helmize1978 (approved):

escattone (approved):
As we discussed earlier, my initial suspicion was that we didn't need this PR, but I was mistaken Tasos! Nice work! 

I tested this on stage using revision 180247 (Greek) and the English revision it's based on (171156) as the target and source texts of the prior translation.  The source text of the prior translation and the new English source text to be translated each contained the following snippet:

> ;{for win10}[[Image:opensearch-win10]]{/for}{for win7, win8, winxp}[[Image:opensearch-win7]]{/for}{for mac}[[Image:opensearch-mac]]{/for}{for linux}[[Image:opensearch-linux]]{/for}

but the target text of the prior translation modified that snippet to:

> ;{for win}[[Image:OpenSearch-el]]{/for}{for mac}[[Image:opensearch-mac]]{/for}{for linux}[[Image:opensearch-linux]]{/for}

I wanted to see if the changes in this PR would make a difference, and they did! Without the changes in this PR, the snippet from the prior translation was copied as is, but with the changes in this PR, it was translated to:

> ;{for win10}[[Image:OpenSearch-el]]{/for}{for win7, win8, winxp}[[Image:OpenSearch-el]]{/for}{for mac}[[Image:opensearch-mac]]{/for}{for linux}[[Image:opensearch-linux]]{/for}

which was brilliant, because it recognized that it shouldn't copy that snippet from the prior translation, but instead freshly translate it while obeying all of the other rules -- which you can see that it did because it replaced `[[Image:opensearch-win10]]` with `[[Image:OpenSearch-el]]`.

helmize1978 (approved):

Issue #7591, opened 2026-06-12
API: /api/2/question/ intermittently returns HTTP 500 with ordering=updated + updated__gt


aatchison:
Hi SUMO team — I'm from **Thunderbird**, working to fix our **Bitergia / GrimoireLab** analytics ingestion of SUMO data. Our harvester pulls questions/answers via the public REST API and has stalled (~7 weeks behind) on intermittent 500s. Sharing a reproducible characterization in case it points at a fixable query/index issue; happy to provide more detail.

## Summary
`GET /api/2/question/` **intermittently returns HTTP 500** (server error page) when **`ordering=updated`** is combined with an **`updated__gt=<datetime>`** range filter over a large result window. Removing either parameter consistently returns 200.

## Reproduction (public, unauthenticated; observed 2026-06-12 UTC; `server: nginx`, `via: … varnish`)
**Returns 200:**
- `GET /api/2/question/?format=json` (count ~770,569)
- `GET /api/2/question/?ordering=updated&format=json`
- `GET /api/2/question/?updated__gt=2026-04-24T10:36:15Z&format=json`  *(no ordering; count ~9,404)*
- `GET /api/2/question/?ordering=updated&updated__gt=2026-06-01T00:00:00Z&format=json`  *(smaller/recent window)*

**Returns HTTP 500 (HTML error page, ~0.5–1s):**
- `GET /api/2/question/?ordering=updated&updated__gt=2026-04-24T10:36:15Z&format=json`
- `GET /api/2/question/?ordering=updated&updated__gt=2026-05-20T00:00:00Z&format=json`

## Observations
- **Intermittent / load-dependent.** The *same* query flips between 500 and 200 within minutes (e.g. `updated__gt=2026-04-24` gave 500, then 200; `2026-05-20` gave 200, then 500). So it doesn't look like a single malformed record — more like a server-side timeout/error under load.
- **Trigger is the combination** of `ordering=updated` and `updated__gt` over a large window; dropping either avoids it. This suggests the `ORDER BY updated` + `updated__gt` range query may be timing out or missing an efficient index for that access pattern.
- We also see **HTTP 429** rate-limiting under normal incremental pagination.

## Impact / ask
[...]

PR #7592, opened 2026-06-12
Make question/answer API profile serialization read-only (#7591)

aatchison:
## Who we are

We're [Thunderbird](https://www.thunderbird.net/) (MZLA). We run a [GrimoireLab/Bitergia](https://github.com/chaoss/grimoirelab) analytics instance at **https://bitergia.thunderbird.net** that ingests SUMO/Kitsune data via the `/api/2/question/` endpoint, which is how we hit this issue (#7591) in production.

## Summary

Fixes the write-on-read N+1 in the question/answer API serializers that is the most likely cause of the intermittent `HTTP 500` on `GET /api/2/question/?ordering=updated&updated__gt=<date>` reported in mozilla/sumo#3124.

`QuestionSerializer` and `AnswerSerializer` resolve each row's `creator`, `updated_by`, `taken_by`, `solution.creator` and every involved answer author through `get_or_create_profile()`, which calls `Profile.objects.get_or_create(user=...)`. `get_or_create` is a **write**, executed here during a **read-only `GET`**, once per related user per row.

## Why this surfaces as an intermittent 500 on the ordered query

- For accounts that already have a `Profile`, `get_or_create` is just a `SELECT`. For accounts **without** one (common for old/imported users), it does an `INSERT` inside a savepoint.
- `?ordering=updated&updated__gt=` walks a window of older questions whose authors are more likely to be profile-less, so a single page can fan out into a burst of per-row writes plus the existing N+1 `SELECT`s.
- The first request that hits such a page does the writes (slow/contended under load → times out → `500`); a retry finds the now-created `Profile`s and succeeds (`200`). That matches the intermittent **500-then-200** behaviour in the issue.

I was not able to reproduce the production `500` directly (it's load/scale dependent), but eliminating writes during `GET` removes a clear timeout/contention source and is correct regardless: **read requests should never write.**

## Change

- `get_or_create_profile()` → `get_profile()`: returns the user's `Profile` or `None`, **never creating one**.
- Serializer methods treat a missing `Profile` as `None`.

This mirrors the read-only graceful fallback the codebase already uses for missing profiles in `profile_avatar()` and `display_name()` (both `try/except` → default, no write), so it preserves the intent of the original *"Gracefully handle missing profiles"* change without the side effect.

### Behaviour change

For the rare profile-less user, the affected fields (`creator`, `updated_by`, `taken_by`, `solved_by`, and entries in `involved`) now serialize as `null`/are omitted, instead of the previous behaviour of silently creating a `Profile` row mid-request. If you'd prefer profiles be guaranteed to exist via a different mechanism (e.g. a signal on user creation / a backfill) rather than returning `null`, I'm happy to adjust — opening early per CONTRIBUTING for discussion.

## Tests

Added regression tests in `kitsune/questions/tests/test_api.py`:

- `test_creator_without_profile_serializes_as_none` — serializing a question whose creator has no `Profile` returns `None` and creates **zero** `Profile` rows.
- `test_list_ordering_by_updated_does_not_write_profiles` — `GET /api/2/question/?ordering=updated&updated__gt=` over profile-less authors returns `200` and creates **zero** `Profile` rows.

Both fail on `main` (the old code re-creates the profile and then errors serializing it) and pass with this change. Full `kitsune.questions.tests.test_api` suite (70 tests) passes locally, `ruff check`/`ruff format` clean.

Refs mozilla/sumo#3124

### DIFF
--- a/kitsune/questions/api.py
+++ b/kitsune/questions/api.py
@@ -36,23 +36,27 @@ from kitsune.users.api import ProfileFKSerializer
 from kitsune.users.models import Profile
 
 
-def get_or_create_profile(user):
+def get_profile(user):
     """
-    Get or create a Profile for a User.
+    Return the Profile for a User without creating one.
 
-    This helper ensures Users without Profiles don't cause serialization failures.
-    Follows Django's get_or_create naming pattern.
+    Serialization happens on read-only (GET) requests, so it must never
+    write. Users without a Profile serialize gracefully as ``None`` (mirroring
+    the read-only fallback already used by ``profile_avatar`` and
+    ``display_name``) rather than triggering a ``get_or_create`` write. That
+    per-row write was an N+1 write-on-read which intermittently timed out
+    (HTTP 500) on large, ordered list queries such as
+    ``?ordering=updated&updated__gt=`` (mozilla/kitsune#7591).
 
     Args:
         user: User instance (can be None for optional fields)
 
     Returns:
-        Profile instance if user is provided, None otherwise
+        Profile instance if the user has one, None otherwise.
     """
     if user is None:
         return None
-    profile, _ = Profile.objects.get_or_create(user=user)
-    return profile
+    return Profile.objects.filter(user=user).first()
 
 
 class QuestionMetaDataSerializer(serializers.ModelSerializer):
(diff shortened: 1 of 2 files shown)


escattone (approved):
@aatchison You're right on. The profile creation was the source of the 500's. I can see them in our Sentry logs. Thanks for your contribution!

escattone:
I'm going to follow-up this PR with one that addresses the N+1 query problems in the question and answer APIs.

PR #7598, opened 2026-06-15
Pass the user selection to llm

akatsoulas:


kitsune/llm/support/classifiers.py
```diff
@@ -122,6 +126,14 @@ def classify_zendesk_submission(submission: SupportTicket) -> dict[str, Any]:
         ),
     }
 
+    # Pass the user's form selection as a hint for topic classification.
+    if submission.category:
+        zd_topic = ZendeskTopic.objects.filter(slug=submission.category).first()
```
escattone: Probably worth adding `.select_related("topic")` to avoid the possible extra query for `zd_topic.topic.title` below.

kitsune/llm/support/classifiers.py
```diff
@@ -64,6 +64,10 @@ def classify_question(question: Question) -> dict[str, Any]:
         ),
     }
 
+    # Pass the user's selected topic as a hint for topic classification.
+    if question.topic_id:
+        payload["selected_topic"] = question.topic.title
```
escattone, line 69: For the case when `_handle_product_reassignment` is called below, and it, in turn, calls `on_reassignment` when the product changes, we should probably clear the `selected_topic` from the payload if it doesn't exist within the new product's taxonomy.

kitsune/llm/support/classifiers.py
```diff
@@ -122,6 +126,14 @@ def classify_zendesk_submission(submission: SupportTicket) -> dict[str, Any]:
         ),
     }
 
+    # Pass the user's form selection as a hint for topic classification.
+    if submission.category:
+        zd_topic = ZendeskTopic.objects.filter(slug=submission.category).first()
+        if zd_topic:
+            payload["selected_topic"] = (
+                zd_topic.topic.title if zd_topic.topic_id else zd_topic.form_title
```
escattone: I don't think it makes sense to fall back to the `form_title` when the `zd_topic` has no associated topic, because if there's no associated topic, doesn't that mean that the `form_title` doesn't exist in the topic taxonomy or doesn't apply to just a single topic? It seems to me that we should only set `selected_topic` if the `zd_topic` has an associated topic, otherwise we should leave it empty.

escattone (approved):

PR #7630, opened 2026-06-18
Make CSP rules stricter

akatsoulas:


kitsune/settings.py
```diff
@@ -1265,14 +1265,12 @@ def setup_logger(name, formatter_str, level=logging.DEBUG):
     "DIRECTIVES": {
         "default-src": [NONE],
         "script-src": [
-            SELF,
```
escattone, line 1268: I think that removing `SELF` from `script-src` will break the Django admin console's load of its own JS files. One option to resolve this is to add something like the following middleware that adds `SELF` but only for the admin URLs:

```python
from csp.constants import SELF
from csp.middleware import CSPMiddleware
...
class AdminCSPMiddleware(CSPMiddleware):
    """
    django-csp middleware that allows same-origin scripts on admin paths.
    """

    def get_policy_parts(self, request, response, report_only=False):
        policy_parts = super().get_policy_parts(request, response, report_only=report_only)
        if not report_only and request.path_info.startswith("/admin/"):
            policy_parts.update = {"script-src": [SELF]}
        return policy_parts

escattone (changes requested):

escattone (approved):

PR #7647, opened 2026-06-23
wrap long usernames on rev history page

escattone:
mozilla/sumo#3023

Before:

Screenshot 2026-06-23 at 11 44 10 AM

After:

image

### DIFF
--- a/kitsune/sumo/static/sumo/scss/components/_wiki.scss
+++ b/kitsune/sumo/static/sumo/scss/components/_wiki.scss
@@ -386,7 +386,6 @@ article {
 
     .creator {
       width: 300px;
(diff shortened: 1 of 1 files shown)


kitsune/sumo/static/sumo/scss/components/_wiki.scss
```diff
@@ -400,6 +399,14 @@ article {
 
   .creator {
     color: #333;
+    // Long display names/usernames without spaces would otherwise expand this
```
akatsoulas, line 402: this is redundant. We should update claude instructions to avoid over-commenting. Merging for now

akatsoulas (approved):

PR #7660, opened 2026-06-29
Drop postcrash app

akatsoulas:


escattone (approved):
r+wc

I think we should also delete the reference to the `kitsune.postcrash` tests in `.circleci/config.yml` (https://github.com/mozilla/kitsune/blob/caf75ad9aeab18ee1b2ec9cff149fd93c40a834f/.circleci/config.yml#L96).

PR #7676, opened 2026-07-02
Fix revision compare RTL alignment

janbrasna:
https://github.com/mozilla/sumo/issues/3148


kitsune/sumo/static/sumo/scss/components/_wiki.scss
```diff
@@ -514,23 +514,6 @@ article {
...
-    padding: 0;
-    text-align: left;
-
-    &.top {
-      background: transparent url(../img/wiki/compare.arrows.png) no-repeat p.$spacing-xs 5px;
-      padding-bottom: 5px;
-    }
-
-    &.bottom {
-      background: transparent url(../img/wiki/compare.arrows.png) no-repeat p.$spacing-xs -40px;
```
janbrasna, line 528: It seems this was the last consumer of that png (current buttons inline its svg directly) — would it be better to also remove the asset to not keep leftovers behind?

janbrasna: Yup, removing here as well.

escattone (changes requested):
This looks great except for the alignment of the top "Compare Selected Revisions" SVG. The fix is simple though. Just remove the `reverse-on-desktop` class from the `div` on line 6 of `revision_list.html`, and this should be good to go.

LTR

Image

RTL

Image

janbrasna:
Thanks for looking into this Ryan! I actually left this in the last time I was touching the line, not familiar enough with its intent to just remove it like the rest earlier. I did not like the effective layout it causes but checking its utility it seems it's neither necessary for mobile vs. desktop there, nor it affects the newly fixed charts so thanks for confirming it can be nuked — I was eyeing that in the previous PR already. 💨 Removed.

escattone (approved):
Thanks @janbrasna! 🙌

PR #7677, opened 2026-07-03
ALways check locale permissions in VisibilityManager.visible()

### LINKED ISSUE #3163: A locale reviewer (added via /locale/kb/locales/) and who holds no global wiki permissions receives a 404 when clicking on the "Pencil" (Edit article based on this revision) icon for an unreviewed first revision
**Preconditions**
* Create a kb article in en-US.
* Create a translation for the en-US article (do not approve it)
* Add a test user as a locale reviewer via the /kb/locales/{locale} page.

**Steps to reproduce**
1. Sign in to SUMO with the user from the preconditions step.
2. Navigate to the unapproved article translation (from the preconditions step)
3. Click on the "Show History" button from the "Editing Tools" side navbar.
4. Click on the "Pencil" icon displayed inside the revisions table.

**Expected behavior**
The reviewer is taken to the translate/edit form, pre-filled with the selected revision as the base.

**Actual behavior**
404


**Desktop:**
 - OS: Windows 11
 - Browser: Firefox


**Additional context**
This issue is reproducible in both `stage` and `prod`.

A comprehensive Root Cause analysis provided by claude:

Root Cause                                                                                                                                                                                                                        
                                                                                                                                                                                                                                    
  The pencil links to wiki.new_revision_based_on → edit_document(revision_id=…) (kitsune/wiki/views.py, revision_list.html:123-124). For a translation, edit_init_and_perms delegates to translate(parent.slug, revision_id)        
  (views.py:544-545), forwarding the uk revision id.                                                                                                                                                                                
                                                                                                                                                                                                                                    
  translate then resolves the base revision without supplying a locale (kitsune/wiki/views.py:1019-1021):                                                                                                                           
  allowed_docs = [parent_doc] if doc is None else [parent_doc, doc]                                                                                                                                                                 
  base_rev = get_visible_revision_or_404(user, pk=revision_id, document__in=allowed_docs)                                                                                                                                           
                                                                                                                                                                                                                                    
  Inside Revision.objects.visible() (kitsune/wiki/managers.py:75-104):                                                                                                                                                              
  - locale = kwargs.get("document__locale") → None, because only pk + document__in were passed (managers.py:84).                                                                                                                    
  - The permission gate can_delete_documents_or_review_revisions(user, locale=None) (managers.py:92-96) therefore skips the _is_leader/_is_reviewer locale-team branch — that branch is guarded by if locale and …                  
  (kitsune/wiki/permissions.py:138) — and falls back to global perms: user.has_perm("wiki.review_revision") or user.has_perm("wiki.delete_document") (permissions.py:142). A locale reviewer has neither.                           
  - So the revision set is restricted to (managers.py:99-102):                                                                                                                                                                      
  Q(document__current_revision__isnull=False) | Q(creator=user)                                                                                                                                                                     
  - The new translation has current_revision = None, and the reviewer didn't author the revision → the revision is filtered out → get_object_or_404 raises 404.
  
  This is the same membership-vs-permission gap noted elsewhere: locale reviewers are authorized via the Locale.reviewers relationship (checked through _is_reviewer), not via Django permissions — but this code path can't apply that check because it never receives the locale.

denyshon:
Resolves https://github.com/mozilla/sumo/issues/3163.

### DIFF
--- a/kitsune/wiki/managers.py
+++ b/kitsune/wiki/managers.py
@@ -1,3 +1,4 @@
+from django.conf import settings
 from django.db import models
 from django.db.models import Exists, OuterRef, Q
 
@@ -94,11 +95,24 @@ class VisibilityManager(models.Manager):
             or in_staff_group(user)
             or can_delete_documents_or_review_revisions(user, locale=locale)
         ):
+            # If locale is not specified, we need to find all the locales
+            # where the user can_delete_documents_or_review_revisions.
+            elevated_locales = []
+            if not locale:
+                elevated_locales = [
+                    candidate_locale
+                    for candidate_locale in settings.SUMO_LANGUAGES
+                    if can_delete_documents_or_review_revisions(
+                        user, locale=candidate_locale
+                    )
+                ]
+
             # Authenticated users without permission to see documents that
             # have no approved content, can only see those they have created.
             return qs.filter(
                 Q(**{f"{prefix}current_revision__isnull": False})
                 | self.get_creator_condition(user)
+                | Q(**{f"{prefix}locale__in": elevated_locales})
             )
 
         return qs
--- a/kitsune/wiki/tests/test_templates.py
+++ b/kitsune/wiki/tests/test_templates.py
@@ -2110,7 +2110,7 @@ class TranslateTests(TestCase):
     @mock.patch.object(EditDocumentEvent, "fire")
     @mock.patch.object(Site.objects, "get_current")
     def test_first_translation_to_locale(self, get_current, edited_fire, ready_fire):
-        """Create the first translation of a doc to new locale."""
+        """Create the first translation of a doc to new locale. Returns the revision."""
         get_current.return_value.domain = "testserver"
 
         url = reverse("wiki.translate", locale="es", args=[self.d.slug])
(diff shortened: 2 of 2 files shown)


escattone (commented):
@denyshon Hi Denys. I think the fix is simpler. I think you can simply add `document__locale=request.LANGUAGE_CODE` to the keyword arguments of the `get_visible_revision_or_404` call [here](https://github.com/mozilla/kitsune/blob/4f9bba8ba5c9c2b469679991baf6664cfd8f61be/kitsune/wiki/views.py#L1020), and that should resolve the problem.

denyshon:
@escattone I tried that at first, but such a change breaks `test_translate_based_on_parent_revision`, which makes sure we can base translations on specific en-US revisions. While I don't think this opportunity is currently exposed in the UI, it sounds like something we'd want to keep (and maybe expose in the future?) And apart from that, it feels semantically right for `visible()` not to require an optional parameter by default

kitsune/wiki/managers.py
```diff
@@ -94,11 +95,24 @@ def visible(self, user=None, **kwargs):
...
+            # where the user can_delete_documents_or_review_revisions.
+            elevated_locales = []
+            if not locale:
+                elevated_locales = [
+                    candidate_locale
+                    for candidate_locale in settings.SUMO_LANGUAGES
+                    if can_delete_documents_or_review_revisions(
+                        user, locale=candidate_locale
+                    )
+                ]
```
escattone, line 108: The `elevated_locales` list is very expensive to calculate this way. Each call to `can_delete_documents_or_review_revisions` entails 2+ database queries, and that's in a loop of about 75 locales (N+1 query problem), so calculating the list this way requires at least 150 database queries.

denyshon: @escattone if we go this way (with checking all the locales), do we really have better options? In theory, we might be able to get elevated_locales from the user (e.g., locales where the user is reviewer/leader), but how reliable is that in terms of permissions? My understanding is that a permission can be granted regardless of groups or anything else?

escattone: By the way, thanks for reminding me that my "simple" suggestion failed to consider the parent case.

I'll put together another option, and have you review it. This is tricky territory.

escattone: @denyshon I created #7741 as another option. Can you take a look at that one?

denyshon: @escattone lgtm and avoids unnecessary requests, so let's go with it

denyshon:
(closing in favor of https://github.com/mozilla/kitsune/pull/7741/)

kitsune/wiki/managers.py
denyshon: (and thank you!)

PR #7688, opened 2026-07-06
Kbox styling updates

akatsoulas:


kitsune/sumo/static/sumo/js/kbox.js
```diff
@@ -108,8 +108,7 @@ KBox.prototype = {
       preClose: false,
       template: TEMPLATE,
       title: self.$el.attr('title') || self.$el.attr('data-title'),
-      windowMargin: self.$el.data('viewport-margin') === undefined ?
-        20 : parseInt(self.$el.data('viewport-margin'), 10),
+      windowMargin: parseInt(self.$el.data('viewport-margin'), 10) || 20,
```
escattone: If `data-viewport-margin="0"` is ever used, I think this would ignore that and default to `20`, since an integer of `0` is falsy.

Maybe the following instead?
```js
parseInt(self.$el.data('viewport-margin') ?? 20, 10)
```

escattone (approved):
r+wc

kitsune/sumo/static/sumo/js/kbox.js
```diff
@@ -306,9 +303,9 @@ KBox.prototype = {
     // Due to max-width, kbox height may have changed.
     rect = self.$kbox[0].getBoundingClientRect();
 
-    if (rect.bottom > windowHeight) {
+    if (rect.bottom > windowHeight - minMargin) {
```
denyshon, line 306: Hey @akatsoulas, thank you for this patch with fixes! Just wanted to highlight that this particular check was intended. I used windowHeight, so that we avoid displaying a scrollbar when there is still space at the bottom (IMO a lesser margin is better than an unnecessary scrollbar). This is particularly relevant to our "needs-change" dynamic resizing, when we cannot place the kbox optimally in advance.

denyshon: submitted https://github.com/mozilla/kitsune/pull/7696 for this

PR #7698, opened 2026-07-07
allow article ordering for topics

escattone:
mozilla/sumo#3140

I went back and forth on this one, but this is what I thought we could consider first. If this is too much, we can hard-wire sorting by "newest" to a set of topic slugs (just `Release notes` for now).

This PR provides configurable KB article ordering per topic for:

1. The articles within the topic's card on the product page (e.g., https://support.mozilla.org/en-US/products/firefox-enterprise), which avoids the following problem:

image

2. The articles on the topic's landing page (e.g., https://support.mozilla.org/en-US/products/firefox-enterprise/whats-new-firefox-enterprise), which avoids the following problem:

image

If this is merged, it requires the additional step of enabling sorting by "newest" for the `Release notes` topic.


kitsune/wiki/facets.py
```diff
@@ -172,24 +193,66 @@ def _documents_for(user, locale, topics=None, products=None):
...
+        # Using the original keeps localized listings and their en-US fallbacks in
+        # the same chronological order as en-US.
+        qs = qs.annotate(
+            first_published_at=Subquery(
+                Document.objects.filter(pk=OuterRef("pk"))
+                .annotate(
+                    first_published_at=Case(
+                        When(
+                            parent__isnull=False,
+                            then=Coalesce(
```
akatsoulas: I wonder if this has any implications similar to what the following PR solved
https://github.com/mozilla/kitsune/pull/6979

kitsune/wiki/facets.py
```diff
@@ -172,24 +193,66 @@ def _documents_for(user, locale, topics=None, products=None):
...
-            }
+    if annotate_first_published:
+        # Annotate each document with the publication date of its original: the
+        # earliest approved revision of the original (parent for translations,
+        # otherwise the document itself). Prefer the review date (when it went
+        # live), falling back to the created date when a revision was approved
+        # without a review timestamp (e.g. auto-created redirects, legacy data).
+        # Using the original keeps localized listings and their en-US fallbacks in
+        # the same chronological order as en-US.
+        qs = qs.annotate(
```
akatsoulas: What happens if `first_published_at` is None? Won't this raise a `TypeError`?

kitsune/wiki/facets.py
escattone: I was concerned about that too -- which led me to consider the option of just ordering by `-id` when we want to sort by "newest" (but that perhaps isn't without concerns as well).

I decided to do some testing in production, comparing the main query with and without the extra `first_published_at` subquery. From my tests using `timeit`, it increases the query time by about 5-10%. Since the result of `_documents_for` is cached, that small penalty will only be paid by one request per cache period (which I think is the default of 5 minutes?).

So, I think this is OK. However, maybe we should consider ordering by `-id` instead, which has no extra cost, but does have the downside of relying on the assumption that a newer document always gets an `id` greater than an older document. Thoughts?

kitsune/wiki/facets.py
escattone: Hm, yeah, I think it is possible, although very rare, for `first_published_at` to be null for translations, because we don't guarantee that the parent has an approved revision. I'll add a fall-back for that case, and then it would be impossible for `first_published_at` to be null.

escattone: The second commit includes an adjustment that ensures that `first_published_at` can never be null.

kitsune/wiki/facets.py
akatsoulas: I am not sure either. Introducing this for just release notes, should be OK from a performance perspective. My concern though is that it can become an expensive feature in the future if more and more topics are using it. The cache helps but it's there only for non authenticated users.

Reversing this, maybe because of the small footprint of the feature, using and `id` should be fine in most cases. I suspect we should order on parent's ID for translations though. Maybe we can start by going with the performant but less accurate in some cases solution (id) and adjust if there are cases where the sorting is not the expected one. Moreover is going to make this feature much simpler with regards to the code.

escattone: With the latest commit, I switched the "newest" sorting to `parent_id or id` (descending).

kitsune/wiki/facets.py
escattone: With the latest commit, this is no longer relevant.

akatsoulas (approved):

PR #7707, opened 2026-07-09
Introduce KB chunker for the retrieval layer

akatsoulas:


escattone (commented):
This was challenging to review. My understanding is that the chunking exists to break an article into meaningful pieces for vectorization/search. What is the ultimate goal when one or more chunks are found based on a vectorized query? Is it to suggest the KB article(s) associated with those chunks, or is it to return chunks of article text that a reader of the KB article(s) would be interested in? It seems to me that this chunking approach is fine if the ultimate goal is to suggest one or more KB articles. However, if the ultimate goal is to return portions of text that a reader of a KB article might find interesting, it seems that this approach doesn't provide the ability to re-assemble chunks into a portion of text that the reader would encounter when reading the KB article?

escattone (approved):
r+wc

I'm going to approve so I don't hold you back from next steps.

akatsoulas:
The idea is to find the right article with that's more accurate for a given query for the case of semantic search. It could be something different for other cases. Chunks are optimized for the embedding process and they are not related with the presentation process. Thanks for the review!

escattone:
By the way, nice work! 🙌 This is a much more complex problem than I anticipated!

PR #7718, opened 2026-07-14
remove `jquery` and `jquery-ui`

escattone:
mozilla/sumo#3101

Sub-issues that this PR resolves:
mozilla/sumo#3216
mozilla/sumo#3218
mozilla/sumo#3219
mozilla/sumo#3220
mozilla/sumo#3223
mozilla/sumo#3225

Replaces `jquery` and `jquery-ui` with vanilla Javascript. Introduces a 40+% increase in our JS unit tests (more than 100 new JS unit tests).


kitsune/sumo/static/sumo/js/gallery.js
```diff
@@ -1,501 +1,564 @@
...
+            return false;
+          }
+          return self.startUpload(input);
+        },
+        onComplete: function (input, content, options) {
+          fireAjaxComplete(input.closest('form'));
+          // Clear just the file input (the old code reset its wrapper form),
+          // leaving the main form's metadata fields intact.
+          input.value = '';
+          if (cancelledUploads.has(input)) {
```
akatsoulas, line 158: Shouldn't we also cancel the upload here as well?

kitsune/sumo/static/sumo/js/questions.js
```diff
@@ -25,62 +27,63 @@ document.addEventListener("DOMContentLoaded", async () => {
...
-          url: url,
-          method: 'POST',
-          beforeSend: function(xhr, settings) {
-            xhr.setRequestHeader('X-CSRFToken', csrftoken);
-          }
-        });
-      }
+    var searchBox = document.querySelector('#support-search input[name=q]');
+    if (searchBox) {
+      searchBox.value = unquote(getCookie('last_search'));
```
akatsoulas: if the cookie is missing this will be `undefined`

kitsune/sumo/static/sumo/js/markup.js
```diff
@@ -25,93 +25,213 @@ import KBox from "sumo/js/kbox";
...
+        e.preventDefault();
+        pick(i);
+      });
+      list.appendChild(li);
+    });
+    activeIndex = -1;
+    list.hidden = false;
+  }
+
+  input.addEventListener("input", function () {
```
akatsoulas, line 129: Pasting review from AI
```
Debouncing prevents rapid requests before they start, but multiple in-flight requests can still
finish out of order. Every response calls the same render function, so an older search can
overwrite suggestions for the current input. jQuery UI guarded responses with a request index.
The replacement needs an equivalent generation check or request cancellation.
```

akatsoulas (approved):
r+wc

kitsune/sumo/static/sumo/js/gallery.js
escattone: Good catch, that's a regression. I'll fix this.

kitsune/sumo/static/sumo/js/questions.js
escattone: Another good catch, I'll fix this.

kitsune/sumo/static/sumo/js/markup.js
escattone: Another good catch. Fixed!

PR #7722, opened 2026-07-15
Do not set a default value for flagging

akatsoulas:


escattone (changes requested):
Can we also reflect the error back to the user when no reason is selected? Currently, if no reason is selected, the user sees:

Image

and has to reload the page in order to try again.

akatsoulas:
@escattone pr updated

escattone (approved):

PR #7724, opened 2026-07-16
Fix search on contribute pages

janbrasna:
Adds a missing id hook to inject the sibling to for:
- https://github.com/mozilla/sumo/issues/1240

### DIFF
--- a/kitsune/landings/jinja2/landings/contribute.html
+++ b/kitsune/landings/jinja2/landings/contribute.html
@@ -7,7 +7,9 @@
 {% endblock %}
 
 {% block contentwrap %}
+<section id="main-content">
 <div id="svelte">
 {% include "pre-render" + path + ".html" %}
 </div>
+</section>
 {% endblock %}


kitsune/landings/jinja2/landings/contribute.html
```diff
@@ -7,7 +7,9 @@
 {% endblock %}
 
 {% block contentwrap %}
+<section id="main-content">
 <div id="svelte">
 {% include "pre-render" + path + ".html" %}
 </div>
+</section>
 {% endblock %}
```
janbrasna, line 15: This is not exactly pretty or smart, but is the least change without touching unrelated whitespace around etc. — the svelte content inside is emitted as a oneliner tag soup anyway, so there's no reason to be… pretty.

There also are definitely better ways to rewrite the hooks that wouldn't need this extra wrapping but again, this seems cheaper without touching existing binding code having a way larger impact surface potentially.

akatsoulas (approved):

kitsune/landings/jinja2/landings/contribute.html
janbrasna: It also fixes the keyboard focus a11y skip link now on these pages. 💪 🎹

PR #7725, opened 2026-07-17
Delete change_list.html

### LINKED ISSUE #3222: Breadcrumbs are broken in the admin panel
Breadcrumbs are partially missing on many pages of the admin panel. That's because our [change_list.html](https://github.com/mozilla/kitsune/blob/9e9402f83e0d57779beb6045c12f4859cc931fd7/kitsune/sumo/templates/admin/change_list.html#L4) (an override of Django's implementation) is out of sync with Django.

denyshon:
Resolves https://github.com/mozilla/sumo/issues/3222.

The purpose of our change_list override is to "add the 'novalidate' attribute to the #changelist-form". However, the upstream Django [already does that](https://github.com/django/django/blob/024c26b1e77ea5b1b158265167ade47927a64c06/django/contrib/admin/templates/admin/change_list.html#L62), so we can safely remove the override (eliminating the need to keep it updated).


akatsoulas:
thanks @denyshon!

akatsoulas (approved):

PR #7734, opened 2026-07-21
Revert "Fix question OP author name when quoting to reply"

escattone:
Reverts mozilla/kitsune#7732

That PR actually breaks forum quoting and doesn't fix AAQ quoting which belongs in `question.js`, so reverting. Also, I think I'll fix this once #7718 lands.

### DIFF
--- a/kitsune/sumo/static/sumo/js/forums.js
+++ b/kitsune/sumo/static/sumo/js/forums.js
@@ -21,7 +21,7 @@ import Marky from "sumo/js/markup";
       var post = $(this).data('post'),
         $post = $('#post-' + post),
         text = $post.find('div.content-raw').text(),
-        user = $post.closest('.thread-post').find('.thread-post--author-name').text().trim(),
+        user = $post.find('.display-name').text(),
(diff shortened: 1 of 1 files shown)


janbrasna:
Lol no idea how that ended up in the wrong file, sorry — I'm actually waiting for you to land #7718 not to conflict with it; and submit a fix once it's beyond jQuery anyways.

escattone:
@janbrasna Hi Jan! No worries! That sounds great. One thing I was going to add in terms of your fix, is that the poster for an answer has `.display-name`, and the OP for the original question has `.thread-post--author-name`, so both will have to be considered depending upon whether the question or one of its answers is being quoted. Thanks.

escattone:
@janbrasna Hi Jan! I merged #7718, so you're all set to create a new PR for this.

janbrasna:
Yup I was following the refactor PR to see if it landed the replacement as–is or made some changes in this area. I've been looking into it for a bit, contemplating both just traversing the tree upwards and sideways, but also perhaps marking up the question differently to contain the author in the ID element…

Now realizing the trigger can come from both ID types I need to check again what's the better approach; will probably do an ugly ternary in JS to cover both cases to minimize the impact.

PR #7745, opened 2026-07-24
avoid clipping on article dropdown for marky links

escattone:
mozilla/sumo#3243

This one was surprisingly challenging. It turns out that the fix needs to do the same thing that `jquery` was doing "under the hood".

After:

image

image

### DIFF
--- a/kitsune/sumo/static/sumo/js/markup.js
+++ b/kitsune/sumo/static/sumo/js/markup.js
@@ -199,6 +199,11 @@ function attachTypeahead(input, source, onSelect) {
   }
 
   function close() {
+    // Drop any search that hasn't started yet, and make sure the responses to
+    // the ones already out there don't get rendered. Otherwise a search from
+    // just before the modal closed can reopen the dropdown after it's gone.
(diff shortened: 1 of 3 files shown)


kitsune/sumo/static/sumo/js/markup.js
```diff
@@ -61,39 +61,175 @@ function isVisible(el) {
...
+      // open.
+      window.addEventListener("scroll", reposition, true);
+      window.addEventListener("resize", reposition);
+    }
+    // Also reposition on later renders, since new results change how tall the
+    // dropdown wants to be.
+    positionList();
+  }
+
   function close() {
```
akatsoulas, line 201: This should also clear the debounce timer and increment the latestRequestId as well to avoid rendering in flight requests when closing the modal.

akatsoulas (approved):
r+wc

kitsune/sumo/static/sumo/js/markup.js
escattone: 👍 Good catch, thanks!

PR #7762, opened 2026-07-29
Avoid an unnecessary new string in ajaxupload.js

denyshon:


escattone (approved):
Good catch, thanks @denyshon!

PR #7763, opened 2026-07-29
Retrieval index scaffolding

akatsoulas:


kitsune/retrieval/eligibility.py
```diff
@@ -0,0 +1,80 @@
...
+    return list(family_documents(document).values_list("id", flat=True))
+
+
+def access_group_ids_for(document) -> list[int]:
+    """Sorted restriction group ids from the original (translations inherit)."""
+    return sorted(group.id for group in document.original.restrict_to_groups.all())
+
+
+def visibility_for(document) -> str:
+    return "group_restricted" if access_group_ids_for(document) else "public"
```
escattone: Should probably use the constants `RESTRICTED_VISIBILITY` and `PUBLIC_VISIBILITY` from `kitsune/retrieval/index.py` instead?

kitsune/settings.py
```diff
@@ -350,6 +350,12 @@ def path(*parts):
 ES_DEFAULT_SQL_CHUNK_SIZE = config("ES_DEFAULT_SQL_CHUNK_SIZE", default=1000, cast=int)
 ES_DEFAULT_ELASTIC_CHUNK_SIZE = config("ES_DEFAULT_ELASTIC_CHUNK_SIZE", default=50, cast=int)
 
+# Retrieval (RAG) embeddings — default to the offline fake backend (safe for dev/CI).
+RETRIEVAL_EMBEDDING_BACKEND = config("RETRIEVAL_EMBEDDING_BACKEND", default="fake")
```
escattone: Maybe default to `""` -- or anything that causes an exception -- in order to guarantee that we don't accidentally use `"fake"` for the `dev`, `staging`, or `prod` environments?

escattone (approved):
r+wc

This is challenging stuff, nice work! I think my main concern is whether we're doing everything possible to ensure that the atoms that chunking generates remain below the token limit of the embedding step.

- Claude mentioned that the `count_tokens` estimator (`len(text) // 4`) is good for Latin-based languages, but too low for CJK and other symbolic languages. Does it make sense to use something off-the-shelf like `vertexai.preview.tokenization` instead?
- What happens if we go over the `budget` in `chunk_kb`?

akatsoulas:
Vertex AI SDK has been deprecated. I made the chunk size count unicode aware for now

PR #7765, opened 2026-07-30
improve spam classifier on unknown features

akatsoulas:


kitsune/products/migrations/0044_new_tab_page_product_metadata.py
```diff
@@ -0,0 +1,76 @@
+from django.db import migrations
+
+DESCRIPTION_ADDITIONS = {
+    "Firefox": (
+        "Opening a new tab, a new window, or the browser itself shows the New Tab page, also"
+        " known as Firefox Home or the start page. It can display a web search box, Shortcuts"
+        " to frequently visited sites, recommended stories, a weather widget, and a wallpaper,"
```
denyshon: There are many widgets, not just the weather one: https://support.mozilla.org/en-US/kb/firefox-new-tab-widgets

kitsune/products/migrations/0044_new_tab_page_product_metadata.py
```diff
@@ -0,0 +1,76 @@
...
+        description = description.strip()
+
+        if any(marker in description.lower() for marker in ALREADY_COVERED_MARKERS):
+            print(
+                f"""Skipped product "{title}" because its description already covers"""
+                " the New Tab page."
+            )
+            continue
+
+        product.metadata = metadata | {"description": f"{description} {addition}"}
```
denyshon, line 65: Just checking, aren't we overwriting user-facing descriptions here (those displayed on the product cards)?

kitsune/products/migrations/0044_new_tab_page_product_metadata.py
escattone: Let's replace the phrase `a weather widget` with `many kinds of widgets (weather, clock, timer, crossword, etc.)`.

kitsune/products/migrations/0044_new_tab_page_product_metadata.py
escattone: @denyshon Hi Denys, no, this "description" in the `metadata` is not public facing. It's only used for LLM purposes.

escattone (approved):
r+wc

PR #7768, opened 2026-07-31
Update ES access architecture

akatsoulas:


escattone (approved):
This was really challenging to review. In the end, I decided that the best thing to do was to try the index commands locally, including changing a document from public to restricted, and back again by deleting the restricting group. Everything worked as expected.

I think this PR also needs to be paired with a `webservices-infra` PR that ensures that we configure Celery workers to pull from the new `retrieval` and `retrieval_bulk` queues.

PR #7773, opened 2026-08-03
Fix reverse restriction render cascade

akatsoulas:


escattone (approved):
Nice catch! While reviewing this one, I realized that I don't think we're covering the case when deleting a group that is being used to restrict one or more documents. We can consider that for another PR.

PR #7781, opened 2026-08-03
re-render on deletion of restricting groups

escattone:
mozilla/sumo#3263


escattone:
Closing this, since I just noticed that #7768 already includes a fix for https://github.com/mozilla/sumo/issues/3263.

PR #7791, opened 2026-08-04
Improve user profile page

emilghittasv:


escattone:
This is so much better than what we have now, thanks @emilghittasv!

image

I have a suggestion though. I think it would be nice to make the `Edit my profile`, `Edit settings`, `Edit contribution areas`, `Manage watch list`, and `My questions` pages show the same left-hand column:

image

and then have the right-side be the content of the various "edit" pages. What do you think?

kitsune/sumo/jinja2/sumo/includes/entity_card.html
escattone, line 1: Should we put this macro in `kitsune/sumo/jinja2/includes/common_macros.html` instead?

escattone (commented):
Some other comments

- The text  of the "decativate" buttons is underlined and lengthy ones are cutoff.

    Image

- Some of the left-side cards don't shrink enough to fit some mobile screens, which activates horizontal scrolling.

    Image

emilghittasv:
> This is so much better than what we have now, thanks @emilghittasv!
> image
> 
> I have a suggestion though. I think it would be nice to make the `Edit my profile`, `Edit settings`, `Edit contribution areas`, `Manage watch list`, and `My questions` pages show the same left-hand column:
> image
> 
> and then have the right-side be the content of the various "edit" pages. What do you think?

Yes I love this idea! I'll tackle this as part of a different PR.



> Some other comments
> 
>     * The text  of the "decativate" buttons is underlined and lengthy ones are cutoff.
>         Image
> 
>     * Some of the left-side cards don't shrink enough to fit some mobile screens, which activates horizontal scrolling.
>         Image

This should be fixed now!

PR #7792, opened 2026-08-04
add retrieval kill switch (default off)

escattone:


akatsoulas:
I would propose to replace this with https://github.com/mozilla/kitsune/pull/7796 which does not allow signals to register but gives us manual control for testing. The reason for proposing this is because it's much simpler and centralized in a single place. Your solution allows us to evict without ingesting but at the same time is distributed and we need to gate keep any new signals/handlers/events/tests which makes it more complex. Given that we actually need a kill switch and not an activation switch, #7796 should suffice - unless I missed something. Thoughts?

escattone:
Closing in favor of #7796

PR #7796, opened 2026-08-05
Add a variable to register retrieval signals

akatsoulas:


escattone (approved):
Much better given that it's only temporary.

PR #7798, opened 2026-08-05
Fix question author name when quoting to reply

janbrasna:
https://github.com/mozilla/sumo/issues/3034

Ryan noticed this has to cater for both selectors, the one within the dataset reply ID which contains the answer author (also done with the same link functinon), and the broken author quoting where the author's elements (question headers) are currently moved outside of its dataset question ID and don't follow the previous class naming anyway.

So this now attempts to reach both element selector possibilities in a failover fashion.

### DIFF
--- a/kitsune/sumo/static/sumo/js/questions.js
+++ b/kitsune/sumo/static/sumo/js/questions.js
@@ -379,9 +379,7 @@ function initReplyToAnswer() {
       var contentEl = document.getElementById(contentId);
       var rawEl = contentEl ? contentEl.querySelector('.content-raw') : null;
       var nameEl = contentEl ? contentEl.querySelector('.display-name') : null;
-      var authorEl = document.querySelector('.thread-post--author-name');
-      var author = authorEl ? authorEl.textContent.trim() : '';
-      var user = nameEl ? nameEl.textContent : author;
+      var user = nameEl ? nameEl.textContent : document.querySelector('.thread-post--author-name')?.textContent.trim();
       var text = rawEl ? rawEl.textContent : '';
       var reply_text = `''<p>${user} [[#${contentId}|${gettext('said')}]]</p>''\n<blockquote>${text}\n</blockquote>\n\n`;
       var textarea = document.getElementById('id_content');
--- a/kitsune/sumo/static/sumo/js/questions.js
+++ b/kitsune/sumo/static/sumo/js/questions.js
@@ -379,8 +379,10 @@ function initReplyToAnswer() {
       var contentEl = document.getElementById(contentId);
       var rawEl = contentEl ? contentEl.querySelector('.content-raw') : null;
       var nameEl = contentEl ? contentEl.querySelector('.display-name') : null;
+      var authorEl = document.querySelector('.thread-post--author-name');
+      var author = authorEl ? authorEl.textContent.trim() : '';
+      var user = nameEl ? nameEl.textContent : author;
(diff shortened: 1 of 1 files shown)


kitsune/sumo/static/sumo/js/questions.js
```diff
@@ -379,8 +379,10 @@ function initReplyToAnswer() {
       var contentEl = document.getElementById(contentId);
       var rawEl = contentEl ? contentEl.querySelector('.content-raw') : null;
       var nameEl = contentEl ? contentEl.querySelector('.display-name') : null;
+      var authorEl = document.querySelector('.thread-post--author-name');
+      var author = authorEl ? authorEl.textContent.trim() : '';
```
escattone: Two small things:
- It'd be nice to only do this work if we have to, so only when we know we're falling back.
- Should we trim the `nameEl.textContent` as well?

Also, can you add a test for this?

escattone (commented):
@janbrasna Hi Jan! Thanks! I just have some minor comments.

kitsune/sumo/static/sumo/js/questions.js
```diff
@@ -379,8 +379,10 @@ function initReplyToAnswer() {
...
+      var user = nameEl ? nameEl.textContent : author;
```
janbrasna: The original jQuery version with tree traversal read a bit different so the above was to keep the style somewhat consistent without either nesting too many things, or chainig too many ternaries. The inline fallback could be something like this:

```suggestion
      var nameEl = contentEl ? contentEl.querySelector('.display-name') : null;
      var user = nameEl ? nameEl.textContent : document.querySelector('.thread-post--author-name')?.textContent.trim();
```

janbrasna: However that only does blind fallback — the correct way would probably be to inspect the dom tree and/or dataset around the event element, and conditionally decide whether it's case "author", or case "user", get the node appropriate for each given case, and fall through to empty string per case… and not mix the content getters just based on absence of that local tree… which however effectively is the practical difference;)

That would hoever be way more operations, not less. So for simplicity/brevity I'd try the above as a way to minimize the querying of selectors, and just based that off of availability in the current parent context. The potential false positive matches do not sound like anything catastrophic;)

janbrasna: _(NB: there's an optional chaining from ES2020, that's hopefully alright, and/or transpiled?)_

janbrasna (commented):
I'm not actually exactly sure about any ES version or style guide to follow, so I'm only making some guesses here:

kitsune/sumo/static/sumo/js/questions.js
janbrasna: I think the old JS mapped to the old HTML alright re: trimming (practically: whitespace in nesting) — it was just the newer code that wouldn't quote correctly without it — it can be added for good measure for posterity, but is not necessary as long as it targets the existing code.

escattone (approved):

janbrasna:
Hey Ryan! I was looking into the test options but did not come up with something interesting/stable. That would probably be Playwright UI automation, but the only validation is reading back the textarea to see if the username is definitely positively in the content after the interaction — not sure if the data used for testing is that stable to be able to rely on string values to be read back in there — but will keep looking.

janbrasna:
Test for this should be in the suite already: https://github.com/mozilla/sumo/issues/3034#issuecomment-5574818773

PR #7806, opened 2026-08-07
add middleware to clean NUL bytes from GET

escattone:
mozilla/sumo#2815

@emilghittasv and I discussed this, so I'm adding him as a reviewer as well.

### DIFF
--- a/kitsune/sumo/middleware.py
+++ b/kitsune/sumo/middleware.py
@@ -274,6 +274,24 @@ class CacheHeadersMiddleware(MiddlewareMixin):
         return response
 
 
+class StripNulCharactersMiddleware(MiddlewareMixin):
+    """Remove NUL characters from the query string."""
+
+    # A NUL arrives percent-encoded, as a raw byte, or as a mix of the two. A "%" that
+    # isn't starting a real escape is a literal percent sign, and it's replaced with
+    # "%25" (the escape for a literal "%") because dropping the NUL next to it would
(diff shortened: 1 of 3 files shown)


kitsune/sumo/middleware.py
```diff
@@ -274,6 +274,40 @@ def process_response(self, request, response):
         return response
 
 
+class StripNulCharactersMiddleware(MiddlewareMixin):
+    """Remove NUL characters from the query string.
```
akatsoulas: This is extremely verbose. Leaving the inline comment could also be enough. I don't have a preference what we are going to keep here but ~25 lines of comments for this functionality is extremely verbose and hurts readability imo

kitsune/sumo/tests/test_middleware.py
```diff
@@ -249,3 +251,135 @@ def test_no_header_sets_no_cookie(self):
         response = HttpResponse()
         response = self.mw.process_response(request, response)
         self.assertNotIn("geoip_country_name", response.cookies)
+
+
+class StripNulCharactersMiddlewareTestCase(TestCase):
```
akatsoulas, line 256: nit: we could reduce the test suite without leaving any cases uncovered. There's currently a bit of overlap

akatsoulas (approved):
r+wc - I also submitted PR https://github.com/mozilla/webservices-infra/pull/12084 which could help reduce some of the overhead (which is though very low)

PR #7825, opened 2026-08-11
Semantic search

akatsoulas:


kitsune/search/parser/__init__.py
```diff
@@ -60,6 +69,22 @@
...
+            elif quoted and character == "\\":
+                escaped = True
+            elif character == '"':
+                quoted = not quoted
+            elif not quoted and character == "(":
+                depth += 1
+                if depth > _MAX_NESTING_DEPTH:
+                    raise ParseException(query, position, "query nesting is too deep")
+            elif not quoted and character == ")":
+                depth = max(0, depth - 1)
```
escattone, line 87: It seems that this is trying to be a quick and dirty check for a recursion violation without running `pyparsing`? However, because this check can't match exactly what `pyparsing` actually does, it can be bypassed and the recursion error triggered (the unhandled `RecursionError` would cause an HTTP 500). For example, I think `?q=" ((((((((((((a))))))))))))` would trigger a `RecursionError`, even though it passes this check.

Because of this, would it make sense to also capture `RecursionError` where `ParseException` is already caught, in order to avoid 500's?

kitsune/retrieval/query.py
```diff
@@ -0,0 +1,823 @@
...
+        or isinstance(approximate_total, bool)
+        or not math.isfinite(approximate_total)
+        or approximate_total < 0
+        or not float(approximate_total).is_integer()
+    ):
+        raise InvalidRetrievalResponse("response has no valid approximate family count")
+
+    candidates = tuple(
+        _decode_hit(hit, rank=offset + rank) for rank, hit in enumerate(hits[:page_size], start=1)
+    )
```
escattone: Would it make sense to prevent a single bad hit -- an individual `InvalidRetrievalResponse` from the `_decode_hit()` call -- from spoiling the entire list of hits, so maybe log the error and move on?

kitsune/retrieval/access.py
```diff
@@ -0,0 +1,274 @@
...
+                    candidate.family_id,
+                    AuthorizedPassage(passage, display),
+                )
+            )
+
+    page_end = page_offset + page_size
+    return RetrievalResult(
+        candidates=tuple(authorized[page_offset:page_end]),
+        approximate_total=result.approximate_total,
+        has_more=len(authorized) > page_end or (use_unvalidated_has_more and result.has_more),
```
escattone: Claude claims that items that the user is not allowed to see can cause visible items to be unreachable:

> `has_more` only asks whether more than one page survived the over-fetch window, so filtering here can hide results the viewer is allowed to see.
> 
> With the defaults (`page_size=5`, `overfetch=5`) and 12 matches where 5 in the middle get filtered out: page 1 shows 5 results, `has_more=False`, and the last 2 visible documents become unreachable.
> 
> Elasticsearch already answered this — `result.has_more` is True — but line 152 gates it off for the KB path. Could we OR it in? A false positive is harmless, since `views.py:118` already redirects an empty page back to page 1. A false negative hides results with no error and no way for the user to get to them.
> 
> Also worth noting: RETRIEVAL_AUTHORIZATION_OVERFETCH=0 makes this unconditional — has_more can then never be true, and c`hecks.py:112` accepts 0 as valid.

kitsune/search/views.py
```diff
@@ -98,24 +100,59 @@ def simple_search(request):
...
+            product_id=product.id if product else None,
+            page=page_number,
+        )
+        page = None
+        results = list(hybrid.results)
+        if page_number > 1 and not results:
+            redirect_params = request.GET.copy()
+            redirect_params["page"] = 1
+            return HttpResponseRedirect(f"{request.path}?{redirect_params.urlencode()}")
+        total = hybrid.approximate_total if results else 0
```
escattone, line 130: It seems that the results number that the user sees differ from what's actually reachable?

escattone (approved):
r+wc

Does RRF tend to favor KB articles over questions, since questions can only appear in the keyword results, but KB articles can appear in both (semantic and keyword results)?

akatsoulas:
Regarding the RRF, yes it could favor KB over AAQ. 
AAQ can contribute only through the lexical leg, KB can contribute both through lexical and semantic. A KB article ranked first in both legs can get from the RRF 1/61 from lexical and 1/61 from semantic whereas AAQ gets 1/61.
KB: 1/61 + 1/61 ~ 0.0328
AAQ: 1/61 ~ 0.0164
So agreement between lexical and semantic retrieval naturally favors authoritative KB content. This is similar to what we are currently doing by favoring KB through weight boosting. Now it happens more naturally. An excellent AAQ result can still beat a KB article that ranks poorly or appears in only one child. If query embedding is unavailable, everything falls back to lexical retrieval and that implicit KB preference disappears.

kitsune/search/views.py
akatsoulas: Yes, this is by design. It's not an exact count because:
- authorization checks only the bounded candidate prefix
- stale indexed candidates can be rejected by the database
- RRF exposes only a bounded ranking window;
- family cardinality itself is approximate. We are collapsing on family_id because ES is matching chunks so the count of family_id is an approximation from ES - usually it spot on but it's not guaranteed

kitsune/retrieval/query.py
akatsoulas: My recommendation is to keep the current behavior initially. A malformed hit means the index, mapping, backfill, or application version is inconsistent. Silently skipping it could conceal a systemic problem. If this
becomes an availability issue, we can introduce “skip + degraded result + structured event” later.

PR #7826, opened 2026-08-12
Fix handling of non-object troubleshooting data

vichekaoeun:
Resolved bug [#3116](https://github.com/mozilla/sumo/issues/3116) causing 500 error when troubleshooting data contains JSON that is not an object. Added tests for non-object JSON on form

### DIFF
--- a/kitsune/questions/forms.py
+++ b/kitsune/questions/forms.py
@@ -186,6 +186,24 @@ class EditQuestionForm(forms.ModelForm):
             raise forms.ValidationError(_("Question content cannot be empty."))
         return content
 
+    def clean_troubleshooting(self):
+        troubleshooting = self.cleaned_data.get("troubleshooting", "")
+        if not troubleshooting:
+            return troubleshooting
+
+        try:
+            parsed = json.loads(troubleshooting)
+        except ValueError:
+            parsed = None
+
+        if not isinstance(parsed, dict):
+            # L10n: An error shown when troubleshooting information isn't a JSON object.
+            raise forms.ValidationError(
+                _("Troubleshooting information must be a valid JSON object.")
+            )
+
+        return troubleshooting
+
     @property
     def metadata_field_keys(self):
         """Returns the keys of the metadata fields for the current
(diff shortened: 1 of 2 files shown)


denyshon:
Should we also add a check for this to `clean_content()` and display an error if the data is not a valid JSON/dict?

vichekaoeun:
yes I think that would make sense but I think it could go into a separate function like clean_troubleshooting since it would mix unrelated issues to the wrong field otherwise.

vichekaoeun:
Hi @denyshon could you review my change

denyshon:
@vichekaoeun sure, but I'm not a part of the [team](https://mozilla.github.io/kitsune/contactus/), so I don't have write access to the repo and can't speak on their behalf

you may want to ping @escattone instead

kitsune/questions/forms.py
```diff
@@ -186,6 +186,24 @@ def clean_content(self):
             raise forms.ValidationError(_("Question content cannot be empty."))
         return content
 
+    def clean_troubleshooting(self):
```
denyshon, line 189: just a neat: let's add a docstring for this method

denyshon (approved):

vichekaoeun:
Hi @escattone could you help review my changes?

kitsune/questions/forms.py
```diff
@@ -186,6 +186,24 @@ def clean_content(self):
...
+        troubleshooting = self.cleaned_data.get("troubleshooting", "")
+        if not troubleshooting:
+            return troubleshooting
+
+        try:
+            parsed = json.loads(troubleshooting)
```
akatsoulas: This will change the current behavior and limit the troubleshooting data only to json. That works for the JS code that pulls json but it will no longer accept manually copied raw data from about:support

akatsoulas (changes requested):

kitsune/questions/forms.py
denyshon: doesn't copied raw data follow the json format?

vichekaoeun: hi @akatsoulas could you review

PR #7828, opened 2026-08-13
Update syntax for docker–compose

janbrasna:
Updates running docs from v1 compose commands to current v2+ while cleaning up some formatting, broken links and the like.

The only material change is fixing a step number mentioned in running the app.

### DIFF
--- a/docs/hacking_howto.md
+++ b/docs/hacking_howto.md
@@ -4,7 +4,7 @@
 
 This chapter helps you get an installation of Kitsune up and running.
 
-If you have any problems getting Kitsune running, let us know. See [Contact us](contactus.md).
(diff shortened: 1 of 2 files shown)


docs/contributors.md
```diff
@@ -21,13 +21,13 @@ guide](https://support.mozilla.org/en-US/get-involved) on the SUMO site.
 
 # Help reporting bugs
 
-Please report any bugs you find with Kitsune on Bugzilla:
```
akatsoulas, line 24: Bugzilla should remain the tracking source for any reported bugs. The sumo repo is used mainly for internal organizational purposes

janbrasna: Dropped, thanks!

akatsoulas (approved):

PR #7830, opened 2026-08-14
Localize DB Topic and Product titles

akatsoulas:


escattone (commented):
I found some product and topic titles, as well as one product description, that were missed, so I'll add those.

PR #7832, opened 2026-08-15
Sort group leaders and members alphabetically by displayed name

### LINKED ISSUE #7270: Sort the display of group members in alpha order of username
https://support.mozilla.org/en-US/groups/forum-moderators/ doesn't list members by username.

This also happens for other groups.

smithellis:
The group profile page listed leaders and members in whatever order the database returned, which also made member pagination unstable across pages. Order both lists case-insensitively by the name the page actually renders: the profile's display name when set, falling back to the username, mirroring Profile.display_name.

Fixes #7270

### DIFF
--- a/kitsune/groups/views.py
+++ b/kitsune/groups/views.py
@@ -5,6 +5,8 @@ from django.contrib import messages
 from django.contrib.auth.models import Group, User
 from django.core.exceptions import PermissionDenied
 from django.core.files.storage import default_storage
+from django.db.models import Value
+from django.db.models.functions import Coalesce, Lower, NullIf
 from django.http import Http404, HttpResponseRedirect
 from django.shortcuts import get_object_or_404, render
 from django.utils.translation import gettext as _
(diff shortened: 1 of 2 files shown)


kitsune/groups/views.py
```diff
@@ -80,8 +85,8 @@ def list(request):
 
 def profile(request, group_slug, member_form=None, leader_form=None):
     prof = _get_group_profile_or_404(request.user, group_slug)
-    leaders = prof.leaders.all().select_related("profile")
-    members_qs = prof.group.user_set.all().select_related("profile")
+    leaders = prof.leaders.all().select_related("profile").order_by(DISPLAY_NAME_ORDER)
+    members_qs = prof.group.user_set.all().select_related("profile").order_by(DISPLAY_NAME_ORDER)
```
escattone, line 89: For both of these, since the profile name doesn't have to be unique, we should guarantee the order for the case when multiple users have the same profile name, by adding a secondary sort column name:

```
.order_by(DISPLAY_NAME_ORDER, "username")
```

and also add a test for that case.

escattone (changes requested):
Hey Smith! I hope all is well! Sorry for the delay in getting to this, but it's been crazy busy lately. This looks great, but just needs a small change and then should be good to go. Thanks for the help!

escattone:
Just so we can get this merged ASAP, I'll merge this as it is, and create a separate PR to address the issue I identified. Thanks again for the help @smithellis!

PR #7855, opened 2026-08-29
Make all hybrid search results reachable through pagination

### LINKED ISSUE #3316: Search pagination caps reachable search results at 50
**Steps to reproduce**
1. Type a search term which returns >50 search results. (e.g `my browser keeps forgetting my logins`)
2. Navigate through the pagination elements in order to view lower ranked search results.

**Expected behavior**
All returned search results can be accessible via the pagination elements.

**Actual behavior**
Since the pagination elements are capped to 5 pages (50 search results) the remaining search results are not accessible.

**Screenshots**
Image

**Desktop:**
 - OS: Windows 11
 - Browser: Firefox

**Additional context**
This issue is reproducible in `stage`.

amannsan:
## Summary

- replace the five-page offset cap with a short-lived, opaque search session
- retrieve bounded RRF segments, exclude every family already encountered, and append new families to immutable page boundaries
- re-authorize cached KB candidates against the primary database before every page is presented
- carry the session through server-rendered and instant-search pagination, while clearing it when the query or filters change
- refresh read-only sessions without writing an older snapshot back over a newly extended ranking

## Why

The displayed count is an approximate family cardinality over the matching query, while the old paginator could only address the first 50 results inside one native-RRF window. Raising the page cap is not sufficient: compound retrievers do not support `search_after`, and increasing `rank_window_size` with the requested page would change earlier ordering and make deep requests scale with the page number.

This change keeps each Elasticsearch request bounded to 50 families plus a has-more probe. Following **Next** progressively extends the cached sequence with a new RRF segment that excludes all previously encountered families. Arbitrary deep page numbers can extend only one continuation segment per request.

Ranks and page boundaries are stable once pinned. RRF scores remain segment-local and are not compared across segment boundaries; appended families receive monotonic global display ranks. If the session expires or no longer matches the query, index generation, ranking settings, or viewer access, the request restarts cleanly at page 1.

If the shared cache cannot persist the session, the response exposes only the currently reachable authorized results and hides broken pagination. Searches that fit on one page do not create a session.

## Verification

- repository pre-commit suite, including Ruff and mypy
- 49 focused Python tests covering retrieval, authorization, complete traversal, session continuation, views, and configuration
- Redis-backed orchestration tests using the production cache serializer
- Elasticsearch 9.2.5 integration test covering native RRF, family collapse, semantic floor, and exclusion-based continuation without duplicates
- instant-search webpack build, ESLint, and focused browser-side tests

Fixes mozilla/sumo#3316


akatsoulas:
Thank you again @amannsan for creating this PR. Really valuable work to better understand the tradeoffs of each approach. I am closing this PR per the explanation provided in the linked ticket

For reference
https://github.com/mozilla/sumo/issues/3316#issuecomment-5522234652

PR #7861, opened 2026-08-31
condition stale machine translations on non-SumoBot revisions as well

escattone:
mozilla/sumo#3325

In addition to the current conditions, stale machine translations will be paused if an unreviewed, up-to-date translation exists that has been created within the review grace period and by someone other than SumBot. This applies to both AI and hybrid locales.

The intention is to provide a grace period for contributors to create, review, and approve a human translation, especially after reviewing and **not** approving a pending machine translation.


kitsune/wiki/services.py
```diff
@@ -25,34 +25,38 @@ class TranslationQueryBuilder:
...
+        created_within_grace_period = Q(
+            created__gte=Now() - timedelta(hours=settings.HYBRID_REVIEW_GRACE_PERIOD)
+        )
+
         return Revision.objects.filter(
             document=OuterRef("pk"),
-            creator=self.sumo_bot,
             is_approved=False,
             reviewed__isnull=True,
             based_on_id=OuterRef("parent__latest_localizable_revision_id"),
```
akatsoulas: Now that we include contributor revisions, maybe we should align with `get_pending_translations` and `get_obsolete_translations` and use `_gte`. A revision based on a newer than `latest_localizable_revision` will not block

kitsune/wiki/services.py
```diff
@@ -25,34 +25,38 @@ class TranslationQueryBuilder:
...
+        A revision qualifies when it's awaiting review, covers the latest localizable
+        English revision, and either we created it or someone else created it recently
+        enough that they may still be reviewed within the grace period.
 
         Returns:
-            Subquery that checks for unreviewed sumo bot revisions
+            Subquery that checks for unreviewed revisions
         """
+        created_within_grace_period = Q(
+            created__gte=Now() - timedelta(hours=settings.HYBRID_REVIEW_GRACE_PERIOD)
```
akatsoulas: Isn't his applied to the AI-locale as well? If yes, let's remove the `HYBRID` part

akatsoulas (approved):
r+wc - just a couple of nits. 
Can we remove the formatter noise from the pr?

PR #7872, opened 2026-09-03
Boost KB title matches

akatsoulas:


escattone (commented):
r+wc

I don't think this works for queries containing advanced syntax, right? That's probably something that would have to be addressed with a different chunk of work anyway.

escattone (approved):

PR #7889, opened 2026-09-07
Ignore empty {for} tags

### LINKED ISSUE #3300: Uncaught TypeError: Cannot read properties of undefined (reading 'split') in show for tags
**Describe the bug**
A wiki-markup {for} with no criteria renders as `<span class="for">` with no data-for attribute. `ShowFor.prototype.initShowFuncs` assumes data-for is always present and `calls .split()` on it unconditionally, throwing a `TypeError. Because new ShowFor()` runs at module scope in document.js, the throw aborts evaluation of the entire module, taking every other feature in that bundle down with it.

**Preconditions**
Sign in to SUMO

**Steps to reproduce**
1. Create or edit a KB article whose content contains the following content: `<nowiki>{for}</nowiki>` inside a heading.
2. Save the changes.

**Expected behavior**
The content is shown (no criteria means no conditions to fail)

**Actual behavior**
Uncaught TypeError: Cannot read properties of undefined (reading 'split')

**Additional context**
* This issue is reproducible in both `stage` and `prod`.
* Maybe we should simply return if !showFor (https://github.com/mozilla/kitsune/blob/c4d8f15837c747906945a7aa30c4a7218c43b9cd/kitsune/sumo/static/sumo/js/showfor.js#L430-L431)

denyshon:
- Resolves https://github.com/mozilla/sumo/issues/3300

Ignore empty {for} tags completely (as they are not valid markup), and also (just in case) add a safeguard to `initShowFuncs` to prevent the crash directly.

### DIFF
--- a/kitsune/sumo/static/sumo/js/showfor.js
+++ b/kitsune/sumo/static/sumo/js/showfor.js
@@ -428,6 +428,9 @@ ShowFor.prototype.initShowFuncs = function () {
   this.showFuncs = new WeakMap();
   this.container.querySelectorAll('.for').forEach(function (elem) {
     var showFor = elem.dataset.for;
+    if (!showFor) {
+      return;
+    }
     var criteria = showFor.split(/\s*,\s*/);
     var showFunc = this.matchesCriteria.bind(this, criteria);
     this.showFuncs.set(elem, showFunc);
--- a/kitsune/wiki/tests/test_parser.py
+++ b/kitsune/wiki/tests/test_parser.py
@@ -634,28 +634,28 @@ class ForWikiTests(TestCase):
         """A {for} set off by itself or wrapping a block-level element should
         be a paragraph or other kind of block-level thing."""
         self.assertWikiHtmlEqual(
-            "Joe\n\n{for}Red{/for}\n\nBlow",
-            '<p>Joe</p><p><span class="for">Red</span></p><p>Blow</p>',
+            "Joe\n\n{for mac}Red{/for}\n\nBlow",
+            '<p>Joe</p><p><span class="for" data-for="mac">Red</span></p><p>Blow</p>',
         )
         self.assertWikiHtmlEqual(
-            "Joe\n\n{for}\n* Red\n{/for}\n\nBlow",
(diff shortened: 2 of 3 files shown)


kitsune/wiki/tests/test_parser.py
```diff
@@ -634,28 +634,28 @@ def test_block(self):
...
-            "Joe\n\nRed {for}riding{/for} hood\n\nBlow",
-            '<p>Joe</p><p>Red <span class="for">riding</span> hood</p><p>Blow</p>',
+            "Joe\n\nRed {for mac}riding{/for} hood\n\nBlow",
+            '<p>Joe</p><p>Red <span class="for" data-for="mac">riding</span> hood</p><p>Blow</p>',
         )
 
     def test_nested(self):
         """{for} tags should be nestable."""
         self.assertWikiHtmlEqual(
-            "{for mac}\nJoe\n\nRed {for}{for}riding\n{/for} hood{/for}\n\nBlow\n{/for}",
```
akatsoulas, line 655: The change will allow an empty `{for}` to leak into the article and the update of this test no longer catches it

kitsune/sumo/static/sumo/js/showfor.js
```diff
@@ -428,6 +428,9 @@ ShowFor.prototype.initShowFuncs = function () {
   this.showFuncs = new WeakMap();
   this.container.querySelectorAll('.for').forEach(function (elem) {
     var showFor = elem.dataset.for;
+    if (!showFor) {
```
akatsoulas, line 431: This should fix the error, why is the update in the regex needed?

denyshon: I want to update the regex so that we ignore empty {for}s completely, without requiring contributors to wrap them in complex escape syntax. They are not valid markup anyway, so there's no point in parsing them

kitsune/wiki/tests/test_parser.py
denyshon: Not sure what you mean... isn't the purpose of this test to check nestableness? If you want, I can create a new test ensuring that empty {for}s make it into the article unchaged, though

PR #7900, opened 2026-09-09
introduce chat token endpoint

escattone:
mozilla/sumo#3349

### DIFF
--- a/kitsune/customercare/urls.py
+++ b/kitsune/customercare/urls.py
@@ -8,4 +8,9 @@ urlpatterns = [
         views.update_topic,
         name="customercare.update_topic",
     ),
+    re_path(
+        r"^support-chat/jwt/(?P<product_slug>[^/]+)$",
+        views.chat_jwt,
+        name="customercare.chat_jwt",
+    ),
 ]
--- a/kitsune/customercare/utils.py
+++ b/kitsune/customercare/utils.py
@@ -40,6 +40,14 @@ def resolve_user_org_group(user) -> GroupProfile | None:
     return _nearest_ancestor_org(user, list(GroupProfile.objects.org_roots()))
 
 
+def is_eligible_for_chat(user, product: Product) -> bool:
+    """Whether this user is allowed to start a live chat about this product.
+
+    Stub. The rules aren't settled yet, so for now everyone is eligible.
+    """
+    return True
+
+
 def fetch_zendesk_ticket_data(zendesk_ticket_id: str):
     """Fetch ticket and comments from Zendesk."""
     client = ZendeskClient()
--- a/kitsune/customercare/views.py
+++ b/kitsune/customercare/views.py
@@ -5,6 +5,7 @@ import json
 import logging
 from datetime import timedelta
 
+import jwt
 import requests
 from django.conf import settings
 from django.contrib.auth.decorators import login_required, permission_required
@@ -17,23 +18,32 @@ from django.utils.dateparse import parse_datetime
 from django.utils.translation import gettext as _
 from django.utils.translation import pgettext
 from django.views import View
+from django.views.decorators.cache import cache_control
 from django.views.decorators.http import require_http_methods, require_POST
 from zenpy.lib.exception import APIException, RecordNotFoundException, ZenpyException
 
 from kitsune.customercare.forms import SupportTicketReplyForm
 from kitsune.customercare.models import SupportTicket
 from kitsune.customercare.tasks import process_zendesk_update
-from kitsune.customercare.utils import generate_classification_tags, sync_ticket_from_zendesk
+from kitsune.customercare.utils import (
+    generate_classification_tags,
+    is_eligible_for_chat,
+    sync_ticket_from_zendesk,
+)
 from kitsune.customercare.zendesk import ZendeskClient
 from kitsune.groups.templatetags.jinja_helpers import group_breadcrumbs
-from kitsune.products.models import Topic
+from kitsune.journal.models import Record
+from kitsune.products.models import Product, Topic
 from kitsune.sumo.urlresolvers import reverse
+from kitsune.sumo.utils import is_ratelimited
 
 log = logging.getLogger("k.customercare")
 
 # Zendesk failures we surface as a notice rather than a 500 (ZenpyException
 # covers client construction, e.g. missing credentials).
 ZENDESK_ERRORS = (APIException, ZenpyException, requests.exceptions.RequestException)
+
+CHAT_JOURNAL_SRC = "customercare.chat"
 
 
 def _ticket_needs_sync(ticket):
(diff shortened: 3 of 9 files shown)


kitsune/settings.py
```diff
@@ -1340,6 +1340,14 @@ def setup_logger(name, formatter_str, level=logging.DEBUG):
 ZENDESK_SYNC_TIMEOUT = config(
     "ZENDESK_SYNC_TIMEOUT", default="5,10", cast=Csv(int, post_process=tuple)
 )
+# Signing key for the Zendesk messaging widget, from Admin Center under
+# Account > Security > End user authentication. The key id goes in the JWT header.
+ZENDESK_CHAT_SIGNING_KEY_ID = config("ZENDESK_CHAT_SIGNING_KEY_ID", default="")
```
akatsoulas, line 1345: I would add another variable here to enable/disable the functionality

kitsune/customercare/utils.py
```diff
@@ -40,6 +40,14 @@ def resolve_user_org_group(user) -> GroupProfile | None:
     return _nearest_ancestor_org(user, list(GroupProfile.objects.org_roots()))
 
 
+def is_eligible_for_chat(user, product: Product) -> bool:
```
akatsoulas: Given that we don't know exactly how this is going to work, maybe we can just add a comment/TODO in the view about it

kitsune/settings.py
akatsoulas: Maybe this could be a waffle switch to control this more easily and checked in the view?

kitsune/customercare/views.py
```diff
@@ -53,6 +63,82 @@ def _reply_placeholder(ticket):
...
+    if not is_eligible_for_chat(request.user, product):
+        Record.objects.info(
+            CHAT_JOURNAL_SRC,
+            "user {username} is not eligible for chat on {slug}",
+            username=request.user.username,
+            slug=product.slug,
+        )
+        return HttpResponse(status=403)
+
+    token = jwt.encode(
```
akatsoulas, line 136: We should check and fail loudly if the signing key is empty - eg the default value of `""`. We could raise an `ImproperlyConfigured` error.

kitsune/customercare/urls.py
```diff
@@ -8,4 +8,9 @@
         views.update_topic,
         name="customercare.update_topic",
     ),
+    re_path(
```
akatsoulas: we should exempt this url from `i18n` patterns

kitsune/customercare/utils.py
akatsoulas: This combined with the waffle switch mentioned in the other comment should render this functionality unreachable in prod

kitsune/settings.py
akatsoulas: The view should return a 404 if the switch is not active - assuming we go with that

kitsune/customercare/views.py
```diff
@@ -53,6 +63,82 @@ def _reply_placeholder(ticket):
...
+    the Origin header.
+    """
+    if chat_jwt_is_ratelimited(request):
+        return HttpResponse(status=429)
+
+    if not request.user.is_authenticated:
+        return HttpResponse(status=401)
+
+    # Reverse one-to-one, so getattr covers a user with no profile row.
+    profile = getattr(request.user, "profile", None)
```
akatsoulas: Assuming an authenticated user reaches this point, I would assume a profile is guaranteed. This check should be redundant. I would even argue that if for some reason we are missing a profile at this stage, we should fail loudly

kitsune/customercare/views.py
```diff
@@ -53,6 +63,82 @@ def _reply_placeholder(ticket):
...
+            username=request.user.username,
+        )
+        return HttpResponse(status=401)
+
+    # Slug is user-supplied, so keep it short enough for the journal message.
+    product = Product.active.filter(slug=product_slug).first()
+    if product is None:
+        Record.objects.info(
+            CHAT_JOURNAL_SRC,
+            "user {username} asked for chat on unknown product {slug}",
```
akatsoulas: `username` can get anonymized. We should enhance this with `user.id` as well

akatsoulas: Given this is going to have a small audience - at least in the beginning - I wonder whether we want to also register the successful issuance of tokens

kitsune/customercare/views.py
```diff
@@ -53,6 +63,82 @@ def _reply_placeholder(ticket):
...
+@cache_control(no_store=True)
+def chat_jwt(request, product_slug):
+    """Sign a short-lived token that identifies the requesting user to Zendesk.
+
+    The messaging widget posts here whenever it needs to prove the requesting
+    user's identity, including when its previous token expires. POST rather than
+    GET so Django's CSRF middleware runs, which checks both the CSRF token and
+    the Origin header.
+    """
+    if chat_jwt_is_ratelimited(request):
```
akatsoulas: This should go after the check about authentication status

akatsoulas (changes requested):

kitsune/customercare/urls.py
escattone: Good catch!

kitsune/customercare/views.py
```diff
@@ -53,6 +60,105 @@ def _reply_placeholder(ticket):
...
+def chat_jwt(request, product_slug):
+    """Sign a short-lived token that identifies the requesting user to Zendesk.
+
+    The messaging widget posts here whenever it needs to prove the requesting
+    user's identity, including when its previous token expires. POST rather than
+    GET so Django's CSRF middleware runs, which checks both the CSRF token and
+    the Origin header.
+    """
+    if not waffle.switch_is_active("zendesk-chat"):
+        return HttpResponse(status=404)
```
akatsoulas: The pattern is to raise the Http404 throughout the project

kitsune/customercare/views.py
```diff
@@ -53,6 +60,105 @@ def _reply_placeholder(ticket):
...
+    if not user.is_authenticated:
+        return HttpResponse(status=401)
+
+    if chat_jwt_is_ratelimited(request):
+        return HttpResponse(status=429)
+
+    profile = user.profile
+    external_id = profile.fxa_uid
+
+    if not external_id:
```
akatsoulas: This is not needed. If a user is logged in - which is checked above - there will be a fxa_uid

kitsune/customercare/views.py
```diff
@@ -53,6 +60,105 @@ def _reply_placeholder(ticket):
...
+            "user {username} (#{user_id}) has no fxa_uid",
+            username=user.username,
+            user_id=user.id,
+        )
+        return HttpResponse(status=401)
+
+    # Slug is user-supplied, so keep it short enough for the journal message.
+    product = Product.active.filter(slug=product_slug).first()
+    if product is None:
+        Record.objects.info(
```
akatsoulas, line 104: nit: There's repetition creating the log. Maybe we can use a dictionary with the entries that the `Record` needs per case, set the status code in a variable and compact the code for the various edge cases

akatsoulas (approved):
r+wc

PR #7901, opened 2026-09-10
Replace hyphens with dashes in urgency choices

denyshon:


JC-Support:
No objections to this change. Hyphen was used because I didn't think about a specific key dash.  Thanks to Deny for calling this out.

akatsoulas (approved):

PR #7904, opened 2026-09-10
Disable semantic search on search terms queries

akatsoulas:


kitsune/search/parser/__init__.py
```diff
@@ -102,3 +102,27 @@ def elastic_query(self, context=None):
...
+    Operators and quoted phrases promise exact lexical behavior, so consumers such
+    as semantic retrieval must not broaden them. A query that fails to parse is not
+    provably plain text and is treated as ineligible rather than broadened.
+    """
+    if not query or not query.strip():
+        return True
+    try:
+        parsed = Parser(query).parsed
+    except ParseException:
+        return False
```
escattone, line 127: I'm not sure why something that fails to parse shouldn't be broadened, because it seems like the failure to parse also indicates a failure to detect/understand an advanced search query. In other words, why is the `ParseException` case different than what `QuestionSearch.is_simple_search` does and what I think `build_lexical_clauses` does?

escattone (approved):
r+wc

kitsune/search/parser/__init__.py
akatsoulas: `is_simple_search` defaults to `True` b/c it basically controls the exclusion of the archived threads. If there's an exception, it means that something went wrong so default to norm and do not search into the archived entries. We are doing the same thing here. If there's an error, then we are just passing the text to ES to figure it out without enabling Semantic. We are basically defaulting to the cheaper option in both cases.

Examples
firefox (sync -> would benefit from semantic but it defaults to lexical b/c it's computationally cheaper
firefox "sync error" ( -> again we have a parse error but ES understands the intention and returns the right results. 

In all cases we are choosing the more conservative approach

PR #7905, opened 2026-09-10
introduce chat widget

escattone:
mozilla/sumo#3352
mozilla/sumo#3353 (the revocation, but not the task of informing the ZD agent)
mozilla/sumo#3354


kitsune/customercare/templatetags/jinja_helpers.py
```diff
@@ -0,0 +1,29 @@
...
+
+@library.global_function
+def chat_is_available(request, products=None) -> bool:
+    """Whether to show the Zendesk chat widget on a page about these products.
+
+    Pass the products the page is about. Any one of them being a chat product is
+    enough. Omit them for a page that isn't about a product, such as the home
+    page, which skips the product check but is still gated on everything else.
+    """
+    if not (waffle.switch_is_active("zendesk-chat") and settings.ZENDESK_CHAT_WIDGET_KEY):
```
akatsoulas: Those are the same checks that the jwt endpoint is performing. Maybe we should have a function in the utils.py which can be used both here and in the jwt endpoint

kitsune/landings/jinja2/landings/home.html
```diff
@@ -3,6 +3,7 @@
 {% from 'includes/common_macros.html' import featured_articles %}
 {% set canonical_url = canonicalize(viewname='home') %}
 {% set ga_content_group = "home" %}
+{% set show_chat = chat_is_available(request) %}
```
akatsoulas: Chat must be available only to the products that have a config that can dynamically add chat based on groups. Products that don't have ticketing support to begin with should not get the widget. This can be a blocker in a future state where we introduce different widget for different products

kitsune/settings.py
```diff
@@ -1346,12 +1346,43 @@ def setup_logger(name, formatter_str, level=logging.DEBUG):
...
-ZENDESK_CHAT_RATELIMITS = config("ZENDESK_CHAT_RATELIMITS", default="10/m,60/h,300/d", cast=Csv())
+# Multi-window limits (see kitsune.customercare.views.chat_jwt_is_ratelimited). The
+# widget asks for a token on every page load, so these have to clear ordinary reading.
+ZENDESK_CHAT_RATELIMITS = config(
+    "ZENDESK_CHAT_RATELIMITS", default="60/m,600/h,2000/d", cast=Csv()
+)
+# Public key from the widget's installation snippet in Admin Center.
+ZENDESK_CHAT_WIDGET_KEY = config("ZENDESK_CHAT_WIDGET_KEY", default="")
+# Show the widget on pages about any of these products.
+ZENDESK_CHAT_PRODUCT_SLUGS = config(
```
akatsoulas: We should not hardcode this but rather get the info from the product configs

kitsune/settings.py
```diff
@@ -1346,12 +1346,43 @@ def setup_logger(name, formatter_str, level=logging.DEBUG):
...
+    "ZENDESK_CHAT_PRODUCT_SLUGS",
+    default="firefox,ios,mobile,firefox-enterprise",
+    cast=Csv(),
+)
+# The product every chat is checked against, whichever page it starts from.
+ZENDESK_CHAT_ELIGIBILITY_PRODUCT_SLUG = config(
```
akatsoulas: Similar to the above

kitsune/settings.py
```diff
@@ -1346,12 +1346,43 @@ def setup_logger(name, formatter_str, level=logging.DEBUG):
...
+    "ZENDESK_CHAT_ELIGIBILITY_PRODUCT_SLUG", default="firefox-enterprise"
+)
+# SUMO locale -> Zendesk locale, for the pages that get the widget. Any other
+# locale doesn't show it. Spell the Zendesk side exactly as it appears in
+# https://support.zendesk.com/api/v2/locales/public.json - that list mixes cases
+# ("en-US" but "pt-br"), the codes don't always agree with ours (SUMO "ne-NP" is
+# Zendesk "ne"), and Zendesk silently ignores a code it doesn't recognise.
+ZENDESK_CHAT_LOCALES = {
```
akatsoulas: Can't we map only the differences? If there's not an entry, we can directly send the locale

kitsune/sumo/static/sumo/js/zendesk-chat.js
```diff
@@ -0,0 +1,59 @@
...
+
+export function signInToChat(jwtUrl, sessionUser) {
+  if (!jwtUrl || !sessionUser || typeof window.zE !== 'function') {
+    return;
+  }
+
+  let signedIn = window.localStorage.getItem(STORAGE_KEY);
+
+  if (signedIn && signedIn !== sessionUser) {
+    // Clear the previous user out. loginUser on its own doesn't replace them -
```
akatsoulas: let's clean up the comments. Those are too verbose with little value. Eg a comment in the code that something was tested in dev is not of great value (when, which dev, how etc). Let's keep comments on what the code does - where it's needed

kitsune/sumo/static/sumo/js/zendesk-chat.js
```diff
@@ -0,0 +1,59 @@
...
+})();
+
+export function signInToChat(jwtUrl, sessionUser) {
+  if (!jwtUrl || !sessionUser || typeof window.zE !== 'function') {
+    return;
+  }
+
+  let signedIn = window.localStorage.getItem(STORAGE_KEY);
+
+  if (signedIn && signedIn !== sessionUser) {
```
akatsoulas: We should also signout the user when there's a logout from sumo, not only when switching users

kitsune/landings/jinja2/landings/home.html
akatsoulas: We should enable the chat widget in the ticket detail page and its parent as well. I also wonder whether the group page of an org that has chat support should enable it

kitsune/sumo/jinja2/base.html
```diff
@@ -18,7 +18,11 @@
   data-matomo-cdn-host="{{ settings.MATOMO_MZLA_CDN_HOST }}"
   {% endif %}
   {% if settings.GA_CONSOLE_LOGGING %}data-ga-console-logging="true"{% endif %}
-  {% if settings.GA_DEBUG_MODE %}data-ga-debug-mode="true"{% endif %}>
+  {% if settings.GA_DEBUG_MODE %}data-ga-debug-mode="true"{% endif %}
+  {% if show_chat %}
+  data-zendesk-chat-jwt-url="{{ url('customercare.chat_jwt', product_slug=settings.ZENDESK_CHAT_ELIGIBILITY_PRODUCT_SLUG) }}"
+  data-zendesk-chat-user="{{ user.username }}"
```
akatsoulas: Username is not a good idenitfier here. We should either use the user.id or the fxa_uid similar to what we do with the jwt endpoint. Username can be altered for the same user and in edge cases could be obtained by a different user

kitsune/settings.py
```diff
@@ -1364,12 +1395,22 @@ def setup_logger(name, formatter_str, level=logging.DEBUG):
             # cdn.matomo.cloud is multi-tenant; a bare host source would defeat NONCE.
             f"https://{MATOMO_MZLA_CDN_HOST}/{MATOMO_MZLA_TRACKER_HOST}/matomo.js",
             "https://pontoon.mozilla.org",
+            *ZENDESK_CHAT_ASSET_HOSTS,
```
akatsoulas: The existing CSP configuration will relax the rules globally. We should only make the rules more relaxed only where the widget is loaded. We already have such an exception with the `AdminCSPMiddleware`

akatsoulas (changes requested):

kitsune/customercare/templatetags/jinja_helpers.py
akatsoulas: some of this checks can be replaced by the functionality in PR https://github.com/mozilla/kitsune/pull/7913

kitsune/settings.py
escattone: I hard-coded it just to get it working, but I don't think we have a way of specifying this in our current product configs. Abby would like to enable chat for products **related** to a product that is enabled for live chat. So, for example, although Firefox for iOS, Firefox desktop and Firefox for Android are not products that will have a product config with support orgs enabled for live chat, they're so closely related to FFE that Abby would like the chat widget enabled for pages related to those products as well (but of course, only if the user is a member of an org enabled for live-chat). How should we handle this?

kitsune/settings.py
escattone: 👍 Yes, that's a better approach.

kitsune/sumo/jinja2/base.html
escattone: By the way, live-chat will still work if the username is changed (there will be a logout followed by a login with the new username), but I agree it would be better to use something that is constant.

kitsune/settings.py
```diff
@@ -1414,12 +1446,15 @@ def setup_logger(name, formatter_str, level=logging.DEBUG):
...
+            *ZENDESK_CHAT_API_HOSTS,
+            # The widget holds a live connection for the conversation.
+            "wss://*.zendesk.com",
         ],
     },
 }
 
 if DEBUG:
-    CONTENT_SECURITY_POLICY["DIRECTIVES"]["style-src"].append(UNSAFE_INLINE)
```
akatsoulas, line 1422: This removal probably breaks GraphiQL

kitsune/customercare/templatetags/jinja_helpers.py
```diff
@@ -0,0 +1,34 @@
...
+from kitsune.customercare.utils import is_chat_enabled, resolve_chat_eligibility
+
+if TYPE_CHECKING:
+    from kitsune.products.models import Product
+
+
+@library.global_function
+def chat_is_available(request, product: Product | None) -> bool:
+    """Whether to show the Zendesk chat widget on a page about this product."""
+    if not product:
```
akatsoulas, line 15: super nit: All 3 checks could be combined in an `any()` statement

kitsune/customercare/utils.py
```diff
@@ -23,6 +25,23 @@
 from kitsune.users.models import Profile
 
 
+def is_chat_enabled() -> bool:
```
akatsoulas: Why don't we fold this into the `resolve_chat_eligibility` to have a centralized place for managing access?

kitsune/products/jinja2/products/documents.html
```diff
@@ -30,6 +30,7 @@
 
 {% set scripts = ('topics',) %}
 {% set ga_topics = "/" + topic.slug + "/" %}
+{% set show_chat = chat_is_available(request, product) %}
```
akatsoulas: Do we want it in the documents? Shall we just remove it and figure session stickiness later?

kitsune/questions/jinja2/questions/question_details.html
```diff
@@ -39,6 +39,7 @@
   {% set ga_products = "/" + product.slug + "/" %}
 {% endif %}
 {% set matomo_mzla = matomo_mzla_enabled([product] if product else []) %}
+{% set show_chat = chat_is_available(request, product) %}
```
akatsoulas: Similar comment with the documents

kitsune/questions/jinja2/questions/question_list.html
```diff
@@ -13,6 +13,8 @@
 {% elif multiple_products %}
   {% set page_title = _('Multiple Products') %}
 {% else %}
+  {% set product = products[0] %}
```
akatsoulas: same comment

kitsune/sumo/static/sumo/js/zendesk-chat.js
```diff
@@ -0,0 +1,87 @@
...
+        if (![401, 403, 404].includes(error.response?.status)) {
+          console.error('Could not get a Zendesk chat token.', error);
+        }
+      });
+  }
+
+  // Login the user on every page load. The widget doesn't restore an authenticated session
+  // on its own, so skipping this leaves the user anonymous with a fresh, empty conversation.
+  window.zE('messenger', 'loginUser', fetchToken, (error) => {
+    if (error) {
```
akatsoulas, line 52: Shouldn't we call `removeChat()` here?

kitsune/products/jinja2/products/documents.html
akatsoulas: Instead I would argue that we need the widget in the ticketing pages (ticket detail and my questions)

kitsune/settings.py
```diff
@@ -1365,12 +1381,22 @@ def setup_logger(name, formatter_str, level=logging.DEBUG):
...
+            # script tag for /sc/faye. Drop this and chat says "Offline. You won't
+            # receive messages." - but only once the fallback is actually used.
+            *ZENDESK_CHAT_API_HOSTS,
             NONCE,
         ],
         "img-src": [
             SELF,
             "blob:",
             "data:",
+            *ZENDESK_CHAT_ASSET_HOSTS,
```
akatsoulas: nit: given that we conditionally relax style-src, shall we also inject the ZD domains only where chat is available?

kitsune/settings.py
```diff
@@ -1347,12 +1347,28 @@ def setup_logger(name, formatter_str, level=logging.DEBUG):
...
 
 # Products that allow un-authenticated users to submit support requests
 LOGIN_EXCEPTIONS = frozenset(["mozilla-account"])
 
+# Zendesk messaging widget. Wildcards because the widget lazy-loads chunks, and
+# because the account subdomain differs per environment.
+# The widget's code and its lazy-loaded chunks.
+ZENDESK_CHAT_ASSET_HOSTS = ["https://*.zdassets.com"]
+# The account's own API, websocket, and JSONP polling; subdomain differs per environment.
+ZENDESK_CHAT_API_HOSTS = ["https://*.zendesk.com"]
```
akatsoulas, line 1370: Shouldn't those be limited to our tenant?

kitsune/settings.py
```diff
@@ -1347,12 +1347,28 @@ def setup_logger(name, formatter_str, level=logging.DEBUG):
...
-ZENDESK_CHAT_RATELIMITS = config("ZENDESK_CHAT_RATELIMITS", default="10/m,60/h,300/d", cast=Csv())
+# Multi-window limits (see kitsune.customercare.views.chat_jwt_is_ratelimited). The
+# widget asks for a token on every page load, so these have to clear ordinary reading.
+ZENDESK_CHAT_RATELIMITS = config(
+    "ZENDESK_CHAT_RATELIMITS", default="60/m,600/h,2000/d", cast=Csv()
+)
+# Public key from the widget's installation snippet in Admin Center.
+ZENDESK_CHAT_WIDGET_KEY = config("ZENDESK_CHAT_WIDGET_KEY", default="")
+# The locales within which the chat widget can be used.
+ZENDESK_CHAT_ENABLED_LOCALES = config(
```
akatsoulas: Do we actually need this? I understand that we need to translate but is this a gate for chatting? Eg can't I have a locale of my choosing but contact support in English? I would just scope the automatic replies in a small listed of supported locales and drop this check altogether

kitsune/customercare/tests/test_jinja_helpers.py
```diff
@@ -0,0 +1,89 @@
...
+
+    def test_user_without_chat_access(self):
+        self.user.groups.clear()
+
+        self.assertFalse(chat_is_available(self._request(), self.product))
+
+    def test_anonymous_user(self):
+        self.assertFalse(chat_is_available(self._request(user=AnonymousUser()), self.product))
+
+    def test_enabled_locale(self):
```
akatsoulas: This could be deleted. Very similar call with `test_entitled_user`

kitsune/customercare/tests/test_jinja_helpers.py
```diff
@@ -0,0 +1,89 @@
...
+        return SimpleNamespace(user=user or self.user, LANGUAGE_CODE=locale)
+
+    def test_entitled_user(self):
+        self.assertTrue(chat_is_available(self._request(), self.product))
+
+    def test_page_that_is_not_about_a_product(self):
+        self.assertFalse(chat_is_available(self._request(), None))
+
+    @override_settings(ZENDESK_CHAT_WIDGET_KEY="")
+    def test_a_page_without_a_product_does_not_care_about_the_configuration(self):
```
akatsoulas: overlap with the test above

kitsune/sumo/static/sumo/js/tests/zendesk-chat-tests.js
```diff
@@ -0,0 +1,279 @@
+import { expect } from "chai";
```
akatsoulas, line 1: From AI review 
- HTTP failures are exercised twice: four tests check cleanup for 401/403/404/429, then another four check logging for those same responses. Combine them into four parameterized cases, each checking cleanup, no token delivery, and the expected logging behavior. 
 - Previous-user replacement is split unnecessarily: `signs the previous user out before signing a different one in` and `forgets the previous user as soon as they are signed out` have identical setup and action. One test can assert both the logout/login sequence and removal of the old stored identity.

kitsune/sumo/static/sumo/js/tests/sign-out-sync-tests.js
```diff
@@ -0,0 +1,53 @@
...
+    document.body.appendChild(form);
+    window.localStorage.removeItem(SIGN_OUT_KEY);
+  });
+
+  afterEach(() => {
+    form.remove();
+    window.localStorage.removeItem(SIGN_OUT_KEY);
+  });
+
+  it("announces a sign-out for the other tabs to hear", () => {
```
akatsoulas: This is almost the same with the next one. A broken implementation that always write `2` passes both

kitsune/sumo/tests/test_middleware.py
```diff
@@ -360,3 +361,78 @@ def test_doubled_percent_nul_in_a_filter(self):
...
+    def test_admin_allows_its_own_scripts(self):
+        directives = self._directives(path="/admin/")
+
+        self.assertIn("'self'", directives["script-src"])
+        self.assertIn("https://*.webservices.mozgcp.net", directives["script-src"])
+
+    def test_other_paths_do_not_allow_admin_scripts(self):
+        self.assertNotIn("'self'", self._directives()["script-src"])
+
+    def test_it_is_the_only_csp_middleware(self):
```
akatsoulas: This just asserts a setting. Can be dropped

akatsoulas (changes requested):

kitsune/customercare/tests/test_jinja_helpers.py
escattone: It's testing that the product check is done prior to the configuration check, but it's not important, so I'll remove it anyway.

kitsune/customercare/utils.py
escattone: If we do that, there are consequences within the `chat_jwt` view. Do we put the `chat_jwt_is_ratelimited` call before or after the `resolve_chat_eligibility` call?

I think we have to rule out placing it after, because the rate-limit protection would be defeated. We'd be making all of the database calls to check eligibility, and then enforcing the rate limit.

However, if we put it before, it establishes some unorthodox behavior for the 404 case. For example, if the waffle switch for chat is not active, we will return 404's until the rate limit is reached, and then start returning 429's.

We can do this, but I think we should return 403's instead of 404's when the waffle switch is not active.

kitsune/products/jinja2/products/documents.html
escattone: Let's remove it and work on stickiness later.

We can add the chat widget to the ticket detail page, because we know the product, but not the `My questions` page because there's no single product implied in order to check if live-chat is allowed. However, I don't see why it makes sense to show the chat widget on the ticket detail page. What's the use case?

kitsune/questions/jinja2/questions/question_details.html
escattone: I'll remove it.

kitsune/questions/jinja2/questions/question_list.html
escattone: I'll remove it.

kitsune/sumo/static/sumo/js/zendesk-chat.js
escattone: Yes, good catch.

kitsune/customercare/views.py
```diff
@@ -92,24 +87,18 @@ def chat_jwt(request, product_slug):
     if chat_jwt_is_ratelimited(request):
         return HttpResponse(status=429)
 
-    user_info = f"user {user.username} (#{user.id})"
+    product = get_object_or_404(Product.active, slug=product_slug)
```
escattone, line 90: I decided against recording a bad product slug, as it provides an attack vector for filling the database with journal records.

kitsune/settings.py
escattone: No, from what I discovered, Zendesk creates and uses sub-domains in the form of `pod-NN-sunco-ws.zendesk.com` where the `NN` chosen by Zendesk. From what I could tell from their docs -- which aren't the greatest on this topic -- they recommend `https://*.zendesk.com`.

kitsune/settings.py
```diff
@@ -1347,12 +1347,54 @@ def setup_logger(name, formatter_str, level=logging.DEBUG):
...
+# Multi-window limits (see kitsune.customercare.views.chat_jwt_is_ratelimited). The
+# widget asks for a token on every page load, so these have to clear ordinary reading.
+ZENDESK_CHAT_RATELIMITS = config(
+    "ZENDESK_CHAT_RATELIMITS", default="60/m,600/h,2000/d", cast=Csv()
+)
+# Public key from the widget's installation snippet in Admin Center.
+ZENDESK_CHAT_WIDGET_KEY = config("ZENDESK_CHAT_WIDGET_KEY", default="")
+# The non-English locales supported by the Zendesk widget and agents.
+ZENDESK_CHAT_SUPPORTED_NON_ENGLISH_LOCALES = config(
+    "ZENDESK_CHAT_SUPPORTED_NON_ENGLISH_LOCALES", default="de", cast=Csv()
```
akatsoulas: Let's default this to either an empty string or `en-US`. I would go for the latter b/c that's what we do in the rest of the site

akatsoulas (approved):
r+wc

PR #7907, opened 2026-09-12
Introduce emails to solution authors

### LINKED ISSUE #2875: Send email notifications to contributors whose replies were marked as solutions
Currently, you only get email notifications if a reply has been posted (or, if you explicitly subscribe to that, if any reply is marked as a solution). However, to incentivize the contributors whose replies actually solved an issue, we should send them dedicated notifications when their reply is marked as a solution. These emails should thank them for helping and prompt them to continue contributing to SUMO.

@kelimuttu @couci I'd appreciate your opinion on this. Also, if you support the idea, could you please draft some text for the emails? For reference, here are the currently used templates: https://github.com/mozilla/kitsune/tree/main/kitsune/questions/jinja2/questions/email

Also, if we add such notifications, do we want to provide an option to disable them? If so, where? A checkmark in the profile settings maybe?

denyshon:
- Resolves https://github.com/mozilla/sumo/issues/2875

Introduce emails to solution authors and a related setting. If a user has opted in to receive these emails or if they are subscribed to notifications about solutions for a particular question thread, send them the new email when their reply is marked as a solution.

Create a dedicated test for the new functionality and update a Playwright test for the profile settings page.

### DIFF
--- a/kitsune/questions/events.py
+++ b/kitsune/questions/events.py
@@ -10,6 +10,7 @@ from kitsune.sumo import email_utils
 from kitsune.sumo.templatetags.jinja_helpers import add_utm, urlparams
 from kitsune.sumo.urlresolvers import reverse
 from kitsune.tidings.events import InstanceEvent
+from kitsune.users.models import Setting
 
 
 class QuestionEvent(InstanceEvent):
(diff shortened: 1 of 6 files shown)


hamzahamidi (commented):
Could we add coverage for a solution author who is already reached through the watcher path with `email_authored_solutions` disabled? That covers the interaction between recipient deduplication and the author-specific preference, and verifies the author receives neither an unwanted notification nor a duplicate.

kitsune/questions/events.py
```diff
@@ -219,6 +240,17 @@ def _make_mail(locale, user, context):
 
             yield _make_mail(locale, u, c)
 
+            if not is_answerer_handled and u.id == answerer_id:
+                is_answerer_handled = True
+
+        # If we haven't sent the email to the solution creator yet, send it now.
+        if not is_answerer_handled:
```
hamzahamidi, line 247: Please add a test where the author is already in `users_and_watches` and has opted out. That is the branch that proves this fallback neither duplicates an existing recipient nor bypasses the preference.

denyshon: I don't think that's necessary

denyshon:
No, I don't think this edge case requires a dedicated test

denyshon:
The setting:

image

denyshon:
Default emails:

Image

Image

denyshon:
Emails when the user is subscribed to solution notifications for a particular question thread (regardless of the new setting):

image

image

denyshon:
@kelimuttu I've raised the remaining questions in the issue, but you can find more screenshots above. The PR is a draft while we are figuring everything out

PR #7912, opened 2026-09-14
Support organization scaffolding

akatsoulas:


kitsune/products/models.py
```diff
@@ -439,6 +440,123 @@ def enable_zendesk_support(self):
...
+        # Local import: products.models -> groups.models -> wiki.parser -> wiki.models
+        # -> products.models.
+        from kitsune.groups.models import GroupProfile
+
+        # Inline admin rows validate against a parent config that may not be saved yet,
+        # so work from the related object rather than from config_id.
+        try:
+            config = self.config
+        except ProductSupportConfig.DoesNotExist:
+            return
```
escattone, line 490: This won't validate the row when adding it to a `ProductSupportConfig` instance that's in the process of being created. Can we perform the `clean` in an inline-formset, so we can use the in-memory `ProductSupportConfig` instance for the validation -- something similar to what we do [here](https://github.com/mozilla/kitsune/blob/d61d9f51c72cff1168cd7ecc4016df2429f04abe/kitsune/questions/admin.py#L47)?

kitsune/products/utils.py
```diff
@@ -1,36 +1,45 @@
...
+    return (
+        GroupProfile.objects.containing(user).filter(slug=settings.ENTERPRISE_GROUP_SLUG).exists()
+    )
+
 
-    cache_key = f"enterprise_hybrid_banner:{user.pk}"
-    result = cache.get(cache_key)
+def should_show_enterprise_banner(user) -> bool:
+    """Show enterprise members the banner when Zendesk is the configured default."""
+    if not is_enterprise_user(user):
```
escattone: Nit. For performance purposes, is it worth re-ordering the cache check for availability and this `is_enterprise_user` call? So do the cache check first and then the `is_enterprise_user` call, just in case we can skip a DB call?

escattone (approved):
r+wc

Also, could `SupportTicketQuerySet.accessible_to` be simplified using `GroupProfileManager.containing`? Maybe not, just wondering.

PR #7913, opened 2026-09-15
Resolve chat eligibility

akatsoulas:


kitsune/customercare/utils.py
```diff
@@ -11,11 +12,61 @@
...
+            return ChatEligibility(reason="ambiguous_multi_org")
+    if organizations.exists():
+        return ChatEligibility(reason="tier_without_chat")
+    if SupportOrganization.objects.filter(
+        group_id__in=member_groups,
+        config__is_active=True,
+        config__zendesk_config__isnull=False,
+        config__product__is_archived=False,
+    ).exists():
+        return ChatEligibility(reason="product_mismatch")
```
escattone: Is this case worth another DB query? I would argue against this, because this is on the most common path. Most users are not a member of a support organization, so most users will trigger this query.

If we do want to keep this query, shouldn't we add `include_live_chat=True` to the `filter` call?

kitsune/customercare/utils.py
```diff
@@ -11,11 +12,61 @@
...
+    if config.subscription_only and not user.profile.products.filter(pk=product.pk).exists():
+        return ChatEligibility(reason="subscription_required")
+
+    member_groups = GroupProfile.objects.containing(user).values("group_id")
+    organizations = config.support_organizations.filter(group_id__in=member_groups)
+    match list(organizations.filter(include_live_chat=True)[:2]):
+        case [org]:
+            return ChatEligibility(org=org)
+        case [_, _]:
+            return ChatEligibility(reason="ambiguous_multi_org")
```
escattone: For the `ambiguous_multi_org` case, I don't understand why we're denying eligibility. I guess it's considered a mis-configuration? Should we include the multiple orgs in the `reason` so we can more quickly fix the configuration?

kitsune/customercare/utils.py
```diff
@@ -11,11 +12,61 @@
...
+    if organizations.exists():
+        return ChatEligibility(reason="tier_without_chat")
```
escattone: It seems like we're introducing a new term here with the word `tier`. What do you think about `"org_without_chat"` instead?

kitsune/customercare/utils.py
```diff
@@ -11,11 +12,61 @@
...
+    if SupportOrganization.objects.filter(
+        group_id__in=member_groups,
+        config__is_active=True,
+        config__zendesk_config__isnull=False,
+        config__product__is_archived=False,
+    ).exists():
```
escattone: Is this case worth another DB query? I would argue against this, because it's on the most common path. Most signed-in users are not a member of a support organization, so most will trigger this query.

If we do want to keep this query, shouldn't we add `include_live_chat=True` to the `filter` call?

escattone (approved):
r+wc

PR #7914, opened 2026-09-16
Optimize tests

akatsoulas:


akatsoulas:
This PR reduces the CI time more than 45%. It cuts almost 15 minutes of running time from ~29 minutes to ~16

pyproject.toml
```diff
@@ -129,6 +129,8 @@ dev = [
     "google-api-python-client>=2.198.0,<3",
     "google-auth-oauthlib>=1.4.0,<2",
     "playwright>=1.61.0",
+    "pyyaml>=6,<7",
```
escattone: I think we can move this into the main dependency list because we're importing it [here](https://github.com/mozilla/kitsune/blob/dab6616fe99674e540416347fd80c485e8e92ecf/kitsune/products/utils.py#L3).

escattone (approved):
Nice! r+wc

PR #7917, opened 2026-09-18
Optimize tests

akatsoulas:


kitsune/retrieval/locks.py
```diff
@@ -27,7 +27,7 @@
 from kitsune.sumo.redis_utils import redis_client
 
 NAMESPACE = "retrieval:"
-KEY_PREFIX = f"{NAMESPACE}lease"
+KEY_PREFIX = settings.RETRIEVAL_LOCK_KEY_PREFIX
```
escattone, line 30: The `RETRIEVAL_LOCK_KEY_PREFIX` is configurable, but it seems that it must start with `NAMESPACE`, which is hard-coded, or the check in `redis_lease()` below will fail. It seems that we need some way to ensure they can't diverge?

escattone (approved):
r+wc

